PR #24, opened 2020-01-09
perf: close sessions async

### LINKED ISSUE #19: Spanner - Closing session pool is slow for large pools
Calling `Spanner#close()` can be slow if the underlying session pool contains a large number of sessions. `Spanner#close()` will close and cleanup all resources that have been used by the instance, and one of these is the session pool. The method will block until all resources have been cleaned up.

The session pool deletes all its sessions when it is closed. This is done asynchronously using a fixed 8-worker thread executor. Each worker thread blocks while waiting for a session to be deleted. This means that closing a session pool with 1,000 sessions will have an execution time of `T * 1000 / 8`, where `T` is the total time needed to delete one session.

Session deletion could be further parallelized for large session pools in order to reduce the total execution time of `Spanner#close()`.

olavloite:
Sessions should be closed using an async gRPC call in order to speed up the closing of a large session pool. Instead of using its own executor, which is limited to 8 worker threads, to execute asynchronous delete session calls, the session pool should use the asynchronous call option in gRPC. This allows a larger number of asynchronous delete session calls to be executed in parallel and speeds up closing a session pool with a large number of sessions.

Testing against a real Cloud Spanner database with a session pool containing 1,000 sessions shows the following performance for closing the session pool:

Without this change (3 test runs):
6603ms
8169ms
8367ms

With this change (3 test runs):
1297ms
1710ms
1851ms

Fixes #19

### DIFF
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolStressTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolStressTest.java
@@ -142,8 +142,6 @@ public class SessionPoolStressTest extends BaseSessionPoolTest {
   private void setupSession(final Session session) {
     ReadContext mockContext = mock(ReadContext.class);
     final ResultSet mockResult = mock(ResultSet.class);
-    //
-    // when(session.asyncClose()).thenReturn(ApiFutures.immediateFuture(Empty.getDefaultInstance()));
     when(session.singleUse(any(TimestampBound.class))).thenReturn(mockContext);
     when(mockContext.executeQuery(any(Statement.class)))
         .thenAnswer(
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolStressTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolStressTest.java
@@ -163,8 +163,9 @@ public class SessionPoolStressTest extends BaseSessionPoolTest {
               public ApiFuture<Empty> answer(InvocationOnMock invocation) throws Throwable {
                 synchronized (lock) {
                   if (expiredSessions.contains(session.getName())) {
-                    throw SpannerExceptionFactory.newSpannerException(
-                        ErrorCode.NOT_FOUND, "Session not found");
+                    return ApiFutures.immediateFailedFuture(
+                        SpannerExceptionFactory.newSpannerException(
+                            ErrorCode.NOT_FOUND, "Session not found"));
                   }
                   if (sessions.remove(session.getName()) == null) {
                     setFailed(closedSessions.get(session.getName()));
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolStressTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolStressTest.java
@@ -22,9 +22,12 @@ import static org.mockito.Mockito.doAnswer;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
+import com.google.api.core.ApiFuture;
+import com.google.api.core.ApiFutures;
 import com.google.cloud.spanner.SessionClient.SessionConsumer;
 import com.google.cloud.spanner.SessionPool.SessionConsumerImpl;
 import com.google.common.util.concurrent.Uninterruptibles;
+import com.google.protobuf.Empty;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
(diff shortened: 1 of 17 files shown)


olavloite:
The binary compatibility check fails because of:

```
[ERROR] 7012: com.google.cloud.spanner.Session: Method 'public com.google.api.core.ApiFuture asyncClose()' has been added to an interface
[ERROR] 7012: com.google.cloud.spanner.spi.v1.SpannerRpc: Method 'public com.google.api.core.ApiFuture asyncDeleteSession(java.lang.String, java.util.Map)' has been added to an interface
```

Both `Session` and `SpannerRpc` are public interfaces, but are only used internally in the client library. There are no public methods returning either a `Session` or a `SpannerRpc`.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolStressTest.java
```diff
@@ -139,6 +142,8 @@ public Void answer(InvocationOnMock invocation) throws Throwable {
   private void setupSession(final Session session) {
     ReadContext mockContext = mock(ReadContext.class);
     final ResultSet mockResult = mock(ResultSet.class);
+    //
+    // when(session.asyncClose()).thenReturn(ApiFutures.immediateFuture(Empty.getDefaultInstance()));
```
skuruppu: Why is this commented out?

olavloite: The commented code has now been removed. It was erroneously added in the first place and caused this test to fail sometimes. While searching for the reason, it was commented out and never removed.

olavloite:
@skuruppu @hengfengli @chingor13 
This PR cannot be merged because the Binary Compatibility test fails. This happens because it adds a method to two public interfaces. These interfaces are not really public (as in: returned to users), but must be public because they reside in different packages within the Spanner client library and are used by different parts of the Spanner client lib.

So:
1. Should we still treat this as a breaking change?
1. Regardless of whether we do treat is a breaking change or not: What action is needed to make the PR mergeable?

skuruppu:
> @skuruppu @hengfengli @chingor13
> This PR cannot be merged because the Binary Compatibility test fails. This happens because it adds a method to two public interfaces. These interfaces are not really public (as in: returned to users), but must be public because they reside in different packages within the Spanner client library and are used by different parts of the Spanner client lib.
> 
> So:
> 
> 1. Should we still treat this as a breaking change?

I don't believe it should be flagged as a breaking change given that we should reserve those when making backwards-incompatible changes that requires users to rebuild their apps.

> 2. Regardless of whether we do treat is a breaking change or not: What action is needed to make the PR mergeable?

I'm not sure actually :( @kolea2 would you be able to advice us on whether it's ok to ignore a Binary Compatibility failure and merge this change.

I don't know why the Java 11 test fails so I triggered a rebuild.

skuruppu:
> > @skuruppu @hengfengli @chingor13
> > This PR cannot be merged because the Binary Compatibility test fails. This happens because it adds a method to two public interfaces. These interfaces are not really public (as in: returned to users), but must be public because they reside in different packages within the Spanner client library and are used by different parts of the Spanner client lib.
> > So:
> > 
> > 1. Should we still treat this as a breaking change?
> 
> I don't believe it should be flagged as a breaking change given that we should reserve those when making backwards-incompatible changes that requires users to rebuild their apps.
> 
> > 1. Regardless of whether we do treat is a breaking change or not: What action is needed to make the PR mergeable?
> 
> I'm not sure actually :( @kolea2 would you be able to advice us on whether it's ok to ignore a Binary Compatibility failure and merge this change.
> 
> I don't know why the Java 11 test fails so I triggered a rebuild.

As discussed offline, it's ok to push this change despite the Binary Compatibility failure since the changes being made are to internal interfaces.

skuruppu:
> > > @skuruppu @hengfengli @chingor13
> > > This PR cannot be merged because the Binary Compatibility test fails. This happens because it adds a method to two public interfaces. These interfaces are not really public (as in: returned to users), but must be public because they reside in different packages within the Spanner client library and are used by different parts of the Spanner client lib.
> > > So:
> > > 
> > > 1. Should we still treat this as a breaking change?
> > 
> > 
> > I don't believe it should be flagged as a breaking change given that we should reserve those when making backwards-incompatible changes that requires users to rebuild their apps.
> > > 1. Regardless of whether we do treat is a breaking change or not: What action is needed to make the PR mergeable?
> > 
> > 
> > I'm not sure actually :( @kolea2 would you be able to advice us on whether it's ok to ignore a Binary Compatibility failure and merge this change.
> > I don't know why the Java 11 test fails so I triggered a rebuild.
> 
> As discussed offline, it's ok to push this change despite the Binary Compatibility failure since the changes being made are to internal interfaces.

Yikes, it turns out the compatibility check just keeps failing for all future PRs as well. I was hoping that  it would stop complaining after it was merged in. Let me check with @kolea2 offline about what we can do to fix this issue.

hengfengli:
Should we revert this change? We can wait until it is resolved so that we can merge it in.

skuruppu:
> Should we revert this change? We can wait until it is resolved so that we can merge it in.

Yeah good point. I'll revert it now.

PR #26, opened 2020-01-10
feat: add support for CallCredentials

### LINKED ISSUE #18: How do I apply a `io.grpc.CallCredentials` to `ServiceOptions`?
I have a private class that generates `io.grpc.CallCredentials` to authorize to Cloud Spanner. Apparently the `ServiceOptions.Builder.setCredentials` only knows about `com.google.auth.Credentials`. I know that the Cloud Spanner communication is done over gRPC. Is there a way to apply my `CallCredentials`?

olavloite:
Adds support for using CallCredentials that could vary per call instead of only static credentials passed in at startup.

Fixes #18

### DIFF
--- a/google-cloud-spanner/pom.xml
+++ b/google-cloud-spanner/pom.xml
@@ -82,6 +82,10 @@
     </dependency>
     <dependency>
       <groupId>io.grpc</groupId>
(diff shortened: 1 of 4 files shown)


Capstan:
LGTM. I believe this would work for our purposes.

PR #34, opened 2020-01-14
fix: stop sending RPCs to deleted database

### LINKED ISSUE #16: spanner: Client library sending a high rate of BeginTransaction requests when database is deleted
Once a database is deleted and if the Java Spanner client is still left open, it starts sending a large number of BeginTransaction requests. The additional requests fail with status NOT_FOUND.

The following error just keeps repeating in the logs:
```
Nov 07, 2019 10:20:16 PM com.google.cloud.spanner.spi.v1.LoggingInterceptor$CallLogger log FINER: google.spanner.v1.Spanner/BeginTransaction[74f0a03d]: Start
Nov 07, 2019 10:20:16 PM com.google.cloud.spanner.spi.v1.LoggingInterceptor$CallLogger log FINER: google.spanner.v1.Spanner/BeginTransaction[74f0a03d]: Send: session: "projects/<project>/instances/<instance>/databases/<database>/sessions/<session>" options { read_write { } }
Nov 07, 2019 10:20:16 PM com.google.cloud.spanner.spi.v1.LoggingInterceptor$CallLogger log FINER: google.spanner.v1.Spanner/BeginTransaction[74f0a03d]: Closed with status Status{code=NOT_FOUND, description=Database not found: projects/<project>/instances/<instance>/databases/<database> resource_type: "type.googleapis.com/google.spanner.admin.database.v1.Database" resource_name: "projects/<project>/instances/<instance>/databases/<database>" description: "Database does not exist." , cause=null} and trailers Metadata(...)
```

I'm guessing this is happening in an attempt to replenish the pool.

olavloite:
DatabaseClients should not continue to try to send RPCs to a database that has been deleted. Instead, the session pool will keep track of whether a DatabaseNotFound error has been returned for a database, and if so, will invalidate itself. All subsequent calls for this database will return a DatabaseNotFoundException without calling a RPC.

If a database is re-created, the user must create a new DatabaseClient with a new session pool in order to resume usage of the database.

Fixes #16

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseNotFoundException.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseNotFoundException.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Google LLC
+ * Copyright 2020 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseNotFoundException.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseNotFoundException.java
@@ -20,7 +20,8 @@ import javax.annotation.Nullable;
 
 /**
  * Exception thrown by Cloud Spanner when an operation detects that the database that is being used
- * no longer exists. This type of error has its own subclass as it is a condition that should cause the client library to stop trying to send RPCs to the backend until the user has taken action.
+ * no longer exists. This type of error has its own subclass as it is a condition that should cause
+ * the client library to stop trying to send RPCs to the backend until the user has taken action.
  */
 public class DatabaseNotFoundException extends SpannerException {
   private static final long serialVersionUID = -6395746612598975751L;
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseNotFoundException.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseNotFoundException.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.spanner;
+
+import javax.annotation.Nullable;
+
+/**
+ * Exception thrown by Cloud Spanner when an operation detects that the database that is being used
+ * no longer exists. This type of error has its own subclass as it is a condition that should cause the client library to stop trying to send RPCs to the backend until the user has taken action.
+ */
+public class DatabaseNotFoundException extends SpannerException {
+  private static final long serialVersionUID = -6395746612598975751L;
+
+  /** Private constructor. Use {@link SpannerExceptionFactory} to create instances. */
+  DatabaseNotFoundException(
+      DoNotConstructDirectly token, @Nullable String message, @Nullable Throwable cause) {
+    super(token, ErrorCode.NOT_FOUND, false, message, cause);
+  }
+}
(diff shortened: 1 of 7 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -775,6 +776,15 @@ public void close() {
       if (lastException != null && isSessionNotFound(lastException)) {
         invalidateSession(this);
       } else {
+        if (lastException != null && isDatabaseNotFound(lastException)) {
```
hengfengli: nits: use `else if`?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerExceptionFactory.java
```diff
@@ -176,6 +182,8 @@ private static SpannerException newSpannerExceptionPreformatted(
       case NOT_FOUND:
         if (message != null && message.contains("Session not found")) {
           return new SessionNotFoundException(token, message, cause);
+        } else if (message != null && DATABASE_NOT_FOUND_MSG_PATTERN.matcher(message).matches()) {
```
hengfengli: Why use a regex matching instead of `message.contains()`? I mean that it is very sensitive to any small text/format change in the message.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
```diff
@@ -51,8 +51,21 @@
 
 @RunWith(JUnit4.class)
 public class DatabaseClientImplTest {
-  private static final String DATABASE_DOES_NOT_EXIST_MSG =
-      "Database not found: projects/<project>/instances/<instance>/databases/<database> resource_type: \"type.googleapis.com/google.spanner.admin.database.v1.Database\" resource_name: \"projects/<project>/instances/<instance>/databases/<database>\" description: \"Database does not exist.\"";
+  private static final String DATABASE_NOT_FOUND_FORMAT =
```
hengfengli: nits: I feel that a hardcoded message is more clear here, because if the message in the code has been mistakenly changed, this test will let it pass instead of finding out the difference.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITDatabaseTest.java
```diff
@@ -43,4 +54,68 @@ public void badDdl() {
...
+    DatabaseNotFoundException notFoundException = null;
+    long millis = 0L;
+    while ((millis = backoff.nextBackOffMillis()) != ExponentialBackOff.STOP) {
+      Thread.sleep(millis);
+      // Queries to this database should eventually return DatabaseNotFoundExceptions.
+      try (ResultSet rs = client.singleUse().executeQuery(Statement.of("SELECT 1"))) {
+        rs.next();
+      } catch (DatabaseNotFoundException e) {
+        // This is what we expect.
+        notFoundException = e;
```
hengfengli: I guess we should break the `while` here once we get the error.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
olavloite: That wouldn't work, as the else block also contains a non-conditional block underneath this `if`-block.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerExceptionFactory.java
olavloite: The reason I added the extensive regex matching for this exception is that it could be thrown by basically any RPC, and I worry about situations where the text 'Database not found' is used as part of the input, and then also returned by the server as part of an error message. I also added an integration test to ensure that the matching does stay in sync with the backend. But maybe I'm a little bit too paranoid here...

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
olavloite: Good point. I changed it back to a hard coded string.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITDatabaseTest.java
olavloite: Yep, good catch.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseNotFoundException.java
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright 2019 Google LLC
```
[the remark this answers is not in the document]
olavloite: Done.

PR #37, opened 2020-01-16
test: [WIP] add Commit Timestamp java integration tests for cloud spanner

agasheesh:
Adds java integration tests which test the cloud spanner functionality around commit timestamps that are written atomically as part of a transaction.


google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITCommitTimestampTest.java
```diff
@@ -0,0 +1,319 @@
...
+                .set("K")
+                .to("a")
+                .set("T1")
+                .to(Value.COMMIT_TIMESTAMP)
+                .set("T2")
+                .to(Value.COMMIT_TIMESTAMP)
+                .build());
+    Struct row = readRow(client, "T", Key.of("a"), "T1", "T2");
+    assertThat(row.getTimestamp(0)).isEqualTo(commitTimestamp);
+    assertThat(row.getTimestamp(1)).isEqualTo(commitTimestamp);
```
olavloite: Could we also add an assertion that the commit timestamp returned by the `write(...)` method is equal to the commit timestamp written to these columns?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITCommitTimestampTest.java
```diff
@@ -0,0 +1,319 @@
...
+    if (ts == Value.COMMIT_TIMESTAMP) {
+      ts = commitTimestamp;
+    }
+    assertThat(readRow(client, "T1", Key.of(ts), "ts").getTimestamp(0)).isEqualTo(ts);
+    assertThat(readRow(client, "T2", Key.of(ts), "ts").getTimestamp(0)).isEqualTo(ts);
+    assertThat(readRow(client, "T3", Key.of(ts), "ts").getTimestamp(0)).isEqualTo(ts);
+  }
+
+  @Test
+  // 1) Write timestamps in the past
```
olavloite: I would prefer these comments in Javadoc style and placed before the `@Test` annotation, i.e.:
```java
/**
 * 1) Write timestamps in the past
 * 2) ...
*/
@Test
public void interleavedTable() throws Exception {
```

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITCommitTimestampTest.java
```diff
@@ -0,0 +1,319 @@
...
+  public void interleavedTable() throws Exception {
+    Database db =
+        testHelper.createTestDatabase(
+            "CREATE TABLE T1 (ts TIMESTAMP) PRIMARY KEY (ts)",
+            "CREATE TABLE T2 (ts TIMESTAMP) PRIMARY KEY (ts), INTERLEAVE IN PARENT T1",
+            "CREATE TABLE T3 (ts TIMESTAMP) PRIMARY KEY (ts), INTERLEAVE IN PARENT T2");
+    DatabaseClient client = testHelper.getDatabaseClient(db);
+    String databaseId = db.getId().getDatabase();
+
+    Timestamp timeNow = Timestamp.ofTimeMicroseconds(Instant.now().toEpochMilli() * 1000);
```
olavloite: Using `TimeUnit.MICROSECONDS.convert(Instant.now().toEpochMilli(), TimeUnit.MILLISECONDS)` instead of `millis * 1000` reduces the chances of ever making a calculation mistake when converting time values between different units.

Also: Using the actual current system time makes this test case depend on the fact that the local system time will not be lagging behind the time of the backend. I think it would be better to use a fixed time here as well, or at least add a negative value to the current local system time to reduce the probability that a small discrepancy between the two would make the test case fail.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITCommitTimestampTest.java
```diff
@@ -0,0 +1,319 @@
...
+    writeAndVerify(client, Value.COMMIT_TIMESTAMP);
+
+    alterColumnOption(databaseId, "T1", "null");
+    alterColumnOption(databaseId, "T2", "null");
+    alterColumnOption(databaseId, "T3", "null");
+    writeAndVerify(client, timeFuture);
+  }
+
+  @Test
+  // In interleaved table, use of commit timestamp in child table is not allowed
```
olavloite: Same as above.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITCommitTimestampTest.java
```diff
@@ -0,0 +1,319 @@
...
+import com.google.cloud.spanner.IntegrationTestEnv;
+import com.google.cloud.spanner.Key;
+import com.google.cloud.spanner.Mutation;
+import com.google.cloud.spanner.SpannerExceptionFactory;
+import com.google.cloud.spanner.Struct;
+import com.google.cloud.spanner.TimestampBound;
+import com.google.cloud.spanner.Value;
+import com.google.cloud.spanner.testing.RemoteSpannerHelper;
+import com.google.common.collect.ImmutableList;
+import java.time.Duration;
```
olavloite: Not supported in Java 7, use `org.threeten.bp.Duration` instead.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITCommitTimestampTest.java
```diff
@@ -0,0 +1,319 @@
...
+import java.time.Instant;
```
olavloite: Same as above.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITCommitTimestampTest.java
```diff
@@ -0,0 +1,319 @@
...
+    expectedException.expect(isSpannerException(ErrorCode.FAILED_PRECONDITION));
+    expectedException.expectMessage(
+        "corresponding shared key columns in this table's interleaved table hierarchy");
+    client.write(
+        Arrays.asList(
+            Mutation.newInsertOrUpdateBuilder("T3").set("ts").to(Value.COMMIT_TIMESTAMP).build()));
+  }
+
+  @Test
+  // In interleaved table, use of commit timestamp in parent table is not
```
olavloite: Same as above.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITCommitTimestampTest.java
agasheesh: Apologies if I missed something here, but line number 102 and 103 are doing exactly that. They are checking that commitTimestamp returned by write method on line 91 is same as that written to the Key "a" for columns "T1" and "T2".

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITCommitTimestampTest.java
agasheesh: Done @ https://github.com/googleapis/java-spanner/pull/37/commits/d920e968233662d9a16c8dd74ed94f5a2ebcb819.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITCommitTimestampTest.java
agasheesh: Done @ https://github.com/googleapis/java-spanner/pull/37/commits/d920e968233662d9a16c8dd74ed94f5a2ebcb819

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITCommitTimestampTest.java
agasheesh: Done @ https://github.com/googleapis/java-spanner/pull/37/commits/d920e968233662d9a16c8dd74ed94f5a2ebcb819.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITCommitTimestampTest.java
agasheesh: I see your point, but this test has been successfully passing in current format since always. Would prefer keeping it this way.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITCommitTimestampTest.java
agasheesh: Done @ 24faebe561fc561308f1587eba85efdcf34fc86b

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITCommitTimestampTest.java
agasheesh: Done @ 24faebe561fc561308f1587eba85efdcf34fc86b

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITCommitTimestampTest.java
olavloite: Sorry, you're completely right, I have no idea where my mind was when I added that comment :-)

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITCommitTimestampTest.java
olavloite: Ack.

PR #40, opened 2020-01-22
Added Resource Based Routing Support

mbril:
@olavloite 
verify fails on compatibility which seems reasonable.
This the same PR previously approved, but, please take a look again that I didn't miss anything.


google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigRpcCache.java
```diff
@@ -0,0 +1,121 @@
...
+import com.google.spanner.admin.instance.v1.GetInstanceRequest;
+import com.google.spanner.admin.instance.v1.Instance;
+import com.google.spanner.admin.instance.v1.InstanceName;
+import com.google.spanner.v1.DatabaseName;
+import com.google.spanner.v1.SessionName;
+import java.util.concurrent.ExecutionException;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public class InstanceConfigRpcCache {
```
olavloite: Would it be possible to move this class to the `com.googlel.cloud.spanner.spi.v1` package and make it package-private? It is not a class that any users should have to deal with, so keeping it package-private whenever possible would be preferable. This also reduces the chances that we run into unnecessary binary compatibility check failures in the future.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigRpcCache.java
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  public void invalidateAll() {
+    for (SpannerRpc rpc : cache.asMap().values()) {
+      if (rpc == this.projectClient) continue;
+      rpc.shutdown();
+    }
+    cache.invalidateAll();
+  }
+}
```
olavloite: nit: missing line ending

Remember to run the code formatter like this:
`mvn com.coveo:fmt-maven-plugin:format`

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceInfo.java
```diff
@@ -76,6 +80,10 @@ static FieldMask toFieldMask(InstanceField... fields) {
 
     public abstract Builder putAllLabels(Map<String, String> labels);
 
+    public abstract Builder addEndpointUri(String endpointUrl);
```
olavloite: Same as above: Would it be possible to make these package-private? Or do they need to be public because it is used by client library code residing in other packages?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Instance.java
```diff
@@ -83,6 +85,18 @@ public Builder putAllLabels(Map<String, String> labels) {
       return this;
     }
 
+    @Override
+    public InstanceInfo.Builder addEndpointUri(String endpointUrl) {
```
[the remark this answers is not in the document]
mbril: I do agree..
But, currently we don't have any package private members in this object.
So, I followed the current style.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigRpcCache.java
mbril: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceInfo.java
mbril: See comment above.. this is to keep same style

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Instance.java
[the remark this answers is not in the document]
olavloite: Sounds reasonable. I just checked, and the method `setState(..)` is also public, which is also clearly a read-only field.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/BatchCreateSessionsTest.java
```diff
@@ -112,13 +117,14 @@ private Spanner createSpanner(int minSessions, int maxSessions) {
...
-        .build()
-        .getService();
+    SpannerOptions options =
+        SpannerOptions.newBuilder()
+            .setProjectId("[PROJECT]")
+            .setChannelProvider(channelProvider)
+            .setSessionPoolOption(sessionPoolOptions)
+            .setCredentials(NoCredentials.getInstance())
+            .build();
+    return options.getService();
```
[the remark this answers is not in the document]
mbril: I think it's leftover from another change.
reverting

skuruppu:
We have decided to implement this functionality on the server side so we no longer need to add this support on the client side.

PR #57, opened 2020-02-04
fix: use resource type to identify type of error

olavloite:
Use the returned resource type instead of the error message to determine the type of error.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerExceptionFactory.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerExceptionFactory.java
@@ -21,12 +21,14 @@ import com.google.api.gax.rpc.ApiException;
 import com.google.cloud.spanner.SpannerException.DoNotConstructDirectly;
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Predicate;
+import com.google.rpc.ResourceInfo;
 import io.grpc.Context;
+import io.grpc.Metadata;
 import io.grpc.Status;
 import io.grpc.StatusRuntimeException;
+import io.grpc.protobuf.ProtoUtils;
 import java.util.concurrent.CancellationException;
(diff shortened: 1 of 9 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerExceptionFactory.java
```diff
@@ -183,10 +193,13 @@ private static SpannerException newSpannerExceptionPreformatted(
       case ABORTED:
         return new AbortedException(token, message, cause);
       case NOT_FOUND:
-        if (message != null && message.contains("Session not found")) {
-          return new SessionNotFoundException(token, message, cause);
-        } else if (message != null && DATABASE_NOT_FOUND_MSG_PATTERN.matcher(message).matches()) {
-          return new DatabaseNotFoundException(token, message, cause);
+        ResourceInfo resourceInfo = extractResourceInfo(cause);
+        if (resourceInfo != null) {
```
snehashah16: what about Instance Not Found ?

olavloite: The client library does not currently handle `InstanceNotFound` specifically, so it is currently not recognized as a separate error. I'll open a separate issue for it, as it could cause the same problems as `DatabaseNotFound` if someone deletes an instance while the client library has an active connection with a database on that instance.

snehashah16:
May be worth considering, what if the resource info is not populated ?
What's the fallback in that case ?

olavloite:
> May be worth considering, what if the resource info is not populated ? What's the fallback in that case ?

If the resource info is not populated, or it is populated with unknown values, the client library will automatically fall back to considering the error as a generic `SpannerException`. The effect of this would be that the client library will not stop sending RPCs for the database client that caused the error. The user will not really notice any difference, as the error is also returned to the user in both cases.

Issue #53, opened 2020-01-31
Instrument Spanner client (session pool) with OpenCensus metrics


mayurkale22:
**Is your feature request related to a problem? Please describe.**
Customers run into the Cloud Spanner session pool tuning issues for their specific workload on a regular basis. Improved visibility into the state of the session pool (or client side metrics) maintained by each client can provide actionable insight and help users take informed decisions.

This will also help customers in identifying a session leaks.

**Describe the solution you'd like**
Instrument Spanner client (session pool) with [OpenCensus](https://opencensus.io/stats/) metrics. we need to capture things including - number of max sessions, number of sessions in use, leaked sessions, count of active spanner requests, count of released and acquired sessions.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

===================================================================
Edit1: Agreed list of useful metrics:

All the metrics will be prefixed by “cloud.google.com/java/spanner/”.
| Metric | Description | Type | Labels |
| ------------- | ------------- | ------------- | ------------- |
| max_in_use_sessions | The maximum number of sessions in use during the last 10 minute interval | Gauge  | database, instance_id, library_version |
| max_allowed_sessions  | The maximum number of sessions allowed. Configurable by the user | Gauge  | database, instance_id, library_version |
| ~in_use_sessions~  | ~The number of sessions currently in use~  | ~Gauge~  | ~database, instance_id, library_version~ |
| get_session_timeouts  | The number of get sessions timeouts due to pool exhaustion  | Counter  | database, instance_id, library_version |
| num_acquired_sessions  | The number of sessions acquired from the session pool | Counter  | database, instance_id, library_version |
| num_released_sessions  | The number of sessions released by the user and pool maintainer  | Counter  | database, instance_id, library_version |
| num_sessions_in_pool | The number of sessions in the pool.  | Gauge  | database, instance_id, library_version, Type |

* Type : num_in_use_sessions | num_sessions_being_prepared | num_read_sessions | num_write_prepared_sessions

/cc @skuruppu

olavloite:
I think that we should consider adding metrics for monitoring whether the user has chosen a good value for `SessionPoolOptions.writeSessionsFraction`. This could include:
* The number of write sessions that are present in the pool by default (i.e. `minSessions * writeSessionsFraction`).
* The current number of write sessions in use.
[...]

PR #65, opened 2020-02-07
feat: add session timeout metric

mayurkale22:
Updates #53 and continuation of #54

New Metric:
`cloud.google.com/java/spanner/get_sessions_timeout` => The number of get session timeouts due to pool exhaustion. This value is continuously increasing, so decided to use `Cumulative` (or `Counter`) type to record it.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/MetricRegistryConstants.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/MetricRegistryConstants.java
@@ -45,7 +45,7 @@ class MetricRegistryConstants {
   static final String MAX_IN_USE_SESSIONS = "cloud.google.com/java/spanner/max_in_use_session";
   static final String MAX_ALLOWED_SESSIONS = "cloud.google.com/java/spanner/max_allowed_sessions";
   static final String IN_USE_SESSIONS = "cloud.google.com/java/spanner/in_use_sessions";
-  static final String SESSIONS_TIMEOUT = "cloud.google.com/java/spanner/sessions_timeout";
+  static final String GET_SESSION_TIMEOUT = "cloud.google.com/java/spanner/get_sessions_timeout";
 
   static final String MAX_IN_USE_SESSIONS_DESCRIPTION =
       "The maximum number of sessions in use during the last 10 minute interval.";
@@ -53,5 +53,5 @@ class MetricRegistryConstants {
       "The maximum number of sessions allowed. Configurable by the user.";
   static final String IN_USE_SESSIONS_DESCRIPTION = "The number of sessions currently in use.";
   static final String SESSIONS_TIMEOUT_DESCRIPTION =
-      "The count of number of sessions timeout due to pool exhaustion";
+      "The number of get session timeouts due to pool exhaustion";
 }
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/MetricRegistryConstants.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/MetricRegistryConstants.java
@@ -42,7 +42,7 @@ class MetricRegistryConstants {
   static final String COUNT = "1";
 
   // The Metric name and description
-  static final String MAX_IN_USE_SESSIONS = "cloud.google.com/java/spanner/max_in_use_session";
+  static final String MAX_IN_USE_SESSIONS = "cloud.google.com/java/spanner/max_in_use_sessions";
   static final String MAX_ALLOWED_SESSIONS = "cloud.google.com/java/spanner/max_allowed_sessions";
   static final String IN_USE_SESSIONS = "cloud.google.com/java/spanner/in_use_sessions";
   static final String GET_SESSION_TIMEOUTS = "cloud.google.com/java/spanner/get_sessions_timeouts";
@@ -53,5 +53,5 @@ class MetricRegistryConstants {
       "The maximum number of sessions allowed. Configurable by the user.";
   static final String IN_USE_SESSIONS_DESCRIPTION = "The number of sessions currently in use.";
   static final String SESSIONS_TIMEOUTS_DESCRIPTION =
-      "The number of get session timeouts due to pool exhaustion";
+      "The number of get sessions timeouts due to pool exhaustion";
 }
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/MetricRegistryConstants.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/MetricRegistryConstants.java
@@ -45,7 +45,7 @@ class MetricRegistryConstants {
   static final String MAX_IN_USE_SESSIONS = "cloud.google.com/java/spanner/max_in_use_sessions";
   static final String MAX_ALLOWED_SESSIONS = "cloud.google.com/java/spanner/max_allowed_sessions";
   static final String IN_USE_SESSIONS = "cloud.google.com/java/spanner/in_use_sessions";
-  static final String GET_SESSIONS_TIMEOUTS = "cloud.google.com/java/spanner/get_sessions_timeouts";
+  static final String GET_SESSION_TIMEOUTS = "cloud.google.com/java/spanner/get_session_timeouts";
 
   static final String MAX_IN_USE_SESSIONS_DESCRIPTION =
       "The maximum number of sessions in use during the last 10 minute interval.";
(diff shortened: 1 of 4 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/MetricRegistryConstants.java
```diff
@@ -45,9 +45,13 @@
   static final String MAX_IN_USE_SESSIONS = "cloud.google.com/java/spanner/max_in_use_session";
   static final String MAX_ALLOWED_SESSIONS = "cloud.google.com/java/spanner/max_allowed_sessions";
   static final String IN_USE_SESSIONS = "cloud.google.com/java/spanner/in_use_sessions";
+  static final String SESSIONS_TIMEOUT = "cloud.google.com/java/spanner/sessions_timeout";
```
olavloite: Suggestion: Call this `GET_SESSION_TIMEOUT`, as it is not a timeout on (a) session(s), but a timeout to get a session.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/MetricRegistryConstants.java
```diff
@@ -45,9 +45,13 @@
...
+
   static final String MAX_IN_USE_SESSIONS_DESCRIPTION =
       "The maximum number of sessions in use during the last 10 minute interval.";
   static final String MAX_ALLOWED_SESSIONS_DESCRIPTION =
       "The maximum number of sessions allowed. Configurable by the user.";
   static final String IN_USE_SESSIONS_DESCRIPTION = "The number of sessions currently in use.";
+  static final String SESSIONS_TIMEOUT_DESCRIPTION =
+      "The count of number of sessions timeout due to pool exhaustion";
```
olavloite: Suggestion: "The number of get session timeouts due to pool exhaustion"

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolTest.java
```diff
@@ -1583,16 +1585,45 @@ public void testSessionMetrics() {
...
+    // Return the checked out session to the pool so the async request will get a session and
+    // finish.
     session2.close();
-    session1.close();
+    // Verify that the async request also succeeds.
+    fut.get(10L, TimeUnit.SECONDS);
+    executor.shutdown();
 
+    session1.close();
+    assertThat(record.getMetrics()).containsEntry(MetricRegistryConstants.SESSIONS_TIMEOUT, 1L);
```
olavloite: I'm a little bit worried that this assert could turn out to be flaky. If I understand the test code correctly, the get session timeout will happen for the first time after 20ms. You poll the number of timeouts every 5ms and break the loop if you have at least 1 timeout. If the following lines of code happen to take more than 15ms to execute, you might have more than 1 timeouts in your metrics. Would it be possible to make this assert check that the number of timeouts is at least 1 (instead of equal to 1)?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/MetricRegistryConstants.java
mayurkale22: Sure, done in 75af67669efbf1eb01cd16118f80c1d170063807

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolTest.java
mayurkale22: I was thinking about it, but tried multiple times locally didn't see the issue. Having said that, I understand your concern, made the change(d1c13d84d35a2e10f4620941649d022d83a8a376) to the test. PTAL when you get a chance. Thanks :)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/MetricRegistryConstants.java
```diff
@@ -45,9 +45,13 @@
   static final String MAX_IN_USE_SESSIONS = "cloud.google.com/java/spanner/max_in_use_session";
   static final String MAX_ALLOWED_SESSIONS = "cloud.google.com/java/spanner/max_allowed_sessions";
   static final String IN_USE_SESSIONS = "cloud.google.com/java/spanner/in_use_sessions";
+  static final String GET_SESSION_TIMEOUT = "cloud.google.com/java/spanner/get_sessions_timeout";
```
rghetia: not sure about the name `get_sessions_timeout`. Maybe `session_open_retries`? This would go well with other metrics `session_opened` and `sessions_closed` if those are implemented.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/MetricRegistryConstants.java
```diff
@@ -45,9 +45,13 @@
...
+
   static final String MAX_IN_USE_SESSIONS_DESCRIPTION =
       "The maximum number of sessions in use during the last 10 minute interval.";
   static final String MAX_ALLOWED_SESSIONS_DESCRIPTION =
       "The maximum number of sessions allowed. Configurable by the user.";
   static final String IN_USE_SESSIONS_DESCRIPTION = "The number of sessions currently in use.";
+  static final String SESSIONS_TIMEOUT_DESCRIPTION =
+      "The number of get session timeouts due to pool exhaustion";
```
rghetia: Based on previous comment this could be changed to
`The number of retries to open a session. Retries occur when session pool is exhausted.`

google-cloud-spanner/src/main/java/com/google/cloud/spanner/MetricRegistryConstants.java
mayurkale22: @skuruppu @olavloite any thoughts?

olavloite: No, this retry is not for opening a session, but for getting (or acquiring) a session from the pool. The maximum number of sessions in the pool has already been reached, and the pool is therefore prohibited from opening any new sessions. Instead, the thread that requested the session is placed in a wait queue where it waits until a session is released into the pool by another thread. The waiter will wake up and increase this counter every now and then in order to be able to log information about the fact that it is still waiting.

rghetia: get_session_timeout still doesn't feel right but I don't have any good suggestion. I think it should be  GET_SESSION_TIMEOUTS though ('s at the end).

mayurkale22:
@skuruppu I would appreciate your reviews and approval here please.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/MetricRegistryConstants.java
```diff
@@ -42,12 +42,16 @@
   static final String COUNT = "1";
 
   // The Metric name and description
-  static final String MAX_IN_USE_SESSIONS = "cloud.google.com/java/spanner/max_in_use_session";
+  static final String MAX_IN_USE_SESSIONS = "cloud.google.com/java/spanner/max_in_use_sessions";
   static final String MAX_ALLOWED_SESSIONS = "cloud.google.com/java/spanner/max_allowed_sessions";
   static final String IN_USE_SESSIONS = "cloud.google.com/java/spanner/in_use_sessions";
+  static final String GET_SESSION_TIMEOUTS = "cloud.google.com/java/spanner/get_sessions_timeouts";
```
skuruppu: There's a discrepancy in the naming here.

```suggestion
  static final String GET_SESSIONS_TIMEOUTS = "cloud.google.com/java/spanner/get_sessions_timeouts";
```

rghetia: I would change get_sessions_timeouts to get_session_timeouts instead.

mayurkale22: I don't have strong preference, but **sessions** seems consistent across other metrics. @skuruppu thoughts?

skuruppu: I think it should be `get_session_timeouts`.

mayurkale22: done, ptal

PR #67, opened 2020-02-10
feat: add metrics to capture acquired and released sessions data

mayurkale22:
Updates #53 and continuation of #54 and #65.

New Metrics:
`cloud.google.com/java/spanner/num_acquired_sessions` => The number of sessions acquired from the session pool.

`cloud.google.com/java/spanner/num_released_sessions` => The number of sessions released (or destroyed) by the user and pool maintainer.

~Another option, instead of having two metrics what if we just include one metric `cloud.google.com/java/spanner/diff_acquired_released_sessions` to indicate diff/gaps between acquired and released sessions. I think stand alone they both (**num_acquired_sessions** and **num_released_sessions**) are adding very less value unless we combine them to see the session leaks or other anomalies.~

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
@@ -800,6 +800,7 @@ final class SessionPool {
     public void close() {
       synchronized (lock) {
         numSessionsInUse--;
+        numSessionsReleased++;
       }
       leakedException = null;
       if (lastException != null && isSessionNotFound(lastException)) {
@@ -1462,6 +1463,7 @@ final class SessionPool {
     if (!options.isFailIfSessionNotFound() && session.allowReplacing) {
       synchronized (lock) {
         numSessionsInUse--;
+        numSessionsReleased++;
       }
       session.leakedException = null;
       invalidateSession(session);
@@ -1482,6 +1484,7 @@ final class SessionPool {
       if (maxSessionsInUse < ++numSessionsInUse) {
         maxSessionsInUse = numSessionsInUse;
       }
+      numSessionsAcquired++;
     }
   }
 
@@ -1523,7 +1526,6 @@ final class SessionPool {
       if (closureFuture != null) {
         return;
       }
-      numSessionsReleased++;
       if (readWaiters.size() == 0 && numSessionsBeingPrepared >= readWriteWaiters.size()) {
         // No pending waiters
         if (shouldPrepareSession()) {
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
@@ -1360,11 +1360,9 @@ final class SessionPool {
           readWaiters.add(waiter);
         } else {
           span.addAnnotation("Acquired read write session");
-          numSessionsAcquired++;
         }
       } else {
         span.addAnnotation("Acquired read only session");
-        numSessionsAcquired++;
       }
     }
     if (waiter != null) {
@@ -1422,7 +1420,6 @@ final class SessionPool {
         if (numSessionsBeingPrepared <= readWriteWaiters.size()) {
           PooledSession readSession = readSessions.poll();
           if (readSession != null) {
-            numSessionsAcquired++;
             span.addAnnotation("Acquired read only session. Preparing for read write transaction");
             prepareSession(readSession);
           } else {
@@ -1433,7 +1430,6 @@ final class SessionPool {
         waiter = new Waiter();
         readWriteWaiters.add(waiter);
       } else {
-        numSessionsAcquired++;
         span.addAnnotation("Acquired read write session");
       }
     }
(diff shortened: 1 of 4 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1346,9 +1359,11 @@ PooledSession getReadSession() throws SpannerException {
           readWaiters.add(waiter);
         } else {
           span.addAnnotation("Acquired read write session");
+          numSessionsAcquired++;
```
mayurkale22: @olavloite I added these counters based on my understanding, let me know if there is a better way to get these (acquired and released sessions) values.

skuruppu:
> Another option, instead of having two metrics what if we just include one metric `cloud.google.com/java/spanner/diff_acquired_released_sessions` to indicate diff/gaps between acquired and released sessions.

I agree that the difference is the most useful. So could we call it something like `cloud.google.com/java/spanner/unreleased_sessions`?

I would also be interested in distinguishing between unreleased sessions held by the pool maintainer vs the user. Just because I want to distinguish between a leak in the client library and a leak by the user.

How would a user distinguish between this metric and `in_use_sessions` though? Logically, I would think of unreleased sessions as being the sessions that I'm using right now.

mayurkale22:
>I would also be interested in distinguishing between unreleased sessions held by the pool maintainer vs the user. Just because I want to distinguish between a leak in the client library and a leak by the user.

SGTM. What is the best way to understand the owner of the session (maintainer/user). I imagine it would be little complicated than keeping variable to track the acquired/released count.

> How would a user distinguish between this metric and in_use_sessions though? Logically, I would think of unreleased sessions as being the sessions that I'm using right now.

I think you're right, users might get confused with these two metrics. Seems like unreleased_sessions are the sum of in_use_sessions and idle/leak sessions and it will be almost equal to (or greater than) in_use_sessions all the time. If the difference between the two is steadily increasing, it is certainly an indication that there is a session leak. 
I think `unreleased_sessions - in_use_sessions == count of session leaks`. Please correct me, this is based on my limited knowledge on session pool.

skuruppu:
> SGTM. What is the best way to understand the owner of the session (maintainer/user). I imagine it would be little complicated than keeping variable to track the acquired/released count.

Yeah it would be more complicated. You would have to go through each of the methods that check out a session an increment a maintainer_checked_out counter or a user_checked_out counter based on whether the pool or the user checks out the session. I'm not too familiar with the code to know how difficult this would be.

> I think unreleased_sessions - in_use_sessions == count of session leaks. Please correct me, this is based on my limited knowledge on session pool.

Yup that's my understanding as well.

mayurkale22:
@olavloite I would like to get your reviews on this.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -795,6 +799,7 @@ public TransactionRunner readWriteTransaction() {
     @Override
     public void close() {
       synchronized (lock) {
+        numSessionsReleased++;
```
skuruppu: I don't think releasing sessions is the same as closing the session pool. Releasing is to do with the user handing the session back to the pool. Closing is shutting down the connection all together.

I think you want to add this to `releaseSession` on line 1522.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1025,6 +1030,7 @@ private void closeIdleSessions(Instant currTime) {
           }
           numSessionsToClose -= sessionsToClose.size();
         }
+        numSessionsReleased += sessionsToClose.size();
```
skuruppu: I think you can remove this as well.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolTest.java
```diff
@@ -1628,7 +1628,7 @@ public Void call() {
     assertThat(record.getMetrics())
         .containsEntry(MetricRegistryConstants.NUM_ACQUIRED_SESSIONS, 0L);
     assertThat(record.getMetrics())
-        .containsEntry(MetricRegistryConstants.NUM_RELEASED_SESSIONS, 3L);
+        .containsEntry(MetricRegistryConstants.NUM_RELEASED_SESSIONS, 5L);
```
mayurkale22: @skuruppu thanks for the suggestions. I made the changes, but somehow test results look fishy to me. After change, NUM_RELEASED_SESSIONS == 5 and NUM_ACQUIRED_SESSIONS == 0.

skuruppu: I think I was partially wrong. It turns out that `releaseSession()` is only used internally. So you can get calls to this function due to the session pool doing keep alive requests (e.g. check out session to send a "select 1" then release it back to the pool). I think this is why you're seeing the number of released sessions being higher than you expected.

I also just realized that `getReadSession()` docs say to call `Session.close()` when you're done with the sessions. It seems that sessions acquired by the user will get closed and sessions acquired by the client library gets released.

@olavloite did I understand this correctly?

skuruppu: I don't understand why `NUM_ACQUIRED_SESSIONS` is zero though. I would've thought calling `getReadSession()` would increment this.

mayurkale22: @olavloite would like to get your eyes on this.

mayurkale22:
@skuruppu I updated the code (341feee4e99379fb3231effea714e0f354564c7d) based on https://github.com/googleapis/java-spanner/pull/67#discussion_r382041439. The released and acquired count looks good to me now, PTAL. Thanks

mayurkale22:
> Consider removing the paragraph in the commit message for this PR regarding the diff metric since we're not implementing that here.

Done.

mayurkale22:
@olavloite Please review and approve when you get a chance. Thanks

mayurkale22:
These metrics have all been tested locally against a real spanner instance and this is what I see:
Screen Shot 2020-02-26 at 7 06 25 PM

mayurkale22:
> I assume that the screenshot of the test run was made before the last minor fix was applied.

Yes, I have changed the screenshot based on current implementations. Thanks for noticing it.

Issue #71, opened 2020-02-12
spanner: "OVERKILLED java.lang.IllegalStateException: refCount dropped below zero" warning


skuruppu:
I have seen a few reports of warning messages polluting error logs.

#### Environment details

1. Specify the API at the beginning of the title. For example, "BigQuery: ...").
Spanner
2. OS type and version: Linux but version unknown
3. Java version: Unknown
4. spanner version(s): 1.43.0

#### Steps to reproduce

I haven't reproduced this yet but I suspect this happens when an error occurs during [`ResumableStreamIterator.computeNext().`](https://github.com/googleapis/java-spanner/blob/1178958eaec5aa6ea80938ad91dfb0b1a688463d/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java#L929) The trace span is ended with [`TraceUtil.endSpanWithFailure(span, e)`](https://github.com/googleapis/java-spanner/blob/1178958eaec5aa6ea80938ad91dfb0b1a688463d/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java#L1000) but then when the stream iterator is closed, the code attempts to [end the span](https://github.com/googleapis/java-spanner/blob/1178958eaec5aa6ea80938ad91dfb0b1a688463d/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java#L924) again.

#### Stack trace
```
WARNING: <span:26a180bf406e2db9, parent:4479af252f884e07, trace:842cc1d5df736326, mask:ffffffff98000080, name:CloudSpannerOperation.ExecuteStreamingQuery>: OVERKILLED
java.lang.IllegalStateException: refCount dropped below zero
	at com.google.tracing.TraceContext.unRef(TraceContext.java:799)
	at com.google.perftools.opencensus.trace.SpanImpl.end(SpanImpl.java:237)
	at io.opencensus.trace.Span.end(Span.java:244)
	at com.google.cloud.spanner.AbstractResultSet$ResumableStreamIterator.close(AbstractResultSet.java:850)
	at com.google.cloud.spanner.AbstractResultSet$GrpcValueIterator.close(AbstractResultSet.java:285)
	at com.google.cloud.spanner.AbstractResultSet$GrpcResultSet.close(AbstractResultSet.java:133)
	at com.google.cloud.spanner.AbstractResultSet$GrpcResultSet.yieldError(AbstractResultSet.java:144)
[...]

PR #74, opened 2020-02-14
WIP: Attempt at reproducing "refCount dropped below zero"

skuruppu:
Added `closedSpanOnNonRetryableError` test. The interesting thing is
that after commenting out all `TraceUtil.endSpanWithFailure()` calls,
something is still closing the span without calling `close()` because
the test passes. Not sure where this could be happening. Might explain
why the refcount drops below zero.

Towards #71.


skuruppu:
Fixe was made in #75.

PR #79, opened 2020-02-18
feat: Added SimpleClient Implementation

mbril:
Simple Client is there to provide ab abstraction layer on top of Cloud Spanner Client.
This is a simplified approach in order to make customers avoid common mistakes which result from specific Cloud Spanner idioms.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SimpleClientImpl.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SimpleClientImpl.java
@@ -1,0 +1,116 @@
+package com.google.cloud.spanner;
+
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public class SimpleClientImpl implements SimpleClient {
+
+  private final DatabaseClient client;
+  private static final Logger logger = Logger.getLogger(SimpleClientImpl.class.getName());
+
+  SimpleClientImpl(DatabaseClient client) {
+    this.client = client;
+  }
+
+  @Override
+  public ResultSet executeSqlQuery(String query) {
+    return executeSqlQuery(query, TimestampBound.strong());
+  }
+
+  @Override
+  public ResultSet executeSnapshotSqlQuery(String query) {
+    return executeSqlQuery(query, TimestampBound.ofExactStaleness(5L, TimeUnit.SECONDS));
+  }
+
+  @Override
+  public ResultSet executeSqlQuery(String query, TimestampBound timestampBound) {
+    return executeSqlQuery(query, null, timestampBound);
+  }
+
+  @Override
+  public ResultSet executeSqlQuery(
+      String query, Map<String, Value> args, TimestampBound timestampBound) {
+    try (ReadOnlyTransaction tx = client.singleUseReadOnlyTransaction(timestampBound)) {
+      Statement statement;
+      if (args != null && !args.isEmpty()) {
+        Statement.Builder builder = Statement.newBuilder(query);
+        for (Map.Entry<String, Value> entry : args.entrySet()) {
+          builder.bind(entry.getKey()).to(entry.getValue());
+        }
+        statement = builder.build();
+      } else {
+        statement = Statement.of(query);
+      }
+      return tx.executeQuery(statement);
+    }
+  }
+
+  @Override
+  public ResultSet executeSqlQuery(String query, Map<String, Value> args) {
+    return executeSqlQuery(query, args, TimestampBound.strong());
+  }
+
+  @Override
+  public ResultSet executeSnapshotSqlQuery(String query, Map<String, Value> args) {
+    return executeSqlQuery(query, args, TimestampBound.ofExactStaleness(5L, TimeUnit.SECONDS));
+  }
+
+  @Override
+  public long executeSqlWrite(String sql, Map<String, Value> args) {
+    Statement.Builder builder = Statement.newBuilder(sql);
+    if (args != null && !args.isEmpty()) {
+      for (Map.Entry<String, Value> entry : args.entrySet()) {
+        builder.bind(entry.getKey()).to(entry.getValue());
+      }
+    }
+    return client.executePartitionedUpdate(builder.build());
+  }
+
+  @Override
+  public boolean runTransaction(final List<Statement> statements) {
+    try {
+      boolean status =
+          client
+              .readWriteTransaction()
+              .run(
+                  new TransactionRunner.TransactionCallable<Boolean>() {
+                    @Override
+                    public Boolean run(TransactionContext transaction) {
+                      try {
+                        transaction.batchUpdate(statements);
+                        return true;
+                      } catch (Exception e) {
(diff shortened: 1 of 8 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SimpleClientImpl.java
```diff
@@ -0,0 +1,131 @@
...
+
+  @Override
+  public long executeSqlWrite(String sql, Map<String, Value> args) {
+    Statement.Builder builder = Statement.newBuilder(sql);
+    if (args != null && !args.isEmpty()) {
+      for (Map.Entry<String, Value> entry : args.entrySet()) {
+        builder.bind(entry.getKey()).to(entry.getValue());
+      }
+    }
+    return client.executePartitionedUpdate(builder.build());
```
olavloite: I don't think you should use Partitioned DML here. Partitioned DML statements are not applied atomically to the entire table (only to each partition), must be partionable, and must be idempotent. Instead, you should just use a normal read/write transaction.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SimpleClientImpl.java
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  @Override
+  public ResultSet executeSqlQuery(String query) {
+    return executeSqlQuery(query, TimestampBound.strong());
+  }
+
+  @Override
+  public ResultSet executeSnapshotSqlQuery(String query) {
+    return executeSqlQuery(query, TimestampBound.ofExactStaleness(5L, TimeUnit.SECONDS));
```
olavloite: I don't really see the use case for this method, The staleness is fixed at exactly 5 seconds, which is not a value that has any special meaning. Reading with an exact staleness means that you are reading the data as they were at one exact moment. AFAIK, this request cannot be served more efficiently than reading with `Timestamp.strong()`, so it also doesn't bring any performance gain over the `executeSqlQuery(String)` method.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SimpleClientImpl.java
```diff
@@ -0,0 +1,131 @@
...
+        builder.bind(entry.getKey()).to(entry.getValue());
+      }
+    }
+    return client.executePartitionedUpdate(builder.build());
+  }
+
+  @Override
+  public boolean runTransaction(final List<Statement> statements) {
+    try {
+      boolean status =
```
olavloite: This doesn't seem to be used, the method always seems to be returning `true`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SimpleClientImpl.java
```diff
@@ -0,0 +1,131 @@
...
+    } catch (Exception e) {
+      throw SpannerExceptionFactory.newSpannerException(e);
+    }
+    return true;
+  }
+
+  @Override
+  public void runTransaction(final Statement readStatement, final OnReadHandler handler) {
+    client
+        .readWriteTransaction()
```
olavloite: You should use a single-use read-only transaction for this as well. Using a read/write transaction instead of a (single use) read-only transaction does not bring any advantages over using a read/write transaction, while it does involve more RPCs and (potentially) more locking.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SimpleClientImpl.java
mbril: Fair point.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SimpleClientImpl.java
mbril: Reading from a snapshot will be more performant. 
I can make the staleness to be 30 sec which should be a more substantial perf improvment.
Let me do that.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SimpleClientImpl.java
mbril: The use case here is read then write
so read write transaction is required.
see IT for example.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SimpleClientImpl.java
mbril: will address

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SimpleClientImpl.java
```diff
@@ -0,0 +1,139 @@
...
+        .run(
+            new TransactionRunner.TransactionCallable<Boolean>() {
+              @Override
+              public Boolean run(TransactionContext transaction) throws Exception {
+                try {
+                  ResultSet resultSet = transaction.executeQuery(readStatement);
+                  handler.handle(resultSet, transaction);
+                  return true;
+                } catch (Exception e) {
+                  logger.log(Level.WARNING, "Failed to complete transaction", e);
```
olavloite: The Spanner client library will commit the transaction if the `TransactionCallable#run(TransactionContext)` method returns normally. As all exceptions are caught by this catch clause, the transaction will always be committed, regardless of the type of exception that might occur during the transaction.

It will also prevent automatic retries of any aborted transaction, as that is based on the `AbortedException` bubbling up from the `run` method.

Catching any exception here will cause the transaction to commit instead of rollback, as the `run` method returns normally. The client library will

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SimpleClientImpl.java
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  @Override
+  public ResultSet executeSqlQuery(String query, Map<String, Value> args) {
+    return executeSqlQuery(query, args, TimestampBound.strong());
+  }
+
+  @Override
+  public ResultSet executeSnapshotSqlQuery(String query, Map<String, Value> args) {
+    return executeSqlQuery(query, args, TimestampBound.ofExactStaleness(30L, TimeUnit.SECONDS));
```
olavloite: Good point about this actually being more performant than using a strong read. May I suggest that we set the staleness value to 15 seconds to keep in sync with the documentation: https://cloud.google.com/spanner/docs/reads

That would prevent any user that might be inspecting the code from asking questions why this client is using a different value than what the documentation is recommending.

mbril: Fair point.
Will change to 15sec

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SimpleClientImpl.java
mbril: change to void
Now not catching exceptions.

skuruppu:
We're not prioritizing this work right now so closing this PR.

PR #90, opened 2020-03-03
feat: add backend query options

olavloite:
Adds the ability to set `QueryOptions` when running Cloud Spanner queries. For now, only setting the `query_optimizer_version` is added.

QueryOptions can be configured through the following mechanisms:

1. Through the `SPANNER_OPTIMIZER_VERSION` environment variable.
1. At `Spanner` level using `SpannerOptions.newBuilder().setDefaultQueryOptions(DatabaseId, QueryOptions)`.
1. At query level using `Statement.newBuilder(String).withQueryOptions(QueryOptions)`.

If the options are configured through multiple mechanisms then:

1. Options set at an environment variable level will override options configured at the `SpannerOptions` level.
1. Options set at a query-level will override options set at either the `SpannerOptions` or environment variable level.

If no options are set, the optimizer version will default to:

1. The optimizer version the database is pinned to.
1. If the database is not pinned to a specific version, then the Cloud Spanner backend will use the "latest" version.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
@@ -450,31 +450,23 @@ abstract class AbstractReadContext
    *   <li>The default {@link SpannerOptions#getDefaultQueryOptions()} specified for the database
    *       where the query is executed.
    * </ol>
-   *
-   * It is possible that the source of each query option in the total set of query options used for
-   * the query is different, for example that the value for {@link
-   * QueryOptions#getOptimizerVersion()} was specified in the list of query options for this query,
-   * while the value for {@link QueryOptions#getOptimizerStatisticsPackage()} was read from an
-   * environment variable because no value for that option was specified for the query.
    */
   @VisibleForTesting
-  QueryOptions buildQueryOptions(Options requestOptions) {
+  QueryOptions buildQueryOptions(QueryOptions requestOptions) {
     // Shortcut for the most common return value.
-    if (defaultQueryOptions.equals(QueryOptions.getDefaultInstance())
-        && !requestOptions.hasBackendQueryOptions()) {
+    if (defaultQueryOptions.equals(QueryOptions.getDefaultInstance()) && requestOptions == null) {
       return QueryOptions.getDefaultInstance();
     }
     // Create a builder based on the default query options.
     QueryOptions.Builder builder = defaultQueryOptions.toBuilder();
     // Then overwrite with specific options for this query.
-    if (requestOptions.hasBackendQueryOptions()) {
-      builder.mergeFrom(requestOptions.backendQueryOptions());
+    if (requestOptions != null) {
+      builder.mergeFrom(requestOptions);
     }
     return builder.build();
   }
 
-  ExecuteSqlRequest.Builder getExecuteSqlRequestBuilder(
-      Statement statement, QueryMode queryMode, Options options) {
+  ExecuteSqlRequest.Builder getExecuteSqlRequestBuilder(Statement statement, QueryMode queryMode) {
     ExecuteSqlRequest.Builder builder =
         ExecuteSqlRequest.newBuilder()
             .setSql(statement.getSql())
@@ -493,7 +485,7 @@ abstract class AbstractReadContext
       builder.setTransaction(selector);
     }
     builder.setSeqno(getSeqNo());
-    builder.setQueryOptions(buildQueryOptions(options));
+    builder.setQueryOptions(buildQueryOptions(statement.getQueryOptions()));
     return builder;
   }
 
@@ -532,8 +524,7 @@ abstract class AbstractReadContext
       Options options,
       ByteString partitionToken) {
     beforeReadOrQuery();
-    final ExecuteSqlRequest.Builder request =
-        getExecuteSqlRequestBuilder(statement, queryMode, options);
+    final ExecuteSqlRequest.Builder request = getExecuteSqlRequestBuilder(statement, queryMode);
     if (partitionToken != null) {
       request.setPartitionToken(partitionToken);
     }
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
@@ -30,11 +30,13 @@ import com.google.cloud.spanner.Options.QueryOption;
 import com.google.cloud.spanner.Options.ReadOption;
 import com.google.cloud.spanner.SessionImpl.SessionTransaction;
 import com.google.cloud.spanner.spi.v1.SpannerRpc;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.protobuf.ByteString;
 import com.google.spanner.v1.BeginTransactionRequest;
 import com.google.spanner.v1.ExecuteBatchDmlRequest;
 import com.google.spanner.v1.ExecuteSqlRequest;
 import com.google.spanner.v1.ExecuteSqlRequest.QueryMode;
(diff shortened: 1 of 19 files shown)


google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
```diff
@@ -701,4 +707,125 @@ public Long run(TransactionContext transaction) throws Exception {
...
+              .singleUse()
+              .executeQuery(
+                  Statement.newBuilder(SELECT1.getSql())
+                      .withQueryOptions(QueryOptions.newBuilder().setOptimizerVersion("1").build())
+                      .build())) {
+        // Just iterate over the results to execute the query.
+        while (rs.next()) {}
+      }
+      // Check that the last query was executed using a custom optimizer version and statistics
+      // package.
```
skuruppu:
```suggestion
      // Check that the last query was executed using a custom optimizer version.
```

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
```diff
@@ -701,4 +707,125 @@ public Long run(TransactionContext transaction) throws Exception {
...
+                Statement.newBuilder(SELECT1.getSql())
+                    .withQueryOptions(QueryOptions.newBuilder().setOptimizerVersion("1").build())
+                    .build(),
+                QueryAnalyzeMode.PROFILE)) {
+          // Just iterate over the results to execute the query.
+          while (rs.next()) {}
+        }
+      }
+      // Check that the last query was executed using a custom optimizer version and statistics
+      // package.
```
skuruppu:
```suggestion
      // Check that the last query was executed using a custom optimizer version.
```

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
```diff
@@ -701,4 +707,125 @@ public Long run(TransactionContext transaction) throws Exception {
...
+              PartitionOptions.newBuilder().setMaxPartitions(10L).build(),
+              Statement.newBuilder(SELECT1.getSql())
+                  .withQueryOptions(QueryOptions.newBuilder().setOptimizerVersion("1").build())
+                  .build());
+      try (ResultSet rs = transaction.execute(partitions.get(0))) {
+        // Just iterate over the results to execute the query.
+        while (rs.next()) {}
+      }
+      // Check that the last query was executed using a custom optimizer version and statistics
+      // package.
```
skuruppu:
```suggestion
      // Check that the last query was executed using a custom optimizer version.
```

hengfengli:
One thing I notice is that PartitionedDML also uses `ExecuteSqlRequest`, which needs to support `QueryOptions` as well. 

https://github.com/googleapis/java-spanner/blob/d8b6438aa3b881c1c9baff584a74813664be4df8/google-cloud-spanner/src/main/java/com/google/cloud/spanner/PartitionedDMLTransaction.java#L73-L78

olavloite:
> LGTM
> 
> A couple of requests:
> 
> 1. I would like to verify that both ExecuteSql and ExecuteStreamingSql requests support query options. I think it does from what I understand of the implementation but thought I'd check.
> 2. If you could update the description of this PR to contain the information about the env variable and the precedence, that would be great. Example [PR](https://github.com/googleapis/google-cloud-dotnet/pull/4512)

1. I've added update statements to `ITQueryOptionsTest`. All update statements are executed by the client library using the `ExecuteSql` RPC. (All queries use `ExecuteStreamingSql`). This verifies that `QueryOptions` work for both RPCs.
1. Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
```diff
@@ -148,27 +247,24 @@ public void onTransactionMetadata(Transaction transaction) {
...
+            builder.bound.getMode() != TimestampBound.Mode.MAX_STALENESS
+                && builder.bound.getMode() != TimestampBound.Mode.MIN_READ_TIMESTAMP,
+            "Bounded staleness mode %s is not supported for multi-use read-only transactions."
+                + " Create a single-use read or read-only transaction instead.",
+            builder.bound.getMode());
+        this.bound = builder.bound;
+      } else {
+        this.timestamp = builder.timestamp;
+        this.transactionId = builder.transactionId;
+      }
```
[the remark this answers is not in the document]
olavloite: I would prefer not to. The reason for introducing this builder pattern in this PR is that this PR requires an additional object to be passed in to the different transaction classes. This would add another parameter to the constructors of these classes, bringing the total number to 6 in this specific case. As a general rule of thumb, a method (or constructor) in Java should not take more than 4 arguments: https://rules.sonarsource.com/java/RSPEC-107

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerImplTest.java
```diff
@@ -88,6 +89,47 @@ public void getDbclientAgainGivesSame() {
...
+          @Override
+          QueryOptions getEnvironmentQueryOptions() {
+            // Override and return default instance to prevent environment variables from
+            // interfering with the test case.
+            return QueryOptions.getDefaultInstance();
+          }
+        }.build();
+
+    try (SpannerImpl implWithQueryOptions = new SpannerImpl(rpc, optionsWithQueryOptions);
+        SpannerImpl implWithouQueryOptions = new SpannerImpl(rpc, optionsWithoutQueryOptions)) {
```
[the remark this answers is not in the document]
olavloite: Done.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerOptionsTest.java
```diff
@@ -427,4 +428,56 @@ public void testSetEmulatorHostWithProtocol() {
...
+        new SpannerOptions.SpannerEnvironment() {
+          @Override
+          public String getOptimizerVersion() {
+            return "";
+          }
+
+          @Override
+          public String getOptimizerStatisticsPackage() {
+            return "";
+          }
```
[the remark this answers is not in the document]
olavloite: Done.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerOptionsTest.java
```diff
@@ -427,4 +428,56 @@ public void testSetEmulatorHostWithProtocol() {
...
+        new SpannerOptions.SpannerEnvironment() {
+          @Override
+          public String getOptimizerVersion() {
+            return "2";
+          }
+
+          @Override
+          public String getOptimizerStatisticsPackage() {
+            return "";
+          }
```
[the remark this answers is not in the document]
olavloite: Done.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
olavloite: Done.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
olavloite: Done.

olavloite:
> One thing I notice is that PartitionedDML also uses `ExecuteSqlRequest`, which needs to support `QueryOptions` as well.
> 

I've added a PartitionedDML statement to `ITQueryOptionsTest` to verify that `QueryOptions` are supported (or actually; currently ignored) for DML and PartitionedDML statements.

skuruppu:
@olavloite you are welcome to merge this in as I've now merged in the generated code.

PR #107, opened 2020-03-15
fix: only register metrics once

### LINKED ISSUE #106: Session pool metrics registered multiple times
The new metrics that have been added for the `SessionPool` are added multiple times if multiple different session pools are created by a `Spanner` instance. This could happen if the user creates database clients for different databases, or if multiple database clients using different `SessionPoolOptions` are requested. This again will lead to an `InvalidArgumentException` being thrown. See for example https://source.cloud.google.com/results/invocations/a8dc0694-336f-483b-bb8f-2ee502506a2b/targets/cloud-devrel%2Fjava%2Fjava-docs-samples%2Fjava8%2Fpresubmit/log

olavloite:
The default MetricsRegistry used by OpenCensus is a singleton that does not allow the same metrics to be registered multiple times. The session pool metrics gathering did not take this into account, and when multiple session pools were created within the same class loader, it would throw
an InvalidArgumentException for the second session pool.

Fixes #106.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/MetricRegistryConstants.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/MetricRegistryConstants.java
@@ -15,6 +15,7 @@
  */
 package com.google.cloud.spanner;
 
+import com.google.api.gax.core.GaxProperties;
 import com.google.common.collect.ImmutableList;
 import io.opencensus.metrics.LabelKey;
 import io.opencensus.metrics.LabelValue;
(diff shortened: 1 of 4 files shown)


mayurkale22:
Although this fix looks good to me, I have implemented fix in OpenCensus (https://github.com/census-instrumentation/opencensus-java/pull/2017) to allow users to register the same metrics multiple times without exception. I am planning to release a cut either on Monday or Tuesday. What do you think about my fix?

olavloite:
> Although this fix looks good to me, I have implemented fix in OpenCensus ([census-instrumentation/opencensus-java#2017](https://github.com/census-instrumentation/opencensus-java/pull/2017)) to allow users to register the same metrics multiple times without exception. I am planning to release a cut either on Monday or Tuesday. What do you think about my fix?

@mayurkale22 Could you elaborate a little on how your fix will work when the same metric is registered multiple times both with the same label values and different label values? In the current implementation in the Spanner client the metrics are given label values to identify the database. The following situations are however possible:
1. Multiple database clients for the same database are created, but with different `SessionPoolOptions`. This will create multiple instances of a `SessionPool`, and each `SessionPool` register the same metric multiple times with the same label value. The measurements will however be different.
1. Multiple database clients for different databases are created. This will also lead to multiple instances of `SessionPool` being created, but this time with different label values. Will the metrics of each `SessionPool` be visible as a separate metric with a different label value in the monitoring?

olavloite:
Closing in favor of https://github.com/census-instrumentation/opencensus-java/pull/2017

PR #117, opened 2020-03-19
fix: add client id to metrics to avoid collisions

### LINKED ISSUE #106: Session pool metrics registered multiple times

olavloite:
This PR adds an automatically generated identifier to all database clients that are created by the client library. This avoids collisions of the same metrics being registered multiple times, and makes it possible to distinguish different clients from each other in the monitoring.

This PR does not allow the user to specify the id, but this could be added in a future change. That would need an API change by adding an overload to the method `getDatabaseClient(DatabaseId)` with an additional `clientId` parameter.

This should also fix the build error in https://github.com/GoogleCloudPlatform/java-docs-samples/pull/2323.

Fixes #106

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClientImpl.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClientImpl.java
@@ -37,9 +37,16 @@ class DatabaseClientImpl implements DatabaseClient {
     READ_WRITE
   }
 
+  @VisibleForTesting final String clientId;
   @VisibleForTesting final SessionPool pool;
 
+  @VisibleForTesting
   DatabaseClientImpl(SessionPool pool) {
+    this("", pool);
+  }
+
+  DatabaseClientImpl(String clientId, SessionPool pool) {
+    this.clientId = clientId;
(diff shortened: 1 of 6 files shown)


olavloite:
> I have verified these changes locally and seems to be working fine.
> 
> If possible, please include unit test to validate the logic.
> 
> 1. Multiple database clients for the same database w/ different `SessionPoolOptions`
> 2. Multiple database clients for different databases.

Thanks for verifying it locally :+1: . I've added a test case for the generated client id.

skuruppu:
I don't understand why cla/google is getting stuck. This has happened each time I've run the presubmits. It's run fine on other PRs.

PR #129, opened 2020-03-26
tests: assign unique id's per test case

### LINKED ISSUE #124: ITDml test is flaky
All tests in ITDMLTest.java relies on the same INSERT_DML statement. Each test executes the INSERT_DML & DELETE_DML statements, except abortOnceShouldSucceedAfterRetry test. All tests executed after abortOnceShouldSucceedAfterRetry will fail with "Row already exists" error. 

Short term fix:
- add the DELETE_DML into the test too

Correct / Long term fix:
- refactor the test cases to not rely on the same set of INSERT/DELETE_dml statements. These tests should be independent (dont try to create a database/create a new table for each test, that will make them slow). Let each test run its own unique INSERT DML that does not interfere with another tests execution.

olavloite:
Fixes #124


olavloite:
> LGTM. These test cases will not be run in parallel, right?

No, the test cases are executed sequentially.

PR #141, opened 2020-04-07
fix: retry non-idempotent long-running RPCs

olavloite:
RPCs returning a long-running operation, such as CreateDatabase, CreateBackup and RestoreDatabase, are non-idempotent and cannot be retried automatically by gax. This means that these RPCs sometimes fail with transient errors, such as `UNAVAILABLE` or `DEADLINE_EXCEEDED`. This change introduces automatic retries of these RPCs using the following logic:
1. Execute the RPC and wait for the operation to be returned.
2. If a transient error occurs while waiting for the operation, the client library queries the backend for the corresponding operation. If the operation is found, the resumes the tracking of the existing operation and returns that to the user.
3. If no corresponding operation is found in step 2, the client library retries the RPC from step 1.

Fixes https://github.com/GoogleCloudPlatform/java-docs-samples/issues/2571

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
@@ -21,6 +21,7 @@ import com.google.api.gax.grpc.GrpcInterceptorProvider;
 import com.google.api.gax.longrunning.OperationSnapshot;
 import com.google.api.gax.longrunning.OperationTimedPollAlgorithm;
 import com.google.api.gax.retrying.RetrySettings;
+import com.google.api.gax.rpc.StatusCode;
 import com.google.api.gax.rpc.TransportChannelProvider;
 import com.google.api.gax.rpc.UnaryCallSettings;
 import com.google.cloud.NoCredentials;
@@ -291,7 +292,8 @@ public class SpannerOptions extends ServiceOptions<Spanner, SpannerOptions> {
           .setRetrySettings(longRunningRetrySettings);
       databaseAdminStubSettingsBuilder
           .updateBackupSettings()
-          .setRetrySettings(longRunningRetrySettings);
+          .setRetrySettings(longRunningRetrySettings)
+          .setRetryableCodes(StatusCode.Code.DEADLINE_EXCEEDED, StatusCode.Code.UNAVAILABLE);
     }
 
     Builder(SpannerOptions options) {
@@ -581,6 +583,7 @@ public class SpannerOptions extends ServiceOptions<Spanner, SpannerOptions> {
       return this;
     }
 
+    @SuppressWarnings("rawtypes")
     @Override
     public SpannerOptions build() {
       // Set the host of emulator has been set.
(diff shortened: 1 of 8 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -584,30 +780,92 @@ public Database getDatabase(String databaseName) throws SpannerException {
...
+              public Timestamp apply(CreateBackupMetadata input) {
+                return input.getProgress().getStartTime();
+              }
+            });
+    return RetryHelper.runWithRetries(
+        callable,
+        databaseAdminStubSettings
+            .createBackupOperationSettings()
+            .getInitialCallSettings()
+            .getRetrySettings(),
```
skuruppu, line 821: I think we don't specify retry settings for non-idempotent RPCs. Do you know what this will default to then?

olavloite: There are defaults defined in the gapic configuration file as well, but those are already overridden by the client library. The defaults are set in the constructor of `SpannerOptions.Builder`:

https://github.com/googleapis/java-spanner/blob/9e5a1cdaacf71147b67681861f063c3276705f44/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java#L242-L295

hengfengli: So does this mean that the Java client library has its own retrying settings which are different from the gapic config file?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -291,7 +292,8 @@ private Builder() {
           .setRetrySettings(longRunningRetrySettings);
       databaseAdminStubSettingsBuilder
           .updateBackupSettings()
-          .setRetrySettings(longRunningRetrySettings);
+          .setRetrySettings(longRunningRetrySettings)
+          .setRetryableCodes(StatusCode.Code.DEADLINE_EXCEEDED, StatusCode.Code.UNAVAILABLE);
```
hengfengli, line 296: Why do we need to manually specify the retryable codes here? In the gapic config file (https://github.com/googleapis/googleapis/blob/d741cd976975c745d0199987aff0e908b8352992/google/spanner/admin/database/v1/spanner_admin_database_grpc_service_config.json#L58-L67), `updateBackup` will retry DEADLINE_EXCEEDED and UNAVAILABLE. Shouldn't this code be auto-generated? 

I guess Java gapic code generator is using a different config file: https://github.com/googleapis/googleapis/blob/master/google/spanner/admin/database/v1/spanner_admin_database_gapic.yaml#L83-L86.

hengfengli: Just curious: why do we manually set retrying settings instead of using the default values from the gapic config file?

olavloite: I had a lot of problems / transient errors using the default values. So I set the values manually to try to find settings that did work, so that we can set them back to the default config.

So the Java client should definitely not continue to use custom retry settings. Once we have verified that these settings work well, we should set these as the default in the config file.

olavloite: And yes, the Java client (and I think also the other clients, but I'm not 100% sure) use the retry settings in https://github.com/googleapis/googleapis/blob/master/google/spanner/admin/database/v1/spanner_admin_database_gapic.yaml. So that means for example that the default config for `UpdateBackup` is [non-idempotent](https://github.com/googleapis/googleapis/blob/master/google/spanner/admin/database/v1/spanner_admin_database_gapic.yaml#L83-L86).

PR #149, opened 2020-04-14
tests: add test for session distribution

olavloite:
Adds an additional test case for the `SessionClient` to verify the following behavior:
* Calling `SessionClient#asyncBatchCreateSesions` with `distributeOverChannels=false` should execute only one RPC using one channel.
* Calling `SessionClient#asyncBatchCreateSesions` with `distributeOverChannels=false` multiple times, should use a new channel hint for each consecutive call.

This ensures that increasing the number of sessions in the pool using consecutive calls to this method will eventually distribute the sessions over all available channels.

### DIFF
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionClientTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionClientTest.java
@@ -31,8 +31,10 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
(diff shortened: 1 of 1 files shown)


hengfengli:
> Adds an additional test case for the `SessionClient` to verify that it does distribute sessions over all available channels when it is called multiple times with `distributeOverChannels=false`. The latter only means that the batch that is created should not be distributed over multiple channels, but consecutive calls to the method should use a new channel hint for each new batch.

I am a bit confused by the first sentence. 

* When `distributeOverChannels=false`, the batch should always use the same channel. 
* If the same method is called again, it should use a new channel from available channels. 

Am I right?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionClientTest.java
```diff
@@ -193,6 +195,73 @@ public void onSessionCreateFailure(Throwable t, int createFailureForSessionCount
...
+        client.asyncBatchCreateSessions(numSessions, false, consumer);
+      }
+    }
+    assertThat(returnedSessionCount.get()).isEqualTo(numSessions * numBatches);
+    assertThat(usedChannelHintss.size()).isEqualTo(spannerOptions.getNumChannels() * 2);
+    List<Long> expectedChannels = new ArrayList<>();
+    for (long l = 0; l < spannerOptions.getNumChannels() * 2; l++) {
+      expectedChannels.add(l);
+    }
+    assertThat(usedChannelHintss).containsExactlyElementsIn(expectedChannels);
```
hengfengli, line 262: Assume that `spannerOptions.getNumChannels()` returns 4. So this means that we only have 4 available channels. 

But why `expectedChannels` can be `[0,1,2,3,...,7]`?

hengfengli: I see. 

https://github.com/googleapis/java-spanner/blob/9e5a1cdaacf71147b67681861f063c3276705f44/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionClient.java#L175

There is an increasing counter in SessionClient. Every time when we call `createSession`, `BatchCreateSessionsRunnable`, and `sessionWithId`, `sessionChannelCounter` increases to make sure they use different channels.

olavloite:
> > Adds an additional test case for the `SessionClient` to verify that it does distribute sessions over all available channels when it is called multiple times with `distributeOverChannels=false`. The latter only means that the batch that is created should not be distributed over multiple channels, but consecutive calls to the method should use a new channel hint for each new batch.
> 
> I am a bit confused by the first sentence.
> 
> * When `distributeOverChannels=false`, the batch should always use the same channel.
> * If the same method is called again with `distributeOverChannels=false`, it should use a new channel from available channels.
> 
> Am I right?

Yes, you're right, and I understand the confusion. I've updated the description to try to make it clearer.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionClientTest.java
olavloite: Java gRPC uses a channel hint to indicate that a call should be linked to a certain channel. That channel hint can be any `long` value, and gRPC will automatically bring that back to a certain channel number. The current implementation is `channelHint % numChannels`, but the `SessionClient` should not really depend on that (or care about that). The API specifies that any `long` value is OK, so the `SessionClient` just assigns a continuously increasing `long`.

PR #153, opened 2020-04-15
fix: fix flaky test and remove warnings

### LINKED ISSUE #146: spanner.it.ITSpannerOptionsTest: testMultipleSpannersFromSameSpannerOptions failed
This test failed!

To configure my behavior, see [the Build Cop Bot documentation](https://github.com/googleapis/repo-automation-bots/tree/master/packages/buildcop).

If I'm commenting on this issue too often, add the `buildcop: quiet` label and
I will stop commenting.

---

commit: c8be7b5d43c01f6170eb02c6e8cc146cb2597ecc
buildURL: [Build Status](https://source.cloud.google.com/results/invocations/a7db9596-6afe-462a-a450-d7cf6245a8d4), [Sponge](http://sponge2/a7db9596-6afe-462a-a450-d7cf6245a8d4)
status: failed

olavloite:
The test should check for at most baseThreadCount instead of equality, as it is possible that the thread pool shuts down one or more of its threads temporarily.

Fixes #146

### DIFF
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITSpannerOptionsTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITSpannerOptionsTest.java
@@ -16,9 +16,7 @@
 
 package com.google.cloud.spanner.it;
 
-import static org.hamcrest.CoreMatchers.equalTo;
-import static org.hamcrest.CoreMatchers.is;
-import static org.hamcrest.MatcherAssert.assertThat;
+import static com.google.common.truth.Truth.assertThat;
 
 import com.google.cloud.spanner.Database;
 import com.google.cloud.spanner.DatabaseAdminClient;
@@ -38,10 +36,8 @@ import java.util.regex.Pattern;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.ClassRule;
-import org.junit.Rule;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
-import org.junit.rules.ExpectedException;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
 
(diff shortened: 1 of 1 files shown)


google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITSpannerOptionsTest.java
```diff
@@ -106,29 +101,26 @@ public void testCloseAllThreadsWhenClosingSpanner() throws InterruptedException
       }
       // Check the number of threads after the query. Doing a request should initialize a thread
       // pool for the underlying SpannerClient.
-      assertThat(
-          getNumberOfThreadsWithName(SPANNER_THREAD_NAME),
-          is(equalTo(DEFAULT_NUM_CHANNELS * NUM_THREADS_PER_CHANNEL + baseThreadCount)));
+      assertThat(getNumberOfThreadsWithName(SPANNER_THREAD_NAME))
+          .isEqualTo(DEFAULT_NUM_CHANNELS * NUM_THREADS_PER_CHANNEL + baseThreadCount);
```
hengfengli, line 105: Why here shouldn't be `isAtMost`?

hengfengli:
Just curious: I keep seeing `Kokoro - Test: Samples` failed over multiple PRs. Any idea?

olavloite:
> Just curious: I keep seeing `Kokoro - Test: Samples` failed over multiple PRs. Any idea?

As far as I know, the plan is to include the Samples test in the build to check that we don't break any samples while making changes, but the configuration is not (yet) correct. So these tests currently fail in all PRs.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITSpannerOptionsTest.java
olavloite: These threads should all be present as long as the `Spanner` instance is opened and being used. Closing it would normally bring the number of threads back to the `baseThreadCount`, but it can happen that additional threads are (temporarily) removed after closing it.

PR #159, opened 2020-04-20
fix: check for null input for Id classes

### LINKED ISSUE #145: InstanceId.of(String, String) accepts null project, but InstanceId.equals doesn't handle null
`InstanceId.equals` threw a NullPointerException when I accidentally set the project to null.  I would have expected the exception to be thrown earlier, in the factory method.

#### Code example

```java
@Test
public void test() {
  InstanceId instance1 = InstanceId.of("project", "instance-1");
  InstanceId instance2 = InstanceId.of(null, "instance-2");
  System.out.println(instance1.equals(instance2));
}
```

#### Stack trace

```
java.lang.NullPointerException
	at com.google.cloud.spanner.InstanceId.equals(InstanceId.java:66)
...
```

EDIT: I used google-cloud-spanner version 1.52.0.

olavloite:
Check and report null input for the static create methods of the InstanceId, InstanceConfigId, DatabaseId, SessionId and BackupId classes.

Fixes #145

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/BackupId.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/BackupId.java
@@ -30,8 +30,8 @@ public final class BackupId {
   private final String backup;
 
   BackupId(InstanceId instanceId, String backup) {
-    this.instanceId = instanceId;
-    this.backup = backup;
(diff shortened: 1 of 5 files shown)


olavloite:
> LGTM. I wonder shouldn't the title be "fix: check for null input for Id classes"? Because null inputs are invalid.

Good point, fixed the header and comment.

PR #169, opened 2020-04-22
fix: client library should use libraries-bom

### LINKED ISSUE #168: Client library should use libraries-bom
The client library should use the `com.google.cloud:libraries-bom` to prevent dependency version conflicts. See also https://github.com/GoogleCloudPlatform/java-docs-samples/issues/2718.

olavloite:
The client library should use the `com.google.cloud:libraries-bom` to prevent version conflicts.

Related to https://github.com/GoogleCloudPlatform/java-docs-samples/issues/2718
Fixes #168


olavloite:
> I can send a PR which demonstrates this.

@chingor13 
Thanks for the review. I would very much appreciate an example for that.

chingor13:
See #172 for what the shared-dependencies import would look like

olavloite:
Closing in favor of #172

PR #180, opened 2020-04-24
fix: create filter in correct order

### LINKED ISSUE #179: ListDatabaseOperations should filter on metadata type first, and then metadata field
`ListDatabaseOperations` requests must specify filter options in a specific order:
* This will fail: `(metadata.source_type=BACKUP) (metadata.@type=type.googleapis.com/google.spanner.admin.database.v1.RestoreDatabaseMetadata)`
* This will succeed: `(metadata.@type=type.googleapis.com/google.spanner.admin.database.v1.RestoreDatabaseMetadata) (metadata.source_type=BACKUP)`

The client library currently creates filters that can contain these fields in the opposite order, which causes RPC's to fail.

olavloite:
Fixes #179


olavloite:
@googlebot I consent.

hengfengli:
We probably should file the order of filter issue to the backend team.

PR #181, opened 2020-04-25
fix: stop preparing session on most errors

### LINKED ISSUE #177: BeginTransaction shouldn't retry indefinitely on FAILED_PRECONDITION
A team reported that they're seeing BeginTransaction retrying indefinitely on FAILED_PRECONDITION errors.

This is an issue similar to #16. For now, just giving up on the retries for this specific error would do.

But perhaps we can consider not retrying for all non-retryable errors as well in a separate PR?

olavloite:
Most errors that occur during preparing a session for a read/write transaction should be considered permanent, and should stop the automatic preparing of sessions. Any subsequent call to get a read/write session will cause an in-process BeginTransaction RPC to be executed. If the problem has been fixed in the meantime, the RPC will succeed and the automatic prepare of sessions will start again. Otherwise, the error is propagated to the user.

Fixes #177

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
@@ -49,6 +49,7 @@ import com.google.common.base.Preconditions;
 import com.google.common.base.Supplier;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.MoreExecutors;
 import com.google.common.util.concurrent.SettableFuture;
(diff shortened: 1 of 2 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1696,25 +1716,30 @@ private void handlePrepareSessionFailure(
...
-          decrementPendingClosures(1);
+        if (isDatabaseOrInstanceNotFound(e)) {
+          // Remove the session from the pool.
+          if (isClosed()) {
+            decrementPendingClosures(1);
+          }
+          allSessions.remove(session);
+          this.resourceNotFoundException =
+              MoreObjects.firstNonNull(
+                  this.resourceNotFoundException, (ResourceNotFoundException) e);
```
skuruppu, line 1739: In the previous logic, we executed this logic also for PERMISSION_DENIED errors. It doesn't anymore. Is this a problem?

olavloite: No, it is still executed for `PERMISSION_DENIED` errors, but that is now covered by the general error handling that also handles other types of errors. The specific errors `DatabaseNotFound` and `InstanceNotFound` are handled specifically, as these should not allow the normal operation of the session pool to be restarted once the error does no longer occur. The reasoning behind that is that a `DatabaseNotFound` exception can only be fixed by creating a new database with the same name as the old one, but that is still a new database.
The other errors, such as `PERMISSION_DENIED` and now also for examle `FAILED_PRECONDITION`, should allow the session pool to restart the automatic preparing of sessions once the error has been fixed. This could be granting the user the necessary permissions or changing the state of database to fulfill all requirements.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
```diff
@@ -469,12 +469,25 @@ public void testDatabaseOrInstanceDoesNotExistOnReplenish() throws Exception {
 
   @Test
   public void testPermissionDeniedOnPrepareSession() throws Exception {
```
olavloite, line 471: This test case covers the `PERMISSION_DENIED` case discussed above.

skuruppu:
Thanks for the clarification @olavloite. Please merge when you're ready.

PR #182, opened 2020-04-25
fix: drop orphaned test databases

olavloite:
Drop orphaned test databases to prevent `RESOURCE_EXHAUSTED` errors on the CI environment.


olavloite:
Closing for now, as `createTime` of a database is apparently not set for databases that are created, only for databases that are restored.

PR #184, opened 2020-04-25
tests: Attempt to fix random timeouts in ITBackup

olavloite:
The `ITBackup` test fails randomly (and lately: consistently) on the CI environment with timeout errors. This seems to be caused by several backup operations that are still in progress and do not finish. Cancelling these before starting the actual backup tests seems to solve the problem (for now).

### DIFF
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITBackupTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITBackupTest.java
@@ -43,6 +43,7 @@ import com.google.common.base.Predicate;
 import com.google.common.base.Stopwatch;
 import com.google.common.collect.Iterables;
 import com.google.longrunning.Operation;
+import com.google.protobuf.InvalidProtocolBufferException;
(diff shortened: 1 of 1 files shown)


google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITBackupTest.java
```diff
@@ -93,6 +100,40 @@ public void setUp() throws Exception {
...
+                        .unpack(CreateBackupMetadata.class)
+                        .getProgress()
+                        .getStartTime());
+            long diffSeconds = currentTime.getSeconds() - startTime.getSeconds();
+            if (TimeUnit.HOURS.convert(diffSeconds, TimeUnit.SECONDS) >= 6L) {
+              logger.warning(
+                  String.format(
+                      "Cancelling test backup operation %s that was started at %s",
+                      operation.getName(), startTime.toString()));
+              dbAdminClient.cancelOperation(operation.getName());
```
skuruppu, line 128: If for any reason in the future, the cancelling still doesn't work, your other option may be to just call delete on those backups. AFAIK, backups in the process of being created can be deleted.

PR #192, opened 2020-05-06
feat: support setting compression option

olavloite:
Adds support for setting the compressor to use for gRPC calls.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
@@ -49,6 +49,8 @@ import com.google.spanner.admin.database.v1.CreateDatabaseRequest;
 import com.google.spanner.admin.database.v1.RestoreDatabaseRequest;
 import com.google.spanner.v1.ExecuteSqlRequest.QueryOptions;
 import io.grpc.CallCredentials;
+import io.grpc.CompressorRegistry;
+import io.grpc.ExperimentalApi;
 import io.grpc.ManagedChannelBuilder;
 import java.io.IOException;
 import java.net.MalformedURLException;
@@ -58,6 +60,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import org.threeten.bp.Duration;
 
 /** Options for the Cloud Spanner service. */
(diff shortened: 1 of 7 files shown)


google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITSpannerOptionsTest.java
```diff
@@ -168,6 +168,27 @@ public void testMultipleSpannersFromSameSpannerOptions() throws InterruptedExcep
...
+    for (String compressorName : new String[] {"gzip", "identity", null}) {
+      SpannerOptions options =
+          env.getTestHelper().getOptions().toBuilder().setCompressorName(compressorName).build();
+      try (Spanner spanner = options.getService()) {
+        DatabaseClient client = spanner.getDatabaseClient(db.getId());
+        try (ResultSet rs =
+            client
+                .singleUse()
+                .executeQuery(Statement.of("SELECT 1 AS COL1 UNION ALL SELECT 2 AS COL2"))) {
+          assertThat(rs.next()).isTrue();
```
rven1, line 182: Would it be possible to test an end to end compression and verify that the compressor is actually used.

skuruppu: Yes absolutely. Once the support is in the staging environment, we can run this IT against it to test the change. @olavloite I'll give you a heads up when the feature is available in staging.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -558,6 +565,28 @@ public Builder setCallCredentialsProvider(CallCredentialsProvider callCredential
...
+    /**
+     * Sets the compression to use for all gRPC calls. The compressor must be a valid name known in
+     * the {@link CompressorRegistry}.
+     *
+     * <p>Supported values are:
+     *
+     * <ul>
+     *   <li>gzip: Enable gzip compression
+     *   <li>identity: Disable compression
+     *   <li><code>null</code>: Use default comporession
```
skuruppu, line 577:
```suggestion
     *   <li><code>null</code>: Use default compression
```

olavloite: Done.

skuruppu:
@olavloite we can merge this change now. If you can resolve the conflict and merge, that would be great.

PR #193, opened 2020-05-06
feat: add support for NUMERIC data type

olavloite:


google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/RandomResultSetGenerator.java
```diff
@@ -41,6 +42,7 @@
         Type.newBuilder().setCode(TypeCode.BOOL).build(),
         Type.newBuilder().setCode(TypeCode.INT64).build(),
         Type.newBuilder().setCode(TypeCode.FLOAT64).build(),
+        //        Type.newBuilder().setCode(TypeCode.NUMERIC).build(),
```
thiagotnunes, line 45: nit: Commented out code. Maybe we forgot to remove it?

olavloite: Ah, good catch. This is not supposed be commented out at all.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITWriteTest.java
```diff
@@ -352,6 +356,21 @@ public void writeDateNull() {
     assertThat(row.isNull(0)).isTrue();
   }
 
+  @Test
```
skuruppu, line 359: @olavloite thanks for adding the integration tests. But I do have one small request. Would you be able to move it to a separate PR?

The reason being that we will merge this PR in before support for the feature is available in prod, which unfortunately mean that the tests will fail at that point.

olavloite: NP, I'll move them to a separate PR.

PR #203, opened 2020-05-13
fix: Partitioned DML timeout was not always respected

### LINKED ISSUE #199: Experienced '504 Deadline Exceeded' error when tried to delete large number rows by partitioned dml
There are 1 billion old records need to be cleaned. According the documents, It seems Partitioned DML should be the good choice. I choose java client because we can set timeout by using setPartitionedDmlTimeout(Duration.ofHours(24L)). However, I still experienced the '504 Deadline Exceeded' after a few hours that shorter than 24 hours I set.  Why setPartitionedDmlTimeout did not work? How should I solve the problem? Is it not possible to delete so large number of data by using Partitioned DML. 

Following are the details.
#### Environment details

1. Spanner 
    3 nodes, 2 billion records in a table and 1 billion old records need to be cleaned
2. OS type and version: MacOs version 10.15.3
3. Java version: java 8
4. spanner version(s): the latest version 

#### Steps to reproduce

  1. Write 2 billion records with timestamp in to the spanner table with dataflow
  2. Run a clean job to delete  the old 1 billion records with Partitioned DML by java client
  3. Set the timeout to be 24 hours by using setPartitionedDmlTimeout(Duration.ofHours(24L))
  4. Run the java code

#### Code example

```java
SpannerOptions options = SpannerOptions.newBuilder().setPartitionedDmlTimeout(Duration.ofHours(24L)).build();

        try {

            DatabaseId db = DatabaseId.of(options.getProjectId(), args[0], args[1]);
            // [END init_client]
            //This will return the default project id based on the environment.
            String clientProject = spanner.getOptions().getProjectId();
            if (!db.getInstanceId().getProject().equals(clientProject)) {
                System.err.println(
                        "Invalid project specified. Project in the database id should match the"
                                + "project name set in the environment variable GOOGLE_CLOUD_PROJECT. Expected: "
                                + clientProject);
            }

            // [START init_client]
            DatabaseClient dbClient = spanner.getDatabaseClient(db);

            SimpleDateFormat formatter = new SimpleDateFormat("dd/MM/yyyy hh:mm:ss.SSS");
            String sql="DELETE FROM MyTable WHERE updateTimestamp<'2020-05-02T03:00:00Z'";
            Date currentDate = new Date();
            System.out.println(String.format("Delete Start:%s",formatter.format(currentDate)));

            long lStartTime = Instant.now().toEpochMilli();
            deleteUsingPartitionedDml(dbClient,sql);
            long lEndTime = Instant.now().toEpochMilli();
            long elaspedTime = lEndTime - lStartTime;
            System.out.println("Elapsed time in : " + elaspedTime/1000);

        } finally {
            spanner.close();
        }
        // [END init_client]
        System.out.println("Closed client");
    }
    private static void deleteUsingPartitionedDml(DatabaseClient dbClient,String sql) {
        long rowCount = dbClient.executePartitionedUpdate(Statement.of(sql));

        System.out.printf("%d records deleted.\n", rowCount);
    }

```

olavloite:
Setting a timeout value for Partitioned DML would not be respected if the timeout value was higher than the timeout value set for the ExecuteSql RPC on the SpannerStub. Lower timeout values would be respected.

Fixes #199

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
@@ -19,6 +19,7 @@ package com.google.cloud.spanner.spi.v1;
 import static com.google.cloud.spanner.SpannerExceptionFactory.newSpannerException;
 
 import com.google.api.core.ApiFuture;
+import com.google.api.core.InternalApi;
 import com.google.api.core.NanoClock;
 import com.google.api.gax.core.CredentialsProvider;
 import com.google.api.gax.core.ExecutorProvider;
@@ -154,6 +155,7 @@ import javax.annotation.Nullable;
 import org.threeten.bp.Duration;
 
 /** Implementation of Cloud Spanner remote calls using Gapic libraries. */
+@InternalApi
 public class GapicSpannerRpc implements SpannerRpc {
   /**
    * {@link ExecutorProvider} that keeps track of the executors that are created and shuts these
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
@@ -54,6 +54,7 @@ import com.google.cloud.spanner.admin.instance.v1.stub.GrpcInstanceAdminStub;
 import com.google.cloud.spanner.admin.instance.v1.stub.InstanceAdminStub;
 import com.google.cloud.spanner.v1.stub.GrpcSpannerStub;
 import com.google.cloud.spanner.v1.stub.SpannerStub;
+import com.google.cloud.spanner.v1.stub.SpannerStubSettings;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Function;
 import com.google.common.base.MoreObjects;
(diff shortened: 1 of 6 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -326,6 +330,22 @@ public GapicSpannerRpc(final SpannerOptions options) {
...
+          .setStreamWatchdogProvider(watchdogProvider)
+          .executeSqlSettings()
+          .setRetrySettings(
+              options
+                  .getSpannerStubSettings()
+                  .executeSqlSettings()
+                  .getRetrySettings()
+                  .toBuilder()
+                  .setInitialRpcTimeout(options.getPartitionedDmlTimeout())
+                  .setMaxRpcTimeout(options.getPartitionedDmlTimeout())
```
skuruppu, line 346: Are PDML requests not meant to be retried? I'm just curious why the initial and max timeouts are the same.

olavloite: Good question.

I would say it's a a matter of probabilities and a little bit opinion:
1. PDML is designed to accept long running update statements. That means that we should not automatically retry the statement if it takes longer than X time, as it would be impossible to find a good global value for X, as we don't know whether the statement is taking a long time because of a network problem or because the statement itself is taking a long time. Re-sending a long-running statement that is still running on the server would only hurt performance. My opinion is that for a PDML statement the chance that the statement is actually taking a long time is more probable than a network problem.
1. The PDML documentation also clearly states that the update statement should be idempotent. This means that we should automatically retry it if the server is unavailable.

I think we should consider adding an additional method to the public API which would allow the users to specify a _lower_ timeout for a specific statement, in addition to the current global setting. That would give the user more control over the timeout setting for statements that are known to take less time than the global setting.

olavloite: I've added an extra test case to ensure that PDML is retried on `UNAVAILABLE`.

PR #214, opened 2020-05-15
fix: test allowed a too old staleness

olavloite:
Fixes an issue that the test would allow a too old staleness (10 seconds), while the test would only wait 100ms before actually start reading.


olavloite:
@skuruppu Sorry, I forgot to add you as a reviewer for this PR. It was added because of a flaky test failure some time ago.

PR #128, opened 2020-03-25
feat: add num_sessions_in_pool metric

mayurkale22:
Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/java-spanner/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Ensure the tests and linter pass
- [x] Code coverage does not decrease (if any source code was changed)
- [x] Appropriate docs were updated (if necessary)

Updates #53, this is based on the last discussion in the thread: https://github.com/googleapis/java-spanner/issues/53#issuecomment-601455550

This is how it looks in Stackdriver UI:
Screen Shot 2020-03-25 at 1 42 39 PM

**Breaking Note:**
The `cloud.google.com/java/spanner/in_use_sessions` metric has been deprecated since v1.55.2, use `cloud.google.com/java/spanner/num_sessions_in_pool` instead.

PR #244, opened 2020-05-27
chore: document OpenCensus Metrics

mayurkale22:
Originates from https://github.com/googleapis/java-spanner/issues/53#issuecomment-628111904

This is just the initial draft, will update after #128.

Feel free to add comments.


skuruppu:
Thanks for this change @mayurkale22.

@chingor13 do you know if we can freely modify the README like this given the recent changes in #173?

PR #252, opened 2020-06-09
fix: backend now supports optimizer version for DML

### LINKED ISSUE #249: com.google.cloud.spanner.it.ITQueryOptionsTest.executeUpdate Fails for 100000
Looking at:

https://github.com/googleapis/java-spanner/blob/f4fa6bfca4bb821cbda426c4cb7bf32f091a2913/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITQueryOptionsTest.java#L168

I see that the test was intended for Version 100,000. But Version 10,000 was provided to the run command.

The comment does say that for version 100,000 the backend should safely ignore it. At least for version 10,000 the backend does not ignore and in fact throws an error:
```
com.google.cloud.spanner.SpannerException: INVALID_ARGUMENT: com.google.api.gax.rpc.InvalidArgumentException: io.grpc.StatusRuntimeException: INVALID_ARGUMENT: Query optimizer version: 10000 is not supported
	at com.google.cloud.spanner.it.ITQueryOptionsTest.executeUpdate(ITQueryOptionsTest.java:157)
Caused by: java.util.concurrent.ExecutionException: com.google.api.gax.rpc.InvalidArgumentException: io.grpc.StatusRuntimeException: INVALID_ARGUMENT: Query optimizer version: 10000 is not supported
	at com.google.cloud.spanner.it.ITQueryOptionsTest.executeUpdate(ITQueryOptionsTest.java:157)
Caused by: com.google.api.gax.rpc.InvalidArgumentException: io.grpc.StatusRuntimeException: INVALID_ARGUMENT: Query optimizer version: 10000 is not supported
Caused by: io.grpc.StatusRuntimeException: INVALID_ARGUMENT: Query optimizer version: 10000 is not supported
```

A fix might be to add a zero to Line 168; but i do not have full context of this test.

olavloite:
Fixes #249 ☕️

### DIFF
--- a/.kokoro/build.sh
+++ b/.kokoro/build.sh
@@ -51,9 +51,7 @@ test)
     RETURN_CODE=$?
     ;;
 lint)
-    mvn \
-      -Penable-samples \
-      com.coveo:fmt-maven-plugin:check
+    mvn com.coveo:fmt-maven-plugin:check
     RETURN_CODE=$?
     ;;
 javadoc)
--- a/.kokoro/presubmit/lint.cfg
+++ b/.kokoro/presubmit/lint.cfg
@@ -4,7 +4,7 @@
 
 env_vars: {
   key: "TRAMPOLINE_IMAGE"
-  value: "gcr.io/cloud-devrel-kokoro-resources/java8"
+  value: "gcr.io/cloud-devrel-kokoro-resources/java11"
 }
 
 env_vars: {
--- a/.kokoro/presubmit/lint.cfg
+++ b/.kokoro/presubmit/lint.cfg
@@ -4,7 +4,7 @@
 
 env_vars: {
   key: "TRAMPOLINE_IMAGE"
-  value: "gcr.io/cloud-devrel-kokoro-resources/java11"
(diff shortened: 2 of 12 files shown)


olavloite:
@skuruppu 
The build error for the code format seems to be caused by a non-related problem. The failing integration test that is fixed by this PR is however breaking all other PRs, so I think we should consider merging this PR even without a successful Code Format check.

skuruppu:
Oooh I have no idea what this error means:
```
java.lang.UnsupportedClassVersionError: com/google/googlejavaformat/java/FormatterException has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
```

Unfortunately, I don't have the permissions to force-merge PRs.

chingor13:
This will be fixed by google-cloud-shared-config 0.7.0 which is releasing now.

skuruppu:
I'm going to merge this in while the build is green. Then I'll get the presubmits running on the other PRs that are blocked.

PR #253, opened 2020-06-09
deps: include test-jar in bom

olavloite:
Including the google-cloud-spanner test-jar in the bom will allow other libraries to use the test jar without having to specify a version. This again should prevent the kind of failures that we are currently seeing in https://github.com/googleapis/java-spanner-jdbc/pull/142

(The jdbc driver currently defines a specific version for the test-jar in the pom [here](https://github.com/googleapis/java-spanner-jdbc/blob/bc7d5bd6205b23c99d01d2895ffb5c48ba423ea3/pom.xml#L62).)

### DIFF
--- a/google-cloud-spanner-bom/pom.xml
+++ b/google-cloud-spanner-bom/pom.xml
@@ -87,6 +87,12 @@
         <version>1.55.1</version><!-- {x-version-update:google-cloud-spanner:current} -->
       </dependency>
       <dependency>
+        <groupId>com.google.cloud</groupId>
+        <artifactId>google-cloud-spanner</artifactId>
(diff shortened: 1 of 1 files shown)


olavloite:
The integration test failures are not related to this change, and will be fixed by #252. The code format failure is as far as I can tell also not related to this change.

skuruppu:
Unfortunately, I don't have the permissions to force-merge if required checks fail :(

chingor13:
The lint problem should be fixed by google-cloud-shared-config 0.7.0 which is releasing now - google-java-format shipped a version incompatible with Java 8

PR #261, opened 2020-06-11
fix: include an explicit version for javax-annotations-api

olavloite:
The Maven flatten plugin transforms the pom of the project from a set of imports and transitive dependencies into a pom that contains all dependencies and versions explicitly. This process did however miss the javax-annotations-api dependency that is only included for JDK9 and higher.

Fixes https://github.com/GoogleCloudPlatform/java-docs-samples/issues/3139


stephaniewang526:
https://github.com/mojohaus/flatten-maven-plugin/pull/155 should fix this problem in the plugin -- we're waiting on the release.

Issue #280, opened 2020-06-18
Java [9, ...) builds were broken but not picked up by tests


olavloite:
Java 9 and higher builds were broken by the maven-flatten-plugin because of this profile in the pom:

```xml
    <profile>
      <id>java9</id>
      <activation>
        <jdk>[9,)</jdk>
      </activation>
      <dependencies>
        <dependency>
          <groupId>javax.annotation</groupId>
          <artifactId>javax.annotation-api</artifactId>
          <version>1.3.2</version>
        </dependency>
      </dependencies>
    </profile>
```

This breakage does not seem to have been picked up by the (samples) tests, while it did break the samples once those were updated to the latest version of the Spanner client. These kinds of errors should be detected by the tests in the client library.

See also https://github.com/GoogleCloudPlatform/java-docs-samples/issues/3182

PR #281, opened 2020-06-18
process: also test deps on Java 11

olavloite:
Run dependency test on both Java 8 and Java 11, as some dependencies are only included in Java versions 9 and higher.

Towards #280 and https://github.com/GoogleCloudPlatform/java-docs-samples/issues/3182


skuruppu:
@olavloite I'll leave this up to you to merge since I don't know which parts of this will be done by the bot.

PR #284, opened 2020-06-19
feat(deps): adopt flatten plugin and google-cloud-shared-dependencies

yangnuoyu:
@stephaniewang526 @chingor13

### DIFF
--- a/google-cloud-spanner/pom.xml
+++ b/google-cloud-spanner/pom.xml
@@ -336,7 +336,6 @@
         <dependency>
           <groupId>javax.annotation</groupId>
           <artifactId>javax.annotation-api</artifactId>
-          <version>1.3.2</version>
         </dependency>
       </dependencies>
     </profile>
--- a/grpc-google-cloud-spanner-v1/pom.xml
+++ b/grpc-google-cloud-spanner-v1/pom.xml
@@ -53,7 +53,6 @@
         <dependency>
           <groupId>javax.annotation</groupId>
           <artifactId>javax.annotation-api</artifactId>
-          <version>1.3.2</version>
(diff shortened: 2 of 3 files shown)


stephaniewang526:
@olavloite could you take a look at the IT failure?
```
expected: 20
but was : 8
	at com.google.cloud.spanner.it.ITSpannerOptionsTest.testCloseAllThreadsWhenClosingSpanner(ITSpannerOptionsTest.java:105)
```

olavloite:
> @olavloite could you take a look at the IT failure?
> 
> ```
> expected: 20
> but was : 8
> 	at com.google.cloud.spanner.it.ITSpannerOptionsTest.testCloseAllThreadsWhenClosingSpanner(ITSpannerOptionsTest.java:105)
> ```

This failure is caused by the changed that is introduced in https://github.com/googleapis/gax-java/commit/cf1bce8b313a3776cbf478eeab821b5f4ef9cbba

Spanner sets an `ExecutorProvider` on the `TransportChannelProvider`. Previously, this meant that an executor would be requested for each channel that was created by the `TransportChannelProvider`. The `setExecutorProvider` has now been deprecated and its implementation has been changed to get one `Executor`  to use for all channels, instead of requesting one `Executor` for each channel. That again means that there are now a lot less threads available for the transport channels than before. I wonder whether that was an intentional part of the change, or more of a side-effect of the change.

I think that we need to add some code changes when upgrading to this latest version to mitigate for this. Otherwise the entire Spanner library will only have 4 threads available for all transport channels for both all the data clients as well as the admin clients.

olavloite:
Superseded by #302

PR #287, opened 2020-06-19
perf: use streaming RPC for PDML

olavloite:
Changes Partitioned DML from using the `ExecuteSql` RPC to `ExecuteStreamingSql`.


google-cloud-spanner/src/main/java/com/google/cloud/spanner/PartitionedDMLTransaction.java
```diff
@@ -60,41 +72,88 @@ private ByteString initTransaction() {
...
+                  foundStats = true;
+                  updateCount += rs.getStats().getRowCountLowerBound();
+                }
+              }
+              break;
+            } catch (UnavailableException e) {
+              // Retry the stream in the same transaction if the stream breaks with
+              // UnavailableException and we have a resume token. Otherwise, we just retry the
+              // entire transaction.
+              if (resumeToken != null && !ByteString.EMPTY.equals(resumeToken)) {
```
skuruppu, line 128: nit: I don't know if you need to check for the `resumeToken != null` condition since you always init it to EMPTY and only update it if it's not null.

olavloite: Good point, done.

PR #290, opened 2020-06-21
feat: async connection api

olavloite:
Adds async support for the connection API.


olavloite:
Closing in favor of #392

PR #295, opened 2020-06-23
Add retry strategy implementation

kurisumakise2011:
By default in case of AbortedTransaction only one strategy is allowed - ChecksumResult set
But in java applications it's usual sutation when result set is changed and you need to retry
a transaction with ignoring checksums.
Add a new particular retry strategy FORCE_RETRY that will do execution anyway and return result set.
By default Checksum is prefferable for compability.


skuruppu:
Since you closed #296, I converted this to a draft. Hope that's ok.

PR #302, opened 2020-06-24
feat(deps): adopt flatten plugin and google-cloud-shared-dependencies and update ExecutorProvider

olavloite:
Replaces #284

### DIFF
--- a/google-cloud-spanner/pom.xml
+++ b/google-cloud-spanner/pom.xml
@@ -336,7 +336,6 @@
         <dependency>
           <groupId>javax.annotation</groupId>
           <artifactId>javax.annotation-api</artifactId>
-          <version>1.3.2</version>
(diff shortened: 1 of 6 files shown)


stephaniewang526:
seeing ``` [ERROR] Failures: 
[ERROR]   SpannerOptionsThreadTest.testCloseAllThreadsWhenClosingSpanner:103 expected: 16```

olavloite:
> seeing ` [ERROR] Failures: [ERROR] SpannerOptionsThreadTest.testCloseAllThreadsWhenClosingSpanner:103 expected: 16`

Yeah, I noticed as well, and I'm looking into it. Locally the test succeeds. My initial guess is that it might be caused by the thread group of the thread that is running the test on the CI env is a different group than group of the gax threads.

PR #305, opened 2020-06-25
test: run integration tests against the emulator

skuruppu:


skuruppu:
@olavloite, how do I get the ITs for the connections API running here as well? Or should that be a separate check?

olavloite:
> @olavloite, how do I get the ITs for the connections API running here as well? Or should that be a separate check?

It seems that the connection API ITs are excluded from the CI builds because they have conflicting categories:
* The abstract super class of all the ITs have the category `IntegrationTest`.
* The concrete sub classes have the category `ParallelIntegrationTest`

Each category is included and excluded from one of the integration test runs, and as exclusions trump inclusions, these were excluded from all CI runs. See https://github.com/googleapis/java-spanner/pull/309 for a fix for this.

skuruppu:
Closing in favor of #310.

PR #310, opened 2020-06-26
test: run integration tests against emulator

olavloite:
Based on #305 , fixes some problems in some of the integration tests.


skuruppu:
Thanks so much for this change @olavloite.

PR #319, opened 2020-07-07
test: clean up system-test setup against emulator

skuruppu:
Use Docker image instead of manually downloading the emulator binary and running it.


olavloite:
The timeout in the integration tests against the emulator is caused by an eternal aborted-retry loop in `ITReadWriteAutocommitSpannerTest`. What happens in the following:

1. Transaction 1 executes `UPDATE TEST SET NAME='test18' WHERE ID=1000`  and then calls `Rollback`.
1. Transaction 2 executes `DELETE FROM TEST WHERE ID=1000`. This statement returns an `Aborted` error. No `Rollback` or `Commit` is called, as the transaction has already been aborted.
1. Transaction 2 is retried and executes `DELETE FROM TEST WHERE ID=1000` in a new transaction. This statement returns an `Aborted` error. No `Rollback` or `Commit` is called, as the transaction has already been aborted.
1. ... this repeats for ever ...

PR #340, opened 2020-07-10
fix: check if emulator is running if env var is set

### LINKED ISSUE #326: Fail Fast on Misconfigured SPANNER_EMULATOR_HOST
**The problem:**
If the `SPANNER_EMULATOR_HOST` is set to an unreachable url when trying to create a database, the client hangs for 1 hour. This is how long we keep retrying, without any feedback to the caller.
This causes confusion, usually, making the caller think that the Spanner client is stuck.

**To reproduce the problem:**
Set the `SPANNER_EMULATOR_HOST` environment variable to `http://localhost:<any port>`. Make sure the emulator is NOT running on this url. Run the following:
```java
DatabaseAdminClient adminClient = spanner.getDatabaseAdminClient();
DatabaseId id = DatabaseId.of("project-name", "instance-name", "database-name");

// Hangs on this call
OperationFuture<Database, CreateDatabaseMetadata> op = adminClient.createDatabase(
    id.getInstanceId().getInstance(),
    id.getDatabase(),
    Collections.<String>emptyList()
);
```

**The proposed solution:**
We should quickly fail if the `SPANNER_EMULATOR_HOST` is misconfigured instead of retrying. We could check for this environment variable's value and verify that the host is not reachable. If the host is not reachable we should present a  message to the caller indicating that the variable has been misconfigured. To be on the safe side, we could do this only if the host is localhost.

olavloite:
Check whether the emulator is actually running if the environment variable has been set to localhost.

Fixes #326

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
@@ -42,10 +42,12 @@ import com.google.api.gax.rpc.ServerStream;
 import com.google.api.gax.rpc.StatusCode;
 import com.google.api.gax.rpc.StreamController;
 import com.google.api.gax.rpc.TransportChannelProvider;
+import com.google.api.gax.rpc.UnavailableException;
 import com.google.api.gax.rpc.WatchdogProvider;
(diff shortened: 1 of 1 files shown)


thiagotnunes:
Thanks for the fix @olavloite, will run some tests on Monday and review the PR!

PR #342, opened 2020-07-10
test: fix flaky ITTransactionRetryTest

### LINKED ISSUE #341: ITTransactionRetryTest can be flaky
https://source.cloud.google.com/results/invocations/a1c1f9e6-a4e6-4e24-bfe5-b1906002bd48/targets/github%2Fjava-spanner%2Fgoogle-cloud-spanner%2Ftarget%2Ffailsafe-reports%2Fintegration%2Fcom.google.cloud.spanner.connection.it.ITTransactionRetryTest/tests

olavloite:
Fixes #341

### DIFF
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/it/ITTransactionRetryTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/it/ITTransactionRetryTest.java
@@ -367,9 +367,6 @@ public class ITTransactionRetryTest extends ITAbstractSpannerTest {
 
       connection.commit();
       assertThat(RETRY_STATISTICS.totalSuccessfulRetries >= 3, is(true));
-      assertThat(
-          RETRY_STATISTICS.totalNestedAborts,
-          is(equalTo(RETRY_STATISTICS.totalSuccessfulRetries - 3)));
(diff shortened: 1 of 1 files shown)


olavloite:
> Thanks for fixing this! I just experienced this flaky one on my builds.
> Could you elaborate on why this would sometimes fail?

The test failed because it wrongly assumed that Cloud Spanner would never itself abort the read/write transaction. The test itself simulates that the transaction was aborted three times through the calls to `AbortInterceptor.setProbability(1.0)`. In addition, the test case did take into consideration that a retry attempt itself could abort. These are counted in the statistic `totalNestedAborts`, but it did not take into account that Cloud Spanner itself could abort the initial transaction any number of times. If for example the query at line 351 would return an `Aborted` error, the test case would fail as that would not be counted as a nested abort, but it would be counted as a successful retry.

olavloite:
@skuruppu Do you have any idea why the cla/google check has started to get stuck lately? I've seen it on a couple of different PRs in the past week.

skuruppu:
That is a good question @olavloite. I've noticed that too but haven't figure out why. I've been toggling the bit and then the bot comes in and fixes it. I can ask around in the chat to see if other teams are experiencing it too.

PR #347, opened 2020-07-13
test: rename integration test

olavloite:
Rename the SessionPoolIntegrationTest to IT... so it is actually picked up by the configuration that executes the integration tests.

### DIFF
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/ITSessionPoolIntegrationTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/ITSessionPoolIntegrationTest.java
@@ -19,6 +19,7 @@ package com.google.cloud.spanner;
 import static com.google.common.truth.Truth.assertThat;
 
 import com.google.cloud.grpc.GrpcTransportOptions.ExecutorFactory;
+import com.google.cloud.spanner.SessionPool.PooledSessionFuture;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.CountDownLatch;
@@ -41,7 +42,7 @@ import org.junit.runners.JUnit4;
  */
 @Category(IntegrationTest.class)
 @RunWith(JUnit4.class)
-public class SessionPoolIntegrationTest {
+public class ITSessionPoolIntegrationTest {
   @ClassRule public static IntegrationTestEnv env = new IntegrationTestEnv();
   private static final String TABLE_NAME = "TestTable";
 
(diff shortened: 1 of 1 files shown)


google-cloud-spanner/src/test/java/com/google/cloud/spanner/ITSessionPoolIntegrationTest.java
```diff
@@ -97,28 +98,28 @@ public ScheduledExecutorService get() {
...
+      assertThat(session.get()).isNotNull();
     }
 
-    try (Session session = pool.getReadSession()) {
-      assertThat(session).isNotNull();
-      Session session2 = pool.getReadSession();
-      assertThat(session2).isNotNull();
+    try (PooledSessionFuture session = pool.getReadSession()) {
+      assertThat(session.get()).isNotNull();
+      PooledSessionFuture session2 = pool.getReadSession();
```
thiagotnunes, line 107: nit: move `session2` declaration into the `try` with resources block

google-cloud-spanner/src/test/java/com/google/cloud/spanner/ITSessionPoolIntegrationTest.java
```diff
@@ -97,28 +98,28 @@ public ScheduledExecutorService get() {
 
   @Test
   public void sessionCreation() {
-    try (Session session = pool.getReadSession()) {
-      assertThat(session).isNotNull();
+    try (PooledSessionFuture session = pool.getReadSession()) {
+      assertThat(session.get()).isNotNull();
```
thiagotnunes, line 102: WDYT about adding timeouts to the `get` calls for the futures (in order not to block indefinitely)?

olavloite: I would prefer not do that, as the client library also calls `PooledSessionFuture.get()` without a timeout, and I would prefer to keep these consistent with the behavior of the client library itself. Instead, the test case itself will eventually timeout if the calls should block indefinitely.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/ITSessionPoolIntegrationTest.java
olavloite: Done.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/ITSessionPoolIntegrationTest.java
thiagotnunes: Cool, thanks for the explanation!

Issue #283, opened 2020-06-18
Run samples test on both Java 8 and Java 11


olavloite:
The Spanner sample tests should be executed on both Java 8 and Java 11, as some dependencies are only included in Java 9 and higher. This has caused broken builds in Java 9 and higher to pass without notice.

See also #280

olavloite:
@chingor13 Is this something you could help with? I do not know how to add an additional Kokoro build to execute the Samples test on both Java 8 and Java 11.

skuruppu:
I can help out with this. I'm trying to understand the Kokoro configs and will get back to you once I learn how this is setup.

thiagotnunes:
@skuruppu @olavloite the sample test badges are not shown in the README file as well, should we add them or is that going to be confusing?

olavloite:
@thiagotnunes Those test badges are automatically generated for all Java clients for the different Google Cloud projects here: https://github.com/googleapis/synthtool/tree/master/synthtool/gcp/templates/java_library

So if we were to change that, we would have to change it for all Java clients.

I don't feel very strongly either way about this.

thiagotnunes:
Got it, thanks @olavloite. I will refrain from changing the badges now due to the impact of the change. I will follow up on this decision with @skuruppu.

olavloite:
@thiagotnunes Can this issue be closed now that #345 has been merged?

thiagotnunes:
@olavloite not yet, I still need to do something in the backend to enable the tests. Hopefully today it will get sorted out and I will close #345.

thiagotnunes:
We should now see the samples tests for java8 and java11 being run in kokoro

PR #348, opened 2020-07-14
refactor: normalises samples config names

### LINKED ISSUE #280: Java [9, ...) builds were broken but not picked up by tests

thiagotnunes:
This PR renames the java specific sample config files for kokoro, so that they are the same as it is done in the BigQuery java client. The filename pattern is: java<version>-samples.cfg (see examples here https://github.com/googleapis/java-bigquery/tree/master/.kokoro/nightly).

Fixes #280 and #283

### DIFF
--- a/.kokoro/presubmit/java8-samples.cfg
+++ b/.kokoro/presubmit/java8-samples.cfg
@@ -1,0 +1,34 @@
+# Format: //devtools/kokoro/config/proto/build.proto
+
+# Configure the docker image for kokoro-trampoline.
+env_vars: {
+  key: "TRAMPOLINE_IMAGE"
+  value: "gcr.io/cloud-devrel-kokoro-resources/java8"
+}
+
+env_vars: {
+  key: "JOB_TYPE"
+  value: "samples"
+}
+
+# TODO: remove this after we've migrated all tests and scripts
+env_vars: {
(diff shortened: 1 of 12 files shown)


.kokoro/presubmit/java8-samples.cfg
```diff
@@ -26,7 +32,3 @@ env_vars: {
   value: "java-docs-samples-service-account"
 }
 
-env_vars: {
```
thiagotnunes, line 29: Please ignore deletion, I do not know why git thinks we moved this file from `.../nightly/samples/common.cfg` to `.../presubmit/java8-samples.cfg`

PR #360, opened 2020-07-20
fix: retry pdml transaction on EOS internal error

thiagotnunes:
The Spanner server can return an internal error closing the connection stream. When this happens an `InternalException` is thrown with a specific `EOS` message. We are seeing this specially in long running PDML tasks.

This case is now retried by the client, either by resuming the existing transaction (when there is a resume token) or by starting a new transaction.


google-cloud-spanner/src/main/java/com/google/cloud/spanner/PartitionedDMLTransaction.java
```diff
@@ -55,23 +56,6 @@
     this.rpc = rpc;
   }
 
-  private ByteString initTransaction() {
```
thiagotnunes, line 58: This method was just moved down (after the public methods).

google-cloud-spanner/src/main/java/com/google/cloud/spanner/PartitionedDMLTransaction.java
```diff
@@ -127,20 +111,24 @@ long executeStreamingPartitionedUpdate(final Statement statement, final Duration
                 }
               }
               break;
-            } catch (UnavailableException e) {
+            } catch (UnavailableException | InternalException e) {
```
thiagotnunes, line 114: In this catch block we retry on EOS exception.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/PartitionedDMLTransaction.java
```diff
@@ -174,4 +162,27 @@ public void invalidate() {
...
+          ErrorCode.INTERNAL,
+          "Failed to init transaction, missing transaction id\n" + session.getName());
+    }
+    return txn.getId();
+  }
+
+  private boolean shouldResumeOrRestartTransaction(Exception e) {
+    return e instanceof UnavailableException
+        || (e instanceof InternalException
+            && e.getMessage().contains("Received unexpected EOS on DATA frame from server"));
```
thiagotnunes, line 186: This is hacky, but unfortunately we do not get a specific exception for this error, so we have to proxy through the exception message.

skuruppu:
@thiagotnunes I'm going to hold off on reviewing this until I better understand why this error happens. Hope it's ok to wait a couple of days until the Spanner team has had a chance to look at the issue.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/PartitionedDMLTransaction.java
olavloite: The same error is also retried for streaming queries, and this check could probably use the existing check in `SpannerExceptionFactory`: https://github.com/googleapis/java-spanner/blob/b2641007c3ef77765ab8d98fba02bc6039321c46/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerExceptionFactory.java#L233. It's still just as hacky, but it would keep it confined to one method.

thiagotnunes:
Thanks for the feedback @olavloite, I did not know about that functionality! I will make a change in this PR to use that.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerExceptionFactory.java
```diff
@@ -257,35 +256,8 @@ private static boolean hasCauseMatching(
   }
 
   private static class Matchers {
-    static final Predicate<Throwable> isRetryableInternalError =
```
thiagotnunes, line 260: I extracted these predicates into their own classes that are now exposed to the `PartitionedDmlTransaction`. This was the cleanest / lowest impact way I could find to de-duplicate the EOS retryable logic.

thiagotnunes:
@olavloite @skuruppu do you mind giving the PR another look? I did the following changes:
- Extracted predicates from the `SpannerExceptionFactory` in order to use them in the `PartitionedDmlTransaction`, thus removing the EOS logic duplication.
- Changes the `PartitionedDmlTransaction` class to use the extracted predicate.
- Added tests for the extracted classes.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/IsRetryableInternalError.java
```diff
@@ -0,0 +1,53 @@
...
+  private static final String CONNECTION_CLOSED_ERROR_MESSAGE =
+      "Connection closed with unknown cause";
+  private static final String EOS_ERROR_MESSAGE =
+      "Received unexpected EOS on DATA frame from server";
+
+  @Override
+  public boolean apply(Throwable cause) {
+    if (isInternalError(cause)) {
+      if (cause.getMessage().contains(HTTP2_ERROR_MESSAGE)) {
+        // See b/25451313.
```
skuruppu, line 36: Here and below, if we could remove the references to internal issues that would be great. I don't know why they were there before. Folks working on this repo are unlikely to have access to those so no point exposing them.

thiagotnunes: No worries, will tackle this in a following PR.

olavloite:
@thiagotnunes Thanks for extracting the predicates for the retryable internal errors.
I have no further comments, this looks good to me.

PR #361, opened 2020-07-21
refactor: removes double loop of pdml transaction

thiagotnunes:
Refactors the PDML transaction class to have a single loop for resumes / retries.
I tried to simplify as much as possible on the retry logic we were doing on the transaction. 

`<rant>`
Unfortunately with Java 7 it is hard to improve this logic by not using exceptions for that without having to write lots of code. I could not also come up with a good immutable approach that would have better readability.
`</rant>`


thiagotnunes:
Since this contains a fix for retrying `InternalException - EOS`, we will wait until the spanner team has reviewed the code before merging.
@skuruppu I could also remove this fix, so that we can merge this right away.

skuruppu:
Do you mean that this also contains the change from #360? If that's the case, yeah it would be ideal if we decouple the two fixes.

thiagotnunes:
@skuruppu will remove the fix from here

thiagotnunes:
@skuruppu the fix is removed from this PR

google-cloud-spanner/src/main/java/com/google/cloud/spanner/PartitionedDmlTransaction.java
```diff
@@ -0,0 +1,196 @@
+/*
+ * Copyright 2019 Google LLC
```
skuruppu, line 2: nit: 2020

thiagotnunes: Will fix on the next PR (#360).

PR #370, opened 2020-07-28
test: add test for blocked BeginTransaction

olavloite:


olavloite:
Test was only needed to show the behavior, it will not be added to the code base.

PR #373, opened 2020-08-03
samples: migrate samples from GoogleCloudPlatform/java-docs-samples /spanner/cloud-client

chingor13:


skuruppu:
> @skuruppu I've updated the samples and pom files so these can be added to the client library repository. The `TracingSample` requires a number of additional dependencies, which means that the pom files also include a number of non-Spanner dependencies. Is this something that we want to keep that way?

Really sorry for the late reply. It looks like the pom files that were modified were specific to the samples. Since the samples won't be packaged up with the client library, I don't think it's a problem that they include non-Spanner dependencies. But let me know if I  misunderstood something.

olavloite:
> It looks like the pom files that were modified were specific to the samples. Since the samples won't be packaged up with the client library, I don't think it's a problem that they include non-Spanner dependencies. But let me know if I misunderstood something.

That is correct, and these dependencies are also imported in the java-docs-samples repo for the samples there, so this shouldn't really be a problem. The only issue that we have had with it in the past, is that as these dependencies are not managed by the `bom`, they do cause some dependency conflict issues sometimes. That is not really related to moving these samples to this repo, but it might be something we should look into later to prevent sudden build failures when upgrading the client library.

lesv:
@googlebot I fixed it.

skuruppu:
@chingor13 do we need a "googlebot I fixed it" from everyone who has a commit in this PR?

chingor13:
For the CLA, we can manually flip it because the CLA was signed for the original commits that are being migrated.

Ideally, let's squash Knut's test fix commits, and use a rebase/merge so we preserve each commit in history for the samples

thiagotnunes:
@chingor13 it seems the only thing missing is the CLA for the commit authors (can not merge without it). Do you know what could we do to fix this?

chingor13:
We are ok to flip the CLA bot as these were already committed to the java-doc-samples repo

thiagotnunes:
@chingor13 sorry, I don't know what you mean by "flip the CLA bot". What should I do to fix this one? (btw, I can not merge, it is disabled because of the CLA errors).

PR #377, opened 2020-08-06
test: skip test on emulator

olavloite:
Skip GapicSpannerRpcTest when an emulator host has been set, as the test interferes with the check whether the emulator is running.

### DIFF
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpcTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpcTest.java
@@ -20,6 +20,7 @@ import static com.google.common.truth.Truth.assertThat;
 import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.Assume.assumeTrue;
 
 import com.google.api.core.ApiFunction;
 import com.google.auth.oauth2.AccessToken;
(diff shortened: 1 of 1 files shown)


skuruppu:
@olavloite I would like to understand this more. Why is it interfering with the check?

olavloite:
@skuruppu 
1. This test connects to a [local mock server using TCP/IP](https://github.com/googleapis/java-spanner/blob/c03b3ec25e8964338a23b7198443e29147f5df42/google-cloud-spanner/src/test/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpcTest.java#L148) instead of one that uses a `LocalChannelProvider`.
1. The client connects to this mock server by calling [setHost(String)](https://github.com/googleapis/java-spanner/blob/c03b3ec25e8964338a23b7198443e29147f5df42/google-cloud-spanner/src/test/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpcTest.java#L358) on `SpannerOptions` to a value in the form `http://localhost:port`.
1. This setting will be overridden by the environment variable `SPANNER_EMULATOR_HOST`, as the environment variable takes precedence above this setting. This again means that this test case will try to connect to the emulator instead of the mock server.
1. This will also [invoke the check whether the emulator is actually running](https://github.com/googleapis/java-spanner/blob/c03b3ec25e8964338a23b7198443e29147f5df42/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java#L425) on `localhost:port`, and if the emulator has not yet started, it will cause an error and test failure.
1. If it is started, it will allow the test case to proceed, but it will use the emulator instead of the mock server, and then cause a test failure at a later moment.

PR #379, opened 2020-08-10
feat: support setting timeout per RPC

### LINKED ISSUE #378: feat: Support timeouts per RPC invocation
The Spanner client allows a user to set custom timeouts while creating a `SpannerOptions` instance, but these timeouts are static  and are applied to all invocations of the RPCs. Sometimes applying a custom timeout to a specific RPC invocation is desirable.

olavloite:
The Spanner client allows a user to set custom timeouts while creating a SpannerOptions instance, but these timeouts are static and are applied to all invocations of the RPCs. This change introduces the possibility to set custom timeouts and other call options on a per-RPC basis.

This feature will also be very useful for the Async Connection API, as it will allow applying the statement timeouts for async calls on a per RPC-invocation basis.

Fixes #378

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
@@ -18,9 +18,11 @@ package com.google.cloud.spanner;
 
 import com.google.api.core.ApiFunction;
 import com.google.api.gax.core.ExecutorProvider;
+import com.google.api.gax.grpc.GrpcCallContext;
 import com.google.api.gax.grpc.GrpcInterceptorProvider;
 import com.google.api.gax.longrunning.OperationTimedPollAlgorithm;
 import com.google.api.gax.retrying.RetrySettings;
+import com.google.api.gax.rpc.ApiCallContext;
 import com.google.api.gax.rpc.TransportChannelProvider;
 import com.google.cloud.NoCredentials;
 import com.google.cloud.ServiceDefaults;
@@ -29,6 +31,8 @@ import com.google.cloud.ServiceRpc;
 import com.google.cloud.TransportOptions;
 import com.google.cloud.grpc.GrpcTransportOptions;
 import com.google.cloud.spanner.Options.QueryOption;
+import com.google.cloud.spanner.SpannerOptions.CallContextConfigurator;
+import com.google.cloud.spanner.SpannerOptions.SpannerCallContextTimeoutConfigurator;
 import com.google.cloud.spanner.admin.database.v1.DatabaseAdminSettings;
 import com.google.cloud.spanner.admin.database.v1.stub.DatabaseAdminStubSettings;
 import com.google.cloud.spanner.admin.instance.v1.InstanceAdminSettings;
@@ -44,11 +48,15 @@ import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
+import com.google.spanner.v1.ExecuteSqlRequest;
 import com.google.spanner.v1.ExecuteSqlRequest.QueryOptions;
+import com.google.spanner.v1.SpannerGrpc;
 import io.grpc.CallCredentials;
 import io.grpc.CompressorRegistry;
+import io.grpc.Context;
 import io.grpc.ExperimentalApi;
 import io.grpc.ManagedChannelBuilder;
+import io.grpc.MethodDescriptor;
 import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URL;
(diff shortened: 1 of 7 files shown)


hengfengli:
@olavloite Do you have any idea why `ci / clirr` fails?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerRpc.java
```diff
@@ -287,8 +287,6 @@ StreamingCall read(
   ApiFuture<ResultSet> executeQueryAsync(
       ExecuteSqlRequest request, @Nullable Map<Option, ?> options);
 
-  ResultSet executePartitionedDml(ExecuteSqlRequest request, @Nullable Map<Option, ?> options);
```
hengfengli, line 290: Why do we remove this? This seems to be related to the failed test: 

```bash
[ERROR] 7002: com.google.cloud.spanner.spi.v1.GapicSpannerRpc: Method 'public com.google.spanner.v1.ResultSet executePartitionedDml(com.google.spanner.v1.ExecuteSqlRequest, java.util.Map)' has been removed
[ERROR] 7002: com.google.cloud.spanner.spi.v1.SpannerRpc: Method 'public com.google.spanner.v1.ResultSet executePartitionedDml(com.google.spanner.v1.ExecuteSqlRequest, java.util.Map)' has been removed
```

google-cloud-spanner/src/test/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpcTest.java
```diff
@@ -336,11 +367,93 @@ public void testNoCallCredentials() {
...
+          @Override
+          public <ReqT, RespT> ApiCallContext configure(
+              ApiCallContext context, ReqT request, MethodDescriptor<ReqT, RespT> method) {
+            // Only configure a timeout for the ExecuteSql method when the request has a ,
+            // as this method is used for executing DML statements.
+            if (request instanceof ExecuteSqlRequest
+                && method.equals(SpannerGrpc.getExecuteSqlMethod())) {
+              ExecuteSqlRequest sqlRequest = (ExecuteSqlRequest) request;
+              // Sequence numbers are only assigned for DML statements, which means that
+              // this is an udate statement.
```
hengfengli, line 401: `udate` to `update`

google-cloud-spanner/src/test/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpcTest.java
```diff
@@ -336,11 +367,93 @@ public void testNoCallCredentials() {
...
+  @Test
+  public void testCallContextTimeout() {
+    // Create a CallContextConfigurator that uses a variable timeout value.
+    final TimeoutHolder timeoutHolder = new TimeoutHolder();
+    CallContextConfigurator configurator =
+        new CallContextConfigurator() {
+          @Override
+          public <ReqT, RespT> ApiCallContext configure(
+              ApiCallContext context, ReqT request, MethodDescriptor<ReqT, RespT> method) {
+            // Only configure a timeout for the ExecuteSql method when the request has a ,
```
hengfengli, line 395: This sentence is not finished yet.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
```diff
@@ -1548,4 +1550,40 @@ public void testReadDoesNotIncludeStatement() {
...
+  public void testSpecificTimeout() {
+    mockSpanner.setExecuteStreamingSqlExecutionTime(
+        SimulatedExecutionTime.ofMinimumAndRandomTime(10000, 0));
+    final DatabaseClient client =
+        spanner.getDatabaseClient(DatabaseId.of(TEST_PROJECT, TEST_INSTANCE, TEST_DATABASE));
+    Context.current()
+        .withValue(
+            SpannerOptions.CALL_CONTEXT_CONFIGURATOR_KEY,
+            SpannerCallContextTimeoutConfigurator.create()
+                .withExecuteQueryTimeout(Duration.ofNanos(1L)))
```
hengfengli, line 1564: So this given timeout is for every ExecuteQuery inside this context? If two ExecuteQuery calls are inside this context, each of them will be limited by the 1ns timeout, right?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerRpc.java
olavloite: This method is no longer used (instead the client library uses the streaming version), but the removal of it should be moved to a separate PR. I've reverted this change. That should also fix the clirr build failure.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpcTest.java
olavloite: Fixed.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
olavloite: Correct.

olavloite:
> Generally, using Context looks okay to me. My concerns are:
> 
> 1. Handwritten code need to be maintained. E.g., we need to call different timeout setting methods to different RPCs. If we need to add a new RPC, we need to add the timeout and its related methods here.

Yes, that is correct, but I'm afraid that there is no real way of circumventing that in the Java client. The Java client really hides away most gRPC details, especially for the data client, so if we want to support RPC specific timeouts, it will always include some handwritten code. Note that the `SpannerCallContextTimeoutConfigurator` class is only a convenience class to support users when setting a timeout for the data client. It is not strictly necessary, and I deliberately skipped the same convenience class for the admin clients to reduce the amount of handwritten code.

> 2. I wonder if there is a way to rely on the underlying `gax-java` and gapic generated code to implement this feature instead of using Context? E.g.,
> 
> ```java
> client
>     .singleUse()
>     .executeQuery(
>         Statement.of(
>             "SELECT SingerId, FirstName, LastName FROM Singers ORDER BY LastName")
>         ),
>         callOptions,      // This includes timeout and retry policy? 
> )
> ```
> 
> The above code is my imagination according to the Ruby version: https://github.com/googleapis/google-cloud-ruby/blob/4dcf29950885d8ac3dec41d219f06b843c638a8e/google-cloud-spanner/lib/google/cloud/spanner/client.rb#L351-L372

This solution does use `CallOptions` and `gax-java` for [configuring the timeout](https://github.com/googleapis/java-spanner/blob/dd2715702423eedc8c9e907e6ccf1e33876f6074/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java#L1403), but the issue here is that the java client does not expose any of that in its public methods, which is why I chose to let the user specify it as a context value. It would also be possible to do it in a similar way as the Ruby client by allowing the user to specify `CallOptions` for each method, but that would mean changing the public interface of virtually every method to allow the user to supply additional `CallOptions`. That feels like a very large change for what is in the end a feature aimed at advanced users.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpcTest.java
olavloite: Done.

olavloite:
> LGTM. Nits: there is an unresolved comment.

Sorry, missed that one, but fixed now.

thiagotnunes (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -121,6 +129,324 @@
...
+    private SpannerCallContextTimeoutConfigurator() {}
+
+    @Override
+    public <ReqT, RespT> ApiCallContext configure(
+        ApiCallContext context, ReqT request, MethodDescriptor<ReqT, RespT> method) {
+      SpannerMethod spannerMethod = SpannerMethod.valueOf(request, method);
+      if (spannerMethod == null) {
+        return null;
+      }
+      switch (SpannerMethod.valueOf(request, method)) {
```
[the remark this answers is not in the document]
olavloite: That's probably a good idea. We would still need some kind of lookup table for the timeout / method combination (i.e. keeping them in a hash map or something), but it might make it somewhat simpler. I'll do that in a separate PR, as it won't change the public interface of it.

PR #386, opened 2020-08-17
fix: sets the project for the integration tests

thiagotnunes:
The emulator uses the project id in order to list the instance configurations. If the none is set, it uses the project set through the `gcloud` tool, which might mismatch the project being used to create the databases in the integration tests.
Here, I explicitly set the project id to be the same as the one being used to create the databases for the integration tests.

### DIFF
--- a/google-cloud-spanner/pom.xml
+++ b/google-cloud-spanner/pom.xml
@@ -69,6 +69,7 @@
           <systemPropertyVariables>
             <spanner.testenv.config.class>com.google.cloud.spanner.GceTestEnvConfig</spanner.testenv.config.class>
             <spanner.testenv.instance>projects/gcloud-devel/instances/spanner-testing</spanner.testenv.instance>
+            <spanner.gce.config.project_id>gcloud-devel</spanner.gce.config.project_id>
           </systemPropertyVariables>
           <forkedProcessTimeoutInSeconds>3000</forkedProcessTimeoutInSeconds>
         </configuration>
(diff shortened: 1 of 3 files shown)


google-cloud-spanner/pom.xml
```diff
@@ -69,6 +69,7 @@
           <systemPropertyVariables>
             <spanner.testenv.config.class>com.google.cloud.spanner.GceTestEnvConfig</spanner.testenv.config.class>
             <spanner.testenv.instance>projects/gcloud-devel/instances/spanner-testing</spanner.testenv.instance>
+            <spanner.gce.config.project_id>gcloud-devel</spanner.gce.config.project_id>
```
olavloite, line 72: Won't this break running the integration tests locally if you don't have permission to access the `gcloud-devel` project, unless you specify a different value for this system property in your maven command? Currently the integration tests use the Google Cloud project that is specified in your environment in combination with the credentials that have also been configured in your environment.

thiagotnunes: I thought so, but the test instance is already hardcoded to `projects/gcloud-devel/instances/spanner-testing` (see property `spanner.testenv.instance`). I think that the only way you can run the IT tests currently is by overriding the two properties. Maybe we should update the CONTRIBUTING.md file here as well.

olavloite: Yep, you're right. My local config overrides the `spanner.testenv.instance` property. This change would mean that I would also need to override the `spanner.gce.config.project_id` property, but that is not really a big ask when the first one needs to be overridden anyways. As you noted, this might be worth noting in the CONTRIBUTING.md.

thiagotnunes: Thanks, updated the `CONTRIBUTING.md` with the new integration tests instructions.

PR #392, opened 2020-08-19
feat!: async connection API

olavloite:
## Description

This PR adds async methods for the Connection API. Note that the Connection API, including the async methods, still assumes the following:
1. A `Connection` instance is not thread safe.
1. At most one statement will be actively executed at any time.

### Technical Implementation

1. The PR adds `Async` versions for all execute and transaction methods of the `Connection` interface.
1. The `ConnectionImpl` class already has a [single-threaded executor](https://github.com/googleapis/java-spanner/blob/c5fba074cdebac48753e9f90e06960b15b8055b6/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java#L104) that will actually execute the statements on the backend. This was already implemented this way in preparation for a future async version of the API, and because it made it possible to cancel statements and make individual statements have a specific timeout.
1. All Async methods are executed using the executor from 2, and are basically the same as the current synchronous version, except the API no longer waits for the statement to finish, but returns the `Future` that is returned by the executor. Some extra logic is added for retrying statements in case of an `AbortedException`.
1. Statements with a timeout used to implement the timeout by calling `Future.get()` with a timeout. That is now [replaced with the new timeout feature](https://github.com/googleapis/java-spanner/blob/c5fba074cdebac48753e9f90e06960b15b8055b6/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/AbstractBaseUnitOfWork.java#L207) that was recently added to the client library. This makes the timeout feature also usable for asynchronous statements.
1. DDL statements cannot use the new timeout feature in the client library, as these execute RPCs on the `DatabaseAdmin` interface instead of the `Spanner` interface, still implement the timeout by calling `Future.get()` [with a timeout](https://github.com/googleapis/java-spanner/blob/c5fba074cdebac48753e9f90e06960b15b8055b6/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/AbstractBaseUnitOfWork.java#L155).
1. The `UnitOfWork` interface is the base interface for all transactions in the Connection API. All the methods in this interface have been changed from sync to async, and the synchronous methods in the `Connection` interface use the async methods under the hood. This ensures that there is only one implementation of each method. It also ensures that all existing (integration) tests also invoke the new async methods.


olavloite:
CC @skuruppu

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/StatementExecutor.java
```diff
@@ -151,18 +180,6 @@ private static ExecutorService createExecutorService() {
...
   }
 
-  /**
-   * Recreates this {@link StatementExecutor} and its {@link ExecutorService}. This can be necessary
-   * if a statement times out or is cancelled, and it cannot be guaranteed that the statement
-   * execution can be terminated. In order to prevent the single threaded {@link ExecutorService} to
-   * continue to block on the timed out/cancelled statement, a new {@link ExecutorService} is
-   * created.
-   */
-  void recreate() {
```
olavloite, line 161: This is no longer needed thanks to the new feature in the client library for setting a timeout per RPC.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/UnitOfWork.java
```diff
@@ -39,9 +41,11 @@
 
   enum UnitOfWorkState {
     STARTED,
+    COMMITTING,
```
olavloite, line 44: State added as a `commitAsync` will initiate a commit, but return before the outcome of the commit is known. Setting this state is needed in order to know that the following statement should be executed in a new transaction.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java
```diff
@@ -275,7 +282,7 @@ public void run() {
               switch (response) {
                 case DONE:
                   state = State.DONE;
-                  closeDelegateResultSet();
```
olavloite, line 278: The delegate `ResultSet` should not be closed by the callback executor, but only by the executor that fetches and buffers the rows (in the `ProduceRowsCallable` at line 380)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/SpannerPool.java
```diff
@@ -236,14 +237,17 @@ public int hashCode() {
   @GuardedBy("this")
   private final Map<SpannerPoolKey, Long> lastConnectionClosedAt = new HashMap<>();
 
+  private final Ticker ticker;
```
olavloite, line 240: Use a `Ticker` to measure time to make it easier to test.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/StatementExecutor.java
```diff
@@ -129,12 +141,13 @@ public boolean hasDuration() {
           .build();
 
   /** Creates an {@link ExecutorService} for a {@link StatementExecutor}. */
-  private static ExecutorService createExecutorService() {
-    return new ThreadPoolExecutor(
-        1, 1, 0L, TimeUnit.MILLISECONDS, new LinkedBlockingQueue<Runnable>(), THREAD_FACTORY);
+  private static ListeningExecutorService createExecutorService() {
```
olavloite, line 144: Create a `ListeningExecutorService`. The `ListenableFuture`s that are returned by this can be converted to `ApiFuture`s.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/UnitOfWork.java
```diff
@@ -39,9 +41,11 @@
...
     COMMITTED,
     COMMIT_FAILED,
     ROLLED_BACK,
+    RUNNING,
```
olavloite, line 48: This state is added to indicate that a batch is running, but not yet finished, and that all subsequent statements should not be included in the batch.

thiagotnunes (approved):
LGTM

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java
```diff
@@ -89,7 +91,7 @@ private State(boolean shouldStop) {
   private final BlockingDeque<Struct> buffer;
   private Struct currentRow;
   /** The underlying synchronous {@link ResultSet} that is producing the rows. */
```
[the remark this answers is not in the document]
olavloite: No, good point. I've changed it to reflect the new situation.

PR #396, opened 2020-08-24
feat: Add experimental DirectPath support

WeiranFang:
Previously DirectPath was enabled by an environment variable. Now since gax-java [v1.57.0](https://github.com/googleapis/gax-java/releases/tag/v1.57.0) DirectPath can be enabled by using `InstantiatingGrpcChannelProvider.Builder.setAttemptDirectPath(true)`.

Some changes are temporary and will be removed once DirectPath goes public using the default spanner endpoint. For now we want to be able to test the DirectPath-enabled endpoint with different client scenarios:
- Client is running on a ipv4-only VM
- Client is running on a ipv6-enabled VM
- Client fallbacks to use normal CFE path

For the purpose of our internal E2E tests, I added a test profile called `spanner-directpath-it` which introduces two new sys props:
- `spanner.attempt_directpath`: this defines whether the test client would call `setAttemptDirectPath(true)` and set the spanner endpoint to use the DirectPath-enabled target `aa423245250f2bbf.sandbox.googleapis.com:443`.
- `spanner.directpath_test_scenario`: this defines which of the three scenarios mentioned above to test. 

And for different scenarios we would have different assertions for the test:
- For ipv4-only VM, we introspect the peer IP addresses and expect they have `DP_IPV4_PREFIX`.
- For ipv6-enabled VM, we introspect the peer IP addresses and expect they have either `DP_IPV4_PREFIX` or `DP_IPV6_PREFIX`.
- For fallback cases, we should allow both DirectPath addresses and CFE addresses.

+cc @chingor13 , @vnorigoog

### DIFF
--- a/google-cloud-spanner/pom.xml
+++ b/google-cloud-spanner/pom.xml
@@ -15,7 +15,9 @@
   </parent>
   <properties>
     <site.installationModule>google-cloud-spanner</site.installationModule>
+    <skipUTs>false</skipUTs>
   </properties>
+
 
   <build>
     <plugins>
@@ -49,6 +51,7 @@
             <id>default-test</id>
             <configuration>
               <excludedGroups>com.google.cloud.spanner.TracerTest,com.google.cloud.spanner.IntegrationTest</excludedGroups>
+              <skipTests>${skipUTs}</skipTests>
             </configuration>
           </execution>
           <execution>
@@ -359,6 +362,27 @@
                 </configuration>
               </execution>
             </executions>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
+    <profile>
+      <id>spanner-directpath-it</id>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-failsafe-plugin</artifactId>
+            <configuration>
+              <systemPropertyVariables>
+                <spanner.testenv.config.class>com.google.cloud.spanner.GceTestEnvConfig</spanner.testenv.config.class>
+                <spanner.testenv.instance>projects/directpath-prod-manual-testing/instances/spanner-testing</spanner.testenv.instance>
+                <spanner.gce.config.project_id>directpath-prod-manual-testing</spanner.gce.config.project_id>
+                <spanner.attempt_directpath>true</spanner.attempt_directpath>
+                <spanner.directpath_test_scenario>ipv4</spanner.directpath_test_scenario>
+              </systemPropertyVariables>
+              <forkedProcessTimeoutInSeconds>3000</forkedProcessTimeoutInSeconds>
+            </configuration>
           </plugin>
         </plugins>
       </build>
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
@@ -320,9 +320,9 @@ public class GapicSpannerRpc implements SpannerRpc {
             // SpannerInterceptorProvider if none is provided
             .setInterceptorProvider(
                 SpannerInterceptorProvider.create(
-                    MoreObjects.firstNonNull(
-                        options.getInterceptorProvider(),
-                        SpannerInterceptorProvider.createDefault()))
+                        MoreObjects.firstNonNull(
+                            options.getInterceptorProvider(),
+                            SpannerInterceptorProvider.createDefault()))
                     .withEncoding(compressorName))
             .setHeaderProvider(mergedHeaderProvider);
 
@@ -333,8 +333,7 @@ public class GapicSpannerRpc implements SpannerRpc {
 
     TransportChannelProvider channelProvider =
         MoreObjects.firstNonNull(
-            options.getChannelProvider(),
-            defaultChannelProviderBuilder.build());
+            options.getChannelProvider(), defaultChannelProviderBuilder.build());
 
     CredentialsProvider credentialsProvider =
         GrpcTransportOptions.setUpCredentialsProvider(options);
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
@@ -223,6 +223,9 @@ public class GapicSpannerRpc implements SpannerRpc {
   private static final int DEFAULT_PERIOD_SECONDS = 10;
   private static final int GRPC_KEEPALIVE_SECONDS = 2 * 60;
 
+  // TODO(weiranf): Remove this temporary endpoint once DirectPath goes to public beta.
+  private static final String DIRECT_PATH_ENDPOINT = "aa423245250f2bbf.sandbox.googleapis.com:443";
+
   private final ManagedInstantiatingExecutorProvider executorProvider;
   private boolean rpcIsClosed;
   private final SpannerStub spannerStub;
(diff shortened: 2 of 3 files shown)


google-cloud-spanner/pom.xml
```diff
@@ -49,6 +51,7 @@
             <id>default-test</id>
             <configuration>
               <excludedGroups>com.google.cloud.spanner.TracerTest,com.google.cloud.spanner.IntegrationTest</excludedGroups>
+              <skipTests>${skipUTs}</skipTests>
```
chingor13, line 54: I only see this being set to `false`. Do we need this configuration?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/GceTestEnvConfig.java
```diff
@@ -51,6 +61,8 @@ public GceTestEnvConfig() {
     double errorProbability =
         Double.parseDouble(System.getProperty(GCE_STREAM_BROKEN_PROBABILITY, "0.0"));
     checkState(errorProbability <= 1.0);
+    boolean attemptDirectPath = Boolean.getBoolean(ATTEMPT_DIRECT_PATH);
```
chingor13, line 64: Should this be
```suggestion
    boolean attemptDirectPath = Boolean.getBoolean(System.getProperty(ATTEMPT_DIRECT_PATH));
```

google-cloud-spanner/src/test/java/com/google/cloud/spanner/GceTestEnvConfig.java
```diff
@@ -41,6 +45,12 @@
   public static final String GCE_CREDENTIALS_FILE = "spanner.gce.config.credentials_file";
   public static final String GCE_STREAM_BROKEN_PROBABILITY =
       "spanner.gce.config.stream_broken_probability";
+  public static final String ATTEMPT_DIRECT_PATH = "spanner.attempt_directpath";
+  public static final String DIRECT_PATH_TEST_SCENARIO = "spanner.directpath_test_scenario";
+
+  // IP address prefixes allocated for DirectPath backends.
+  public static final String DP_IPV6_PREFIX = "2001:4860:8040";
+  public static final String DP_IPV4_PREFIX = "34.126";
```
chingor13, line 53: Will this work outside of GCP? i.e. if you run the system tests with the direct path profile from your local machine?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -300,31 +303,38 @@ public GapicSpannerRpc(final SpannerOptions options) {
...
+            .setInterceptorProvider(
+                SpannerInterceptorProvider.create(
+                    MoreObjects.firstNonNull(
+                        options.getInterceptorProvider(),
+                        SpannerInterceptorProvider.createDefault()))
+                    .withEncoding(compressorName))
+            .setHeaderProvider(mergedHeaderProvider);
+
+    // TODO(weiranf): Set to true by default once DirectPath goes to public beta.
+    if (shouldAttemptDirectPath()) {
```
thiagotnunes, line 330: I think that currently we are mostly enabling features / configurations through environment variables (which should be standardised through all of our clients) or through the `SpannerOptions` class.

I imagine we could add an option like `useDirectPath()` in the `SpannerOptions.Builder` class. I guess that even if we do not use an environment variable as well (and go with the property), it would be nice to concentrated the configuration directly in the `SpannerOptions` class.

Wdyt?

thiagotnunes (commented):
Thanks for adding this change!

google-cloud-spanner/pom.xml
WeiranFang: Hi Jeff, I added this configuration just for our internal test to be able to use `-DskipUTs=true`. Currently we are using `mvn verify -am -pl google-cloud-spanner -B -Penable-integration-tests,spanner-directpath-it -DskipUTs=true -D... ...` for our internal E2E DP tests, and meanwhile I don't want to change the behavior for other tests. Do you think this is the right way for approaching this purpose?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/GceTestEnvConfig.java
WeiranFang: Good point, actually that's what I initially did, but after I checked the [doc](https://docs.oracle.com/javase/8/docs/api/java/lang/Boolean.html#getBoolean-java.lang.String-), it seems this method would call getProperty for us:
> Returns true if and only if the system property named by the argument exists and is equal to the string "true". (Beginning with version 1.0.2 of the JavaTM platform, the test of this string is case insensitive.) A system property is accessible through getProperty, a method defined by the System class.
If there is no property with the specified name, or if the specified name is empty or null, then false is returned.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/GceTestEnvConfig.java
WeiranFang: Good question. The short answer is no. DirectPath is designed for clients calling from GCE VM directly, so we actually require that the client need to use the default service account provided by VM itself (see [InstantiatingGrpcChannelProvider](https://github.com/googleapis/gax-java/blob/master/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java#L253)). If user uses other type of credentials (e.g. set GOOGLE_APPLICATIONS_CREDENTIALS with a key file and run client from outside of GCP), then the client would just use CFE path instead of DirectPath.

For testing purpose, I defined this `directpath_test_scenario` property to differentiate three cases that could happen when client opts in to use DirectPath: 1) client is using DP over IPv4, 2) client is using DP over IPv6 and 3) client uses CFE instead. And for cases like running client from my local machine, it should fall in to scenario 3)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
WeiranFang: Hi Thiago, thanks for the comment! So our original idea is, the DirectPath is a network side feature, it should be agnostic to the end users. Currently we add this `shouldAttemptDirectPath` temporarily so our test can start exercising DirectPath. Once DirectPath is going public, we will remove these temporary checker and set the client to use DirectPath by default. And correct me if I'm wrong, this SpannerOptions seems like a configurable option provided to end users, but in fact we don't actually want users to config DirectPath attempt, which should only be controlled by service owner/SREs via Access Control List. WDYT?

thiagotnunes: That makes sense, thanks for the response.

A couple of questions. on the approach: 
1. Aren't we enabling this to a few customers at first (before being the default)? If so, is the expectation that they would have to execute with the property set?
2. Should we override the endpoint even when the user provides a custom channel provider? I don't know what is best here, but I imagine that it might be confusing if the user executes their program with the property set and a custom channel provider, but do not get the benefits of the direct path.

Thanks!

WeiranFang: Good questions! To answer them:
Actually we would need to remove these temporary checker and enable DirectPath with a client lib release before we engage real customers, because DirectPath should be fully controlled by the ACL config (i.e. initially only a few customers will get ALLOW from ACL check, all others will get DENY). So this property set should only be used by our tests. 

This sandbox endpoint is only used for our testing purposes (as right now DirectPath-ipv4 is only available in certain cell in prod) Once we've done enough tests and DirectPath is fully ready, Spanner will make its default endpoint to be capable of handling DirectPath traffic. So by that time, we will use the same endpoint no matter the client is using DirectPath or Not.

thiagotnunes (approved):

google-cloud-spanner/pom.xml
chingor13: I might add a note somewhere so we don't inadvertently break this.

We might also want to add something like this to our shared setup.

WeiranFang:
Hi @thiagotnunes and @chingor13, is there anything else need to be done before this can be merge?

thiagotnunes:
@WeiranFang nothing from side. Can you merge?

WeiranFang:
Hi @thiagotnunes Actually I don't have access to merge this PR:
> "Only those with write access to this repository can merge pull requests."

 is it OK to add me write access to this repo? Thanks!

thiagotnunes:
@WeiranFang you are right, I will merge it for you and follow up if I can give you write access.

PR #398, opened 2020-08-26
fix: remove potential infinite loop in administrative requests

olavloite:
The automatic retryer for non-idempotent administrative requests did not keep track of the nextPageToken correctly when querying the backend for running operations. This could cause an infinite loop if a resource (database/backup) with the same name had been created repeatedly on the same instance.


thiagotnunes (approved):
Nice catch!

PR #411, opened 2020-09-01
Remove Guava ImmutableList from API surface.

### LINKED ISSUE #410: Using Guava classes in AsyncResultSet exposes it on API surface
The AsyncResultSet interface has two members that return a Future<ImmutableList<>>. and have a Guava Function as a parameter

https://github.com/googleapis/java-spanner/blob/77398f2771d5531b8f703c17df694dc77e42f35f/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSet.java#L206-L207

This exposes these two Guava classes in the API surface of the spanner client library, which it did not before.

Good java practice is to refer to objects by their interfaces (Effective Java (Blosch), item 64)
_"If appropriate interface types exist, then parameters, return values,
variables, and fields should all be declared using interface types."_

So these functions should return Future<List<>> and take a java.util.function.Function

But java.util.function.Function does not appear until JDK8, so Guava Function still needs to be used.

nielm:
Replace com.google.common.collect.ImmutableList
with java.util.List

Fixes #410 ☕️

### DIFF
--- a/google-cloud-spanner/clirr-ignored-differences.xml
+++ b/google-cloud-spanner/clirr-ignored-differences.xml
@@ -227,7 +227,21 @@
     <className>com/google/cloud/spanner/connection/Connection</className>
     <method>* executeQueryAsync(*)</method>
   </difference>
-  
+
+  <!-- Use List interface in Async API (compared to 1.60.0)-->
+  <difference>
+    <differenceType>7006</differenceType>
+    <className>com/google/cloud/spanner/AsyncResultSet</className>
+    <method>com.google.common.collect.ImmutableList toList(com.google.common.base.Function)</method>
+    <to>java.util.List</to>
+  </difference>
+  <difference>
+    <differenceType>7006</differenceType>
+    <className>com/google/cloud/spanner/ForwardingAsyncResultSet</className>
+    <method>com.google.common.collect.ImmutableList toList(com.google.common.base.Function)</method>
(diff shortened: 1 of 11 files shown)


nielm:
@olavloite as he wrote the code I am modifying. 
I don't know if the benefits of changing this API surface outweigh the cost of breaking binary compatibility with older library versions...

nielm:
Updated to only replace ImmutableList -> List to keep JDK 1.7 compatibility.

olavloite:
@nielm You need to add an entry to https://github.com/googleapis/java-spanner/blob/master/google-cloud-spanner/clirr-ignored-differences.xml make the clirr plugin stop complaining. Otherwise this LGTM.

nielm:
> @nielm You need to add an entry to https://github.com/googleapis/java-spanner/blob/master/google-cloud-spanner/clirr-ignored-differences.xml make the clirr plugin stop complaining. Otherwise this LGTM.

Thanks, Knut, Done!

olavloite (approved):

thiagotnunes:
@nielm could you amend your commit to have the "breaking change" footer or "!" in the commit message (https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-description-and-breaking-change-footer). This will make sure that the automation releases a major version.

thiagotnunes (approved):
LGTM

PR #413, opened 2020-09-04
test: add integration tests for numeric

olavloite:
Adds integration tests for `NUMERIC` data type.

### DIFF
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITWriteTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITWriteTest.java
@@ -90,7 +90,7 @@ public class ITWriteTest {
                     + "  BytesArrayValue     ARRAY<BYTES(MAX)>,"
                     + "  TimestampArrayValue ARRAY<TIMESTAMP>,"
                     + "  DateArrayValue      ARRAY<DATE>,"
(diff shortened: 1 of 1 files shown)


olavloite:
@skuruppu Could it be that the emulator does not yet support `NUMERIC`?

skuruppu (approved):

skuruppu:
@olavloite yes, Emulator doesn't yet support NUMERIC :( So you will have to maintain a separate copy of the schema for emulator vs non-emulator tests and the skips for the numeric-specific tests. It's a bit messy but the emulator support is coming soon so we can clean it up quickly.

olavloite:
(The build failure for `Kokoro - Test: Samples Java 8` is an unrelated transient `[ERROR] error reading /root/.m2/repository/com/google/protobuf/protobuf-java/3.12.4/protobuf-java-3.12.4.jar; zip file is empty` error)

olavloite:
> @olavloite yes, Emulator doesn't yet support NUMERIC :( So you will have to maintain a separate copy of the schema for emulator vs non-emulator tests and the skips for the numeric-specific tests. It's a bit messy but the emulator support is coming soon so we can clean it up quickly.

I've made the integration test check whether it is running on the emulator, and based on that, it decides whether or not to include a `NUMERIC` column.

PR #416, opened 2020-09-08
fix: iterate over async result set in sync

olavloite:
Iterating over an AsyncResultSet using a standard `while (rs.next())` loop would fail after the last row was consumed.


thiagotnunes (approved):
LGTM

PR #421, opened 2020-09-10
fix: fix aborted handling of batchUpdateAsync

olavloite:
If a transaction is aborted before a `batchUpdateAsync` call, the RPC will return an `Aborted` status. This status was not translated to a `com.google.cloud.spanner.AbortedException`, but was instead returned as a `com.google.api.gax.rpc.AbortedException`, which could cause retry handling to fail.

Fixes https://github.com/GoogleCloudPlatform/java-docs-samples/issues/3634


thiagotnunes (approved):
LGTM

PR #423, opened 2020-09-12
feat: add lazy initializer

olavloite:
Adds a `LazyInitializer` for `Spanner` instances that can be used in environments where it is desirable to initialize one instance once, and reuse this for multiple purposes, such as for example Google Cloud Functions.

This change would allow us to simplify the [Google Cloud Functions Java sample](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/functions/spanner/src/main/java/functions/HelloSpanner.java).

Towards https://github.com/GoogleCloudPlatform/java-docs-samples/issues/2862.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractLazyInitializer.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractLazyInitializer.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractLazyInitializer.java
```diff
@@ -0,0 +1,55 @@
...
+ * limitations under the License.
+ */
+
+package com.google.cloud.spanner;
+
+/**
+ * Generic {@link AbstractLazyInitializer} for any heavy-weight object that might throw an exception
+ * during initialization. The underlying object is initialized at most once.
+ */
+public abstract class AbstractLazyInitializer<T> {
```
thiagotnunes, line 23: Could we use composition instead of inheritance (pass in the initialiser function instead of overriding it)?

olavloite: That would be possible, but as we have to support Java 7, that would mean that we would have to use `com.google.cloud.base.Function` in the public interface. I think that it is more idiomatic to use inheritance than composition for Java 7 (once we drop Java 7 support, I would agree that using composition would be a lot nicer in this case).

thiagotnunes (approved):
LGTM

PR #424, opened 2020-09-14
fix: improve numeric range checks

olavloite:
Adds additional checks on `NUMERIC` values before sending them to Cloud Spanner in order to return sensible error messages as soon as possible, and adds documentation on how to handle `NUMERIC` values with trailing zeros in the fractional part.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/Value.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/Value.java
@@ -123,11 +123,37 @@ public abstract class Value implements Serializable {
   }
 
   /**
-   * Returns a {@code NUMERIC} value.
+   * Returns a {@code NUMERIC} value. The valid value range for the whole component of the {@link
+   * BigDecimal} is from -9,999,999,999,999,999,999,999,999 to +9,999,999,999,999,999,999,999,999
+   * (both inclusive), i.e. the max length of the whole component is 29 digits. The max length of
+   * the fractional part is 9 digits. Trailing zeros in the fractional part are not considered and
+   * will be lost, as Cloud Spanner does not preserve the precision of a numeric value.
+   *
+   * <p>If you set a numeric value of a record to for example 0.10, Cloud Spanner will return this
+   * value as 0.1 in subsequent queries. Use {@link BigDecimal#stripTrailingZeros()} to compare
+   * inserted values with retrieved values if your application might insert numeric values with
+   * trailing zeros.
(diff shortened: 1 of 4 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/Value.java
```diff
@@ -123,11 +123,37 @@ public static Value float64(double v) {
...
+   * Returns a {@code NUMERIC} value. The valid value range for the whole component of the {@link
+   * BigDecimal} is from -9,999,999,999,999,999,999,999,999 to +9,999,999,999,999,999,999,999,999
+   * (both inclusive), i.e. the max length of the whole component is 29 digits. The max length of
+   * the fractional part is 9 digits. Trailing zeros in the fractional part are not considered and
+   * will be lost, as Cloud Spanner does not preserve the precision of a numeric value.
+   *
+   * <p>If you set a numeric value of a record to for example 0.10, Cloud Spanner will return this
+   * value as 0.1 in subsequent queries. Use {@link BigDecimal#stripTrailingZeros()} to compare
+   * inserted values with retrieved values if your application might insert numeric values with
+   * trailing zeros.
```
skuruppu, line 135: Is this all necessary because BigDecimal has arbitrary precision but NUMERIC doesn't?

Just as an FYI, we chose not to do this explicit check in other clients just because we wanted to make sure that the backend threw errors in case they decide to change the precision later on. In practice, they wouldn't change it without notifying us so it's not a big deal.

olavloite: I added the check because the error messages from the backend are not very clear (at the moment). If you send an out-of-range `NUMERIC` value to Cloud Spanner, it will respond with a generic `io.grpc.StatusRuntimeException: INVALID_ARGUMENT: Invalid value for bind parameter p1: Expected NUMERIC`. That already confused one of the engineers at the Java frameworks team as he was sending a `NUMERIC` value, but it was out of range.

thiagotnunes (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Value.java
skuruppu: Thanks @olavloite. @thiagotnunes has informed the backend team asking them to return a more meaningful error. But in the meantime, it's ok for us to do this in the client. We can remove it when the backend returns a better error message.

PR #425, opened 2020-09-14
fix: do not close delegate rs in callback runnable

olavloite:
`AsyncResultSetImpl` should not close its delegate `ResultSet` in the `CallbackRunnable`, but should leave that to the `ProduceRowsCallable` and instead only set the flag `cursorReturnedDoneOrException`. Otherwise, the main loop that fetches and produces rows could get stuck in an infinite loop.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java
@@ -275,7 +275,7 @@ class AsyncResultSetImpl extends ForwardingStructReader implements ListenableAsy
               switch (response) {
                 case DONE:
                   state = State.DONE;
-                  closeDelegateResultSet();
+                  cursorReturnedDoneOrException = true;
                   return;
                 case PAUSE:
                   state = State.PAUSED;
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/AsyncResultSetImplStressTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/AsyncResultSetImplStressTest.java
@@ -285,6 +285,47 @@ public class AsyncResultSetImplStressTest {
   }
 
   @Test
+  public void returnDoneBeforeEnd() throws Exception {
+    ExecutorProvider executorProvider = SpannerOptions.createDefaultAsyncExecutorProvider();
+    final Random random = new Random();
+    for (Executor executor :
+        new Executor[] {
+          MoreExecutors.directExecutor(), createExecService(), createExecService(32)
+        }) {
+      for (int bufferSize = 1; bufferSize < resultSetSize * 2; bufferSize *= 2) {
+        for (int i = 0; i < TEST_RUNS; i++) {
+          try (AsyncResultSetImpl impl =
+              new AsyncResultSetImpl(executorProvider, createResultSet(), bufferSize)) {
+            ApiFuture<Void> res =
+                impl.setCallback(
+                    executor,
+                    new ReadyCallback() {
+                      @Override
+                      public CallbackResponse cursorReady(AsyncResultSet resultSet) {
+                        switch (resultSet.tryNext()) {
+                          case DONE:
+                            return CallbackResponse.DONE;
+                          case NOT_READY:
+                            return random.nextBoolean()
+                                ? CallbackResponse.DONE
+                                : CallbackResponse.CONTINUE;
+                          case OK:
+                            return random.nextInt(resultSetSize) <= 2
+                                ? CallbackResponse.DONE
+                                : CallbackResponse.CONTINUE;
+                          default:
+                            throw new IllegalStateException();
+                        }
+                      }
+                    });
+            assertThat(res.get(10L, TimeUnit.SECONDS)).isNull();
+          }
+        }
+      }
+    }
+  }
+
+  @Test
   public void pauseResume() throws Exception {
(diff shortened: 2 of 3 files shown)


thiagotnunes:
I would like to understand a bit more about this before approving. Could you point me to where the infinite loop would be?

olavloite:
> I would like to understand a bit more about this before approving. Could you point me to where the infinite loop would be?

Yeah, I should have included that in the initial commit. The problem is as follows:

1. The `ProduceRowsCallable` has two loops.
2. The [first loop](https://github.com/googleapis/java-spanner/blob/fd63673eafdf63b22ffb1013fcf6104717d1513d/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java#L335) iterates over the underlying `ResultSet` and fetches and buffers the rows, and initiates callbacks if one is needed. It will stop if the underlying `ResultSet` does not have any further rows, if the `AsyncResultSet` [reaches a stoppable state](https://github.com/googleapis/java-spanner/blob/fd63673eafdf63b22ffb1013fcf6104717d1513d/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java#L338) (DONE, CANCELLED), or [if an error occurs](https://github.com/googleapis/java-spanner/blob/fd63673eafdf63b22ffb1013fcf6104717d1513d/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java#L367).
3. The [second loop](https://github.com/googleapis/java-spanner/blob/fd63673eafdf63b22ffb1013fcf6104717d1513d/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java#L382) initiates callbacks for the remaining rows in the buffer until the variable `cursorReturnedDoneOrException` [is true](https://github.com/googleapis/java-spanner/blob/fd63673eafdf63b22ffb1013fcf6104717d1513d/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java#L388). This will be set to true in the following cases:
3.1. If the `AsyncResultSet` [is cancelled](https://github.com/googleapis/java-spanner/blob/fd63673eafdf63b22ffb1013fcf6104717d1513d/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java#L199).
3.2. If an [error occurred](https://github.com/googleapis/java-spanner/blob/fd63673eafdf63b22ffb1013fcf6104717d1513d/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java#L203) during a callback.
3.3. If `tryNext()` [returned DONE to the callback](https://github.com/googleapis/java-spanner/blob/fd63673eafdf63b22ffb1013fcf6104717d1513d/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java#L214).
3.4. It was however NOT set if the [callback returned DONE](https://github.com/googleapis/java-spanner/blob/fd63673eafdf63b22ffb1013fcf6104717d1513d/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java#L278) before all rows had been consumed.
4. The second loop will only finish if `cursorReturnedDoneOrException` set to true, which is where the infinite loop will occur.

This problem was introduced by https://github.com/googleapis/java-spanner/commit/fb26abe692f8eeaa208c4421887547d1bc575f9c#diff-df5cefcf33d6f54a3dbfc8ac52ee50f3. Before that change, the loop would also finish if the `state` of the `AsyncResultSet` was `DONE`, but that was removed as it could cause the `AsyncResultSet` to miss a callback if it was cancelled. When an `AsyncResultSet` is cancelled, the contract is that the callback will be called one last time and that the `tryNext()` method will throw a `SpannerException` with code `CANCELLED` to ensure that the client application is notified and can handle it as an exception.

thiagotnunes:
Thanks for the explanation @olavloite, I think I understand it a bit better now.

thiagotnunes (approved):

thiagotnunes:
Maybe unrelated to this PR, but shouldn't we be checking the `state.shouldStop` after this line as well: https://github.com/googleapis/java-spanner/blob/fd63673eafdf63b22ffb1013fcf6104717d1513d/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java#L358?

olavloite:
> Maybe unrelated to this PR, but shouldn't we be updating the `stop` variable with the `state.shouldStop` inside a lock after this line as well:
> 
> https://github.com/googleapis/java-spanner/blob/fd63673eafdf63b22ffb1013fcf6104717d1513d/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java#L358

No, I would say that that would be superfluous. It has always been updated almost directly before that:

1. [Here](https://github.com/googleapis/java-spanner/blob/fd63673eafdf63b22ffb1013fcf6104717d1513d/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java#L338), and if one of the following `if` or `while` conditions are false, it falls right through to line 358.
1. [Here](https://github.com/googleapis/java-spanner/blob/fd63673eafdf63b22ffb1013fcf6104717d1513d/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java#L355) if it enters the `while` loop to wait for buffer capacity.

The worst that could happen is that one extra row is fetched and buffered. The `startCallbackIfNecessary()` will check whether it is allowed to do a callback (which is not equal to whether the further traversal and buffering results should continue, as the callback should also be called if the state is `PAUSED`, but it should continue to buffer results until the buffer is full).

PR #426, opened 2020-09-15
fix: uses old version of gax-grpc method

thiagotnunes:
Internally, we are using version 1.47.1 of the gax-grpc library, which does not provide a method currently being used (setExecutor).

In here we use a method that is available in 1.47.1 and achieves the same goal (setExecutorProvider).


skuruppu (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -317,7 +316,11 @@ public GapicSpannerRpc(final SpannerOptions options) {
             .setMaxInboundMessageSize(MAX_MESSAGE_SIZE)
             .setMaxInboundMetadataSize(MAX_METADATA_SIZE)
             .setPoolSize(options.getNumChannels())
-            .setExecutor(executorProvider.getExecutor())
+
+            // Before updating this method to setExecutor, please make sure to verify the version
+            // of gax-grpc used internally. At the moment we are using 1.47.1, which does not
+            // support the setExecutor variant.
```
skuruppu, line 322: This comment may not give enough context for folks outside of Google. So instead reword to something like:

"Before updating this method to setExecutor, please verify with a code owner on the lowest version of gax-grpc that needs to be supported. Currently v1.47.17, which doesn't support the setExecutor variant."

thiagotnunes: Done.

PR #427, opened 2020-09-15
feat: adds benchmark utils class

thiagotnunes:
Adds a class specifically for benchmarking tasks, which should NOT be used in production.

The only functionality currently present is to wait for the session pool to be ready and write prepared.


thiagotnunes:
Will create a branch in the repository itself.

PR #430, opened 2020-09-15
fix: use credentials key in pool

olavloite:
The `SpannerPool` uses among other fields the `credentials` of a `ConnectionOptions` to determine whether an existing `Spanner` instance can be reused or not. Certain credentials instances, most notably `UserCredentials`, are mutable and the result of `equals(Object)` and `hashCode()` of an instance may change during its lifetime. This could cause the `SpannerPool` to inadvertently close a `Spanner` instance that is still in use.

This change replaces the usage of the actual credentials as part of the pool key with a key that is based on the method that was used to determine the credentials. This ensures that if two connections request a `Spanner` instance using the same `credentialsUrl`, that they will be considered equal. The same also applies to two connections that both do not set any specific credentials options and use the application default credentials.

Fixes https://github.com/googleapis/java-spanner-jdbc/issues/206 and https://github.com/GoogleCloudPlatform/google-cloud-spanner-hibernate/issues/202

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
@@ -384,6 +384,7 @@ public class ConnectionOptions {
   private final String uri;
   private final String credentialsUrl;
   private final String oauthToken;
+  private final Credentials fixedCredentials;
 
   private final boolean usePlainText;
   private final String host;
@@ -413,6 +414,7 @@ public class ConnectionOptions {
         builder.credentialsUrl != null ? builder.credentialsUrl : parseCredentials(builder.uri);
     this.oauthToken =
         builder.oauthToken != null ? builder.oauthToken : parseOAuthToken(builder.uri);
+    this.fixedCredentials = builder.credentials;
     // Check that not both credentials and an OAuth token have been specified.
     Preconditions.checkArgument(
         (builder.credentials == null && this.credentialsUrl == null) || this.oauthToken == null,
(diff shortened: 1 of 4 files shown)


olavloite:
@skuruppu @thiagotnunes Friendly ping for this PR. It fixes a bug in the JDBC driver that can cause a `Pooled has been closed` exception to occur.

thiagotnunes:
hey @olavloite, I think I am missing how does assigning the `credentials` to a new field `fixedCredentials` will prevent mutability (since `equals` will compare the new `CredentialsKey`, which might have a `credential`, which might be mutable).

olavloite:
> hey @olavloite, I think I am missing how does assigning the `credentials` to a new field `fixedCredentials` will prevent mutability (since `equals` will compare the new `CredentialsKey`, which might have a `credential`, which might be mutable).

Thanks for the quick response @thiagotnunes 

Assigning the `credentials` to the new field `fixedCredentials` will not fix the problem of mutability of the `fixedCredentials`. The problem however was that the methods of assigning a `credentials` instance and a `credentialsUrl` to the `ConnectionOptions`  were not separated properly.

You have a number of options for setting the credentials when creating a `ConnectionOptions` instance:
1. You can assign an OAuthToken that you have previously obtained.
1. You can assign a `Credentials` instance.
1. You set set a URL that points to a file that contains the credentials.
1. You can let the client library pick the default credentials of the environment by not setting any of the above options.

Option 2 and 3 internally both set the `ConnectionOptions.credentials` field and when determining whether a `Spanner` instance was present in the pool, this field was used to determine equality. This change separates the two options so that:
1. When a `credentials` instance has been assigned to the `ConnectionOptions` the field `fixedCredentials` will be used for the equality check.
1. When a `credentialsUrl` has been set on the `ConnectionOptions`, the URL will be used for the equality check.

Setting a `credentials` instance directly when creating a `ConnectionOptions` will still suffer from any possible mutability of the credentials instance (or use of mutable fields in its `equals` method). It is however not currently used by any part of the client library or JDBC driver.

Setting a `credentialsUrl` is used by the JDBC driver and will no longer suffer from the possible mutability of the credentials that are chosen.

What also might be worth noting:
* This problem occurs when using `UserCredentials`, as this class calls `super.equals(Object)` as part of its `equals(Object)` method. The super implementation (`OAuth2Credentials`) will compare the current OAuth tokens of this and other.
* This problem does not occur when using `ServiceAccountCredentials`, as this class does not include any mutable fields in the `equals(Object)` implementation, and does not call the super implementation.

thiagotnunes:
@olavloite thanks for the detailed explanation. LGTM.

thiagotnunes (approved):

olavloite:
@thiagotnunes Apparently you have not (yet?) been added as a member of https://github.com/orgs/googleapis/teams/yoshi-java/members, which means that merging this PR is still blocked until a member of that team approves the PR. Maybe one of the maintainers could add you as a member to the team.

thiagotnunes:
@olavloite sorry about that, I have requested membership

elharo (approved):

PR #452, opened 2020-09-23
feat!: Point In Time Recovery (PITR)

thiagotnunes:
Implements the PITR-lite support. 

With this functionality users will be able to set the retention period for their databases (through the `version_retention_period` database field). Other than that, user should be able to specify the timestamp for a consistent copy of a database through the `version_time` field when creating a backup.


google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITPitrUpdateDatabaseTest.java
```diff
@@ -0,0 +1,195 @@
...
+      if (hasVersionRetentionPeriodStatement) {
+        break;
+      }
+    }
+    assertThat(hasVersionRetentionPeriodStatement).isTrue();
+  }
+
+  @Test
+  public void returnsTheVersionRetentionPeriodSetThroughInformationSchema() {
+    final ResultSet rs =
```
olavloite, line 123: This query will list the `OPTION_VALUE` of all databases on the test instance. If this test is executed against an existing instance with multiple databases having a `version_retention_period`, it could fail.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITPitrUpdateDatabaseTest.java
```diff
@@ -0,0 +1,195 @@
...
+        dbClient
+            .singleUse()
+            .executeQuery(
+                Statement.of(
+                    "SELECT OPTION_VALUE AS version_retention_period "
+                        + "FROM INFORMATION_SCHEMA.DATABASE_OPTIONS "
+                        + "WHERE SCHEMA_NAME = '' AND OPTION_NAME = 'version_retention_period'"));
+
+    String versionRetentionPeriod = null;
+    while (rs.next()) {
```
olavloite, line 133: nit: The `ResultSet` is created using a single-use read context and it is fully consumed, so it will be closed, but I would say that it is good practice to always close the `ResultSet` when you are done with it.

olavloite (changes requested):
This looks generally good to me. There is one integration test case that needs a small change.

There are also a number of generated code changes in this PR. I assume that those will be added through a separate PR, and removed from this.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITPitrUpdateDatabaseTest.java
thiagotnunes: I am wondering how could this happen, since I am using the `dbClient` for which we need to specify a database to create. Or maybe I am missing something?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITPitrUpdateDatabaseTest.java
thiagotnunes: Good point, added the `close` call.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITPitrUpdateDatabaseTest.java
olavloite: Sorry, you're completely right. I forgot that the INFORMATION_SCHEMA is always database-specific.

olavloite (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseInfo.java
```diff
@@ -129,6 +155,16 @@ public Timestamp getCreateTime() {
     return createTime;
   }
 
+  /** Returns the version retention period of the database. */
```
olavloite, line 158: nit: Could we have a more elaborate comment on this method? And also for `getEarliestVersionTime`. I don't think everyone will understand what these fields mean.

olavloite (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseInfo.java
thiagotnunes: Added more information to both params, let me know if there are satisfactory.

olavloite (approved):
LGTM

thiagotnunes:
@olavloite Could you take a look at the last 7 commits? I added the missing functionality for PITR (Backup and Restore).

olavloite (approved):
LGTM, with a couple of small (nit) questions.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BackupInfo.java
```diff
@@ -55,6 +67,7 @@
     protected final BackupId id;
     private State state = State.UNSPECIFIED;
     private Timestamp expireTime;
+    private Timestamp versionTime;
```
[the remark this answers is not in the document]
thiagotnunes: I am trying to keep the naming one to one with the proto fields. I did try to add some javadoc in the `setVersionTime` that explains more about the field.

Let me know if this is not clear enough though, and I could make this change.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Database.java
```diff
@@ -118,8 +118,17 @@ public boolean exists() {
...
-    return dbClient.createBackup(
-        instance(), backup.getId().getBackup(), database(), backup.getExpireTime());
+
+    final BackupInfo.Builder backupBuilder =
+        dbClient
+            .newBackupBuilder(backup.getId())
+            .setDatabase(getId())
+            .setExpireTime(backup.getExpireTime());
+    if (backup.getVersionTime() != null) {
+      backupBuilder.setVersionTime(backup.getVersionTime());
```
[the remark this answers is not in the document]
thiagotnunes: Good point, made this change.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BackupInfo.java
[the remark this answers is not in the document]
olavloite: > I am trying to keep the naming one to one with the proto fields.

That is reasonable and also makes it easier for users to search for documentation online.

olavloite:
LGTM.

PR #460, opened 2020-09-25
test: show metadata filter failure

olavloite:
There seems to be a problem with filtering database operations using metadata values, which now causes some tests to fail. This also suddenly started to happen in the NodeJS library, which seems to indicate that it is related to a change in the backend.

Updates #456


olavloite:
@skuruppu There seems to be a problem with filtering (database) operations based on metadata fields, which is now causing some test failures. A similar problem also started around the same time in the NodeJS library (see links in #456 )

This PR just adds a very simple test case that should show the failure.

CC @AVaksman

AVaksman:
Confirm.
Same failure have been noticed in PHP as well.

olavloite:
No longer needed, the tests that were skipped have been re-enabled in #634

PR #461, opened 2020-09-29
fix: fixed visibility of LazyInit methods

olavloite:
The visibility of the two methods in `AbstractLazyInitializer` were simply wrong:
* `initialize()`: Should be `protected` as it is only intended for sub classes to override if needed.
* `get()`: Should be `public` as it is intended for consumers to use to get an instance of the heavy-weight object.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractLazyInitializer.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractLazyInitializer.java
@@ -27,7 +27,7 @@ public abstract class AbstractLazyInitializer<T> {
   private volatile Exception error;
 
   /** Returns an initialized instance of T. */
-  T get() throws Exception {
+  public T get() throws Exception {
     // First check without a lock to improve performance.
     if (!initialized) {
       synchronized (lock) {
@@ -51,5 +51,5 @@ public abstract class AbstractLazyInitializer<T> {
    * Initializes the actual object that should be returned. Is called once the first time an
    * instance of T is required.
    */
-  public abstract T initialize() throws Exception;
+  protected abstract T initialize() throws Exception;
 }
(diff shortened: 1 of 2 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractLazyInitializer.java
```diff
@@ -51,5 +51,5 @@ T get() throws Exception {
    * Initializes the actual object that should be returned. Is called once the first time an
    * instance of T is required.
    */
-  public abstract T initialize() throws Exception;
+  protected abstract T initialize() throws Exception;
```
elharo, line 54: It's probably too late to make this change without a major version bump.

olavloite: I agree that it would normally require a major version bump to make such a change. I would however argue that in this specific case we should consider relaxing that requirement for the following reasons:
* The current implementation marks the `get()` method as package-private, which means that the class can currently only be (purposefully) used by the client library itself, as any outside user would not be able to get an instance of the object that is created.
* Reducing the visibility of the `initialize()` method would break any application that is directly calling the method, but that is not the intention of the method (and would also make no sense).
* Reducing the visibility of the `initialize()` method does not break any application that extends the class and overrides this method using public visibility, as increasing the visibility of an overridden method is allowed.

So I would say that the probability of this change actually breaking source compatibility with any users is negligent. The change obviously does break binary compatibility.

All that being said: I would rather drop this specific visibility change than do a major version bump just for this. So if my arguments are not convincing enough (which I can perfectly understand from an API maintenance perspective), I'll revert this part of the change.

elharo: The get method change is fine. As you point out, it's OK to increase visibility. If that's sent separately, I'm fine with that much.

olavloite:
Closing in favor of #486

PR #462, opened 2020-09-29
chore: adds api-spanner-java group as codeowners

thiagotnunes:
Adds the group api-spanner-java as codeowners.


skuruppu (approved):

chingor13 (changes requested):
Put them on the same line with a space separating them

chingor13 (changes requested):
Also, this file is generated. There's an optional codeowner_team option to set in .repo-metadata.json

chingor13 (approved):

PR #486, opened 2020-09-30
fix: increase visibility of #get()

olavloite:
Increases the visibility of `AbstractLazyInitializer#get()` to make it usable outside the client library.

Replaces #461.


thiagotnunes (approved):
LGTM, do we have to cut a major for this? @olavloite

olavloite:
> LGTM, do we have to cut a major for this? @olavloite

I would say no. This only adds visibility, so it should not break anything for anyone.

I would also like to reduce the visibility of the [initialize method](https://github.com/googleapis/java-spanner/blob/b916df631652340eaf7e66bce875566cada5774b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractLazyInitializer.java#L54) to `protected`, which would be a breaking change, but I would prefer to include that change with other breaking changes instead of having to cut a major release just for that.

elharo (approved):

PR #487, opened 2020-10-01
fix: ResultSet#close() should not throw exceptions from session creation

olavloite:
If a client application requested a session from the session pool, and that request required a new BatchCreateSessions request that would fail, then the error from the BatchCreateSessions RPC would be propagated to both the ReadContext#executeQuery(..) method as well as the ResultSet#close() method. The latter should not happen, as the close method should silently ignore any errors from the session creation.


thiagotnunes (approved):
LGTM

PR #494, opened 2020-10-02
fix: remove dependency on commons-lang

### LINKED ISSUE #493: Remove depedency on commons-lang3
Why are we using commons-lang3 at all? There's no hard rule against but we do have a strong preference for Guava here.

There's just one line of code:  final QueryOption[] internalOptions = ArrayUtils.remove(options, i);

We shouldn't pull in Apache commons for that

https://jlbp.dev/JLBP-1

elharo:
fixes #493 @thiagotnunes


olavloite (approved):
LGTM

PR #501, opened 2020-10-07
fix: close executor when closing pool

olavloite:
Shutdown the executor that maintains the pool when the pool is closing.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/SpannerPool.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/SpannerPool.java
@@ -390,26 +390,34 @@ public class SpannerPool {
           keysStillInUse.add(entry.getKey());
         }
       }
-      if (keysStillInUse.isEmpty() || mode == CheckAndCloseSpannersMode.WARN) {
-        if (!keysStillInUse.isEmpty()) {
+      try {
+        if (keysStillInUse.isEmpty() || mode == CheckAndCloseSpannersMode.WARN) {
+          if (!keysStillInUse.isEmpty()) {
+            logLeakedConnections(keysStillInUse);
+            logger.log(
+                Level.WARNING,
+                "There is/are "
+                    + keysStillInUse.size()
+                    + " connection(s) still open."
+                    + " Close all connections before stopping the application");
+          }
+          // Force close all Spanner instances by passing in a value that will always be less than
+          // the
+          // difference between the current time and the close time of a connection.
+          closeUnusedSpanners(Long.MIN_VALUE);
+        } else {
           logLeakedConnections(keysStillInUse);
-          logger.log(
-              Level.WARNING,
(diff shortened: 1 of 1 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/SpannerPool.java
```diff
@@ -390,26 +390,34 @@ void checkAndCloseSpanners(CheckAndCloseSpannersMode mode) {
...
-        // difference between the current time and the close time of a connection.
-        closeUnusedSpanners(Long.MIN_VALUE);
-      } else {
-        logLeakedConnections(keysStillInUse);
-        throw SpannerExceptionFactory.newSpannerException(
-            ErrorCode.FAILED_PRECONDITION,
-            "There is/are "
-                + keysStillInUse.size()
-                + " connection(s) still open. Close all connections before calling closeSpanner()");
+      } finally {
```
olavloite, line 416: The only real change here is the addition of the `finally` block, but GitHub thinks there is a lot more changed because of the changed indentation and formatting.

olavloite:
cc @jjfox15

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/SpannerPool.java
```diff
@@ -390,26 +390,34 @@ void checkAndCloseSpanners(CheckAndCloseSpannersMode mode) {
...
+        if (closerService != null) {
+          closerService.shutdown();
```
thiagotnunes, line 418: Can this throw an exception?

olavloite: Normally, no. It does not declare any checked exceptions, and it also does not wait until it has actually been shutdown, so the chance that anything goes wrong is relatively low.

The method can be invoked in two ways:
* If a client application calls `SpannerPool#closeSpannerPool()` to explicitly close the pool. Any exception from `closerService.shutdown()` would in that case be propagated to the client application. Considering the fact that it would not be a checked exception, I would say that that is a reasonable thing to do, as it might indicate a problem somewhere, for example linked to the state of the client application at that moment.
* It is called from the shutdown hook that closes the pool automatically when the application terminates. The shutdown hook [catches and ignores any errors](https://github.com/googleapis/java-spanner/blob/87884ac2abd055bee9da303380e2e96fcd1c9e69/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/SpannerPool.java#L95), as any errors at that moment are (probably) not something that a client application could do anything with. The alternative would be to bubble it up, which would cause the JVM (by default) to print the error to `System.err`.

thiagotnunes: Cool, that is fine, I was just worried that we could bubble up exceptions that we previously did not.

thiagotnunes (approved):

PR #505, opened 2020-10-09
fix: AsyncTransactionManager should rollback on close

### LINKED ISSUE #504: AsyncTransactionManager needs to roll back active transaction when closed
When testing the new R2DBC driver with the Cloud Spanner emulator, I noticed that the active transaction was not being rolled back despite `AsyncTransactionManager.close()` being called.

The [contract for `close()`](https://github.com/googleapis/java-spanner/blob/462839b625e58e235581b8ba10b398e1d222eaaf/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncTransactionManager.java#L198) is to roll back the active transaction when closed, but I wonder if `close()` was not meant to be implemented as `closeAsync()` and kind of got left out.

#### Environment details
Spanner client library version: 1.59.0 

#### Steps to reproduce

  1. Start emulator and set the emulator environment variable (`SPANNER_EMULATOR_HOST=localhost:9010`). The issue has nothing to do with the emulator, but it's really easy to observe when using one.
  2. Update "project", "instance", "database" to real values.
  3. Set up the following table definition:
      `gcloud spanner databases ddl update database --ddl="CREATE TABLE test  ( value INT64 ) PRIMARY KEY (value)" --instance=instance`
  4. Run `transactionManagerRollsBackTransactionWhenClosed` _twice_; observe that everything works as expected.
  3. Run `asyncTransactionManagerDoesNotRollBackTransactionWhenClosed` _twice_; observe that the first run succeeds, but the second fails with error `io.grpc.StatusRuntimeException: ABORTED: Transaction NNN aborted due to active transaction MMM. The emulator only supports one transaction at a time.`

```java
  DatabaseId databaseId = DatabaseId.of("project", "instance", "database");
  DatabaseClient dbClient = null;
  TransactionManager txnManager = null;
  AsyncTransactionManager asyncTxnManager = null;

  @Test
  public void transactionManagerRollsBackTransactionWhenClosed() throws Exception {

    try {
      dbClient = spanner.getDatabaseClient(databaseId);
      txnManager = dbClient.transactionManager();
      TransactionContext ctx = txnManager.begin();
      ctx.executeUpdate(Statement.newBuilder("INSERT INTO test (value) VALUES (1234567)").build());

      // MISSING COMMIT; expecting rollback
      // tm.commit();

    } finally {

      if (txnManager != null) {
        System.out.println("************ CLOSING TRANSACTION MANAGER");
        txnManager.close();
      }
    }
  }

  @Test
  public void asyncTransactionManagerDoesNotRollBackTransactionWhenClosed() throws Exception {

    try {
      dbClient = spanner.getDatabaseClient(databaseId);
      asyncTxnManager = dbClient.transactionManagerAsync();
      TransactionContext ctx = asyncTxnManager.beginAsync().get();
      ctx.executeUpdateAsync(Statement.newBuilder("INSERT INTO test (value) VALUES (1234567)").build())
        .get();

      // MISSING COMMIT; expecting rollback upon transaction manager closing
      // tm.commit();

    } finally {

      if (asyncTxnManager != null) {
        asyncTxnManager.close();
      }
    }
  }
```

olavloite:
AsyncTransctionManager should rollback the transaction if close is called while the transaction is still in state STARTED. Failing to do so, will keep the transaction open on the backend for longer than necessary, holding on to locks until it is garbage collected after 10 seconds.

Fixes #504

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncTransactionManagerImpl.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncTransactionManagerImpl.java
@@ -24,6 +24,7 @@ import com.google.cloud.Timestamp;
 import com.google.cloud.spanner.SessionImpl.SessionTransaction;
 import com.google.cloud.spanner.TransactionContextFutureImpl.CommittableAsyncTransactionManager;
 import com.google.cloud.spanner.TransactionManager.TransactionState;
+import com.google.common.base.MoreObjects;
 import com.google.common.base.Preconditions;
 import com.google.common.util.concurrent.MoreExecutors;
 import io.opencensus.trace.Span;
@@ -54,10 +55,17 @@ final class AsyncTransactionManagerImpl
 
   @Override
   public void close() {
+    closeAsync();
+  }
+
+  @Override
+  public ApiFuture<Void> closeAsync() {
+    ApiFuture<Void> res = null;
     if (txnState == TransactionState.STARTED) {
-      rollbackAsync();
+      res = rollbackAsync();
     }
     txn.close();
+    return MoreObjects.firstNonNull(res, ApiFutures.<Void>immediateFuture(null));
   }
 
   @Override
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncTransactionManagerImpl.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncTransactionManagerImpl.java
@@ -54,6 +54,9 @@ final class AsyncTransactionManagerImpl
 
   @Override
   public void close() {
+    if (txnState == TransactionState.STARTED) {
+      rollbackAsync();
+    }
     txn.close();
   }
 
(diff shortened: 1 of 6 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncTransactionManagerImpl.java
```diff
@@ -54,6 +54,9 @@ public void setSpan(Span span) {
 
   @Override
   public void close() {
+    if (txnState == TransactionState.STARTED) {
+      rollbackAsync();
```
elefeint, line 58: Can there be a version of `close()` that propagates the `ApiFutrue` from `rollbackAsync()` back to the caller? Otherwise the application could shut down (or, in my case, the test terminate) before the rollback happens.

elefeint (changes requested):
Thank you for getting to it so fast!

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncTransactionManagerImpl.java
olavloite: I've added a `closeAsync()` method to the `AsyncTransactionManager` interface that returns an `ApiFuture` that is done when everything has been rolled back and released. PTAL.

elefeint (approved):
Thank you!

PR #516, opened 2020-10-15
fix: AsyncTransactionManager did not propagate statement errors

### LINKED ISSUE #514: Exceptions not propagated from async function back to async transaction
#### Environment details

1. Java version: 11
1. spanner version(s): 1.59.0

#### Steps to reproduce

In the Async API, `TransactionContext.executeUpdateAsync()` correctly propagates errors through `ApiFuture` callbacks. However, in an asynchronous transaction, the `.then()` operator does not get the same exceptions propagated to it from the transaction function.

#### Code example

The following test (just add a custom `databaseId`) hangs because neither the success nor failure callback is called, and the latch is never released. If the SQL statement were corrected, the test would pass, as success is propagated correctly.

```java
@Test
  public void clientLibraryTransactionNotPropagatingFailure()
      throws ExecutionException, InterruptedException {
    Executor executor = Executors.newFixedThreadPool(4);

    DatabaseClient dbClient = options.getService().getDatabaseClient(databaseId);
    AsyncTransactionManager transactionManager = dbClient.transactionManagerAsync();
    TransactionContextFuture txnContextFuture = transactionManager.beginAsync();


    // Statement has garbled SQL syntax intentionally
    Statement statement =
        Statement.newBuilder("INSERT INTO BOOKS (UUID, TITLE) VALUES ('123', 'Test book')jljlk")
            .build();
    ApiFuture<Long> updateFuture =
        txnContextFuture.then(
            (txnContext, unusedVoid) -> txnContext.executeUpdateAsync(statement),
            executor);

    CountDownLatch latch = new CountDownLatch(1);
    ApiFutures.addCallback(
        updateFuture,
        new ApiFutureCallback<Long>() {
          @Override
          public void onFailure(Throwable throwable) {
            // Expecting the failure callback to be called, but it's not, so the latch hangs.
            System.out.println("Exception: " + throwable);
            latch.countDown();
          }

          @Override
          public void onSuccess(Long aLong) {
            System.out.println("Success: " + aLong);
            latch.countDown();
          }
        }, executor);

    latch.await();
  }
```

Thanks!

olavloite:
Invalid statements or other statements that would cause an error would not cause the returned ApiFuture to fail.

Fixes #514

### DIFF
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/AsyncTransactionManagerTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/AsyncTransactionManagerTest.java
@@ -1117,14 +1117,6 @@ public class AsyncTransactionManagerTest extends AbstractAsyncTransactionTest {
   @Test
   public void asyncTransactionManager_shouldPropagateStatementFailure()
       throws ExecutionException, InterruptedException, TimeoutException {
-    final Statement garbledStatement =
-        Statement.newBuilder("INSERT INTO BOOKS (UUID, TITLE) VALUES ('123', 'Test book')jljlk")
-            .build();
-    mockSpanner.putStatementResult(
-        StatementResult.exception(
-            garbledStatement,
-            Status.INVALID_ARGUMENT.withDescription("Garbled SQL").asRuntimeException()));
-
     DatabaseClient dbClient = client();
     try (AsyncTransactionManager transactionManager = dbClient.transactionManagerAsync()) {
       TransactionContextFuture txnContextFuture = transactionManager.beginAsync();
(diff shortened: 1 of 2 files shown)


google-cloud-spanner/src/test/java/com/google/cloud/spanner/AsyncTransactionManagerTest.java
```diff
@@ -1112,4 +1113,56 @@ public void asyncTransactionManagerQuery() throws Exception {
       }
     }
   }
+
+  @Test
+  public void asyncTransactionManager_shouldPropagateStatementFailure()
+      throws ExecutionException, InterruptedException, TimeoutException {
+    final Statement garbledStatement =
+        Statement.newBuilder("INSERT INTO BOOKS (UUID, TITLE) VALUES ('123', 'Test book')jljlk")
```
elefeint, line 1121: Optional: use a table more commonly found in client library integration test, so future readers won't have to wonder where `BOOKS` came from.

elefeint (approved):
Thank you!

elefeint (approved):

olavloite:
The failing samples test (Java 8) is a known transient error that frequently happens with Kokoro builds.

PR #522, opened 2020-10-18
feat!: add support for CommitStats

olavloite:
Adds support for `CommitStats`.


.github/workflows/approve-readme.yaml
```diff
@@ -0,0 +1,54 @@
+on:
```
thiagotnunes, line 1: Should this file be part of this PR?

.kokoro/build.sh
```diff
@@ -69,9 +69,16 @@ integration)
     RETURN_CODE=$?
```
thiagotnunes, line 69: Should this file be part of this PR?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java
```diff
@@ -83,6 +83,30 @@
...
+
+    /** The commit statistics of the transaction. */
+    CommitStats getCommitStats();
+  }
+
+  /**
+   * Same as {@link #write(Iterable)}, but requests the backend to return both a commit timestamp
+   * and {@link CommitStats}.
+   */
+  WriteResponse writeWithCommitStats(Iterable<Mutation> mutations) throws SpannerException;
```
thiagotnunes, line 101: Can't we add a breaking change to the existing `write` methods in order to avoid the overload here?

.github/workflows/approve-readme.yaml
olavloite: No, it's part of the generated changes by running the synth tool. It should be removed before we merge it.

.kokoro/build.sh
olavloite: No, it's part of the generated changes by running the synth tool. It should be removed before we merge it.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java
olavloite: That is possible, but it will be a compile time breaking change which will require users to change existing code if they use the commit timestamp that was returned by the `write` methods. I've implemented it that way now, and it does seem like a better solution to me as well, assuming that we do accept this breaking change.
The advantage of this solution is that we can introduce a `TransactionOption` option in the same way as that we already have `ReadOption` and `QueryOption`, and use that for all transaction types that need to support `CommitStats`. This also adds additional possibilities for future additions to read/write transactions and/or commit requests.

WDYT?

thiagotnunes: Hey @olavloite, we had an internal discussion and decided not change the existing methods (sorry about that). Instead, we decided to create two new methods, `writeWithOptions` and `writeAtLeastOnceWithOptions` that would have the different return type, as well as accept `WriteOption`s.

Since this modification will be required here, and for another feature that we are developing, we decided to put the API change in a separate PR. This way we can merge that in first and use it both here and for the other feature. You can take a look at the PR in https://github.com/googleapis/java-spanner/pull/531.

olavloite:
Replaced by #544

PR #524, opened 2020-10-19
test: remove explicit CreateInstance timeout

skuruppu:
To prevent test flakiness.

### DIFF
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/IntegrationTestEnv.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/IntegrationTestEnv.java
@@ -24,7 +24,6 @@ import com.google.common.collect.Iterators;
 import com.google.spanner.admin.instance.v1.CreateInstanceMetadata;
 import io.grpc.Status;
 import java.util.Random;
-import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import org.junit.rules.ExternalResource;
@@ -126,7 +125,7 @@ public class IntegrationTestEnv extends ExternalResource {
         instanceAdminClient.createInstance(instance);
     Instance createdInstance;
     try {
-      createdInstance = op.get(30000L, TimeUnit.MILLISECONDS);
+      createdInstance = op.get();
     } catch (Exception e) {
       boolean cancelled = false;
       try {


google-cloud-spanner/src/test/java/com/google/cloud/spanner/IntegrationTestEnv.java
```diff
@@ -126,7 +126,7 @@ private void initializeInstance(InstanceId instanceId) {
         instanceAdminClient.createInstance(instance);
     Instance createdInstance;
     try {
-      createdInstance = op.get(30000L, TimeUnit.MILLISECONDS);
+      createdInstance = op.get(3600000L, TimeUnit.MILLISECONDS);
```
skuruppu, line 129: This is to fix the flakiness on the internal runs of these integration tests.

I do wonder if this timeout is even necessary. @olavloite do you know if these long-running operation settings are respected by the OperationFuture?

https://github.com/googleapis/googleapis/blob/master/google/spanner/admin/instance/v1/spanner_admin_instance_gapic.yaml#L23-L27

thiagotnunes: Won't we have to wait 1 hour if the test hang for some reason (or do you think this should not happen)? Maybe a few minutes could be enough?

skuruppu: Yeah I agree, there may be other test timeouts that get triggered if we make it so long. I will check with the team on what they think is an appropriate value.

olavloite: > I do wonder if this timeout is even necessary. @olavloite do you know if these long-running operation settings are respected by the OperationFuture?

Yes, as far as I can remember they are. The total timeout in the config is however 24 hours, so it will take a very long time before it actually times out with the default settings.

skuruppu: Yup, I think it is set here:
https://github.com/googleapis/java-spanner/blob/60dd9fbc419ac959a300b64fef84ef5f801548e5/google-cloud-spanner/src/main/java/com/google/cloud/spanner/admin/instance/v1/stub/InstanceAdminStubSettings.java#L626-L636

So I will remove this line altogether :)

skuruppu:
PTAL @thiagotnunes and @olavloite. I would like to get this in before the next release as it's required to fix some internal flaky tests.

thiagotnunes (approved):

olavloite (approved):

PR #531, opened 2020-10-21
feat: adds options to the write operations

thiagotnunes:
Adds the possibility of adding options to the write operations and encapsulates the write response into its own class so that we can augment the response with more fields than the commit timestamp.

At the moment, the new methods just delegate to the existing write methods. Once we add write option implementations we should change the definition of the added methods.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java
@@ -17,6 +17,7 @@
 package com.google.cloud.spanner;
 
 import com.google.cloud.Timestamp;
+import com.google.cloud.spanner.Options.WriteOption;
 
 /**
  * Interface for all the APIs that are used to read/write data into a Cloud Spanner database. An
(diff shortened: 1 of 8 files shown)


olavloite (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java
```diff
@@ -52,6 +53,35 @@
...
+   *         .to("Billy")
+   *         .set("LastName")
+   *         .to("Joel")
+   *         .build();
+   * dbClient.writeWithOptions(Collections.singletonList(mutation));
+   * }</pre>
+   *
+   * @return a response with the timestamp at which the write was committed
+   */
+  CommitResponse writeWithOptions(Iterable<Mutation> mutations, TransactionOption... options)
```
mayurkale22, line 82: What do you think about `WriteOption` in place of `TransactionOption`?

olavloite: Thiago and I had a short discussion about this, and we think `TransactionOption` is better in this case, as we intend to also use this for options that we will need for normal read/write transactions.

mayurkale22: Ok sgtm

mayurkale22 (approved):

PR #536, opened 2020-10-21
fix!: initialize should be protected

olavloite:
The `AbstractLazyInitializer#initialize()` method should be `protected`.


olavloite:
@thiagotnunes It would be good to include this change in the release with the other breaking changes.

thiagotnunes (approved):

PR #544, opened 2020-10-24
feat!: add support for CommitStats

olavloite:
Adds support for returning `CommitStats` from read/write transactions.

Replaces #522


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolAsyncTransactionManager.java
```diff
@@ -239,4 +244,22 @@ public TransactionState getState() {
       return txnState;
     }
   }
+
+  public ApiFuture<CommitResponse> getCommitResponse() {
+    synchronized (lock) {
+      Preconditions.checkState(
+          txnState == TransactionState.COMMITTED,
+          "commit can only be invoked if the transaction was successfully committed");
```
skuruppu, line 252: Should this be "getCommitResponse can only be invoked..."?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionManagerImpl.java
```diff
@@ -115,7 +117,14 @@ public Timestamp getCommitTimestamp() {
...
         txnState == TransactionState.COMMITTED,
         "getCommitTimestamp can only be invoked if the transaction committed successfully");
-    return txn.commitTimestamp();
+    return txn.getCommitResponse().getCommitTimestamp();
+  }
+
+  public CommitResponse getCommitResponse() {
+    Preconditions.checkState(
+        txnState == TransactionState.COMMITTED,
+        "getCommitTimestamp can only be invoked if the transaction committed successfully");
```
skuruppu, line 126:
```suggestion
        "getCommitResponse can only be invoked if the transaction committed successfully");
```

skuruppu (approved):
I'm not quite sure if we need the breaking change here. If it's just because of the interface update, I wouldn't mark it as such. But if there is an actual breaking change where users will have to update their code in order to use the next release, then let me know.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolAsyncTransactionManager.java
olavloite: No, this is in the actual commit method. A client application could for example call `commit()` after calling `rollback` or before calling `begin`. This error will be thrown in those cases.

olavloite:
> I'm not quite sure if we need the breaking change here. If it's just because of the interface update, I wouldn't mark it as such. But if there is an actual breaking change where users will have to update their code in order to use the next release, then let me know.

There are no changes in this PR that will break existing code that just uses these interfaces. Binary compatibility with the previous version will be broken, and code that actually implements the `DatabaseClient` interface would need to be updated.

google-cloud-spanner/clirr-ignored-differences.xml
```diff
@@ -406,4 +406,62 @@
     <className>com/google/cloud/spanner/AbstractLazyInitializer</className>
     <method>java.lang.Object initialize()</method>
   </difference>
+  
+  <!-- Support for CommitStats added -->
```
elharo, line 410: another one where a major version bump is required

elharo (changes requested):

skuruppu:
@olavloite, I just merged in #817 so you can rebase this PR. In light of the discussions about binary incompatibility, please take a look and see if the implementation needs to be changed to avoid a breaking change or if we'll just have to do a semver updgrade.

olavloite:
> @olavloite, I just merged in #817 so you can rebase this PR. In light of the discussions about binary incompatibility, please take a look and see if the implementation needs to be changed to avoid a breaking change or if we'll just have to do a semver updgrade.

@skuruppu One possible way to change this into a non-breaking change would be to return the `CommitStats` through the options object instead of through the `TransactionRunner`. In the current implementation requesting and getting the `CommitStats` looks like this:

```java
    TransactionRunner runner = client.readWriteTransaction(Options.commitStats());
    runner.run(
        new TransactionCallable<Void>() {
          @Override
          public Void run(TransactionContext transaction) throws Exception {
            transaction.buffer(Mutation.delete("FOO", Key.of("foo")));
            return null;
          }
        });
    System.out.println(runner.getCommitResponse().getCommitStats());
```

That could be changed to the following:

```java
    CommitStatsOption option = Options.commitStats();
    TransactionRunner runner = client.readWriteTransaction(option);
    runner.run(
        new TransactionCallable<Void>() {
          @Override
          public Void run(TransactionContext transaction) throws Exception {
            transaction.buffer(Mutation.delete("FOO", Key.of("foo")));
            return null;
          }
        });
    System.out.println(option.getCommitStats());
```

In the latter case we don't need to add the new method `TransactionRunner#getCommitResponse` which is the source of the breaking change in this case. Instead, we make the `CommitStatsOption` class public (it is currently package-private) and add a `getCommitStats()` method to it.

WDYT?

skuruppu:
> > @olavloite, I just merged in #817 so you can rebase this PR. In light of the discussions about binary incompatibility, please take a look and see if the implementation needs to be changed to avoid a breaking change or if we'll just have to do a semver updgrade.
> 
> @skuruppu One possible way to change this into a non-breaking change would be to return the `CommitStats` through the options object instead of through the `TransactionRunner`. In the current implementation requesting and getting the `CommitStats` looks like this:
> 
> ```java
>     TransactionRunner runner = client.readWriteTransaction(Options.commitStats());
>     runner.run(
>         new TransactionCallable<Void>() {
>           @Override
>           public Void run(TransactionContext transaction) throws Exception {
>             transaction.buffer(Mutation.delete("FOO", Key.of("foo")));
>             return null;
>           }
>         });
>     System.out.println(runner.getCommitResponse().getCommitStats());
> ```
> 
> That could be changed to the following:
> 
> ```java
>     CommitStatsOption option = Options.commitStats();
>     TransactionRunner runner = client.readWriteTransaction(option);
>     runner.run(
>         new TransactionCallable<Void>() {
>           @Override
>           public Void run(TransactionContext transaction) throws Exception {
>             transaction.buffer(Mutation.delete("FOO", Key.of("foo")));
>             return null;
>           }
>         });
>     System.out.println(option.getCommitStats());
> ```
> 
> In the latter case we don't need to add the new method `TransactionRunner#getCommitResponse` which is the source of the breaking change in this case. Instead, we make the `CommitStatsOption` class public (it is currently package-private) and add a `getCommitStats()` method to it.
> 
> WDYT?

I think that's a great idea. It is an option after all and as long as it's accessible when required, then it works. @thiagotnunes FYI.

thiagotnunes:
@olavloite hey Knut, after talking with @skuruppu we have decided not to change the design right now, but to do a major release instead.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITAsyncAPITest.java
```diff
@@ -302,4 +316,65 @@ public void asyncRunnerFireAndForgetInvalidUpdate() throws Exception {
...
+    assumeFalse("Emulator does not return commit statistics", isUsingEmulator());
+    try (AsyncTransactionManager mgr = client.transactionManagerAsync(Options.commitStats())) {
+      TransactionContextFuture ctx = mgr.beginAsync();
+      while (true) {
+        try {
+          get(
+              ctx.then(
+                      new AsyncTransactionFunction<Void, Void>() {
+                        @Override
+                        public ApiFuture<Void> apply(TransactionContext txn, Void input)
```
elharo, line 355: txn --> transaction per Google style guide

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITAsyncAPITest.java
```diff
@@ -302,4 +316,65 @@ public void asyncRunnerFireAndForgetInvalidUpdate() throws Exception {
...
+    assertThat(get(runner.getCommitResponse()).getCommitStats()).isNotNull();
+    // MutationCount = 2 columns + 2 secondary indexes.
+    assertThat(get(runner.getCommitResponse()).getCommitStats().getMutationCount()).isEqualTo(4L);
+  }
+
+  @Test
+  public void asyncTransactionManagerReturnsCommitStats() throws InterruptedException {
+    assumeFalse("Emulator does not return commit statistics", isUsingEmulator());
+    try (AsyncTransactionManager mgr = client.transactionManagerAsync(Options.commitStats())) {
+      TransactionContextFuture ctx = mgr.beginAsync();
```
elharo, line 348: ctx --> context per Google style guide

elharo (commented):
It's always acceptable to bump the major version. I think you'll want to add a ! after feat in the PR description (i.e. feat! ) so the release automation notices this. It might also be useful to update the snapshot version in this PR.

After this release goes out, you can remove the ignored differences file and point it at the newly released major version.

thiagotnunes:
@elharo Thanks, updated the PR description.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
```diff
@@ -253,6 +277,9 @@ public boolean equals(Object o) {
   @Override
   public int hashCode() {
     int result = 31;
+    if (withCommitStats) {
+      result = 31 * result + Boolean.valueOf(withCommitStats).hashCode();
```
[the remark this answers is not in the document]
olavloite: Replaced with constant.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/OptionsTest.java
```diff
@@ -237,6 +240,15 @@ public void testTransactionOptionsEquality() {
 
     o2 = Options.fromReadOptions(Options.prefetchChunks(1));
     assertThat(o1.equals(o2)).isFalse();
+
+    o1 = Options.fromTransactionOptions(Options.commitStats());
```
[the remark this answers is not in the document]
olavloite: Separated into separate test methods.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITAsyncAPITest.java
olavloite: Done

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITAsyncAPITest.java
olavloite: Done.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/OptionsTest.java
```diff
@@ -224,6 +224,9 @@ public void queryEquality() {
   public void testFromTransactionOptions() {
     Options opts = Options.fromTransactionOptions();
     assertThat(opts.toString()).isEqualTo("");
+
+    opts = Options.fromTransactionOptions(Options.commitStats());
```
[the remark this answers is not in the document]
olavloite: Separated into separate test methods.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITTransactionTest.java
```diff
@@ -630,4 +631,23 @@ public Long run(TransactionContext transaction) throws Exception {
       assertThat(e.getErrorCode()).isEqualTo(ErrorCode.INVALID_ARGUMENT);
     }
   }
+
+  @Test
+  public void transactionRunnerReturnsCommitStats() {
+    assumeFalse("Emulator does not return commit statistics", isUsingEmulator());
```
elharo, line 637: The assume here is surprising. Why would this be sometimes be true and sometimes not true in this one test method?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITTransactionTest.java
```diff
@@ -630,4 +631,23 @@ public Long run(TransactionContext transaction) throws Exception {
...
+    runner.run(
+        new TransactionCallable<Void>() {
+          @Override
+          public Void run(TransactionContext transaction) throws Exception {
+            transaction.buffer(
+                Mutation.newInsertBuilder("T").set("K").to(key).set("V").to(0).build());
+            return null;
+          }
+        });
+    assertThat(runner.getCommitResponse().getCommitStats()).isNotNull();
```
elharo, line 649: you can use assertNotNull here. Truth doesn't add anything for simple cases.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITTransactionTest.java
olavloite: `isUsingEmulator()` checks whether the environment variable `SPANNER_EMULATOR_HOST` has been set to a non-empty value. If so, the integration tests are running against the emulator. The emulator does not support all features of Cloud Spanner, which means we need to skip some specific tests. The CI environments runs the tests both against the emulator and Cloud Spanner.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITTransactionTest.java
olavloite: done.

skuruppu:
@olavloite this can now be merged when you're ready.

skuruppu:
@elharo friendly ping. This feature is launching soon so we need to start merging in the PRs to make sure that the samples are ready to ingested into the Cloud Docs.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITTransactionManagerTest.java
```diff
@@ -212,4 +214,31 @@ public void abortAndRetry() throws InterruptedException {
...
     }
   }
+
+  @SuppressWarnings("resource")
+  @Test
+  public void testTransactionManagerReturnsCommitStats() throws InterruptedException {
+    assumeFalse("Emulator does not return commit statistics", isUsingEmulator());
+    try (TransactionManager manager = client.transactionManager(Options.commitStats())) {
+      TransactionContext transaction = manager.begin();
+      while (true) {
```
elharo, line 224: This one scares me since it looks like an infinite loop Could you add a counter with a maximum number of retries before failure?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITTransactionManagerTest.java
```diff
@@ -212,4 +214,31 @@ public void abortAndRetry() throws InterruptedException {
...
+                .set("BoolValue")
+                .to(true)
+                .build());
+        try {
+          manager.commit();
+          assertNotNull(manager.getCommitResponse().getCommitStats());
+          assertEquals(2L, manager.getCommitResponse().getCommitStats().getMutationCount());
+          break;
+        } catch (AbortedException e) {
+          Thread.sleep(e.getRetryDelayInMillis() / 1000);
```
elharo, line 238: BUG! Thread.sleep takes milliseconds. No need to divide by 1000.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITTransactionManagerTest.java
olavloite: This is the recommended way to use `TransactionManager` according to the [documentation](https://github.com/googleapis/java-spanner/blob/5af36739532037360dfd504a4a0988562550526c/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java#L324). It is also the way it is already used in multiple other test cases. I would rather either:
1. Keep this and all other instances as they are.
2. Change this and other instances + the documentation in a separate PR.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncRunnerImpl.java
```diff
@@ -66,16 +74,30 @@ public R run(TransactionContext transaction) throws Exception {
         });
   }
 
-  private void setCommitTimestamp() {
+  private void setCommitResponse() {
     try {
-      commitTimestamp.set(delegate.getCommitTimestamp());
+      commitResponse.set(delegate.getCommitResponse());
     } catch (Throwable t) {
```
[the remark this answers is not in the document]
olavloite: Added issue: https://github.com/googleapis/java-spanner/issues/875

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncRunner.java
```diff
@@ -56,4 +56,10 @@
    * {@link ExecutionException} if the transaction did not commit.
    */
   ApiFuture<Timestamp> getCommitTimestamp();
+
+  /**
+   * Returns the {@link CommitResponse} of this transaction. {@link ApiFuture#get()} will throw an
```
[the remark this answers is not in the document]
olavloite: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncTransactionManagerImpl.java
```diff
@@ -132,29 +133,37 @@ public void onError(Throwable t) {
           SpannerExceptionFactory.newSpannerException(
               ErrorCode.ABORTED, "Transaction already aborted"));
     }
-    ApiFuture<Timestamp> res = txn.commitAsync();
+    ApiFuture<CommitResponse> res = txn.commitAsync();
```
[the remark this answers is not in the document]
olavloite: Changed to `commitResponseFuture`

google-cloud-spanner/src/main/java/com/google/cloud/spanner/CommitStats.java
```diff
@@ -0,0 +1,54 @@
...
+
+  /**
+   * The number of mutations that were executed by the transaction. Insert and update operations
+   * count with the multiplicity of the number of columns they affect. For example, inserting a new
+   * record may count as five mutations, if values are inserted into five columns. Delete and delete
+   * range operations count as one mutation regardless of the number of columns affected. Deleting a
+   * row from a parent table that has the ON DELETE CASCADE annotation is also counted as one
+   * mutation regardless of the number of interleaved child rows present. The exception to this is
+   * if there are secondary indexes defined on rows being deleted, then the changes to the secondary
+   * indexes will be counted individually. For example, if a table has 2 secondary indexes, deleting
```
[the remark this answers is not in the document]
olavloite: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/CommitStats.java
```diff
@@ -0,0 +1,54 @@
...
+   * a range of rows in the table will count as 1 mutation for the table, plus 2 mutations for each
```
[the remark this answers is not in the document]
olavloite: Done

google-cloud-spanner/src/test/java/com/google/cloud/spanner/AsyncTransactionManagerImplTest.java
```diff
@@ -0,0 +1,60 @@
...
+import com.google.api.core.ApiFutures;
+import com.google.cloud.Timestamp;
+import io.opencensus.trace.Span;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import org.mockito.Mock;
+
+@RunWith(JUnit4.class)
```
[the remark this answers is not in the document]
olavloite: Done

google-cloud-spanner/src/test/java/com/google/cloud/spanner/OptionsTest.java
```diff
@@ -221,22 +225,57 @@ public void queryEquality() {
...
+    Options option1 = Options.fromTransactionOptions(Options.commitStats());
+    Options option2 = Options.fromTransactionOptions(Options.commitStats());
+    assertTrue(option1.equals(option2));
+  }
+
+  @Test
+  public void testTransactionOptions_withCommitStatsAndOtherOptionAreNotEqual() {
+    Options option1 = Options.fromTransactionOptions(Options.commitStats());
+    Options option2 = Options.fromQueryOptions(Options.prefetchChunks(10));
+    assertFalse(option1.equals(option2));
```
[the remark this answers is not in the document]
olavloite: Done

google-cloud-spanner/src/test/java/com/google/cloud/spanner/OptionsTest.java
```diff
@@ -221,22 +225,57 @@ public void queryEquality() {
...
+    assertTrue(option1.equals(option2));
+  }
 
-    o2 = Options.fromReadOptions(Options.prefetchChunks(1));
-    assertThat(o1.equals(o2)).isFalse();
+  @Test
+  public void testTransactionOptions_withCommitStatsAreEqual() {
+    Options option1 = Options.fromTransactionOptions(Options.commitStats());
+    Options option2 = Options.fromTransactionOptions(Options.commitStats());
+    assertTrue(option1.equals(option2));
```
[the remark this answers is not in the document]
olavloite: Done

google-cloud-spanner/src/test/java/com/google/cloud/spanner/OptionsTest.java
```diff
@@ -221,22 +225,57 @@ public void queryEquality() {
...
+  }
 
-    o1 = Options.fromTransactionOptions();
-    o2 = Options.fromTransactionOptions();
-    assertThat(o1.equals(o2)).isTrue();
+  @Test
+  public void testTransactionOptions_noOptionsAreEqual() {
+    Options option1 = Options.fromTransactionOptions();
+    Options option2 = Options.fromTransactionOptions();
+    assertTrue(option1.equals(option2));
```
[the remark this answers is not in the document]
olavloite: Done

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITTransactionManagerTest.java
olavloite: Good point. Changed and [filed a bug](https://github.com/googleapis/java-spanner/issues/874) for the documentation that also includes this.

PR #552, opened 2020-10-26
fix: SessionNotFound was not retried for AsyncTransactionManager

olavloite:
`SessionNotFoundException` was not retried if these happened when using `AsyncTransactionManager`. These exceptions are now handled explicitly in the following way:
1. The underlying session that is used for the transaction is replaced with a new session from the pool.
1. The `SessionNotFoundException` is translated to an `AbortedException`, which again will be retried, assuming that the client application has implemented a normal retry loop, as is required when using an `AsyncTransactionManager`.


thiagotnunes (approved):
LGTM

PR #559, opened 2020-10-27
fix: adds assembly descriptor to snippets samples

thiagotnunes:
Creates a jar with dependencies for the snippets samples. This is how we executed the samples in the other repository, so we are mimicking the process here.


samples/snippets/pom.xml
```diff
@@ -112,4 +114,45 @@
...
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-failsafe-plugin</artifactId>
+        <version>3.0.0-M5</version>
+        <configuration>
+          <systemPropertyVariables>
+            <spanner.test.instance>default-instance</spanner.test.instance>
+            <spanner.sample.database>mysample</spanner.sample.database>
+            <spanner.quickstart.database>quickstart-db
+            </spanner.quickstart.database>
```
olavloite, line 152: nit: could this be one line

olavloite (approved):

samples/snippets/pom.xml
thiagotnunes: Oh this was broken, because of the 80 characters per column checkstyle rule.

PR #568, opened 2020-10-31
test: allow partial statements on the mock server

olavloite:
Allow a partial statement to be registered as the key for a statement result on the mock server. This makes it easier to mock results for statements that are generated by a framework with random or time based values, as only the first (static) part of the statement needs to be registered on the mock server.

Fixes this TODO: https://github.com/cloudspannerecosystem/liquibase-spanner/blob/31c04dcaeb8105009774a51cf41a1a10bc1c6685/src/test/java/com/google/cloud/spanner/TempMockSpannerServiceImpl.java#L188


thiagotnunes (approved):
LGTM

PR #579, opened 2020-11-04
fix: make enums in the Connection API public

olavloite:
Some enums were defined as package-private, but these were used in the public API of the Connection API. That made the specific methods in the Connection API unusable from outside the Spanner client library.

Towards https://github.com/googleapis/java-spanner-jdbc/issues/253


thiagotnunes (approved):
LGTM

olavloite:
Build failures for the samples are not related to this change.

PR #608, opened 2020-11-06
feat!: add CommitStats to Connection API

olavloite:
Adds support for `CommitStats` to the Connection API.


google-cloud-spanner/pom.xml
```diff
@@ -377,5 +377,38 @@
         </plugins>
       </build>
     </profile>
+    <profile>
```
olavloite, line 380: This was a missing piece of configuration that was still in the JDBC repository, but that belongs in the Connection API. It is a Maven profile for automatically generating tests for new client side SQL statements.

thiagotnunes: Could we add this as a separate PR? Otherwise we won't have this until the commit stats is merged into master.

thiagotnunes (commented):
LGTM, just a few questions

google-cloud-spanner/src/test/resources/com/google/cloud/spanner/connection/ClientSideStatementsTest.sql
```diff
@@ -1931,9148 +1931,11005 @@ NEW_CONNECTION;
 @EXPECT EXCEPTION INVALID_ARGUMENT
 show variable/-read_only_staleness;
 NEW_CONNECTION;
-begin;
+show variable optimizer_version;
```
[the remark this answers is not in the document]
olavloite: I forgot to regenerate the test sql files for the `optimizer_version` feature, so that's why that is included here. The generator does not support generating tests for only some features, so it's impossible to keep that change separate from the changes for `CommitStats` without completely rewriting the generator.

google-cloud-spanner/pom.xml
olavloite: Good point, I've added #611 for that, and I'll remove it from this PR.

thiagotnunes (approved):

elharo (changes requested):
probably requires a major version bump

google-cloud-spanner/src/main/java/com/google/cloud/spanner/CommitResponse.java
```diff
@@ -34,6 +34,11 @@ public Timestamp getCommitTimestamp() {
     return Timestamp.fromProto(proto.getCommitTimestamp());
   }
 
+  /** @return true if the {@link CommitResponse} includes {@link CommitStats}. */
+  public boolean hasCommitStats() {
```
elharo, line 38: worth a test

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java
```diff
@@ -170,6 +181,26 @@ public StatementResult statementShowCommitTimestamp() {
         "COMMIT_TIMESTAMP", getConnection().getCommitTimestampOrNull(), SHOW_COMMIT_TIMESTAMP);
   }
 
+  @Override
+  public StatementResult statementShowCommitResponse() {
+    CommitResponse response = getConnection().getCommitResponseOrNull();
+    CommitStats stats =
+        response == null || !response.hasCommitStats() ? null : response.getCommitStats();
```
elharo, line 188: This is a little too complex to read as a ternary. Consider using if blocks.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java
```diff
@@ -170,6 +181,26 @@ public StatementResult statementShowCommitTimestamp() {
...
+    ResultSet rs =
```
elharo, line 189: rs --> resultSet per google style

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/DdlBatch.java
```diff
@@ -168,6 +169,17 @@ public Timestamp getCommitTimestampOrNull() {
...
   }
 
+  @Override
+  public CommitResponse getCommitResponse() {
+    throw SpannerExceptionFactory.newSpannerException(
+        ErrorCode.FAILED_PRECONDITION, "There is no commit response available for DDL batches.");
+  }
+
+  @Override
+  public CommitResponse getCommitResponseOrNull() {
```
elharo, line 179: needs a test

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/DdlBatch.java
```diff
@@ -168,6 +169,17 @@ public Timestamp getCommitTimestampOrNull() {
     return null;
   }
 
+  @Override
+  public CommitResponse getCommitResponse() {
+    throw SpannerExceptionFactory.newSpannerException(
+        ErrorCode.FAILED_PRECONDITION, "There is no commit response available for DDL batches.");
```
elharo, line 175: This is a code smell suggests that perhaps getCommitResponse does not belong in the superclass/interface in the first place since it doesn't apply to all subclasses.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/DmlBatch.java
```diff
@@ -119,6 +120,17 @@ public Timestamp getCommitTimestampOrNull() {
     return null;
   }
 
+  @Override
+  public CommitResponse getCommitResponse() {
```
elharo, line 124: Doubles my suspicion that this doesn't belong in the superclass.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ReadOnlyTransaction.java
```diff
@@ -136,6 +137,18 @@ public Timestamp getCommitTimestampOrNull() {
     return null;
   }
 
+  @Override
+  public CommitResponse getCommitResponse() {
```
elharo, line 141: Now there are three instances where this method doesn't make sense. 

Maybe you can construct a commit response for these cases where the server doesn't return one, but maybe the API isn't in the right place.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ReadWriteTransaction.java
```diff
@@ -254,19 +265,30 @@ public Timestamp getReadTimestampOrNull() {
     return null;
   }
 
-  private boolean hasCommitTimestamp() {
-    return commitTimestampFuture != null;
+  private boolean hasCommitResponse() {
```
elharo, line 268: I originally thought this meant the response wasn't available yet, but if it means this object doesn't support commit responses, then I'm again suspecting the API needs some further thought.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/UnitOfWork.java
```diff
@@ -142,6 +143,18 @@ public boolean isActive() {
   /** @return the commit timestamp of this transaction or null if there is no commit timestamp. */
   Timestamp getCommitTimestampOrNull();
 
+  /**
+   * @return the {@link CommitResponse} of this transaction. Will throw a {@link SpannerException}
```
elharo, line 147: Will throw --> Throws

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/UnitOfWork.java
```diff
@@ -142,6 +143,18 @@ public boolean isActive() {
...
+   *     if there is no {@link CommitResponse}.
+   */
+  CommitResponse getCommitResponse();
+
+  /**
+   * @return the {@link CommitResponse} of this transaction or null if there is no {@link
+   *     CommitResponse}.
```
elharo, line 154: no period

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/UnitOfWork.java
```diff
@@ -142,6 +143,18 @@ public boolean isActive() {
...
+   */
+  CommitResponse getCommitResponseOrNull();
```
elharo, line 156: You don't need both these methods. Even if these methods are left here, all we need is a single getCommitResponse method that returns a commit response or null.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/AbstractConnectionImplTest.java
```diff
@@ -735,6 +735,22 @@ public void testGetCommitTimestamp() {
     }
   }
 
+  @Test
+  public void testGetCommitResponse() {
+    try (Connection connection = getConnection()) {
+      if (isGetCommitTimestampAllowed()) {
```
elharo, line 741: if blocks in tests are code smells. Only one path is taken and tested. Does isGetCommitTimestampAllowed() return true or false here?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
```diff
@@ -620,6 +630,68 @@ public void testExecuteGetOptimizerVersion() {
...
+
+      StatementResult res = subject.execute(Statement.of("show variable return_commit_stats"));
+      assertThat(res.getResultType(), is(equalTo(ResultType.RESULT_SET)));
+      assertThat(res.getResultSet().next(), is(true));
+      assertThat(res.getResultSet().getBoolean("RETURN_COMMIT_STATS"), is(equalTo(false)));
+
+      subject.execute(Statement.of("set return_commit_stats=true"));
+      res = subject.execute(Statement.of("show variable return_commit_stats"));
+      assertThat(res.getResultType(), is(equalTo(ResultType.RESULT_SET)));
+      assertThat(res.getResultSet().next(), is(true));
```
elharo, line 690: assertTrue is simpler

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
```diff
@@ -620,6 +630,68 @@ public void testExecuteGetOptimizerVersion() {
...
+
+  @Test
+  public void testExecuteGetReturnCommitStats() {
+    try (ConnectionImpl subject =
+        createConnection(
+            ConnectionOptions.newBuilder()
+                .setCredentials(NoCredentials.getInstance())
+                .setUri(URI)
+                .build())) {
+      assertThat(subject.isReturnCommitStats(), is(equalTo(false)));
```
elharo, line 680: assertFalse

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
```diff
@@ -620,6 +630,68 @@ public void testExecuteGetOptimizerVersion() {
...
+
+      StatementResult res = subject.execute(Statement.of("show variable return_commit_stats"));
```
elharo, line 682: res --> result

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
```diff
@@ -620,6 +630,68 @@ public void testExecuteGetOptimizerVersion() {
...
+      assertThat(res.getResultType(), is(equalTo(ResultType.RESULT_SET)));
+      assertThat(res.getResultSet().next(), is(true));
+      assertThat(res.getResultSet().getBoolean("RETURN_COMMIT_STATS"), is(equalTo(false)));
+
+      subject.execute(Statement.of("set return_commit_stats=true"));
+      res = subject.execute(Statement.of("show variable return_commit_stats"));
```
elharo, line 688: avoid reassigning local variables

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
```diff
@@ -749,6 +821,43 @@ public void testExecuteGetCommitTimestamp() {
...
+                .setCredentials(NoCredentials.getInstance())
+                .setUri(URI)
+                .build())) {
+      subject.beginTransaction();
+      subject.executeQuery(Statement.of(AbstractConnectionImplTest.SELECT)).next();
+      subject.commit();
+      StatementResult res = subject.execute(Statement.of("show variable commit_response"));
+      assertThat(res.getResultType(), is(equalTo(ResultType.RESULT_SET)));
+      assertThat(res.getResultSet().next(), is(true));
+      assertThat(res.getResultSet().getTimestamp("COMMIT_TIMESTAMP"), is(notNullValue()));
```
elharo, line 838: assertNotNull

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
```diff
@@ -749,6 +821,43 @@ public void testExecuteGetCommitTimestamp() {
...
+    try (ConnectionImpl subject =
+        createConnection(
+            ConnectionOptions.newBuilder()
+                .setCredentials(NoCredentials.getInstance())
+                .setUri(URI + ";returnCommitStats=true")
+                .build())) {
+      subject.beginTransaction();
+      subject.executeQuery(Statement.of(AbstractConnectionImplTest.SELECT)).next();
+      subject.commit();
+      StatementResult res = subject.execute(Statement.of("show variable commit_response"));
```
elharo, line 852: result

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
```diff
@@ -749,6 +821,43 @@ public void testExecuteGetCommitTimestamp() {
...
+      assertThat(res.getResultType(), is(equalTo(ResultType.RESULT_SET)));
+      assertThat(res.getResultSet().next(), is(true));
+      assertThat(res.getResultSet().getTimestamp("COMMIT_TIMESTAMP"), is(notNullValue()));
+      assertThat(res.getResultSet().isNull("MUTATION_COUNT"), is(false));
```
elharo, line 856: assertNull

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/it/ITCommitResponseTest.java
```diff
@@ -0,0 +1,149 @@
+/*
+ * Copyright 2020 Google LLC
```
elharo, line 2: 2021

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/it/ITCommitResponseTest.java
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  @Test
+  public void testDefaultNoCommitStats() {
+    try (ITConnection connection = createConnection()) {
+      connection.bufferedWrite(
+          Mutation.newInsertBuilder("TEST").set("ID").to(1L).set("NAME").to("TEST").build());
+      connection.commit();
+      assertThat(connection.getCommitResponse()).isNotNull();
+      assertThat(connection.getCommitResponse().getCommitTimestamp()).isNotNull();
```
elharo, line 61: assertNotNull

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/it/ITCommitResponseTest.java
```diff
@@ -0,0 +1,149 @@
...
+    try (ITConnection connection = createConnection()) {
+      connection.execute(Statement.of("SET AUTOCOMMIT=TRUE"));
+      connection.execute(Statement.of("SET RETURN_COMMIT_STATS=TRUE"));
+      connection.write(
+          Mutation.newInsertBuilder("TEST").set("ID").to(1L).set("NAME").to("TEST").build());
+      assertThat(connection.getCommitResponse()).isNotNull();
+      assertThat(connection.getCommitResponse().getCommitTimestamp()).isNotNull();
+      assertThat(connection.getCommitResponse().hasCommitStats()).isTrue();
+      assertThat(connection.getCommitResponse().getCommitStats().getMutationCount()).isEqualTo(2L);
+      try (ResultSet rs =
```
elharo, line 139: rs --> resultSet

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/it/ITCommitResponseTest.java
```diff
@@ -0,0 +1,149 @@
...
+          connection.execute(Statement.of("SHOW VARIABLE COMMIT_RESPONSE")).getResultSet()) {
+        assertThat(rs.next()).isTrue();
```
elharo, line 141: in general, avoid Truth for assertions that are simpler in JUnit

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITTransactionManagerAsyncTest.java
```diff
@@ -224,6 +222,8 @@ public void testRollback() throws InterruptedException {
     }
   }
 
+  @Ignore(
```
elharo, line 225: Changes in this class don't immediately appear specific to this PR. Can you break this into a separate PR?

elharo (changes requested):
I have substantive concerns about the proposed API here, beyond issues of backwards compatibility

google-cloud-spanner/src/main/java/com/google/cloud/spanner/CommitResponse.java
olavloite: done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java
olavloite: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java
olavloite: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/DdlBatch.java
olavloite: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ReadWriteTransaction.java
olavloite: For this class your initial thought is correct. This indicates that the commit response is not yet available, because the transaction has not yet committed. Once the transaction has been committed, a commit response will be available.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/DdlBatch.java
olavloite: This is a package private class (extending from a package private interface), so it is not something that client applications will deal with. Client applications only interact with the [Connection](https://github.com/googleapis/java-spanner/blob/1b077e457d40dd1f64c20a9f7d897313f78ea4e9/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/Connection.java#L141) interface. (And client applications in this case is only intended to be frameworks /drivers that implement a declarative connection-type interface, such as JDBC.)
These declarative style APIs often mean that certain methods are only applicable in certain states, and this is what is also reflected in these classes. The method that client applications will interact with is [Connection#getCommitResponse()](https://github.com/googleapis/java-spanner/blob/1b077e457d40dd1f64c20a9f7d897313f78ea4e9/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/Connection.java#L654). This method will delegate the actual response to the at that moment active `UnitOfWork` on the `Connection`. The different `UnitOfWork` implementations will respond based on what they support, which may also be that they do not support it.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/DmlBatch.java
olavloite: See above.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ReadOnlyTransaction.java
olavloite: See above.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/UnitOfWork.java
olavloite: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/UnitOfWork.java
olavloite: This is not intended for public use (it's a package private interface). The reason that there are two different versions is that some APIs require a method to throw an error if the connection / transaction is not in a valid state for a specific method call, and in other cases it expects it to return `null`.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
olavloite: Done.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
olavloite: Done.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
olavloite: Done.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
olavloite: Done.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
olavloite: Done.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
olavloite: Done.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
olavloite: Done.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/it/ITCommitResponseTest.java
olavloite: Done.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/it/ITCommitResponseTest.java
olavloite: Done.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/it/ITCommitResponseTest.java
olavloite: Done.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/it/ITCommitResponseTest.java
olavloite: Replace all in this file.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/AbstractConnectionImplTest.java
olavloite: This and the other methods in this abstract test class are used to test that a `Connection` responds correctly in all possible states that it can be. The specific state of a `Connection` is implemented in a concrete subclass, and that determines whether `isGetCommitTimestampAllowed()` will return `true` or `false`.

Example 1: [ConnectionImplAutocommitReadWriteNoActionsTest](ConnectionImplAutocommitReadWriteNoActionsTest) verifies that a connection that is in autocommit read/write mode and has not yet executed any statements, responds correctly to the state-dependent methods. [isGetCommitTimestampAllowed()](https://github.com/googleapis/java-spanner/blob/f9ac29cd05e775c8ea17e0e227eb925df88a8dba/google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplAutocommitReadWriteTest.java#L140) returns false, as the connection has not executed a commit. `getCommitResponse()` should therefore not return a result / throw an exception.

Example 2: [ConnectionImplAutocommitReadWriteAfterTemporaryTransactionTest](https://github.com/googleapis/java-spanner/blob/f9ac29cd05e775c8ea17e0e227eb925df88a8dba/google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplAutocommitReadWriteTest.java#L894) verifies that a connection in autocommit read/write mode that has just successfully executed an ad-hoc transaction does return a commit response. [isGetCommitTimestampAllowed()](https://github.com/googleapis/java-spanner/blob/f9ac29cd05e775c8ea17e0e227eb925df88a8dba/google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplAutocommitReadWriteTest.java#L1004) therefore returns true for this implementation.

This set of test classes is used to generate a [SQL script](https://github.com/googleapis/java-spanner/blob/master/google-cloud-spanner/src/test/resources/com/google/cloud/spanner/connection/ConnectionImplGeneratedSqlScriptTest.sql) that can also be used for the same Connection API in other client languages.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITTransactionManagerAsyncTest.java
olavloite: It seems that that had already been done. I've brought this and the `commit-stats2` up to date with the latest changes from `master`, and there is are differences anymore in this class between this branch and `master`.

olavloite:
> I have substantive concerns about the proposed API here, beyond issues of backwards compatibility

The only public API changes in this PR are the ones that are listed in `clirr-ignored-differences.xml`. The other changes that are discussed above are all in internal classes and interfaces.

In case the intention behind this specific package is not known: This `Connection API` is intended as a common base for the implementation of declarative style connection APIs such as JDBC and R2DBC. This is also stated in its public [classes](https://github.com/googleapis/java-spanner/blob/f9ac29cd05e775c8ea17e0e227eb925df88a8dba/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java#L48) and [interfaces](https://github.com/googleapis/java-spanner/blob/f9ac29cd05e775c8ea17e0e227eb925df88a8dba/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/Connection.java#L40), and in the [package-info](https://github.com/googleapis/java-spanner/blob/master/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/package-info.java).

The (public) API is therefore mainly aimed at making the implementation of the JDBC driver and any other drivers that may use this in the future as thin as possible.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/CommitResponse.java
```diff
@@ -41,6 +41,11 @@ public Timestamp getCommitTimestamp() {
     return Timestamp.fromProto(proto.getCommitTimestamp());
   }
 
+  /** @return true if the {@link CommitResponse} includes {@link CommitStats}. */
```
elharo, line 44: nit: no period

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/Connection.java
```diff
@@ -438,6 +439,15 @@
    */
   String getOptimizerVersion();
 
+  /**
+   * Sets whether this connection should request commit statistics from Cloud Spanner for read/write
+   * transactions and DML statements in autocommit mode.
+   */
+  void setReturnCommitStats(boolean returnCommitStats);
+
+  /** @return true if this connection requests commit statistics from Cloud Spanner. */
```
elharo, line 448: nit: no period

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/Connection.java
```diff
@@ -632,6 +642,17 @@
    */
   Timestamp getCommitTimestamp();
 
+  /**
+   * @return the {@link CommitResponse} of the last {@link TransactionMode#READ_WRITE_TRANSACTION}
+   *     transaction. This method will throw a {@link SpannerException} if there is no last {@link
```
elharo, line 647: will throw --> throws

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/Connection.java
```diff
@@ -632,6 +642,17 @@
...
+   *     TransactionMode#READ_WRITE_TRANSACTION} transaction (i.e. the last transaction was a {@link
```
elharo, line 648: per google style avoid Latin i.e. Perhaps rewrite as 

TransactionMode#READ_WRITE_TRANSACTION} transaction. That is, if the last transaction was a {@link

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
```diff
@@ -574,6 +578,31 @@ Timestamp getCommitTimestampOrNull() {
...
 
+  @Override
+  public CommitResponse getCommitResponse() {
+    ConnectionPreconditions.checkState(!isClosed(), CLOSED_ERROR_MSG);
+    ConnectionPreconditions.checkState(
+        this.currentUnitOfWork != null, "There is no transaction on this connection");
+    return this.currentUnitOfWork.getCommitResponse();
+  }
+
+  CommitResponse getCommitResponseOrNull() {
```
elharo, line 589: I still don't see why both getCommitResponseOrNull and getCommitResponse exist. Why not just getCommitResponse that returns null if there's no unit of work?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
```diff
@@ -574,6 +578,31 @@ Timestamp getCommitTimestampOrNull() {
...
+    ConnectionPreconditions.checkState(!isClosed(), CLOSED_ERROR_MSG);
+    return this.currentUnitOfWork == null ? null : this.currentUnitOfWork.getCommitResponseOrNull();
+  }
+
+  @Override
+  public void setReturnCommitStats(boolean returnCommitStats) {
+    ConnectionPreconditions.checkState(!isClosed(), CLOSED_ERROR_MSG);
```
elharo, line 596: Is it useful to check the connection state here? What breaks if this field is set or read on a closed connection?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
```diff
@@ -184,6 +185,8 @@ public String getDefaultValue() {
   private static final String USER_AGENT_PROPERTY_NAME = "userAgent";
   /** Query optimizer version to use for a connection. */
   private static final String OPTIMIZER_VERSION_PROPERTY_NAME = "optimizerVersion";
+  /** Query optimizer version to use for a connection. */
+  private static final String RETURN_COMMIT_STATS_PROPERTY_NAME = "returnCommitStats";
```
elharo, line 189: The constant is less clear than the string itself and doesn't help. It's only used in one obvious place.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
```diff
@@ -184,6 +185,8 @@ public String getDefaultValue() {
   private static final String USER_AGENT_PROPERTY_NAME = "userAgent";
   /** Query optimizer version to use for a connection. */
   private static final String OPTIMIZER_VERSION_PROPERTY_NAME = "optimizerVersion";
+  /** Query optimizer version to use for a connection. */
```
elharo, line 188: copy pasta

olavloite:
@skuruppu I assume it is OK to merge this one as well once it is approved?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
```diff
@@ -634,6 +641,12 @@ static String parseOptimizerVersion(String uri) {
     return value != null ? value : DEFAULT_OPTIMIZER_VERSION;
   }
 
+  @VisibleForTesting
+  static boolean parseReturnCommitStats(String uri) {
+    String value = parseUriProperty(uri, RETURN_COMMIT_STATS_PROPERTY_NAME);
+    return value != null ? Boolean.valueOf(value) : DEFAULT_RETURN_COMMIT_STATS;
```
elharo, line 647: `false` is clearer than a constant whose value is 600 lines up.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/UnitOfWork.java
elharo: Which APIs? It sounds like the responsibility for interpreting and handling a null return or an exception belongs in those methods.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
```diff
@@ -103,6 +110,12 @@ public void commit() {
       Timestamp commitTimestamp = Timestamp.now();
       commitResponse = mock(CommitResponse.class);
       when(commitResponse.getCommitTimestamp()).thenReturn(commitTimestamp);
+      if (returnCommitStats) {
+        CommitStats stats = mock(CommitStats.class);
```
elharo, line 114: Isn't this just a value class? You don't need to mock it.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
```diff
@@ -620,6 +633,70 @@ public void testExecuteGetOptimizerVersion() {
...
 
+  @Test
+  public void testExecuteSetReturnCommitStats() {
+    try (ConnectionImpl subject =
+        createConnection(
+            ConnectionOptions.newBuilder()
+                .setCredentials(NoCredentials.getInstance())
+                .setUri(URI)
+                .build())) {
+      assertThat(subject.isReturnCommitStats(), is(equalTo(false)));
```
elharo, line 644: assertFalse is simpler and easier to read

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
```diff
@@ -620,6 +633,70 @@ public void testExecuteGetOptimizerVersion() {
...
+
+      StatementResult res = subject.execute(Statement.of("set return_commit_stats=true"));
+      assertThat(res.getResultType(), is(equalTo(ResultType.NO_RESULT)));
+      assertThat(subject.isReturnCommitStats(), is(equalTo(true)));
```
elharo, line 648: assertTrue

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
```diff
@@ -620,6 +633,70 @@ public void testExecuteGetOptimizerVersion() {
...
+
+      res = subject.execute(Statement.of("set return_commit_stats=false"));
+      assertThat(res.getResultType(), is(equalTo(ResultType.NO_RESULT)));
+      assertThat(subject.isReturnCommitStats(), is(equalTo(false)));
```
elharo, line 652: assertFalse is simpler and easier to read

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
```diff
@@ -620,6 +633,70 @@ public void testExecuteGetOptimizerVersion() {
...
+  public void testExecuteSetReturnCommitStats() {
+    try (ConnectionImpl subject =
+        createConnection(
+            ConnectionOptions.newBuilder()
+                .setCredentials(NoCredentials.getInstance())
+                .setUri(URI)
+                .build())) {
+      assertThat(subject.isReturnCommitStats(), is(equalTo(false)));
+
+      StatementResult res = subject.execute(Statement.of("set return_commit_stats=true"));
```
elharo, line 646: res --> result

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
```diff
@@ -620,6 +633,70 @@ public void testExecuteGetOptimizerVersion() {
...
+
+  @Test
+  public void testExecuteSetReturnCommitStatsInvalidValue() {
+    try (ConnectionImpl subject =
+        createConnection(
+            ConnectionOptions.newBuilder()
+                .setCredentials(NoCredentials.getInstance())
+                .setUri(URI)
+                .build())) {
+      assertThat(subject.isReturnCommitStats(), is(equalTo(false)));
```
elharo, line 664: assertFalse is simpler and easier to read

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
```diff
@@ -620,6 +633,70 @@ public void testExecuteGetOptimizerVersion() {
...
+
+      try {
+        subject.execute(Statement.of("set return_commit_stats=yes"));
+        fail("Missing expected exception");
+      } catch (SpannerException e) {
+        assertThat(e.getErrorCode(), is(equalTo(ErrorCode.INVALID_ARGUMENT)));
```
elharo, line 670: If you muse use Truth, then the way to write this is

assertThat(e.getErrorCode()).isEqualTo(ErrorCode.INVALID_ARGUMENT

but frankly you're better off not even using truth for basic equality, same as, null, non-null, and true and false comparisons. Truth is better reserved for more complex comparisons that JUnit doesn't naively support.

elharo (changes requested):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
olavloite: These two methods serve two different APIs that behave differently:
* The `Connection` interface contains several methods that return the current state of the connection or of the last statement. The chosen pattern for that API is to throw an exception if the `Connection` is not in a valid state for the method to be called. I don't think we should deviate from that pattern for one method. This also somewhat reflects the standard in JDBC, where most methods also throw an exception if the `Connection` is not in a valid state for the method. See for example the [commit()](https://docs.oracle.com/javase/8/docs/api/java/sql/Connection.html#commit--) method. (Although this is not 100% true in all cases, as some other JDBC methods have a different contract, and allows you to call it in what seems to be an illogical state. Calling [setAutoCommit](https://docs.oracle.com/javase/8/docs/api/java/sql/Connection.html#setAutoCommit-boolean-) is for example allowed during a transaction, and will automatically commit the transaction __if__ the auto commit mode).
* The Connection API also contains a simple SQL parser that can be used to get and set the state of a connection. This 'API' is lenient, and will return `null` when a value is requested that is not available at that moment. This SQL parser uses the `getCommitResponseOrNull` and other similar methods.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
olavloite: This also follows the standard of the other methods in this class, as well as the default behavior in JDBC, where all methods throw an exception if it is called on a closed connection. See for example [getAutoCommit()](https://docs.oracle.com/javase/8/docs/api/java/sql/Connection.html#getAutoCommit--).

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
olavloite: It's used in two places (in the `parseReturnCommitStats` and to create the set of valid properties). But it's not something that can change in the future, as it would break existing connection URLs. I would prefer to change this in a separate PR, so the entire list of constants can be changed at once, instead of having one parameter that handled differently from the rest.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
olavloite: Fixed.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/Connection.java
olavloite: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/Connection.java
olavloite: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/Connection.java
olavloite: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/CommitResponse.java
olavloite: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
olavloite: While it would make it clearer here, it also makes it harder to get an overview of what the all the default values of a new connection is, as you would have to scroll through the different parse methods. Now the default values for all connection properties are grouped in one place.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
olavloite: True, it is just a value class, but its [constructor and static create method](https://github.com/googleapis/java-spanner/blob/66c5f88b55836abf4912e1031e504ef5d2abe9c3/google-cloud-spanner/src/main/java/com/google/cloud/spanner/CommitStats.java#L32) are not accessible from here. This test class is in the `.connection` sub-package, while `CommitStats` is in `com.google.cloud.spanner`.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
olavloite: Done.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
olavloite: Done.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
olavloite: Done.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
olavloite: Done.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
olavloite: Done.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
olavloite: This is actually not Truth, but [Hamcrest](https://github.com/googleapis/java-spanner/blob/b351d08d7567bc2a84213f1106b7e7f7fa56886e/google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java#L23). That was what this entire client library initially used, and most classes have gradually been migrated to Truth. This class has not.

I'll stick to JUnit for the simpler comparisons from now on.

(And changed this specific one to JUnit)

skuruppu:
> @skuruppu I assume it is OK to merge this one as well once it is approved?

Yes absolutely, please merge once you have all approvals.

elharo (commented):
Looks like you need to manually merge and resolve this one.

releaseplease is not yet signalling a major version update:

https://github.com/googleapis/java-spanner/pull/850

I'm checking how that works.

olavloite:
> Looks like you need to manually merge and resolve this one.
> 
> releaseplease is not yet signalling a major version update:
> 
> #850
> 
> I'm checking how that works.

I'll look into the merge conflict right away.

Could it be that the issue with releaseplease is that this PR is currently based on the branch `commit-stats2` and not `master`, as it depends on the changes that are made in the client library for `CommitStats`?

skuruppu:
@olavloite @elharo is there any update on this? The feature is getting launched in a couple of days so it will be really good to have this merged so that the JDBC driver change can be merged asap.

I guess we're going to run into an issue with the BOM dependency.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/UnitOfWork.java
olavloite: The two APIs being served are:
1. The `Connection` interface with its `getCommitResponse()` and other similar methods. These all throw an exception if they are called while the `Connection` is in an invalid state for the method.
2. The interpreted SQL interface which will return a `ResultSet` with a `null` value if there is no commit response available at that moment.

The `Connection` interface could use the `getCommitResponseOrNull()` method, check for a null value and then throw an exception. The disadvantage of that would be that the `Connection` does not know why the value is null, and can therefore only throw an exception with a generic message. By delegating the exception that should be thrown to the `UnitOfWork`, the `UnitOfWork` can throw an exception with a more specific message indicating why there is no commit response.

olavloite:
> @olavloite @elharo is there any update on this? The feature is getting launched in a couple of days so it will be really good to have this merged so that the JDBC driver change can be merged asap.
> 
> I guess we're going to run into an issue with the BOM dependency.

I've resolved all conflicts and updated the PR.

A note regarding the breaking changes: This PR adds three methods to the `com.google.cloud.spanner.connection.Connection` interface. That interface (and the entire package where it is located) is marked with `@InternalApi` and is [noted in the documentation](https://github.com/googleapis/java-spanner/blob/master/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/package-info.java) that it may make breaking changes without prior notice.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
elharo: That sounds like a good case for docs or comments.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
```diff
@@ -823,6 +836,13 @@ QueryOptions getQueryOptions() {
     return queryOptions;
   }
 
+  /**
+   * The initial returnCommitStats value for connections created by this {@link ConnectionOptions}
```
elharo, line 840: Only the initial value? If it changes, this still returns the initial value? That's surprising. Hmm, looks like it's final so no need to say initial. probably rewrite as, "whether connections created by this {@link ConnectionOptions} return commit stats"

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/DdlBatch.java
elharo: Have you considered returning an empty object or null instead?

If clients really won't see this, fine. However it's not enough that this class is non-public. They could still invoke this method if they get an instance of it, even while only knowing its supertype.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ReadWriteTransaction.java
```diff
@@ -254,19 +265,30 @@ public Timestamp getReadTimestampOrNull() {
...
+    return commitResponseFuture != null;
   }
 
   @Override
   public Timestamp getCommitTimestamp() {
-    ConnectionPreconditions.checkState(hasCommitTimestamp(), "This transaction has not committed.");
-    return get(commitTimestampFuture);
+    ConnectionPreconditions.checkState(hasCommitResponse(), "This transaction has not committed.");
```
elharo, line 274: not been committed

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ReadWriteTransaction.java
```diff
@@ -254,19 +265,30 @@ public Timestamp getReadTimestampOrNull() {
...
 
   @Override
   public Timestamp getCommitTimestampOrNull() {
-    return hasCommitTimestamp() ? get(commitTimestampFuture) : null;
+    return hasCommitResponse() ? get(commitResponseFuture).getCommitTimestamp() : null;
+  }
+
+  @Override
+  public CommitResponse getCommitResponse() {
+    ConnectionPreconditions.checkState(hasCommitResponse(), "This transaction has not committed.");
```
elharo, line 285: not been committed

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
olavloite: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
olavloite: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/DdlBatch.java
olavloite: Returning null or an empty object here would only move the logic to the `ConnectionImpl` class, which should throw an exception if it is not in a valid state to return `CommitStats` at that moment (for example because a DDL batch is active). This logic is therefore delegated to the concrete implementations of `UnitOfWork`.

This class and its super types are all package private and or not accessible from outside the library.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ReadWriteTransaction.java
olavloite: done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ReadWriteTransaction.java
olavloite: done

PR #609, opened 2020-11-08
docs: add descriptions for connection URL properties

olavloite:
Adds descriptions for the supported connection URL properties. These can be picked up by the JDBC driver and shown as descriptions for java.sql.DriverPropertyInfo.


thiagotnunes (approved):
Thanks for fixing this!

PR #611, opened 2020-11-09
test: add profile for generating test sql scripts

olavloite:
Adds a profile for regenerating the SQL test files for the Connection API. This was forgotten when moving the Connection API code from the jdbc repository to the client library repository.


thiagotnunes (approved):
LGTM

PR #616, opened 2020-11-11
fix: session retry could cause infinite wait

### LINKED ISSUE #605: RetryOnInvalidatedSessionTest seems to get stuck sometimes
```
[INFO] Running com.google.cloud.spanner.RetryOnInvalidatedSessionTest

ERROR: Aborting VM command due to timeout of 10800 seconds
```

olavloite:
A "Session not found" when using an AsyncTransactionManager could cause an infinite wait for an `ApiFuture` that would never be done.

Fixes #605

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncTransactionManagerImpl.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncTransactionManagerImpl.java
@@ -98,6 +98,7 @@ final class AsyncTransactionManagerImpl
         new ApiFutureCallback<Void>() {
           @Override
           public void onFailure(Throwable t) {
+            onError(t);
             res.setException(SpannerExceptionFactory.newSpannerException(t));
           }
 
@@ -130,6 +131,7 @@ final class AsyncTransactionManagerImpl
     }
     ApiFuture<Timestamp> res = txn.commitAsync();
     txnState = TransactionState.COMMITTED;
+
     ApiFutures.addCallback(
         res,
         new ApiFutureCallback<Timestamp>() {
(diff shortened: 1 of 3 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionContextFutureImpl.java
```diff
@@ -109,6 +109,7 @@ public Timestamp get(long timeout, TimeUnit unit)
             @Override
             public void onFailure(Throwable t) {
               mgr.onError(t);
+              statementResult.setException(t);
```
olavloite, line 112: This missing statement was what could cause the infinite wait. The result of this specific statement would never be done if the previous statement (`input`) would fail.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncTransactionManagerImpl.java
```diff
@@ -174,10 +176,6 @@ public void onSuccess(Timestamp result) {
 
   @Override
   public TransactionContextFuture resetForRetryAsync() {
-    if (txn == null || (!txn.isAborted() && txnState != TransactionState.ABORTED)) {
```
olavloite, line 177: This check is no longer needed here as it is checked in the `SessionPoolAsyncTransactionManager`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolAsyncTransactionManager.java
```diff
@@ -167,9 +171,12 @@ public void onError(Throwable t) {
   public ApiFuture<Timestamp> commitAsync() {
     synchronized (lock) {
       Preconditions.checkState(
-          txnState == TransactionState.STARTED,
+          txnState == TransactionState.STARTED || txnState == TransactionState.ABORTED,
           "commit can only be invoked if the transaction is in progress. Current state: "
               + txnState);
+      if (txnState == TransactionState.ABORTED) {
```
olavloite, line 177: This makes sure that the `AsyncTransactionManager` will return an `ApiFuture` with exception `Aborted` if the transaction has already been aborted by an asynchronous statement that has not yet been checked by the application, e.g. (simplified):
```java
txn.updateAsync(...); // This aborts the transaction, but the status of the returned ApiFuture is not checked by the application.
...
txn.commitAsync();
```

thiagotnunes (approved):

PR #623, opened 2020-11-13
feat: add support for tagging to Connection API

olavloite:
Adds support for tagging in the Connection API. This is needed to support tagging in the JDBC driver.

This is a binary breaking change, as it adds 4 methods to the `Connection` interface. That interface is marked as `@InternalApi` and warns that it may make breaking changes without prior notice.


google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
```diff
@@ -959,6 +1034,7 @@ private UnitOfWork createNewUnitOfWork() {
               .setReadOnlyStaleness(readOnlyStaleness)
               .setStatementTimeout(statementTimeout)
               .withStatementExecutor(statementExecutor)
+              .setTransactionTag(transactionTag)
```
thiagotnunes, line 1037: How will this behave if the user never set a transaction tag?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
```diff
@@ -476,6 +482,42 @@ public void setTransactionMode(TransactionMode transactionMode) {
...
+
+  @Override
+  public String getStatementTag() {
+    ConnectionPreconditions.checkState(!isClosed(), CLOSED_ERROR_MSG);
+    ConnectionPreconditions.checkState(!isDdlBatchActive(), "This connection is in a DDL batch");
+    return statementTag;
+  }
+
+  @Override
+  public void setStatementTag(String tag) {
```
thiagotnunes, line 514: Out of curiosity, could we prevent the calling of this for a `COMMIT`? The user should not be able to do a statement tag in that case.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
olavloite: It is nullable, so that is not a problem. It is verified by [this test case](https://github.com/googleapis/java-spanner/blob/6fcbdd022d9bf610a437f237814e4147cf1106db/google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/TaggingTest.java#L36).

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
olavloite: Good question. We can't prevent it before the `COMMIT`, but we can throw an exception if `COMMIT` is called after a statement tag has been set, and I think that makes sense as we are quite strict in checking the order of other statements (e.g. you are only allowed to get a commit timestamp if you actually committed etc.).

olavloite: I've changed the implementation to check for this, and to throw an error if the application tries to set a statement tag for a `COMMIT` or `ROLLBACK` statement. This change also adds a change relating to DML batches: DML batches can only include one statement tag, as the statements are sent as one `ExecuteBatchDmlRequest`, which only allows one statement tag. This statement tag must be set before calling `START BATCH DML`, e.g.

```sql
SET STATEMENT_TAG = 'tag-1';
START BATCH DML;
INSERT INTO Singers (SingerId, Name) VALUES (1, 'Morrison');
INSERT INTO Singers (SingerId, Name) VALUES (2, 'Pieterson');
RUN BATCH;
```

thiagotnunes (approved):

olavloite:
@thiagotnunes Would you mind taking a final look before I merge? I've rebased on master and realigned with the last choice that were made for tagging in the client library. Also: This is a binary breaking change as it adds methods to the `Connection` interface. Do we need a version bump for this?

PR #664, opened 2020-11-19
test: fix flaky InlineBeginTransactionTest

olavloite:
The query_ThenUpdate_ThenConsumeResultSet did not take into account that the session pool initialization will execute 4 BatchCreateSessions requests. These will normally be executed before the test transaction, but sometimes at least one of those requests might arrive after the transaction has started. That could cause the last 3 requests on the mock server to be
different from what the test expected.


thiagotnunes (approved):
Thanks for this fix!

PR #666, opened 2020-11-23
feat!: customer-managed encryption keys for Spanner

thiagotnunes:
Implements CMEK functionality for Spanner.


google-cloud-spanner/src/main/java/com/google/cloud/spanner/BackupInfo.java
```diff
@@ -18,6 +18,8 @@
 
 import com.google.api.client.util.Preconditions;
 import com.google.cloud.Timestamp;
+import com.google.spanner.admin.database.v1.EncryptionConfig;
```
olavloite, line 21: Shouldn't we be using the wrapper classes here instead of the generated classes?

olavloite (approved):
Small question on whether we should use the wrapper or proto class, otherwise LGTM.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BackupInfo.java
thiagotnunes: Oh thanks for pointing that out, I fixed it and created a wrapper class for the EncryptionInfo as well.

google-cloud-spanner/clirr-ignored-differences.xml
```diff
@@ -453,4 +453,71 @@
     <className>com/google/cloud/spanner/TransactionContext</className>
     <method>com.google.api.core.ApiFuture executeUpdateAsync(com.google.cloud.spanner.Statement)</method>
   </difference>
+
+  <!-- Support creating encrypted databases -->
+  <difference>
+    <differenceType>7004</differenceType>
```
elharo, line 459: This change will break customers and requires a major version bump. This is not optional.

elharo (changes requested):

google-cloud-spanner/clirr-ignored-differences.xml
thiagotnunes: Modified the clirr differences to include only the interface / implementation changes required for this feature. Since we need to set the encryption info we will need to break the interface here (meaning we will need a major version bump).

thiagotnunes: Marked the PR as breaking.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BackupInfo.java
```diff
@@ -28,8 +28,22 @@
 
     abstract Builder setSize(long size);
 
+    /**
+     * Output only.
+     *
+     * <p>The customer-manager encryption key version used to encrypt the backup.
```
elharo, line 34: not the key itself but only the version of the key?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BackupInfo.java
```diff
@@ -28,8 +28,22 @@
 
     abstract Builder setSize(long size);
 
+    /**
+     * Output only.
```
elharo, line 32: What does this mean?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BackupInfo.java
```diff
@@ -152,6 +186,24 @@ public long getSize() {
     return size;
   }
 
+  /**
+   * Returns the {@link EncryptionConfigInfo} to be used to encrypt the backup during it's creation.
```
elharo, line 190: delete "to be"
it's --> its

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BackupInfo.java
```diff
@@ -152,6 +186,24 @@ public long getSize() {
...
+   */
+  public EncryptionConfigInfo getEncryptionConfigInfo() {
+    return encryptionConfigInfo;
+  }
+
+  /**
+   * Returns the {@link EncryptionInfo} of the backup if the backup is encrypted, or <code>null
+   * </code> if this backup is not encrypted.
+   *
+   * @return
```
elharo, line 201: delete

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BackupInfo.java
```diff
@@ -152,6 +186,24 @@ public long getSize() {
     return size;
   }
 
+  /**
+   * Returns the {@link EncryptionConfigInfo} to be used to encrypt the backup during it's creation.
+   * Returns <code>null</code> if no customer-managed encryption key should be used.
+   */
+  public EncryptionConfigInfo getEncryptionConfigInfo() {
```
elharo, line 193: Instead of returning null is there a config info that indicates no encryption?

elharo (commented):
Is the actual key data ending up in memory here? Perhaps even dumped by a toString method? 

That is, at any point are cleartext keys in memory or on disk?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BackupInfo.java
thiagotnunes: Updated the comment to the proto comment.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BackupInfo.java
thiagotnunes: Updated the comment to the proto comment.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BackupInfo.java
thiagotnunes: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BackupInfo.java
thiagotnunes: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BackupInfo.java
thiagotnunes: Not at the moment.

thiagotnunes:
@elharo No keys are kept in memory, the new classes introduced are `EncryptionConfigInfo` and `EncryptionInfo` (from the protobuf definition), which specify a URL for the key location (`projects/*/locations/*/keyRings/*/cryptoKeys/*/cryptoKeyVersions/*`). The former is used as a request object, while the latter is a response object (this was defined in the API protos, the client library change just expose them to the user).

elharo (approved):

elharo (commented):
Integration tests failed:

Caused by: com.google.cloud.spanner.SpannerException: UNIMPLEMENTED: io.grpc.StatusRuntimeException: UNIMPLEMENTED
	at com.google.cloud.spanner.SpannerExceptionFactory.newSpannerExceptionPreformatted(SpannerExceptionFactory.java:283)
	at com.google.cloud.spanner.SpannerExceptionFactory.newSpannerException(SpannerExceptionFactory.java:61)
	at com.google.cloud.spanner.SpannerExceptionFactory.fromApiException(SpannerExceptionFactory.java:298)
	at com.google.cloud.spanner.SpannerExceptionFactory.newSpannerException(SpannerExceptionFactory.java:174)
	at com.google.cloud.spanner.SpannerExceptionFactory.newSpannerException(SpannerExceptionFactory.java:110)
	at com.google.cloud.spanner.DatabaseAdminClientImpl$4.apply(DatabaseAdminClientImpl.java:174)
	at com.google.cloud.spanner.DatabaseAdminClientImpl$4.apply(DatabaseAdminClientImpl.java:171)
	at com.google.api.core.ApiFutures$GaxFunctionToGuavaFunction.apply(ApiFutures.java:240)
	at com.google.common.util.concurrent.AbstractCatchingFuture$CatchingFuture.doFallback(AbstractCatchingFuture.java:224)
	at com.google.common.util.con

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BackupInfo.java
```diff
@@ -174,6 +217,23 @@ public long getSize() {
     return size;
   }
 
+  /**
+   * Returns the {@link BackupEncryptionConfig} to encrypt the backup during its creation. Returns
+   * <code>
+   * null</code> if no customer-managed encryption key should be used.
```
olavloite, line 223: nit: This formatting seems strange

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseInfo.java
```diff
@@ -34,6 +35,15 @@
 
     abstract Builder setEarliestVersionTime(Timestamp earliestVersionTime);
 
+    /**
+     * Optional for creating a new backup.
+     *
+     * <p>The encryption configuration to be used for the database. The only encryption, other than
+     * google's default encryption, is a customer managed encryption with a provided key. If no
```
olavloite, line 42: nit:
```suggestion
     * Google's default encryption, is a customer managed encryption with a provided key. If no
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseInfo.java
```diff
@@ -34,6 +35,15 @@
...
+     * encryption is provided, google's default encryption will be used.
```
olavloite, line 43: nit:
```suggestion
     * encryption is provided, Google's default encryption will be used.
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Restore.java
```diff
@@ -0,0 +1,109 @@
+/*
+ * Copyright 2020 Google LLC
```
olavloite, line 2:
```suggestion
 * Copyright 2021 Google LLC
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/encryption/EncryptionConfig.java
```diff
@@ -0,0 +1,26 @@
...
+package com.google.cloud.spanner.encryption;
+
+import com.google.api.core.InternalApi;
+
+/**
+ * Marker interface for encryption configurations that can be applied on databases, backups and
+ * restores.
+ */
+@InternalApi
+public interface EncryptionConfig {}
```
olavloite, line 26: I don't think we need this interface. It seems that it is only referenced by the other interfaces that extend this, but this specific marker interface is never used.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/encryption/EncryptionInfo.java
```diff
@@ -0,0 +1,92 @@
+/*
+ * Copyright 2020 Google LLC
```
olavloite, line 2:
```suggestion
 * Copyright 2021 Google LLC
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -1145,15 +1152,31 @@ public Database call() throws Exception {
...
-      throws SpannerException {
-    CreateBackupRequest request =
+      final com.google.cloud.spanner.Backup backupInfo) throws SpannerException {
+    final String instanceName = backupInfo.getInstanceId().getName();
+    final String databaseName = backupInfo.getDatabase().getName();
+    final String backupId = backupInfo.getId().getBackup();
+    final Backup.Builder backupBuilder =
+        com.google.spanner.admin.database.v1.Backup.newBuilder()
+            .setDatabase(databaseName)
+            .setExpireTime(backupInfo.getExpireTime().toProto());
```
olavloite, line 1162: I don't think it is guaranteed that `backupInfo.getExpireTime()` will always be set, so this could cause a `NullPointerException`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerRpc.java
```diff
@@ -216,26 +220,22 @@ public String getNextPageToken() {
...
   /**
-   * Creates a new backup from the source database specified in the {@link Backup} instance.
+   * Creates a new backup from the source database specified in the {@link
+   * com.google.cloud.spanner.Backup} instance.
    *
-   * @param instanceName the name of the instance where the backup should be created.
-   * @param backupId the id of the backup to create.
-   * @param backup the backup to create. The database and expireTime fields of the backup must be
-   *     filled.
+   * @param backupInfo the backup to create. The instance, database and expireTime fields of the
```
olavloite, line 226: I think we should check for these conditions in the builder for `BackupInfo` to ensure that they are actually filled. So that would mean that the `build()` method of the builder should fail if these fields have not been set.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/encryption/EncryptionConfigProtoMapperTest.java
```diff
@@ -0,0 +1,142 @@
...
+    final RestoreDatabaseEncryptionConfig actual =
+        EncryptionConfigProtoMapper.restoreDatabaseEncryptionConfig(UseBackupEncryption.INSTANCE);
+
+    assertEquals(expected, actual);
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void testRestoreDatabaseConfigInvalidEncryption() {
+    EncryptionConfigProtoMapper.restoreDatabaseEncryptionConfig(null);
+    fail("Restore database encryption config with invalid encryption should fail");
```
olavloite, line 140: nit: This line can be removed

google-cloud-spanner/src/test/java/com/google/cloud/spanner/encryption/EncryptionConfigProtoMapperTest.java
```diff
@@ -0,0 +1,142 @@
...
+    final CreateBackupEncryptionConfig actual =
+        EncryptionConfigProtoMapper.createBackupEncryptionConfig(UseDatabaseEncryption.INSTANCE);
+
+    assertEquals(expected, actual);
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void testCreateBackupInvalidEncryption() {
+    EncryptionConfigProtoMapper.createBackupEncryptionConfig(null);
+    fail("Create backup encryption config with invalid encryption should fail");
```
olavloite, line 88: nit: This line can be removed

google-cloud-spanner/src/test/java/com/google/cloud/spanner/encryption/EncryptionConfigsTest.java
```diff
@@ -0,0 +1,57 @@
...
+    final CustomerManagedEncryption actual =
+        EncryptionConfigs.customerManagedEncryption("kms-key-name");
+
+    assertEquals(expected, actual);
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void testCustomerManagedEncryptionNullKeyName() {
+    EncryptionConfigs.customerManagedEncryption(null);
+    fail("Customer managed encryption with null key should fail");
```
olavloite, line 40: nit: This line can be removed

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BackupInfo.java
thiagotnunes: Fixed.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/encryption/EncryptionConfig.java
thiagotnunes: Removed.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
thiagotnunes: We shouldn't be able to create a backup without an expire time (https://github.com/googleapis/java-spanner/blob/32fdd606f392bc97dab7f37b1c566b3954839f7e/google-cloud-spanner/src/main/java/com/google/cloud/spanner/Backup.java#L64-L65). I moved these checks into the database client instead (with illegal argument exception, instead of illegal state exceptions), so that we prevent the creation of a Backup without the necessary parameters.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/encryption/EncryptionConfigProtoMapperTest.java
thiagotnunes: Removed.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/encryption/EncryptionConfigProtoMapperTest.java
thiagotnunes: Removed.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/encryption/EncryptionConfigsTest.java
thiagotnunes: Removed.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerRpc.java
thiagotnunes: We can not do this in the builder itself, because there are times where the backup can be built without some of the necessary information. For instance:

- [backup.reload()](https://github.com/googleapis/java-spanner/blob/master/google-cloud-spanner/src/test/java/com/google/cloud/spanner/BackupTest.java#L199-L207)
- [backup.delete()](https://github.com/googleapis/java-spanner/blob/master/google-cloud-spanner/src/test/java/com/google/cloud/spanner/BackupTest.java#L209-L217)
- [backup.updateExpireTime()](https://github.com/googleapis/java-spanner/blob/master/google-cloud-spanner/src/test/java/com/google/cloud/spanner/BackupTest.java#L219-L229)

For this reason I moved the checks into the `DatabaseAdminClientImpl.createBackup` instead.

PR #675, opened 2020-11-30
chore: add DirectPath fallback integration test

mohanli-ml:
Add a DirectPath fallback test for spanner. The test is basically the same as what we have for bigtable (https://github.com/googleapis/java-bigtable/blob/master/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/it/DirectPathFallbackIT.java). The logic of the test is:

1. First wait until DirectPath traffic is observed;
2. Blackhole DirectPath ipv4 and ipv6 IP, and wait until the client fallback to use CFE;
3. Unblackhole DirectPath ipv4 and ipv6 traffic, and the client should upgrade back to use DirectPath again.

The DirectPath and CFE traffic are distinguished by peer IP.

### DIFF
--- a/google-cloud-spanner/pom.xml
+++ b/google-cloud-spanner/pom.xml
@@ -144,6 +144,16 @@
     </dependency>
     <dependency>
       <groupId>io.grpc</groupId>
+      <artifactId>grpc-alts</artifactId>
+      <scope>runtime</scope>
+    </dependency>
+    <!--
+      grpc-stub is needed directly by our tests and transitively by grpc-alts at runtime.
+      So it has to be declared as a direct dependency and to avoid overriding grpc-alts'
+      runtime requirement it has to be promoted to the runtime scope.
+    -->
+    <dependency>
+      <groupId>io.grpc</groupId>
       <artifactId>grpc-stub</artifactId>
     </dependency>
     <dependency>
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITDirectPathFallback.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITDirectPathFallback.java
@@ -1,0 +1,304 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.spanner.it;
+
+import static com.google.common.truth.Truth.assertWithMessage;
+import static com.google.common.truth.TruthJUnit.assume;
+
+import com.google.api.core.ApiFunction;
+import com.google.api.gax.core.FixedCredentialsProvider;
+import com.google.api.gax.grpc.InstantiatingGrpcChannelProvider;
+import com.google.auth.oauth2.ComputeEngineCredentials;
+import com.google.cloud.spanner.Database;
+import com.google.cloud.spanner.DatabaseClient;
+import com.google.cloud.spanner.IntegrationTestEnv;
+import com.google.cloud.spanner.Key;
+import com.google.cloud.spanner.Mutation;
+import com.google.cloud.spanner.ParallelIntegrationTest;
+import com.google.cloud.spanner.SpannerOptions;
+import com.google.cloud.spanner.TimestampBound;
+import com.google.cloud.spanner.testing.RemoteSpannerHelper;
+import com.google.common.base.Stopwatch;
+import io.grpc.ManagedChannelBuilder;
+import io.grpc.alts.ComputeEngineChannelBuilder;
+import io.grpc.netty.shaded.io.grpc.netty.NettyChannelBuilder;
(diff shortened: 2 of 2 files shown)


google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITDirectPathFallback.java
```diff
@@ -0,0 +1,304 @@
...
+    assertWithMessage("Failed to upgrade back to DirectPath").that(exerciseDirectPath()).isTrue();
+  }
+
+  private boolean exerciseDirectPath() throws InterruptedException, TimeoutException {
+    Stopwatch stopwatch = Stopwatch.createStarted();
+    numDpAddrRead.set(0);
+
+    boolean seenEnough = false;
+
+    while (!seenEnough && stopwatch.elapsed(TimeUnit.MINUTES) < 2) {
```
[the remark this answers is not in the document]
mohanli-ml: The while loop will wait at most 2 minutes to get directpath traffic. In my manual test, the client should be able to get directpath traffic within a few seconds, then `seenEnough` becomes true, and the while loop will stop.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITDirectPathFallback.java
```diff
@@ -0,0 +1,304 @@
...
+                })
+            .build());
+    // Forcefully ignore GOOGLE_APPLICATION_CREDENTIALS
+    builder.setCredentials(
+        FixedCredentialsProvider.create(ComputeEngineCredentials.create()).getCredentials());
+
+    // Create a new testHelper with the instrumented transport provider
+    try {
+      testHelper = RemoteSpannerHelper.create(builder.build(), env.getTestHelper().getInstanceId());
+    } catch (Throwable t) {
```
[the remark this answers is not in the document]
mohanli-ml: The original function https://github.com/googleapis/java-spanner/blob/master/google-cloud-spanner/src/main/java/com/google/cloud/spanner/testing/RemoteSpannerHelper.java#L158 throws a Throwable. Since we do not want to catch a Throwable, I directly throw it.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITDirectPathFallback.java
```diff
@@ -0,0 +1,305 @@
+/*
+ * Copyright 2017 Google LLC
```
olavloite, line 2: nit: 2020

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITDirectPathFallback.java
```diff
@@ -0,0 +1,305 @@
...
+  private ChannelFactory<NioSocketChannel> channelFactory;
+  private EventLoopGroup eventLoopGroup;
+  private RemoteSpannerHelper testHelper;
+
+  private static final String TABLE_NAME = "TestTable";
+  private static final List<String> ALL_COLUMNS = Arrays.asList("Key", "StringValue");
+  private static Database db;
+  private static DatabaseClient client;
+
+  private static final String DIRECT_PATH_ENDPOINT = "aa423245250f2bbf.sandbox.googleapis.com:443";
```
olavloite, line 102: I assume that this is temporary and will be replaced by a environment variable or similar at a later moment?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITDirectPathFallback.java
```diff
@@ -0,0 +1,305 @@
...
+    // Create a new testHelper with the instrumented transport provider
+    testHelper = RemoteSpannerHelper.create(builder.build(), env.getTestHelper().getInstanceId());
+
+    db =
+        testHelper.createTestDatabase(
+            "CREATE TABLE TestTable ("
+                + "  Key                STRING(MAX) NOT NULL,"
+                + "  StringValue        STRING(MAX),"
+                + ") PRIMARY KEY (Key)",
+            "CREATE INDEX TestTableByValue ON TestTable(StringValue)",
```
olavloite, line 150: nit: Could we remove these indexes? I don't think they are relevant for this test, and creating and maintaining secondary indexes are expensive operations.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITDirectPathFallback.java
mohanli-ml: Done. Thanks!

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITDirectPathFallback.java
mohanli-ml: Yes this is temporary. It will be removed once DirectPath goes to public beta, and this test will use the default Spanner endpoint. I added a TODO here.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITDirectPathFallback.java
mohanli-ml: Done. Thanks!

thiagotnunes (approved):
LGTM

olavloite (approved):

google-cloud-spanner/pom.xml
```diff
@@ -142,6 +142,16 @@
       <groupId>io.grpc</groupId>
       <artifactId>grpc-protobuf</artifactId>
     </dependency>
+    <dependency>
```
olavloite, line 145: It seems that this change is now causing this problem further down the line with the `flatten-maven-plugin`: https://github.com/googleapis/java-spanner-jdbc/pull/293#issuecomment-742471649

PR #576, opened 2020-11-03
feat: add support for tagging

mayurkale22:
## Description
Add support to enable users to set `tags` on database operations such as ***reads***, ***queries*** and ***transactions***.

## Background
The statistics of queries, reads and transactions in Cloud Spanner are stored. Providing tags for database operations will allow these statistics to be grouped by a tag and makes queries/transactions easily searchable by tag. This will help make the information provided by the statistics more useful.

# TODO
- [x] Rebase when `RequestOptions` proto changes are published.

Note: samples will be handled in a separate PR.

PR #676, opened 2020-12-01
feat: support RPC priority

olavloite:
Adds support for setting RPC priority.


olavloite:
@thiagotnunes The implementation of this feature should be synchronized with #576 (Tagging), as they both set `RequestOptions` for both individual statements as well as transactions. The most important question regarding this is whether we want to use `TransactionOption` to pass in the options for a read/write transaction, or whether we want to add a `withSomeOption` on `TransactionContext` for the different options.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
```diff
@@ -67,6 +101,11 @@ public static ReadAndQueryOption bufferRows(int bufferRows) {
     return new BufferRowsOption(bufferRows);
   }
 
+  /** Specifies the priority to use for the RPC. */
+  public static ReadQueryUpdateTransactionOption priority(RpcPriority priority) {
```
thiagotnunes, line 105: Should we add the same helper method in the tagging feature? If so, could you add this feedback in the PR?

thiagotnunes (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
olavloite: I'm not sure whether you mean literally this helper method or a similar one for tagging. The latter is already in the PR:
https://github.com/googleapis/java-spanner/blob/c3be7ce852a8d88655fe295fda080143ff40123a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java#L81

But I don't think we should add this specific helper method to the other PR. The two PR's are going to conflict, but I think that is something we should fix once one of them has been merged.

thiagotnunes: Oh sorry, I missed that from the other PR.

google-cloud-spanner/clirr-ignored-differences.xml
```diff
@@ -406,4 +406,51 @@
     <className>com/google/cloud/spanner/AbstractLazyInitializer</className>
     <method>java.lang.Object initialize()</method>
   </difference>
+  
+  <!-- RPC Priority -->
+  <difference>
```
elharo, line 411: This change will break customers and requires a major version bump. This is not optional.

elharo (changes requested):

google-cloud-spanner/clirr-ignored-differences.xml
olavloite: This is no longer a breaking change, as the breaking changes that were needed for this and other changes were grouped into a separate change to avoid having to bump the version multiple times. That change added `Options... options` varargs to the methods that need to support a variety of query / update / transaction options.

thiagotnunes (approved):

olavloite:
@elharo Would you mind taking another look at this PR? As your review requested changes, this PR is currently blocked from merging.

elharo (approved):

PR #693, opened 2020-12-09
fix: temporarily disables emulator tests

thiagotnunes:
Temporarily disables tests that are not working with the emulator.


olavloite:
@thiagotnunes Good catch. I looked into this a little bit, and there was a fundamental problem with that test case for both the emulator and real Cloud Spanner. The reason that it could go unnoticed for some time was that the test class was not annotated with a `@Category`, which meant that it was not picked up by the CI environment. I've created #706 which should fix this problem for both the emulator and real Spanner.

thiagotnunes:
Thanks for that @olavloite

PR #706, opened 2020-12-10
test: fix ITTransactionManagerAsyncTest

olavloite:
The ITTransactionManagerAsyncTest had a couple of problems:
1. The Spanner client was closed after each test, which would make only the first test to succeed.
1. The test class was not tagged with a category, which meant that it would not be executed by the CI builds. That is also the reason that the problem in 1 was not noticed.

Replaces #693

### DIFF
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITTransactionManagerAsyncTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITTransactionManagerAsyncTest.java
@@ -36,6 +36,7 @@ import com.google.cloud.spanner.IntegrationTestEnv;
 import com.google.cloud.spanner.Key;
 import com.google.cloud.spanner.KeySet;
 import com.google.cloud.spanner.Mutation;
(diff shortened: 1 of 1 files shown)


google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITTransactionManagerAsyncTest.java
```diff
@@ -224,6 +222,8 @@ public void testRollback() throws InterruptedException {
     }
   }
 
+  @Ignore(
+      "Cloud Spanner now seems to return CANCELLED instead of ABORTED when a transaction is invalidated by a later transaction in the same session")
```
olavloite, line 226: This seems to be a slight behavioral change in Cloud Spanner: Starting a new read/write transaction on a session that already has a read/write transaction will invalidate the first transaction. As far as I can tell, that would previously return an `ABORTED` error. That seems to have been changed to a `CANCELLED` error, which does make sense as it is an indication of an application error rather than a transient server error.

thiagotnunes (approved):

PR #669, opened 2020-11-24
feat: retry admin request limit exceeded error

olavloite:
Automatically retry requests that fail because the admin requests per seconds limit has been exceeded using an exponential backoff.

Fixes #655 and others

PR #707, opened 2020-12-10
docs: add sample for timeout for one RPC

olavloite:
Adds a sample for setting a timeout for a single RPC.

Fixes https://github.com/GoogleCloudPlatform/java-docs-samples/issues/3805

### DIFF
--- a/samples/snippets/src/main/java/com/example/spanner/StatementTimeoutExample.java
+++ b/samples/snippets/src/main/java/com/example/spanner/StatementTimeoutExample.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright 2020 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.example.spanner;
+
+import com.google.api.gax.grpc.GrpcCallContext;
+import com.google.api.gax.rpc.ApiCallContext;
+import com.google.cloud.spanner.DatabaseClient;
+import com.google.cloud.spanner.DatabaseId;
+import com.google.cloud.spanner.Spanner;
+import com.google.cloud.spanner.SpannerOptions;
+import com.google.cloud.spanner.SpannerOptions.CallContextConfigurator;
+import com.google.cloud.spanner.Statement;
+import com.google.cloud.spanner.TransactionContext;
+import com.google.cloud.spanner.TransactionRunner.TransactionCallable;
+import com.google.spanner.v1.SpannerGrpc;
+import java.util.concurrent.TimeUnit;
+import io.grpc.CallOptions;
+import io.grpc.Context;
+import io.grpc.MethodDescriptor;
+
+class StatementTimeoutExample {
+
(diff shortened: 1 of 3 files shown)


eaball35:
Screen Shot 2020-12-10 at 9 00 28 AM

olavloite:
@eaball35 These `RESOURCE_EXHAUSTED` errors have been mitigated by https://github.com/googleapis/java-spanner/pull/669, which is included in version 3.1.0. The sample builds are currently using version 3.0.4. I'm afraid the only possible mitigation at this moment is retrying the test run.

thiagotnunes (approved):

samples/snippets/src/main/java/com/example/spanner/StatementTimeoutExample.java
```diff
@@ -0,0 +1,82 @@
...
+    Context context =
+        Context.current().withValue(SpannerOptions.CALL_CONTEXT_CONFIGURATOR_KEY, configurator);
+    // Run the transaction in the custom context.
+    context.run(new Runnable() {
+      public void run() {
+        client.readWriteTransaction().run(new TransactionCallable<long[]>() {
+          public long[] run(TransactionContext transaction) throws Exception {
+            String sql = "INSERT Singers (SingerId, FirstName, LastName)\n"
+                + "VALUES (20, 'George', 'Washington')";
+            long rowCount = transaction.executeUpdate(Statement.of(sql));
```
lesv, line 74: How does this work in a multiprocessing environment when several processes are trying to test this at the same time?

lesv (changes requested):
Testing can't possibly ever work.

samples/snippets/src/main/java/com/example/spanner/StatementTimeoutExample.java
olavloite: Each test run always creates a database that is only used for that test run:

https://github.com/googleapis/java-spanner/blob/f794dd43665b618e51002204b651049ea5aa013b/samples/snippets/src/test/java/com/example/spanner/SpannerStandaloneExamplesIT.java#L80-L95

eaball35: Should work to create and drop db before and after each test, but dbs need be unique so they don't overlap. Will dbs be unique with current use of instanceid/databseId?

olavloite: Yes, the `databaseId` is uniquely generated for each test run by taking the system property as a base id and appending a UUID:
1. Getting the base `databaseId`: https://github.com/googleapis/java-spanner/blob/e6a34d924dff94e2f238eec63b9e62366d3cb506/samples/snippets/src/test/java/com/example/spanner/SpannerStandaloneExamplesIT.java#L52
2. Appending a UUID: https://github.com/googleapis/java-spanner/blob/2b3775a02466176695d7b88312b17c1aeedfbc16/samples/snippets/src/test/java/com/example/spanner/SpannerSampleIT.java#L439-L441

eaball35 (approved):

olavloite:
@lesv Would you mind taking a second look at this PR? The PR is currently blocked from merging as you requested changes.

lesv:
@olavloite  Apologies for the delay, what I'm concerned about is that we might test with many PR's / java version tests running at the same time, I'm concerned that you will have collisons.

I was probably stuck on table names of Singers & Venues, but that's not really the problem since you are creating an instance and then deleting it.  The code in `createTestDatabase` appears to use EnvVar, it's unclear if that's unique.  It also falls back to using the last instanceId, but that looks like it should at least have a while in there somewhere, and actually go to a next id rather than the last one to have something unique.  

I'm going to release my objection so you can move forward - again I missed the email or I'd have responded earlier.  Apologies.

If you believe I'm wrong in my concern, feel free to merge, if not could you fix it and explain where my understanding is flawed?

lesv (approved):

lesv:
I found the discussion. LGTM

PR #716, opened 2020-12-11
feat: introduce TransactionOptions and UpdateOptions

olavloite:
Adds TransactionOptions and UpdateOptions for read/write transactions and statements. These can be used in the future to specify options to affect how transactions and statements will be executed.

These changes can be used as the basis for #676 and #576 to ensure both changes use the same approach, and to prevent enormous merge conflicts.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
@@ -30,11 +30,18 @@ public final class Options implements Serializable {
   /** Marker interface to mark options applicable to read operation */
   public interface ReadOption {}
 
+  /** Marker interface to mark options applicable to Read, Query, Update and Write operations */
+  public interface ReadQueryUpdateTransactionOption
+      extends ReadOption, QueryOption, UpdateOption, TransactionOption {}
+
   /** Marker interface to mark options applicable to query operation. */
   public interface QueryOption {}
 
   /** Marker interface to mark options applicable to write operations */
   public interface TransactionOption {}
+
+  /** Marker interface to mark options applicable to update operation. */
+  public interface UpdateOption {}
 
   /** Marker interface to mark options applicable to list operations in admin API. */
   public interface ListOption {}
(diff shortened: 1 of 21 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
```diff
@@ -287,6 +294,26 @@ static Options fromQueryOptions(QueryOption... options) {
     return readOptions;
   }
 
+  static Options fromUpdateOptions(UpdateOption... options) {
```
mayurkale22, line 297: Can we add test for this?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
```diff
@@ -287,6 +294,26 @@ static Options fromQueryOptions(QueryOption... options) {
...
+    Options updateOptions = new Options();
+    for (UpdateOption option : options) {
+      if (option instanceof InternalOption) {
+        ((InternalOption) option).appendToOptions(updateOptions);
+      }
+    }
+    return updateOptions;
+  }
+
+  static Options fromTransactionOptions(TransactionOption... options) {
```
mayurkale22, line 307: Same here.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionImpl.java
```diff
@@ -340,10 +343,11 @@ public void run() {
     return res;
   }
 
-  TransactionContextImpl newTransaction() {
+  TransactionContextImpl newTransaction(Options options) {
```
mayurkale22, line 346: Can this be just `TransactionOption` instead of `Options`?

mayurkale22 (commented):
I like this PR, this will certainly simplify #676 and #576. Thank you so much for doing this!

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
olavloite: I've added a couple of (placeholder) tests for it. As this implementation does not contain any actual `UpdateOption` yet, the test case cannot check very much.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
olavloite: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionImpl.java
olavloite: That would be possible (assuming you mean `TransactionOption...`), but I would prefer to keep this as the `Options` object. This makes it a required argument, making it less likely to be missed by accident as it will cause a compile error if you don't supply an `Options` object. If we change it to `TransactionOption...` it would not cause a compile time error if we forget to pass in the options somewhere.

mayurkale22: Hmm, makes sense.

mayurkale22 (approved):

thiagotnunes (approved):
LGTM

olavloite:
@skuruppu Do you want to weigh in on this as well? It is technically a (binary) breaking change as we are adding `vararg` arguments to a number of public interface methods. Existing client application code does not need to be changed, but it breaks binary compatibility, so it needs to be recompiled.

skuruppu:
I think previously, we haven't marked changes that break binary compatibility as breaking changes so this is fine.

PR #757, opened 2020-12-16
deps: grpc-alts is only used for tests

olavloite:
Adding `grpc-alts` as a runtime dependency causes the `maven-flatten-plugin` to add the compile time dependencies in the flattened pom. As far as I can tell, `grpc-alts` is (currently) only used for testing, so adding it as a `test` scoped dependency seems to make more sense to me.

Fixes the build error in https://github.com/googleapis/java-spanner-jdbc/pull/293

### DIFF
--- a/google-cloud-spanner/pom.xml
+++ b/google-cloud-spanner/pom.xml
@@ -144,16 +144,6 @@
     </dependency>
     <dependency>
       <groupId>io.grpc</groupId>
-      <artifactId>grpc-alts</artifactId>
-      <scope>runtime</scope>
-    </dependency>
-    <!--
-      grpc-stub is needed directly by our tests and transitively by grpc-alts at runtime.
-      So it has to be declared as a direct dependency and to avoid overriding grpc-alts'
-      runtime requirement it has to be promoted to the runtime scope.
-    -->
-    <dependency>
-      <groupId>io.grpc</groupId>
       <artifactId>grpc-stub</artifactId>
     </dependency>
     <dependency>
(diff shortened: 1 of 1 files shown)


olavloite:
@mohanli-ml Is there a specific reason that `grpc-alts` was added as a `runtime` dependency instead of a `test` dependency?

thiagotnunes (commented):
@mohanli-ml Won't we need alts for the usage of direct path in production? Or is this only necessary for the tests?

olavloite:
> @mohanli-ml Won't we need alts for the usage of direct path in production? Or is this only necessary for the tests?

If it is needed for production: Why does it need to explicitly be added as a `runtime` dependency, as it is already a `compile` dependency (it's already included transitively by a couple of other dependencies)?

thiagotnunes:
@olavloite I wonder why didn't we add it in the `test` scope instead.

olavloite:
> @olavloite I wonder why didn't we add it in the `test` scope instead.

@thiagotnunes I'm not exactly sure what you mean in this case. This PR puts it in the `test` scope. The current situation is that it is explicitly added as a `runtime` dependency, and that is what is causing some trouble downstream with the JDBC driver.

thiagotnunes:
Yes, sorry, what I meant was why didn't we use the `test` scope in the first place instead of the `runtime` scope.

mohanli-ml:
@olavloite @thiagotnunes DirectPath use ALTS for authentication, but in this repo we only use it for tests, so I think use test scope should be good. Thanks!

thiagotnunes (approved):

PR #762, opened 2020-12-21
fix: throw already exists exception when op exists

thiagotnunes:
All the other libraries and the rest API return / throw an error when the same operation id is used. In java, we used to piggy back on the existing operation instead. Here, we change the behaviour to be consistent with the other implementations.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
@@ -1068,15 +1068,12 @@ public class GapicSpannerRpc implements SpannerRpc {
             } catch (ExecutionException e) {
               Throwable t = e.getCause();
               SpannerException se = SpannerExceptionFactory.asSpannerException(t);
+              if (t instanceof AlreadyExistsException) {
+                throw se;
+              }
               if (se instanceof AdminRequestsPerMinuteExceededException) {
                 // Propagate this to trigger a retry.
                 throw se;
-              }
-              if (t instanceof AlreadyExistsException) {
-                String operationName =
-                    OPERATION_NAME_TEMPLATE.instantiate(
-                        "database", databaseName, "operation", updateId);
-                return callable.resumeFutureCall(operationName, context);
               }
             }
             return operationFuture;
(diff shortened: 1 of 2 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -1068,16 +1068,13 @@ public Timestamp apply(Operation input) {
...
               Throwable t = e.getCause();
               SpannerException se = SpannerExceptionFactory.asSpannerException(t);
+              if (t instanceof AlreadyExistsException) {
+                throw se;
+              }
               if (se instanceof AdminRequestsPerMinuteExceededException) {
                 // Propagate this to trigger a retry.
                 throw se;
               }
-              if (t instanceof AlreadyExistsException) {
```
olavloite, line 1075: I agree that this looks strange, but this is actually an intended feature that has been in the client library for a very long time. The following is what is going on here:
1. End users can supply an operation ID manually in the public API: https://github.com/googleapis/java-spanner/blob/250921133e4a425e6ed872aa3dbd9e5a7dde2937/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseAdminClient.java#L194
2. The generated code and retry settings make the `UpdateDatabaseDdl` method retryable: https://github.com/googleapis/java-spanner/blob/250921133e4a425e6ed872aa3dbd9e5a7dde2937/google-cloud-spanner/src/main/java/com/google/cloud/spanner/admin/database/v1/stub/DatabaseAdminStubSettings.java#L865
3. This means that gax will automatically retry the RPC if it fails with for example `UNAVAILABLE`. If the user specified an operation ID and the `UNAVAILABLE` error occurred because the response from the server did not reach the client, the operation will already have started on the server and would cause an `AlreadyExists` error.
4. The `AlreadyExists` error from 3 should therefore not be propagated, but used to resume the tracking of the already existing operation.

Changing this would therefore technically be a breaking change as users might be relying on this behavior.

olavloite (changes requested):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
thiagotnunes: Thanks for the clarification @olavloite. Will discuss it with the team on how to proceed.

thiagotnunes:
Closing in favour of https://github.com/googleapis/java-spanner/pull/767

PR #763, opened 2020-12-23
test: add option to track the statement that starts a transaction

olavloite:
Adds an option to track the statement that starts a read/write transaction by inlining a `BeginTransaction` option with the request. This can be used to track down statements that do not return a transaction id in a timely fashon.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
@@ -60,6 +60,8 @@ import java.util.List;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -76,6 +78,7 @@ class TransactionRunnerImpl implements SessionTransaction, TransactionRunner {
     static class Builder extends AbstractReadContext.Builder<Builder, TransactionContextImpl> {
       private ByteString transactionId;
       private Options options;
+      private boolean trackTransactionStarter;
 
       private Builder() {}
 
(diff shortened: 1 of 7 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
```diff
@@ -432,6 +445,9 @@ TransactionSelector getTransactionSelector() {
             // transactionIdFuture until an actual transactionId is available.
             if (transactionIdFuture == null) {
               transactionIdFuture = SettableApiFuture.create();
+              if (trackTransactionStarter) {
+                transactionStarter = new Exception("Requesting new transaction");
```
olavloite, line 449: This will create an `Exception` with the call stack of the statement that _should_ return a transaction id. We keep track of this to be able to include it in any exception that could be thrown by later statements that timeout while waiting for this transaction id.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
```diff
@@ -447,7 +463,13 @@ TransactionSelector getTransactionSelector() {
             // Aborted error if the call that included the BeginTransaction option fails. The
             // Aborted error will cause the entire transaction to be retried, and the retry will use
             // a separate BeginTransaction RPC.
-            TransactionSelector.newBuilder().setId(tx.get()).build();
+            if (trackTransactionStarter) {
+              TransactionSelector.newBuilder()
+                  .setId(tx.get(waitForTransactionTimeoutMillis, TimeUnit.MILLISECONDS))
```
olavloite, line 468: Use a timeout if we are tracking the statement that should return a transaction id.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
```diff
@@ -456,6 +478,17 @@ TransactionSelector getTransactionSelector() {
...
           throw SpannerExceptionFactory.newSpannerException(e.getCause());
+        } catch (TimeoutException e) {
+          SpannerException se =
+              SpannerExceptionFactory.newSpannerException(
+                  ErrorCode.DEADLINE_EXCEEDED,
+                  "Timeout while waiting for a transaction to be returned by another statement. "
+                      + "See the suppressed exception for the stacktrace of the caller that should return a transaction",
+                  e);
+          if (transactionStarter != null) {
+            se.addSuppressed(transactionStarter);
```
olavloite, line 489: Include the stacktrace of the statement that should have returned the transaction in the exception that we are throwing.

thiagotnunes (approved):

PR #766, opened 2020-12-31
fix: remove time series before adding it

### LINKED ISSUE #202: Attempting to read from a non-existent database can create many distinct client IDs in the session metrics
I noticed that calling `Spanner.getDatabaseClient` with a non-existent database and then trying to read from the database can cause new client IDs to appear in the OpenCensus session metrics.  The client IDs for the non-existent database seem problematic because they can cause the size of the metrics to grow.

Here is a simplified reproduction:

google-cloud-spanner version: 1.54.0
opencensus-java version: 0.26.0

```java
import io.opencensus.exporter.stats.stackdriver.StackdriverStatsExporter;
import java.util.Arrays;
import org.junit.Test;

public class ClientIdTest {
  @Test
  public void testClientId() throws Exception {
    StackdriverStatsExporter.createAndRegister();
    spanner
        .getInstanceAdminClient()
        .createInstance(
                .setNodeCount(1)
                .build())
        .get();

    for (int i = 0; i < 100; i++) {
      try {
        spanner
            .singleUse()
      } catch (Exception e) {
        // ignore
      }
    }
    Thread.sleep(300_000);
  }
}
```

The metrics exported to Stackdriver contain 100 client IDs.  For example, here is `in_use_sessions`:

session-metrics

session-metrics-2

olavloite:
Adding the same time series multiple times will cause an error in OpenCensus. A time series could be added multiple times for the same Database client id if a database client was invalidated, for example because if was created for a database that did
not exist.

This works for OpenCensusImpl 0.26 but not for 0.24, but that seems to be the case both with and without this change. With this change the problem is solved for 0.26. In version 0.24 OpenCensusImpl will always throw an error if you try to add a metric with the same name twice.

Fixes #202

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
@@ -2370,6 +2370,7 @@ class SessionPool {
 
     // The value of a maxSessionsInUse is observed from a callback function. This function is
     // invoked whenever metrics are collected.
+    maxInUseSessionsMetric.removeTimeSeries(labelValues);
     maxInUseSessionsMetric.createTimeSeries(
         labelValues,
         this,
@@ -2382,6 +2383,7 @@ class SessionPool {
 
     // The value of a maxSessions is observed from a callback function. This function is invoked
     // whenever metrics are collected.
+    maxAllowedSessionsMetric.removeTimeSeries(labelValues);
     maxAllowedSessionsMetric.createTimeSeries(
         labelValues,
         options,
@@ -2394,6 +2396,7 @@ class SessionPool {
 
     // The value of a numWaiterTimeouts is observed from a callback function. This function is
     // invoked whenever metrics are collected.
+    sessionsTimeouts.removeTimeSeries(labelValues);
     sessionsTimeouts.createTimeSeries(
         labelValues,
         this,
@@ -2404,6 +2407,7 @@ class SessionPool {
           }
         });
 
+    numAcquiredSessionsMetric.removeTimeSeries(labelValues);
     numAcquiredSessionsMetric.createTimeSeries(
         labelValues,
         this,
@@ -2414,6 +2418,7 @@ class SessionPool {
           }
         });
 
+    numReleasedSessionsMetric.removeTimeSeries(labelValues);
     numReleasedSessionsMetric.createTimeSeries(
         labelValues,
         this,
@@ -2426,6 +2431,7 @@ class SessionPool {
 
     List<LabelValue> labelValuesWithBeingPreparedType = new ArrayList<>(labelValues);
     labelValuesWithBeingPreparedType.add(NUM_SESSIONS_BEING_PREPARED);
+    numSessionsInPoolMetric.removeTimeSeries(labelValuesWithBeingPreparedType);
     numSessionsInPoolMetric.createTimeSeries(
         labelValuesWithBeingPreparedType,
         this,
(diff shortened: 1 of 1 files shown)


olavloite:
This seems to work for OpenCensus 0.26 but not for 0.24...

sebright:
@olavloite Thanks for fixing issue #202!  I tested this PR with my application, and it worked with OpenCensus 0.26.0 as well as 0.28.0.

sebright (approved):
If I understand correctly, the issue with increasing client IDs could still occur if a different exception were thrown from https://github.com/googleapis/java-spanner/blob/c99294beb43ce1bd67cc3d12e4104641efab6710/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java#L225.  It might help to also make a separate change to the error handling in SpannerImpl.getDatabaseClient.

olavloite:
> If I understand correctly, the issue with increasing client IDs could still occur if a different exception were thrown from
> 
> https://github.com/googleapis/java-spanner/blob/c99294beb43ce1bd67cc3d12e4104641efab6710/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java#L225

That is correct.

> 
> . It might help to also make a separate change to the error handling in SpannerImpl.getDatabaseClient.

I'm somewhat reluctant to do that. As I see it, the real problem here is not that the client IDs are increasing, but the fact that `createPool` was throwing an exception. It is a method that should normally not throw any exceptions, and adding additional error handling in `SpannerImpl#getDatabaseClient(DatabaseId)` might mask other unexpected errors. Note that if the user has specified for example an invalid database name, `createPool` will not throw an exception. Instead, that error will be delayed until the user tries to get a session from the pool. Such errors that are caused by invalid input arguments will not cause the problem with ever-increasing client IDs.

sebright:
That makes sense.  I didn't realize that createPool was designed to not throw exceptions.

thiagotnunes:
@olavloite Just for my understanding, won't this remove historical data (time series data) if the user redeploys their application?

olavloite:
> @olavloite Just for my understanding, won't this remove historical data (time series data) if the user redeploys their application?

Good question. It won't be a problem after a redeploy, as these time-series are all in-memory on the local client, so these data structures will be all empty after a redeploy. But I'll take an extra look to check that it won't be a problem if a `SessionPool` is used for a while, then invalidated because of for example a deleted database, and then a new one with the same name is created.

olavloite:
> @olavloite Just for my understanding, won't this remove historical data (time series data) if the user redeploys their application?

I've been stepping through the OpenCensus implementation, and my conclusion based on that is:
1. Removing a `TimeSeries` or other metric that has not already been added in the current local deployment is a no-op. This means that in the normal case where a new `SessionPool` is created, the `removeTimeSeries(..)` calls are no-ops.
2. Removing a `TimeSeries` that has been previously added in the current JVM, will remove the local data structure from memory. This means that any data that has not yet been pushed to the server, will be lost. It will not remove any data on the server.
3. How much metrics that might be lost, depends on the exporter and how often this exports data to the server.

It is however important to realize that any in-mem sampling data will only be lost in the case that:
1. A `SessionPool` is created and can operate normally.
2. The `SessionPool` is invalidated because the underlying database or instance is deleted.
3. A new `SessionPool` is created in the same JVM for the same database that was deleted in step 2.

thiagotnunes:
Thanks for the analysis @olavloite. LGTM

thiagotnunes (approved):

PR #784, opened 2021-01-07
fix: closes pool maintainer on invalidation

### LINKED ISSUE #768: Reading from a non-existent database causes the client to send many BatchCreateSessions RPCs
I noticed that reads to a non-existent database can be very slow to fail.  I looked into the issue further and found that the client continues to send BatchCreateSessions RPCs after the read fails, so the additional BatchCreateSessions RPCs may be the cause of the slowness.

google-cloud-spanner: 3.1.2 and HEAD
grpc-java: 1.34.0

#### Reproducing slow reads

When I ran this test, the reads started out taking about 300 milliseconds.  After five minutes, the logging showed that the reads had slowed down to about ten seconds.

```java
import java.time.Duration;
import java.time.Instant;
import java.util.Arrays;
import java.util.logging.Level;
import java.util.logging.Logger;
import org.junit.Test;

public class BatchCreateSessionsTest {
  private static final Logger logger = Logger.getLogger(BatchCreateSessionsTest.class.getName());

  @Test
  public void testBatchCreateSessions() throws Exception {
    spanner
        .getInstanceAdminClient()
        .createInstance(
                .setNodeCount(1)
                .build())
        .get();

    while (true) {
      Instant start = Instant.now();
      try {
        spanner
            .singleUse()
      } catch (Exception e) {
        logger.log(Level.INFO, e.getMessage());
      } finally {
        Instant end = Instant.now();
        logger.log(Level.INFO, "Duration: " + Duration.between(start, end));
      }
    }
  }
}
```

#### Reproducing excessive calls to BatchCreateSessions

I initially noticed the high rate of BatchCreateSessions RPCs in OpenCensus RPC stats.  However, I simplified the reproduction to avoid OpenCensus and issue #202 by instead adding a log statement before https://github.com/googleapis/java-spanner/blob/90ad9d614bc1950f46d148930e06bde93aeb2098/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionClient.java#L296.  Then I modified the test case above with this diff:

```diff
@@ -30,21 +30,18 @@ public class BatchCreateSessionsTest {
                 .setNodeCount(1)
                 .build())
         .get();
     DatabaseId databaseId = DatabaseId.of(instanceId, "my-database");
 
-    while (true) {
-      Instant start = Instant.now();
+    for (int i = 0; i < 2; i++) {
       try {
         spanner
             .getDatabaseClient(databaseId)
             .singleUse()
             .readRow("MyTable", Key.of(0), Arrays.asList("MyColumn"));
       } catch (Exception e) {
         logger.log(Level.INFO, e.getMessage());
-      } finally {
-        Instant end = Instant.now();
-        logger.log(Level.INFO, "Duration: " + Duration.between(start, end));
       }
     }
+    Thread.sleep(300_000);
   }
 }
```

The output seems to show that the client sends one BatchCreateSessions RPC every ten seconds for each failed read:

```
Running BatchCreateSessionsTest
Jan 04, 2021 10:14:37 PM com.google.auth.oauth2.DefaultCredentialsProvider warnAboutProblematicCredentials
WARNING: Your application has authenticated using end user credentials from Google Cloud SDK. We recommend that most server applications use service accounts instead. If your application continues to use end user credentials from Cloud SDK, you might receive a "quota exceeded" or "API not enabled" error. For more information about service accounts, see https://cloud.google.com/docs/authentication/.
Jan 04, 2021 10:14:58 PM com.google.cloud.spanner.SessionClient internalBatchCreateSessions
INFO: Calling BatchCreateSessions with sessionCount=25
Jan 04, 2021 10:14:58 PM com.google.cloud.spanner.SessionClient internalBatchCreateSessions
INFO: Calling BatchCreateSessions with sessionCount=25
Jan 04, 2021 10:14:58 PM com.google.cloud.spanner.SessionClient internalBatchCreateSessions
INFO: Calling BatchCreateSessions with sessionCount=25
Jan 04, 2021 10:14:58 PM com.google.cloud.spanner.SessionClient internalBatchCreateSessions
INFO: Calling BatchCreateSessions with sessionCount=25
Jan 04, 2021 10:14:59 PM BatchCreateSessionsTest testBatchCreateSessions
INFO: NOT_FOUND: com.google.api.gax.rpc.NotFoundException: io.grpc.StatusRuntimeException: NOT_FOUND: Database not found: projects/<project-id>/instances/my-instance/databases/my-database
resource_type: "type.googleapis.com/google.spanner.admin.database.v1.Database"
resource_name: "projects/<project-id>/instances/my-instance/databases/my-database"
description: "Database does not exist."

Jan 04, 2021 10:14:59 PM com.google.cloud.spanner.SessionClient internalBatchCreateSessions
INFO: Calling BatchCreateSessions with sessionCount=25
Jan 04, 2021 10:14:59 PM com.google.cloud.spanner.SessionClient internalBatchCreateSessions
INFO: Calling BatchCreateSessions with sessionCount=25
Jan 04, 2021 10:14:59 PM com.google.cloud.spanner.SessionClient internalBatchCreateSessions
INFO: Calling BatchCreateSessions with sessionCount=25
Jan 04, 2021 10:14:59 PM com.google.cloud.spanner.SessionClient internalBatchCreateSessions
INFO: Calling BatchCreateSessions with sessionCount=25
Jan 04, 2021 10:14:59 PM BatchCreateSessionsTest testBatchCreateSessions
INFO: NOT_FOUND: com.google.api.gax.rpc.NotFoundException: io.grpc.StatusRuntimeException: NOT_FOUND: Database not found: projects/<project-id>/instances/my-instance/databases/my-database
resource_type: "type.googleapis.com/google.spanner.admin.database.v1.Database"
resource_name: "projects/<project-id>/instances/my-instance/databases/my-database"
description: "Database does not exist."

Jan 04, 2021 10:15:08 PM com.google.cloud.spanner.SessionClient internalBatchCreateSessions
INFO: Calling BatchCreateSessions with sessionCount=100
Jan 04, 2021 10:15:09 PM com.google.cloud.spanner.SessionClient internalBatchCreateSessions
INFO: Calling BatchCreateSessions with sessionCount=100
Jan 04, 2021 10:15:18 PM com.google.cloud.spanner.SessionClient internalBatchCreateSessions
INFO: Calling BatchCreateSessions with sessionCount=100
Jan 04, 2021 10:15:19 PM com.google.cloud.spanner.SessionClient internalBatchCreateSessions
INFO: Calling BatchCreateSessions with sessionCount=100
Jan 04, 2021 10:15:28 PM com.google.cloud.spanner.SessionClient internalBatchCreateSessions
INFO: Calling BatchCreateSessions with sessionCount=100
Jan 04, 2021 10:15:29 PM com.google.cloud.spanner.SessionClient internalBatchCreateSessions
INFO: Calling BatchCreateSessions with sessionCount=100
Jan 04, 2021 10:15:38 PM com.google.cloud.spanner.SessionClient internalBatchCreateSessions
INFO: Calling BatchCreateSessions with sessionCount=100
Jan 04, 2021 10:15:39 PM com.google.cloud.spanner.SessionClient internalBatchCreateSessions
INFO: Calling BatchCreateSessions with sessionCount=100
Jan 04, 2021 10:15:48 PM com.google.cloud.spanner.SessionClient internalBatchCreateSessions
INFO: Calling BatchCreateSessions with sessionCount=100
Jan 04, 2021 10:15:49 PM com.google.cloud.spanner.SessionClient internalBatchCreateSessions
INFO: Calling BatchCreateSessions with sessionCount=100
Jan 04, 2021 10:15:58 PM com.google.cloud.spanner.SessionClient internalBatchCreateSessions
INFO: Calling BatchCreateSessions with sessionCount=100
Jan 04, 2021 10:15:59 PM com.google.cloud.spanner.SessionClient internalBatchCreateSessions
INFO: Calling BatchCreateSessions with sessionCount=100
Jan 04, 2021 10:16:08 PM com.google.cloud.spanner.SessionClient internalBatchCreateSessions
INFO: Calling BatchCreateSessions with sessionCount=100
Jan 04, 2021 10:16:09 PM com.google.cloud.spanner.SessionClient internalBatchCreateSessions
INFO: Calling BatchCreateSessions with sessionCount=100
...
```

thiagotnunes:
When the session pool is marked as invalid, we immediately close the pool maintainer in order to keep it from trying to replenish the pool. This way we prevent useless batch create sessions requests.

Fixes #768

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
@@ -2129,7 +2129,6 @@ class SessionPool {
       }
       if (isDatabaseOrInstanceNotFound(e)) {
         setResourceNotFoundException((ResourceNotFoundException) e);
-        pendingClosure += 1;
         poolMaintainer.close();
       }
     }
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
@@ -2121,6 +2121,7 @@ class SessionPool {
       }
       if (isDatabaseOrInstanceNotFound(e)) {
         setResourceNotFoundException((ResourceNotFoundException) e);
+        poolMaintainer.close();
       }
     }
   }
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
@@ -1590,6 +1590,7 @@ class SessionPool {
     Instant lastResetTime = Instant.ofEpochMilli(0);
     int numSessionsToClose = 0;
     int sessionsToClosePerLoop = 0;
+    boolean closed = false;
 
     @GuardedBy("lock")
     ScheduledFuture<?> scheduledFuture;
@@ -1616,6 +1617,7 @@ class SessionPool {
 
     void close() {
       synchronized (lock) {
+        closed = true;
         scheduledFuture.cancel(false);
         if (!running) {
           decrementPendingClosures(1);
@@ -1623,10 +1625,16 @@ class SessionPool {
       }
     }
 
+    boolean isClosed() {
+      synchronized (lock) {
+        return closed;
+      }
+    }
+
     // Does various pool maintenance activities.
     void maintainPool() {
       synchronized (lock) {
-        if (isClosed()) {
+        if (SessionPool.this.isClosed()) {
           return;
         }
         running = true;
@@ -1638,7 +1646,7 @@ class SessionPool {
       replenishPool();
       synchronized (lock) {
         running = false;
-        if (isClosed()) {
+        if (SessionPool.this.isClosed()) {
           decrementPendingClosures(1);
         }
       }
@@ -2157,10 +2165,14 @@ class SessionPool {
       }
       closureFuture = SettableFuture.create();
       retFuture = closureFuture;
-      pendingClosure =
-          totalSessions() + numSessionsBeingCreated + 1 /* For pool maintenance thread */;
 
-      poolMaintainer.close();
+      pendingClosure = totalSessions() + numSessionsBeingCreated;
+
+      if (!poolMaintainer.isClosed()) {
+        pendingClosure += 1 /* For pool maintenance thread */;
+        poolMaintainer.close();
+      }
+
       sessions.clear();
       for (PooledSessionFuture session : checkedOutSessions) {
         if (session.leakedException != null) {
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
@@ -2129,6 +2129,7 @@ class SessionPool {
       }
       if (isDatabaseOrInstanceNotFound(e)) {
         setResourceNotFoundException((ResourceNotFoundException) e);
+        pendingClosure += 1;
         poolMaintainer.close();
       }
     }
@@ -2169,7 +2170,7 @@ class SessionPool {
       pendingClosure = totalSessions() + numSessionsBeingCreated;
 
       if (!poolMaintainer.isClosed()) {
-        pendingClosure += 1 /* For pool maintenance thread */;
+        pendingClosure += 1; // For pool maintenance thread
         poolMaintainer.close();
       }
 
@@ -2189,6 +2190,12 @@ class SessionPool {
         }
       }
     }
+
+    // Nothing to be closed, mark as complete
+    if (pendingClosure == 0) {
+      closureFuture.set(null);
+    }
+
     retFuture.addListener(
         new Runnable() {
           @Override
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
@@ -1617,10 +1617,12 @@ class SessionPool {
 
     void close() {
       synchronized (lock) {
-        closed = true;
-        scheduledFuture.cancel(false);
-        if (!running) {
-          decrementPendingClosures(1);
+        if (!closed) {
+          closed = true;
+          scheduledFuture.cancel(false);
+          if (!running) {
+            decrementPendingClosures(1);
+          }
         }
       }
     }
@@ -2188,11 +2190,11 @@ class SessionPool {
           closeSessionAsync(session);
         }
       }
-    }
 
-    // Nothing to be closed, mark as complete
-    if (pendingClosure == 0) {
-      closureFuture.set(null);
+      // Nothing to be closed, mark as complete
+      if (pendingClosure == 0) {
+        closureFuture.set(null);
+      }
     }
 
     retFuture.addListener(


thiagotnunes:
This fixes the following problem

1. When a resource not found error occurs we set the exception in the session pool [here](https://github.com/googleapis/java-spanner/blob/master/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java#L2123).
2. Because the exception was set in #1, the session pool becomes invalid as seen [here](https://github.com/googleapis/java-spanner/blob/master/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java#L1967).
3. Since the session pool is invalid, a new one will be created if we try to get the database client (see [here](https://github.com/googleapis/java-spanner/blob/master/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java#L205-L210) and [here](https://github.com/googleapis/java-spanner/blob/master/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java#L224-L226)).
4. On the invalid database client we delay closing it and buffer it to be closed when the spanner client is closed (see [here](https://github.com/googleapis/java-spanner/blob/master/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java#L206-L208)).
5. Because the invalid database client is not closed, the underlying pool maintainer is never closed ([this method](https://github.com/googleapis/java-spanner/blob/master/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java#L1617-L1624) is not called).
6. Since the method in #5 is never called the pool maintainer will try to keep maintaining the sessions (as seen [here](https://github.com/googleapis/java-spanner/blob/master/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java#L1705-L1714)) for the invalid databases.
7. The pool maintainer will try to keep creating sessions, even for the invalid session pools.
8. Since we are repeatedly calling getDatabaseClient and creating new session pools, we end up accumulating more and more pool maintainers which are always invalid and will keep trying to create new sessions.
9. The invalid session pool maintainers build up making the getDatabaseClient slower and slower.

olavloite (approved):
Nice catch!

olavloite:
@thiagotnunes  I just thought of one small catch with this:
1. When the session pool is being closed we keep track of the number of sessions and other resources that need to be closed.
2. This is [set here](https://github.com/googleapis/java-spanner/blob/3361fcefb10fae756b43be7d04c7c9b1bd145e25/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java#L2159) while executing the close method. This number also includes one for the session pool maintainer.
3. The session pool maintainer (and other resources) will [decrement](https://github.com/googleapis/java-spanner/blob/3361fcefb10fae756b43be7d04c7c9b1bd145e25/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java#L1642) the `pendingClosures` count when they actually close.
4. The `ApiFuture` that is returned by `SessionPool#close()` method will be done when `pendingClosures` reach zero. That does not happen if the pool maintainer has already been closed, so the `SessionPool` is never reported as closed.

The above is probably also the reason that a number of the tests were cancelled, because they never finished (still waiting for the session pool to close).

The solution would be to only conditionally do +1 in step 2 above for the pool maintainer. If it has already been closed because the session pool is invalid, then we should not wait for it to close.

olavloite (changes requested):
(Sorry for the flip-flopping, see my previous comment)

thiagotnunes:
@olavloite Thanks for spotting this, will work on the fix.

thiagotnunes:
@olavloite Fixed the implementation, now we only close the `poolMaintainer` if it has not been closed before. Other than that, I had to add a check at the end of the `closeAsync` method to complete the closing future, if there is nothing left to be closed (in the previous implementation that was handled by the `PoolMaintainer.close`, since it called `decrementPendingClosures`).

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2121,6 +2129,8 @@ private void handleCreateSessionsFailure(SpannerException e, int count) {
       }
       if (isDatabaseOrInstanceNotFound(e)) {
         setResourceNotFoundException((ResourceNotFoundException) e);
+        pendingClosure += 1;
```
olavloite, line 2132: This is AFAICT redundant. It will be overwritten on line 2170 when the pool is closed.

olavloite (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
thiagotnunes: The issue is that if I don't increment the `pendingClosure`, when the `PoolMaintainer.close()` executes and `running` is marked as false, it won't go into the branch on line 2145 (`closureFuture.set(null)`), since the `pendingClosure` will be `-1`. I did not want to change the condition to `pendingClosure <= 0` instead of `pendingClosure == 0` in line 2144, I preferred to increment the closures here.

olavloite: I've been trying to go through it once more, and I actually think that it is not only redundant, but also possibly a source of a `NullPointerException`. What's important here is that `closureFuture` is only defined and only relevant once the `SessionPool` is closing. It is only instantiated during the `closeAsync()` method [here](https://github.com/googleapis/java-spanner/blob/6cec2918d33d61a19772b40ea665618228a20243/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java#L2167).

So what might happen in this case is:
1. A `DatabaseNotFoundException` is thrown during the creation of new sessions.
2. The `handleCreateSessionsFailure` method sets the `pendingClosure` is to 1.
2. `poolMaintainer.close()` is called. If `running` is false, `pendingClosure` is decreased and `closureFuture.set(null)` is called. However, `closureFuture` will be `null` at this moment, unless the `SessionPool` has already been closed.

The reason that I thought it was redundant is that we are initializing the value of `pendingClosure` one [line 2170](https://github.com/googleapis/java-spanner/blob/6cec2918d33d61a19772b40ea665618228a20243/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java#L2170) and then increasing the `pendingClosure` value on [line 2173](https://github.com/googleapis/java-spanner/blob/6cec2918d33d61a19772b40ea665618228a20243/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java#L2173) if the `PoolMaintainer` is still running. So this part will ensure that `pendingClosure` will have the correct value and will reach zero once everything is closed.

Or am I missing something else here?

thiagotnunes: You are correct, an NPE is thrown in that case (verified with `ReadAsyncTest#invalidDatabase`). Removed the `pendingClosure += 1` from the code.

The weird thing is that the exception there was being swollen, because it was part of `SessionClient.BatchCreateSessionsRunnable#run` method, so I never got to see it.

thiagotnunes:
Some tests seem to be getting stuck eventually, I think I introduced a bug somewhere. Will mark it as do not merge until I had the time to investigate.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1616,17 +1617,24 @@ public void run() {
 
     void close() {
       synchronized (lock) {
+        closed = true;
         scheduledFuture.cancel(false);
         if (!running) {
           decrementPendingClosures(1);
         }
```
olavloite, line 1624: This needs to wrapped in an if-not-closed block to prevent it from being called multiple times.

```suggestion
        if (!closed) {
          closed = true;
          scheduledFuture.cancel(false);
          if (!running) {
            decrementPendingClosures(1);
          }
        }
```

The above `decrementPendingClosures(1)` call could be executed multiple times from the `handleCreateSessionsFailure` method. The problem that is associated with that manifests itself in the following case (but can also cause problems in production environments):
1. Set the `SPANNER_EMULATOR_HOST` to a valid value and start an emulator instance.
2. Execute the tests that try to use a non-existing database and then close the client.

What could happen in that case was:
1. A database client with a session pool is created.
2. The session pool tries to create sessions on the emulator. These are now in flight on the emulator (i.e. the client has not received any responses yet).
3. The client application closes the database client and session pool. As the session pool has not yet received any answers from the emulator, it assumes that 100 sessions are still being created (MinSessions=100) and that the maintainer is valid.
4. The `pendingClosure` count is therefore set to 101 during the `closeAsync` call.
5. The session creation fails and responds with a `DatabaseNotFoundException`. The `createSessions` method decreases the `pendingClosure` count with 25 for the sessions and then the `handleCreateSessionsFailure` decreases it with 1 for the pool maintainer.
6. Repeat the previous step three more times, and the value of `pendingClosure` is never exactly zero (but does reach a negative value).

olavloite (changes requested):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2176,6 +2189,12 @@ private void decrementPendingClosures(int count) {
...
       }
     }
+
+    // Nothing to be closed, mark as complete
+    if (pendingClosure == 0) {
```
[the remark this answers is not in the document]
thiagotnunes: Well spot!

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
thiagotnunes: Thanks for the detailed explanation. I have updated the code accordingly.

olavloite (approved):

PR #791, opened 2021-01-11
fix: mark transaction as invalid if no tx is returned before close

olavloite:
If a query requests the begin of a new transaction, the transaction id is returned by the first call to `ResultSet#next()`. If the `ResultSet` is closed by another thread before the first result has been returned, or before that result has been consumed internally to set the transaction id, no transaction id will be set. This will cause any subsequent statement in the same transaction
to timeout while waiting for a transaction to be returned.


thiagotnunes (approved):
Thanks so much for fixing this!

PR #797, opened 2021-01-14
fix: blanks span for session keepAlive traces

thiagotnunes:
The keepAlive traces were being tracked along with the parent span set by the client, which clutters the tracing stack. Here, we set the tracing to blank before issuing the keepAlive query.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
@@ -77,6 +77,7 @@ import io.opencensus.metrics.MetricRegistry;
 import io.opencensus.metrics.Metrics;
 import io.opencensus.trace.Annotation;
 import io.opencensus.trace.AttributeValue;
+import io.opencensus.trace.BlankSpan;
 import io.opencensus.trace.Span;
 import io.opencensus.trace.Status;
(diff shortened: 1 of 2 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1472,11 +1473,15 @@ public void prepareReadWriteTransaction() {
 
     private void keepAlive() {
       markUsed();
+      final Span previousSpan = delegate.getCurrentSpan();
```
thiagotnunes, line 1476: We could, instead of doing this, set the current span on the delegate to blank during close. I think that this is the least intrusive approach, that is why I went for it.

c24t (approved):
LGTM, and suppressing tracing for keep-alive calls seems like a clear improvement. It may be worth adding a test to check that this works as intended, and we don't export spans for these calls.

Any risk that we'll fail to trace a real `executeQuery` call (e.g. from another thread) in between calls to `setCurrentSpan` in `keepAlive`?

thiagotnunes:
@c24t good question. So, when the pool maintainer is running the keepAlive method, it removes the session from the pool temporarily. Because of this, no other thread can be using the same Session, until the keepAlive has finished.

olavloite (approved):

PR #798, opened 2021-01-14
test: fixes DatabaseClientImpl stuck test

thiagotnunes:
The DatabaseClientImpl#singleUseBoundAsync is sometimes getting stuck during the execution. This is simple to reproduce by executing the test class multiple times.
This is due to the static single threaded executor that is shared among all the test cases.
In this fix, we create an executor in the setup method, avoiding the sharing of the same among different tests.

### DIFF
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
@@ -92,7 +92,7 @@ public class DatabaseClientImplTest {
   private static final long UPDATE_COUNT = 1L;
   private Spanner spanner;
   private Spanner spannerWithEmptySessionPool;
-  private ExecutorService executor;
+  private static ExecutorService executor;
 
   @BeforeClass
   public static void startStaticServer() throws IOException {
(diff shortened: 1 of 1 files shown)


thiagotnunes:
@olavloite could you let me know if this makes sense or was the intent of the test to exercise a single shared executor among all the test cases?

olavloite:
> @olavloite could you let me know if this makes sense or was the intent of the test to exercise a single shared executor among all the test cases?

There's no specific reason that the test was using a single shared executor, but at the same time I also don't see any specific reason that it should be necessary to create a new executor for each test, which makes me worry that this fix is actually just covering some other underlying problem.

olavloite:
@thiagotnunes How are you running the test when it gets stuck? I am not able to reproduce it locally, both when running it through Maven (`mvn test -Dtest=DatabaseClientImplTest`) and when running it in Eclipse. I changed it to a [parameterized test](https://stackoverflow.com/questions/1492856/easy-way-of-running-the-same-junit-test-over-and-over) to make it easier to run it several times, but even after several hundred executions I have not been able to get it stuck.

thiagotnunes:
@olavloite that is interesting. I am running it through Intellij by changing the runner to repeat the test execution until Failure (it does not actually fail, but it gets stuck).

I can consistently reproduce this, on the second execution of the test. I will see if I have the time to do some more investigation on what's going on.

olavloite (approved):
I tried it out in the test runner in IntelliJ in the way that you suggested, and I can also reproduce it there. The problem seems to be that IntelliJ in that case will never unload the class, which again means that the executor is only initialized during the first run and then shutdown at the end. So the problem is not some underlying bug, and this fix is safe.

I would however suggest that we consider changing it in a slightly different way: Keep the executor `static` (but not `final`) and initialize it in the `startStaticServer()` method. Then we can also keep the `executor.shutdown()` in the `stopServer()` method.

thiagotnunes:
@olavloite good suggestion, this way we cover if the thread could be hanging somewhere in between tests

PR #824, opened 2021-01-28
feat: Point In Time Recovery (PITR) samples

thiagotnunes:
Adds samples for PITR.

### DIFF
--- a/samples/snippets/src/main/java/com/example/spanner/SpannerSample.java
+++ b/samples/snippets/src/main/java/com/example/spanner/SpannerSample.java
@@ -296,6 +296,44 @@ public class SpannerSample {
   }
   // [END spanner_create_database]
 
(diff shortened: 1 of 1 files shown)


thiagotnunes:
This should only be merged after we release a client library version with the PITR feature. (Before that we need to rebase over master).

olavloite:
@thiagotnunes New samples should be added as standalone samples and follow the guideline in https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/SAMPLE_FORMAT.md. The [AddNumericColumn](https://github.com/googleapis/java-spanner/blob/master/samples/snippets/src/main/java/com/example/spanner/AddNumericColumnSample.java) sample could be used as an example.

thiagotnunes:
Closes in favour of https://github.com/googleapis/java-spanner/pull/837

PR #834, opened 2021-01-30
test: unflake testCloseResultSetWhileRequestInFlight

olavloite:
* Unflakes `InlineBeginTransactionTest.testCloseResultSetWhileRequestInFlight`
* Refactors `InlineBeginTransactionTest` so that only those tests that require an executor are parameterized. This prevents a lot of tests from being executed 3 times without any difference between each test run.
* Reduces wait time while waiting for requests to be found on mock server to reduce build time.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
@@ -786,7 +786,7 @@ abstract class AbstractResultSet<R> extends AbstractStructReader implements Resu
     private final Statement statement;
 
     private SpannerRpc.StreamingCall call;
-    private boolean withBeginTransaction;
+    private volatile boolean withBeginTransaction;
     private SpannerException error;
 
     @VisibleForTesting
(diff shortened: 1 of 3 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
```diff
@@ -786,7 +786,7 @@ protected Float64Array getDoubleListInternal(int columnIndex) {
     private final Statement statement;
 
     private SpannerRpc.StreamingCall call;
-    private boolean withBeginTransaction;
+    private volatile boolean withBeginTransaction;
```
olavloite, line 789: This value is read by multiple threads (in the tests) and could therefore sometimes return a stale value without the `volatile` keyword.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
```diff
@@ -889,12 +889,6 @@ public void onError(SpannerException e) {
         error = e;
         addToStream(END_OF_STREAM);
       }
-
```
olavloite, line 892: Unused code.

thiagotnunes (approved):

PR #835, opened 2021-01-30
test: reduce wait time for BackendExhaustedTest

olavloite:
The BackendExhaustedTest deliberately causes requests to get stuck on the mock server. This causes the Spanner instance to refuse to shutdown nicely and causes it to wait for 10 seconds. This can be circumvented by using shutdownNow() to force the closure of all gRPC transport channels. This reduces the overall build time with approx 10 seconds.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
@@ -1718,6 +1718,23 @@ public class GapicSpannerRpc implements SpannerRpc {
     }
   }
 
+  public void shutdownNow() {
+    this.rpcIsClosed = true;
+    this.spannerStub.close();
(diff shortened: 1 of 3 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -1718,6 +1718,23 @@ public void shutdown() {
...
+    this.rpcIsClosed = true;
+    this.spannerStub.close();
+    this.partitionedDmlStub.close();
+    this.instanceAdminStub.close();
+    this.databaseAdminStub.close();
+    this.spannerWatchdog.shutdown();
+    this.executorProvider.shutdown();
+
+    this.spannerStub.shutdownNow();
+    ;
```
thiagotnunes, line 1731:
```suggestion

```

thiagotnunes (approved):

PR #837, opened 2021-02-01
feat: adds samples for PITR

thiagotnunes:
Adds sample for creating a database with a version retention period. Updates sample for creating a backup with a version time. Updates sample for restoring a backup, printing the backup version time.


thiagotnunes:
This should only be merged after we release a client library version with the PITR feature. (Before that we need to rebase over master).

samples/snippets/src/main/java/com/example/spanner/CreateDatabaseWithVersionRetentionPeriodSample.java
```diff
@@ -0,0 +1,83 @@
...
+    String databaseId = "my-database";
+
+    try (Spanner spanner =
+        SpannerOptions.newBuilder().setProjectId(projectId).build().getService()) {
+      DatabaseAdminClient adminClient = spanner.getDatabaseAdminClient();
+      createDatabaseWithVersionRetentionPeriod(adminClient, instanceId, databaseId);
+    }
+  }
+
+  static void createDatabaseWithVersionRetentionPeriod(DatabaseAdminClient adminClient,
```
olavloite, line 46: Should we add an integration test for this?

samples/snippets/src/main/java/com/example/spanner/CreateDatabaseWithVersionRetentionPeriodSample.java
```diff
@@ -0,0 +1,83 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.example.spanner;
+
+// [START spanner_create_database_with_version_retention_period]
```
mmmarklu, line 19: Typically, the start region comes just before the method definition. In this case, it should be line 33, so users don't have to read all the import statements.

olavloite: The formatting guideline states that `import` statements should be included within the region tags: https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/SAMPLE_FORMAT.md#class-structure

mmmarklu: Thanks Knut. Did this guidance change recently? If so, I'm curious what the rationale was. If not, why are so many of the existing samples different?

samples/snippets/src/main/java/com/example/spanner/CreateDatabaseWithVersionRetentionPeriodSample.java
thiagotnunes: Yes, I will add one

samples/snippets/src/main/java/com/example/spanner/CreateDatabaseWithVersionRetentionPeriodSample.java
olavloite: Not (very) recently. I think it has been like this for at least the last 7-8 months, and maybe also longer. But there are a lot of older samples in the Spanner repository that were created before this styleguide was available (or enforced), so I think that that is the reason that there are many samples in this repository that don't adhere to the guide. I don't know whether that is also the case for other repositories, though.

samples/snippets/src/main/java/com/example/spanner/CreateDatabaseWithVersionRetentionPeriodSample.java
thiagotnunes: Done.

olavloite (approved):

samples/snippets/src/main/java/com/example/spanner/SpannerSample.java
```diff
@@ -1589,11 +1589,14 @@ static void createBackup(
     // Set expire time to 14 days from now.
     Timestamp expireTime = Timestamp.ofTimeMicroseconds(TimeUnit.MICROSECONDS.convert(
         System.currentTimeMillis() + TimeUnit.DAYS.toMillis(14), TimeUnit.MILLISECONDS));
+    // Sets the version time to the current time.
+    Timestamp versionTime = Timestamp.now();
```
olavloite, line 1593: Considering the issues we've seen for .NET and Go, we should change this to a timestamp that we know is valid on the backend, such as `Database.getEarliestVersionTime()`.

thiagotnunes: Makes sense, I will update the samples

thiagotnunes: Done.

olavloite (approved):

thiagotnunes:
We need to wait for the snippets BOM to be updated to the version that contains the PITR changes (>= 4.0.0) before we can merge this PR.

thiagotnunes:
Do not delete the this branch, it is still referred by the docs.

PR #841, opened 2021-02-01
test: suppress logging a known warning

olavloite:
Suppress the logging of a warning to the console that a connection has been leaked, as that is exactly what the test is checking.


elharo (approved):
Thanks!

thiagotnunes (approved):

PR #842, opened 2021-02-01
test: execute unit tests in parallel when possible

olavloite:
(Experimental)

Parallelizes the execution of unit tests for the tests that support that.


google-cloud-spanner/pom.xml
```diff
@@ -50,7 +50,10 @@
           <execution>
             <id>default-test</id>
             <configuration>
-              <excludedGroups>com.google.cloud.spanner.TracerTest,com.google.cloud.spanner.IntegrationTest</excludedGroups>
+              <excludedGroups>com.google.cloud.spanner.TracerTest,com.google.cloud.spanner.IntegrationTest,com.google.cloud.spanner.SerialUnitTest</excludedGroups>
```
thiagotnunes, line 53: I wonder if we could execute some integration tests concurrently as well.

olavloite: All integration tests marked with `@ParallelIntegrationTest` are executed in parallel. See for example [ITWriteTest](https://github.com/googleapis/java-spanner/blob/58cebd85a9d82bd1526b9eae98892181f1a022f1/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITWriteTest.java#L63). It might be worthwhile to take another pass to see if there's any integration test that is currently not marked with this, that could be marked with it.

olavloite:
Superseded by #1151.

PR #871, opened 2021-02-15
fix: allows user-agent header with header provider

thiagotnunes:
A bug was introduced, where if the caller tried to set a custom user agent with a header provider an exception would be thrown (for duplicate keys). Here, we merge the user agent set by the client along with the one set by the library, instead of throwing such exception.

Fixes https://github.com/googleapis/java-spanner/issues/880

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
@@ -77,7 +77,6 @@ import com.google.common.base.Function;
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.util.concurrent.RateLimiter;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
@@ -161,6 +160,7 @@ import java.io.UnsupportedEncodingException;
 import java.net.URLDecoder;
 import java.nio.charset.StandardCharsets;
 import java.util.Comparator;
+import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


thiagotnunes:
cc @allenpradeep @mohanli-ml

mohanli-ml:
LGTM

google-cloud-spanner/src/test/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpcTest.java
```diff
@@ -502,6 +505,30 @@ public void testAdminRequestsLimitExceededRetryAlgorithm() {
...
+    final SpannerOptions options =
+        createSpannerOptions().toBuilder().setHeaderProvider(userAgentHeaderProvider).build();
+    final Spanner spanner = options.getService();
+    final DatabaseClient databaseClient =
+        spanner.getDatabaseClient(DatabaseId.of("[PROJECT]", "[INSTANCE]", "[DATABASE]"));
+    try (final ResultSet rs = databaseClient.singleUse().executeQuery(SELECT1AND2)) {
+      rs.next();
+    }
+
+    assertThat(seenHeaders.get(Key.of("user-agent", Metadata.ASCII_STRING_MARSHALLER)))
```
olavloite, line 528: nit: could we add an assertion that the default header provider is also present in the seen headers, so we prevent any silent regression from the current behavior?

olavloite (approved):

google-cloud-spanner/src/test/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpcTest.java
thiagotnunes: Good idea, I will add that.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -494,6 +489,16 @@ public GapicSpannerRpc(final SpannerOptions options) {
     }
   }
 
+  private static HeaderProvider headerProviderWithUserAgentFrom(HeaderProvider headerProvider) {
+    final Map<String, String> headersWithUserAgent = new HashMap<>(headerProvider.getHeaders());
+    final String userAgent = headersWithUserAgent.get(USER_AGENT_KEY);
```
meltsufin, line 494: @thiagotnunes `user-agent` and `User-Agent` should be equivalent.
This needs case insensitivity.

thiagotnunes: Hey there, thanks for pointing this out.

This has been fixed in the following versions:

1. [3.3.3](https://github.com/googleapis/java-spanner/blob/3.3.x/CHANGELOG.md#333-2021-05-07)
1. [4.0.3](https://github.com/googleapis/java-spanner/blob/4.0.x/CHANGELOG.md#403-2021-05-03)
1. [5.2.1](https://github.com/googleapis/java-spanner/blob/5.2.x/CHANGELOG.md#521-2021-05-03)
1. [6.4.0](https://github.com/googleapis/java-spanner/blob/master/CHANGELOG.md#640-2021-04-29)

PR #872, opened 2021-02-16
feat: add support for JSON data type

zoercai:
Allow users to read and write to Cloud Spanner databases using the JSON type through the client libraries.

Integration tests here: https://github.com/zoercai/java-spanner/pull/1

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
@@ -378,6 +378,9 @@ abstract class AbstractResultSet<R> extends AbstractStructReader implements Resu
           case STRING:
             builder.set(fieldName).to((String) value);
             break;
+          case JSON:
+            builder.set(fieldName).to(Value.json((String) value));
+            break;
           case BYTES:
             builder.set(fieldName).to((ByteArray) value);
             break;
@@ -403,6 +406,9 @@ abstract class AbstractResultSet<R> extends AbstractStructReader implements Resu
                 break;
               case STRING:
                 builder.set(fieldName).toStringArray((Iterable<String>) value);
+                break;
+              case JSON:
+                builder.set(fieldName).toJsonArray((Iterable<String>) value);
                 break;
               case BYTES:
                 builder.set(fieldName).toBytesArray((Iterable<ByteArray>) value);
@@ -480,6 +486,7 @@ abstract class AbstractResultSet<R> extends AbstractStructReader implements Resu
         case NUMERIC:
           return new BigDecimal(proto.getStringValue());
         case STRING:
+        case JSON:
           checkType(fieldType, proto, KindCase.STRING_VALUE);
           return proto.getStringValue();
         case BYTES:
@@ -543,6 +550,7 @@ abstract class AbstractResultSet<R> extends AbstractStructReader implements Resu
             return list;
           }
         case STRING:
+        case JSON:
           return Lists.transform(
               listValue.getValuesList(),
               input -> input.getKindCase() == KindCase.NULL_VALUE ? null : input.getStringValue());
@@ -651,6 +659,11 @@ abstract class AbstractResultSet<R> extends AbstractStructReader implements Resu
 
     @Override
     protected String getStringInternal(int columnIndex) {
+      return (String) rowData.get(columnIndex);
+    }
+
+    @Override
+    protected String getJsonInternal(int columnIndex) {
       return (String) rowData.get(columnIndex);
     }
 
@@ -779,6 +792,12 @@ abstract class AbstractResultSet<R> extends AbstractStructReader implements Resu
     @Override
     @SuppressWarnings("unchecked") // We know ARRAY<STRING> produces a List<String>.
     protected List<String> getStringListInternal(int columnIndex) {
+      return Collections.unmodifiableList((List<String>) rowData.get(columnIndex));
+    }
+
+    @Override
+    @SuppressWarnings("unchecked") // We know ARRAY<String> produces a List<String>.
+    protected List<String> getJsonListInternal(int columnIndex) {
       return Collections.unmodifiableList((List<String>) rowData.get(columnIndex));
     }
 
@@ -1309,6 +1328,11 @@ abstract class AbstractResultSet<R> extends AbstractStructReader implements Resu
   }
 
   @Override
+  protected String getJsonInternal(int columnIndex) {
+    return currRow().getJsonInternal(columnIndex);
+  }
+
+  @Override
   protected ByteArray getBytesInternal(int columnIndex) {
     return currRow().getBytesInternal(columnIndex);
   }
@@ -1366,6 +1390,11 @@ abstract class AbstractResultSet<R> extends AbstractStructReader implements Resu
   @Override
   protected List<String> getStringListInternal(int columnIndex) {
     return currRow().getStringListInternal(columnIndex);
+  }
+
+  @Override
+  protected List<String> getJsonListInternal(int columnIndex) {
+    return currRow().getJsonListInternal(columnIndex);
   }
 
   @Override
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractStructReader.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractStructReader.java
@@ -43,7 +43,9 @@ public abstract class AbstractStructReader implements StructReader {
 
   protected abstract String getStringInternal(int columnIndex);
 
-  protected abstract String getJsonInternal(int columnIndex);
+  protected String getJsonInternal(int columnIndex) {
+    throw new UnsupportedOperationException("Not implemented");
+  }
 
   protected abstract ByteArray getBytesInternal(int columnIndex);
 
@@ -71,7 +73,9 @@ public abstract class AbstractStructReader implements StructReader {
 
   protected abstract List<String> getStringListInternal(int columnIndex);
 
-  protected abstract List<String> getJsonListInternal(int columnIndex);
+  protected List<String> getJsonListInternal(int columnIndex) {
+    throw new UnsupportedOperationException("Not implemented");
+  }
 
   protected abstract List<ByteArray> getBytesListInternal(int columnIndex);
 
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractStructReader.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractStructReader.java
@@ -43,6 +43,8 @@ public abstract class AbstractStructReader implements StructReader {
 
   protected abstract String getStringInternal(int columnIndex);
 
+  protected abstract String getJsonInternal(int columnIndex);
+
   protected abstract ByteArray getBytesInternal(int columnIndex);
 
   protected abstract Timestamp getTimestampInternal(int columnIndex);
@@ -68,6 +70,8 @@ public abstract class AbstractStructReader implements StructReader {
   protected abstract List<BigDecimal> getBigDecimalListInternal(int columnIndex);
 
   protected abstract List<String> getStringListInternal(int columnIndex);
+
+  protected abstract List<String> getJsonListInternal(int columnIndex);
 
   protected abstract List<ByteArray> getBytesListInternal(int columnIndex);
 
@@ -160,6 +164,19 @@ public abstract class AbstractStructReader implements StructReader {
     int columnIndex = getColumnIndex(columnName);
     checkNonNullOfType(columnIndex, Type.string(), columnName);
     return getStringInternal(columnIndex);
+  }
+
+  @Override
+  public String getJson(int columnIndex) {
+    checkNonNullOfType(columnIndex, Type.json(), columnIndex);
+    return getJsonInternal(columnIndex);
+  }
+
+  @Override
+  public String getJson(String columnName) {
+    int columnIndex = getColumnIndex(columnName);
+    checkNonNullOfType(columnIndex, Type.json(), columnName);
+    return getJsonInternal(columnIndex);
   }
 
   @Override
@@ -315,6 +332,19 @@ public abstract class AbstractStructReader implements StructReader {
     int columnIndex = getColumnIndex(columnName);
     checkNonNullOfType(columnIndex, Type.array(Type.string()), columnName);
     return getStringListInternal(columnIndex);
+  }
+
+  @Override
+  public List<String> getJsonList(int columnIndex) {
+    checkNonNullOfType(columnIndex, Type.array(Type.json()), columnIndex);
+    return getJsonListInternal(columnIndex);
+  }
+
+  @Override
+  public List<String> getJsonList(String columnName) {
+    int columnIndex = getColumnIndex(columnName);
+    checkNonNullOfType(columnIndex, Type.array(Type.json()), columnName);
+    return getJsonListInternal(columnIndex);
   }
 
   @Override
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/Type.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/Type.java
@@ -46,6 +46,7 @@ public final class Type implements Serializable {
   private static final Type TYPE_FLOAT64 = new Type(Code.FLOAT64, null, null);
   private static final Type TYPE_NUMERIC = new Type(Code.NUMERIC, null, null);
   private static final Type TYPE_STRING = new Type(Code.STRING, null, null);
+  private static final Type TYPE_JSON = new Type(Code.JSON, null, null);
   private static final Type TYPE_BYTES = new Type(Code.BYTES, null, null);
   private static final Type TYPE_TIMESTAMP = new Type(Code.TIMESTAMP, null, null);
   private static final Type TYPE_DATE = new Type(Code.DATE, null, null);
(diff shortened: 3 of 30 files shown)


skuruppu:
@olavloite we're not sure why the `InlineBeginTransactionWithoutExecutorTest` is failing. Maybe a flake but not sure.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/GrpcResultSetTest.java
```diff
@@ -764,6 +764,25 @@ public void getTimestamp() {
         .isEqualTo(Timestamp.parseTimestamp("0001-01-01T00:00:00Z"));
   }
 
+  @Test
+  public void getJson() {
+    consumer.onPartialResultSet(
+        PartialResultSet.newBuilder()
+            .setMetadata(makeMetadata(Type.struct(Type.StructField.of("f", Type.json()))))
+            .addValues(Value.json(new Json("{\"color\":\"red\",\"value\":\"#f00\"}")).toProto())
```
thiagotnunes, line 772: Should we also test special UTF-8 characters?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Json.java
```diff
@@ -0,0 +1,56 @@
...
+ * limitations under the License.
+ */
+
+package com.google.cloud.spanner;
+
+import java.io.Serializable;
+import java.util.Objects;
+import javax.annotation.Nullable;
+
+public class Json implements Comparable<Json>, Serializable {
```
thiagotnunes, line 23: I think we need to add some documentation here. Especially noting that this is a tiny type and no json parsing is done, so if an input string is given it could be completely invalid JSON.

thiagotnunes:
@skuruppu @zoercai the failed integration test is a flake: https://github.com/googleapis/java-spanner/issues/869

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Json.java
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(value);
+  }
+
+  @Override
+  public int compareTo(Json other) {
+    return value.compareTo(other.value);
```
olavloite, line 54: This will throw a `NullPointerException` if `this.value` is null. The specification of `compareTo()` allows the method to throw a `NullPointerException` if `other` is null, but not if the internal value of `this` is null.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Json.java
```diff
@@ -0,0 +1,56 @@
...
+  public String toString() {
+    return this.value;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
```
olavloite, line 40: You should use `!(o instanceof Json)` instead of `getClass() != o.getClass()`. This implementation will prevent any subclass of `Json` to use `super.equals(Object)` purposefully, as it will always return false.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Json.java
```diff
@@ -0,0 +1,56 @@
...
+
+import java.io.Serializable;
+import java.util.Objects;
+import javax.annotation.Nullable;
+
+public class Json implements Comparable<Json>, Serializable {
+  public String value;
+
+  public Json(@Nullable String value) {
+    this.value = value;
```
olavloite, line 27: Should this be nullable? What is the difference between a `Json` instance with a null value, and using `null` where a Json object is expected?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Type.java
```diff
@@ -94,6 +96,10 @@ public static Type string() {
     return TYPE_STRING;
   }
 
+  public static Type json() {
```
olavloite, line 99: nit: public method should have a Javadoc.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ChecksumResultSet.java
```diff
@@ -354,6 +364,11 @@ private void funnelArray(
             into.putInt(stringValue.length());
             into.putUnencodedChars(stringValue);
             break;
+          case JSON:
+            Json json = (Json) value;
+            into.putInt(json.value.length());
```
olavloite, line 369: This will throw a `NullPointerException` if `json.value` is null (see also above)

google-cloud-spanner/src/test/java/com/google/cloud/spanner/GrpcResultSetTest.java
```diff
@@ -867,4 +886,22 @@ public void getDateList() {
     assertThat(resultSet.next()).isTrue();
     assertThat(resultSet.getDateList(0)).isEqualTo(dateList);
   }
+
+  @Test
+  public void getJsonList() {
+    List<Json> jsonList = new ArrayList<>();
+    jsonList.add(new Json("{\"color\":\"red\",\"value\":\"#f00\"}"));
+    jsonList.add(new Json("[]"));
```
olavloite, line 894: I think it would be good to also add a null value to the list to ensure that decoding and comparing null values inside of an array is also handled correctly.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Json.java
zoercai: Yeah I think it doesn't make sense for the value to be nullable. I have removed the Nullable tag and added a null check to the constructor.

zoercai:
Hi @olavloite I've removed the custom `Json` class as there were concerns with inconsistencies in the type of JSON documents that could be parsed by gson vs other languages' parsers. Now we are no longer binding the result value type to `Json` (we're just returning `String`), and for the input value type the user would need to specify `to(Value.json("json"))`. This way only the backend controls the parsing and the normalization.
Could you please take another look and let me know if you think this approach could cause issues for any use cases?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
```diff
@@ -736,6 +759,12 @@ protected Float64Array getDoubleListInternal(int columnIndex) {
       return Collections.unmodifiableList((List<String>) rowData.get(columnIndex));
     }
 
+    @Override
+    @SuppressWarnings("unchecked") // We know ARRAY<JSON> produces a List<JSON>.
```
zoercai, line 758: TODO: need to check and update the comments containing the word "JSON" still

google-cloud-spanner/src/test/java/com/google/cloud/spanner/ValueBinderTest.java
```diff
@@ -110,15 +110,23 @@ public void reflection() throws InvocationTargetException, IllegalAccessExceptio
         // Test unary null.
         if (!binderMethod.getParameterTypes()[0].isPrimitive()) {
           Value expected = (Value) method.invoke(Value.class, (Object) null);
+          // Special case for json as it maps to ValueBinder.to(String)
+          if (method.getName().toLowerCase().equals("json")) {
```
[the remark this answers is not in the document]
zoercai: Did you see https://github.com/googleapis/java-spanner/pull/872#issuecomment-817587906? :slightly_smiling_face: 

Since the type of the input into the DB will be of type Json rather than String, the backend wouldn't be affected. So it's just down to the customers' usage. The result type for JSON will be String so there won't be a difference between string and json values being returned, do you think that would cause issues for the customers?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/AbstractStructReaderTypesTest.java
```diff
@@ -238,6 +255,13 @@ public boolean isNull(int columnIndex) {
             "getStringList",
             null
           },
+          {
+            Type.array(Type.json()),
+            "getJsonListInternal",
+            Arrays.asList("{}", "{\"color\":\"red\",\"value\":\"#f00\"}", "[]"),
```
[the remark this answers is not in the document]
zoercai: I think so! It's an empty json array

google-cloud-spanner/src/test/java/com/google/cloud/spanner/ValueBinderTest.java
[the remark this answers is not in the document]
olavloite: Yes, but maybe I didn't understand it correctly (or maybe I'm confused by the `Special case for json as it maps to ValueBinder.to(String)` comment in this method). According to https://github.com/googleapis/java-spanner/pull/872#issuecomment-817587906 a user should use `to(Value.json("json"))` for JSON parameters. According to the comment in this test method, it maps to `ValueBinder.to(String)`, which would mean that the parameter would have type `String`. My worry was that that would mean that the backend would also not check that the parameter value is valid JSON.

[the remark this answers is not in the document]
zoercai: Ah I see what you mean now. The special case of `ValueBinder.to(String)` for JSON only applies to this test, because this test uses reflection to get and invoke the methods. It's comparing the output of `method` (`public static Value json(java.lang.String)`) with the output of `binderMethod` (`public Object ValueBinder.to(String)`, instead of `ValueBinder.to(Value.json(String))`) since the test uses the parameter type of the `method` to determine the `binderMethod`. This wouldn't happen when used by actual users though, as the users should do `ValueBinder.to(Value.json(String))`. 

I changed the test just now, to actually use the `to(Value)` binderMethod for the JSON case. Hopefully this makes what I'm trying to clearer. And it actually tests the JSON value binding now rather than testing String input again.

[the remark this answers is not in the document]
olavloite: Thanks for the explanation, now I understand what is going on.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/ValueBinderTest.java
```diff
@@ -110,15 +110,23 @@ public void reflection() throws InvocationTargetException, IllegalAccessExceptio
...
 
           assertThat(binder.to(expected)).isEqualTo(lastReturnValue);
           assertThat(lastValue).isEqualTo(expected);
-        }
-        // Test unary non-null.
+        } // Test unary non-null.
         Object defaultObject = DefaultValues.getDefault(method.getGenericParameterTypes()[0]);
         Value expected = (Value) method.invoke(Value.class, defaultObject);
+        // Special case for json as it maps to ValueBinder.to(String)
+        if (method.getName().toLowerCase().equals("json")) {
```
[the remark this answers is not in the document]
olavloite: (This has now been resolved)

olavloite (commented):
This looks good to me. I'll wait with the approve until the backend is also ready, so we can test it with the backend as well.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/ValueBinderTest.java
```diff
@@ -109,17 +111,32 @@ public void reflection() throws InvocationTargetException, IllegalAccessExceptio
       } else if (binderMethod.getParameterTypes().length == 1) {
         // Test unary null.
         if (!binderMethod.getParameterTypes()[0].isPrimitive()) {
+          if (method.getName().toLowerCase().equals("json")) {
```
thiagotnunes, line 114: nit: Can you extract a constant for `"json"`? Maybe `JSON_METHOD_NAME` or something alike?

thiagotnunes (commented):
LGTM, but I think we should not merge until we have some integration tests in place (when the backend is ready).

zoercai:
@thiagotnunes @olavloite Could you please take another look? I've added the integration tests [here](https://github.com/zoercai/java-spanner/pull/1), they passed when I ran them locally against staging.

olavloite:
> @thiagotnunes @olavloite Could you please take another look? I've added the integration tests [here](https://github.com/zoercai/java-spanner/pull/1), they passed when I ran them locally against staging (since it's based on this branch, I couldn't assign you as reviewers)

@zoercai 
I've added my review to the other PR.

thiagotnunes:
@zoercai also added comments on the other PR

google-cloud-spanner/src/test/java/com/google/cloud/spanner/ValueBinderTest.java
```diff
@@ -111,17 +115,32 @@ public void reflection() throws InvocationTargetException, IllegalAccessExceptio
       } else if (binderMethod.getParameterTypes().length == 1) {
         // Test unary null.
         if (!binderMethod.getParameterTypes()[0].isPrimitive()) {
+          if (method.getName().toLowerCase().equals(JSON_METHOD_NAME)) {
```
thiagotnunes, line 118: nit: `method.getName().equalsIgnoreCase(JSON_METHOD_NAME)`

google-cloud-spanner/src/test/java/com/google/cloud/spanner/ValueBinderTest.java
```diff
@@ -111,17 +115,32 @@ public void reflection() throws InvocationTargetException, IllegalAccessExceptio
...
-          assertThat(binderMethod.invoke(binder, (Object) null)).isEqualTo(lastReturnValue);
           assertThat(lastValue).isEqualTo(expected);
 
           assertThat(binder.to(expected)).isEqualTo(lastReturnValue);
           assertThat(lastValue).isEqualTo(expected);
         }
         // Test unary non-null.
-        Object defaultObject = DefaultValues.getDefault(method.getGenericParameterTypes()[0]);
+        Object defaultObject;
+        if (method.getName().toLowerCase().equals(JSON_METHOD_NAME)) {
```
thiagotnunes, line 134: nit: `method.getName().equalsIgnoreCase(JSON_METHOD_NAME)`

google-cloud-spanner/src/test/java/com/google/cloud/spanner/ValueBinderTest.java
```diff
@@ -239,6 +262,10 @@ public static Date defaultDate() {
       return Arrays.asList("a", "b");
     }
 
+    public static Iterable<String> defaultJsonIterable() {
```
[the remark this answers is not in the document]
zoercai: I don't think so, but according to IntelliJ most of the methods aren't being used anywhere (all the ones in grey):

image

so added this just in case. Maybe we can remove these in another cleanup PR?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractStructReader.java
```diff
@@ -43,6 +43,8 @@
 
   protected abstract String getStringInternal(int columnIndex);
 
+  protected abstract String getJsonInternal(int columnIndex);
```
thiagotnunes, line 46:
```suggestion
  protected String getJsonInternal(int columnIndex) {
    throw new UnsupportedOperationException("Not implemented");
  }
```

thiagotnunes: We should provide a default implementation not to break any users who have their own implementation of the interface.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractStructReader.java
```diff
@@ -69,6 +71,8 @@ protected Value getValueInternal(int columnIndex) {
 
   protected abstract List<String> getStringListInternal(int columnIndex);
 
+  protected abstract List<String> getJsonListInternal(int columnIndex);
```
thiagotnunes, line 74:
```suggestion
  protected String getJsonListInternal(int columnIndex) {
    throw new UnsupportedOperationException("Not implemented");
  }
```

thiagotnunes: We should provide a default implementation not to break any users who have their own implementation of the interface.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Value.java
```diff
@@ -441,6 +460,13 @@ private Value() {}
    */
   public abstract String getString();
 
+  /**
+   * Returns the value of a {@code JSON}-typed instance.
+   *
+   * @throws IllegalStateException if {@code isNull()} or the value is not of the expected type
+   */
+  public abstract String getJson();
```
thiagotnunes, line 468:
```suggestion
  public String getJson() {
    throw new UnsupportedOperationException("Not implemented");
  }
```

thiagotnunes: We should provide a default implementation not to break any users who have their own implementation of the interface.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Value.java
```diff
@@ -513,6 +539,14 @@ private Value() {}
    */
   public abstract List<String> getStringArray();
 
+  /**
+   * Returns the value of an {@code ARRAY<JSON>}-typed instance. While the returned list itself will
+   * never be {@code null}, elements of that list may be null.
+   *
+   * @throws IllegalStateException if {@code isNull()} or the value is not of the expected type
+   */
+  public abstract List<String> getJsonArray();
```
thiagotnunes, line 548:
```suggestion
  public List<String> getJsonArray() {
    throw new UnsupportedOperationException("Not implemented");
  }
```

thiagotnunes: We should provide a default implementation not to break any users who have their own implementation of the interface.

thiagotnunes (commented):
I think it is a good idea for us to add `JSON` support in the `MockSpannerServiceImpl` as well. See https://github.com/googleapis/java-spanner/blob/master/google-cloud-spanner/src/test/java/com/google/cloud/spanner/MockSpannerServiceImpl.java#L1175

thiagotnunes (approved):
LGTM

olavloite (approved):

PR #885, opened 2021-02-17
fix: wrong use of getRetryDelayInMillis() / 1000 in documentation and retry loops

### LINKED ISSUE #874: Thread.sleep is one 1000 times too fast
The example usage for `TransactionManager` contains an error in the retry part: The returned timeout is already in milliseconds, but the example [divides it by 1000](https://github.com/googleapis/java-spanner/blob/5af36739532037360dfd504a4a0988562550526c/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java#L335) before passing it to `Thread.sleep`.

olavloite:
Both the documentation for `TransactionManager` as well as some internal retry loops wrongly used `SpannerException#getRetryDelayInMillis() / 1000` as input for `Thread.sleep(long)`. The retry delay is already in milliseconds and should not be modified.

Fixes #874

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
@@ -533,7 +533,7 @@ class TransactionRunnerImpl implements SessionTransaction, TransactionRunner {
             SpannerExceptionFactory.newSpannerException(
                 ErrorCode.ABORTED,
                 "Aborted due to failed initial statement",
-                SpannerExceptionFactory.createAbortedExceptionWithRetry(null, e, 0, 1)));
+                SpannerExceptionFactory.createAbortedExceptionWithRetryDelay(null, e, 0, 1)));
       }
 
       if (e.getErrorCode() == ErrorCode.ABORTED) {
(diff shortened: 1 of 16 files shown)


elharo (approved):

elharo (approved):
Ick! This is worth a very quick release of 4.0.1

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
```diff
@@ -531,7 +531,9 @@ public void onError(SpannerException e, boolean withBeginTransaction) {
         // Simulate an aborted transaction to force a retry with a new transaction.
         this.transactionIdFuture.setException(
             SpannerExceptionFactory.newSpannerException(
-                ErrorCode.ABORTED, "Aborted due to failed initial statement", e));
+                ErrorCode.ABORTED,
+                "Aborted due to failed initial statement",
+                SpannerExceptionFactory.createAbortedExceptionWithRetryDelay(null, e, 0, 1)));
```
thiagotnunes, line 536: Should we add a non null message here?

thiagotnunes (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
olavloite: Yes, that's probably better to keep it in line with the other manually constructed `Aborted` errors.

PR #893, opened 2021-02-19
samples: commit stats samples

thiagotnunes:
Adds commit stats samples.

### DIFF
--- a/samples/snippets/src/main/java/com/example/spanner/GetCommitStatsSample.java
+++ b/samples/snippets/src/main/java/com/example/spanner/GetCommitStatsSample.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2021 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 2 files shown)


samples/snippets/src/main/java/com/example/spanner/GetCommitStatsSample.java
```diff
@@ -0,0 +1,71 @@
...
+
+    try (final Spanner spanner =
+        SpannerOptions.newBuilder().setProjectId(projectId).build().getService()) {
+      final DatabaseClient databaseClient = spanner
+          .getDatabaseClient(DatabaseId.of(projectId, instanceId, databaseId));
+      getCommitStats(databaseClient, table);
+    }
+  }
+
+  static void getCommitStats(DatabaseClient databaseClient, String table) {
```
olavloite, line 47: nit: I think it would be better to not have `table` as an input parameter, but to set it in the method itself. See for example https://cloud.google.com/spanner/docs/samples/spanner-update-data#spanner_update_data-java

samples/snippets/src/test/java/com/example/spanner/GetCommitStatsSampleIT.java
```diff
@@ -0,0 +1,148 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/**
+ * Integration tests for {@link GetCommitStatsSample}
+ */
+@RunWith(JUnit4.class)
+public class GetCommitStatsSampleIT {
+  private static String instanceId = System.getProperty("spanner.test.instance");
+  private static String databaseId = formatForTest(System.getProperty("spanner.sample.database", "commitstatssample"));
```
olavloite, line 50: nit: This line is longer than 100 characters and needs to be wrapped to get the format check to pass.

olavloite (approved):

thiagotnunes:
@olavloite moved to https://github.com/googleapis/java-spanner/pull/899

PR #921, opened 2021-03-01
samples: PITR samples backup fix

thiagotnunes:
Instead of using the earliest version time of the database, uses the current time (from spanner). If we used the earliest version time of the database instead we would be creating an empty backup, since the database we are backing up is not 1 hour old (default version retention period).

### DIFF
--- a/samples/snippets/src/main/java/com/example/spanner/SpannerSample.java
+++ b/samples/snippets/src/main/java/com/example/spanner/SpannerSample.java
@@ -1585,14 +1585,20 @@ public class SpannerSample {
 
   // [START spanner_create_backup]
   static void createBackup(
-      DatabaseAdminClient dbAdminClient, DatabaseId databaseId, BackupId backupId) {
+      DatabaseClient databaseClient, DatabaseAdminClient dbAdminClient, DatabaseId databaseId,
+      BackupId backupId) {
     Database databaseToBackup = dbAdminClient
         .getDatabase(databaseId.getInstanceId().getInstance(), databaseId.getDatabase());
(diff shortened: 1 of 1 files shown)


olavloite (approved):
Out of curiosity: Why are we changing this? Is this something that should also be changed for the other client libraries / tests?

samples/snippets/src/main/java/com/example/spanner/SpannerSample.java
```diff
@@ -1585,14 +1585,20 @@ static void queryWithQueryOptions(DatabaseClient dbClient) {
 
   // [START spanner_create_backup]
   static void createBackup(
-      DatabaseAdminClient dbAdminClient, DatabaseId databaseId, BackupId backupId) {
+      DatabaseClient databaseClient, DatabaseAdminClient dbAdminClient, DatabaseId databaseId,
+      BackupId backupId) {
     Database databaseToBackup = dbAdminClient
         .getDatabase(databaseId.getInstanceId().getInstance(), databaseId.getDatabase());
```
devbww, line 1591: It looks like this is unused now.

thiagotnunes:
@olavloite This is because some sample tests verify that the backup created has data. If we use the `earliest_version_time` the backup will be created empty (since the database is younger than 1 hour in the test). I missed this in the initial implementation of this sample.

samples/snippets/src/main/java/com/example/spanner/SpannerSample.java
thiagotnunes: Well spotted, fixed.

PR #950, opened 2021-03-11
fix: all throwables should be ignored in shutdown hook

### LINKED ISSUE #949: SpannerPool shutdown hook can fail with ClassNotFoundException
The shutdown hook in `SpannerPool` can in some cases fail with a `ClassNotFoundException`, for example when running Liquibase in Maven. This seems to be caused by a difference in class loader during startup and shutdown.

```
Exception in thread "SpannerPool shutdown hook" java.lang.NoClassDefFoundError: com/google/cloud/spanner/connection/SpannerPool$CheckAndCloseSpannersMode
	at com.google.cloud.spanner.connection.SpannerPool$CloseSpannerRunnable.run(SpannerPool.java:96)
	at java.base/java.lang.Thread.run(Thread.java:834)
Caused by: java.lang.ClassNotFoundException: com.google.cloud.spanner.connection.SpannerPool$CheckAndCloseSpannersMode
	at org.codehaus.plexus.classworlds.strategy.SelfFirstStrategy.loadClass(SelfFirstStrategy.java:50)
	at org.codehaus.plexus.classworlds.realm.ClassRealm.unsynchronizedLoadClass(ClassRealm.java:271)
	at org.codehaus.plexus.classworlds.realm.ClassRealm.loadClass(ClassRealm.java:247)
	at org.codehaus.plexus.classworlds.realm.ClassRealm.loadClass(ClassRealm.java:239)
	... 2 more
```

olavloite:
All throwables (and not just exceptions) should be ignored in the shutdown hook. Failing to close these resources during shutdown is not a major problem, as they will be garbage collected by the backend anyways. Without this wide catch, some
applications will log a ClassNotFoundException when shutting down, which can be confusing for end users.

Fixes #949

Also fixes https://github.com/cloudspannerecosystem/liquibase-spanner/issues/66#issuecomment-794765002

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/SpannerPool.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/SpannerPool.java
@@ -104,7 +104,7 @@ public class SpannerPool {
     public void run() {
       try {
         checkAndCloseSpanners(CheckAndCloseSpannersMode.WARN);
-      } catch (Exception e) {
+      } catch (Throwable e) {
         // ignore
       }
     }


thiagotnunes:
Why is the class loader different during initialization vs finalization?

olavloite:
> Why is the class loader different during initialization vs finalization?

I'm not sure exactly what happens, but globally speaking it seems to be caused by the fact that the shutdown hook is invoked when Maven is being shutdown, and not when the plugin is shutdown. So my hunch is the following:
1. Maven starts and runs the Liquibase plugin.
2. The Liquibase plugin dynamically loads the JDBC drivers and other services that it can find and then does its work.
3. The Liquibase plugin shuts down, but that does not shutdown the JVM.
4. Maven shuts down, which also shuts down the JVM, and that invokes the shutdown hook of the JDBC driver. But Maven does not know that the JDBC driver had been dynamically loaded by Liquibase.

Note: This is not a `finalizer` of a class, but a JVM shutdown hook, so it is only invoked when the entire JVM is shut down.

thiagotnunes:
Thanks for the explanation @olavloite

thiagotnunes (approved):

PR #951, opened 2021-03-11
fix: retry restore operation on too many pending restores

olavloite:
Retry the test restore operation if it fails because another restore operation is pending.

(Should fix) #948


thiagotnunes:
@olavloite I created https://github.com/googleapis/java-spanner/pull/956 to move the PITR backup tests with the existing backup tests. I think we can close this one.

olavloite:
> @olavloite I created #956 to move the PITR backup tests with the existing backup tests. I think we can close this one.

Yes, that's better.

PR #956, opened 2021-03-15
test: moves PITR backup tests

thiagotnunes:
Since the backup tests are very expensive, here we exercise the PITR functionality along with the existing backup tests. We also remove the dedicated backup tests for PITR.

Fixes https://github.com/googleapis/java-spanner/issues/948


google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITBackupTest.java
```diff
@@ -349,14 +352,79 @@ public void testBackups() throws InterruptedException, ExecutionException {
...
+
+    getOrThrow(dbAdminClient.createBackup(backupToCreate));
+  }
+
+  private <T> T getOrThrow(OperationFuture<T, ?> operation)
+      throws InterruptedException, ExecutionException {
+    try {
+      return operation.get();
+    } catch (ExecutionException e) {
+      if (e.getCause() != null && e.getCause() instanceof SpannerException) {
```
olavloite, line 406: nit: null-check is redundant, as `null instanceof whatever` always returns false

olavloite (approved):

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITBackupTest.java
thiagotnunes: Thanks, fixed it.

PR #959, opened 2021-03-15
Trying Guava JRE flavor to see build failure

suztomo:
Background: I'm checking repositories of the libraries in LTS can be built with other libraries in the same LTS. During preparation, I found that I cannot build this repository (java-spanner) with Guava's JRE flavor (which is to be added in LTS).

### DIFF
--- a/google-cloud-spanner/pom.xml
+++ b/google-cloud-spanner/pom.xml
@@ -222,6 +222,12 @@
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
       <version>30.1-jre</version>
+      <exclusions>
+        <exclusion>
+          <groupId>org.checkerframework</groupId>
+          <artifactId>checker-qual</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>com.google.api</groupId>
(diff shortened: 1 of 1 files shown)


suztomo:
It fails somehow https://github.com/googleapis/java-spanner/pull/959/checks?check_run_id=2117302006

```
Error:  COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
Error:  /home/runner/work/java-spanner/java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/IsRetryableInternalError.java:[24,8] 
com.google.cloud.spanner.IsRetryableInternalError is not abstract and does not override abstract method 
test(java.lang.Throwable) in java.util.function.Predicate
Error:  /home/runner/work/java-spanner/java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/IsSslHandshakeException.java:[22,8] 
com.google.cloud.spanner.IsSslHandshakeException is not abstract and does not override abstract method 
test(java.lang.Throwable) in java.util.function.Predicate
```

But this does not make sense to me because Guava's Predicate has [`default` test method](https://guava.dev/releases/30.1-jre/api/docs/com/google/common/base/Predicate.html#test(T)).

PR #984, opened 2021-03-17
feat!: add closeAsync() method to Connection

olavloite:
Adds a non-blocking close method to `Connection` to support APIs that should offer only non-blocking methods to consumers (e.g. R2DBC).

The `Connection` interface is marked with `InternalApi` and documented with `Internal connection API for Google Cloud Spanner. This interface may introduce breaking changes without prior notice.` 

cc @elefeint


elefeint (approved):
Thank you!

thiagotnunes (approved):

PR #989, opened 2021-03-18
samples: adds samples for CMEK

thiagotnunes:
Adds samples to create an encrypted database, to create an encrypted backup and to restore to an encrypted database.


samples/snippets/src/main/java/com/example/spanner/CreateBackupWithEncryptionKey.java
```diff
@@ -0,0 +1,119 @@
...
+
+public class CreateBackupWithEncryptionKey {
+
+  static void createBackupWithEncryptionKey() throws InterruptedException {
+    // TODO(developer): Replace these variables before running the sample.
+    String projectId = "my-project";
+    String instanceId = "my-instance";
+    String databaseId = "my-database";
+    String backupId = "my-backup";
+    String kmsKeyName = "my-key";
```
larkee, line 48: `kmsKeyName` is a resource path rather than an ID like the above variables. Should the placeholder be updated to reflect the expected format? For example:
```
String kmsKeyName = "projects/<project>/locations/<location>/keyRings/<keyRing>/cryptoKeys/<keyId>";
```

samples/snippets/src/main/java/com/example/spanner/CreateBackupWithEncryptionKey.java
```diff
@@ -0,0 +1,119 @@
...
+      throw SpannerExceptionFactory.propagateInterrupt(e);
+    } catch (TimeoutException e) {
+      // If the operation timed out propagates the timeout
+      throw SpannerExceptionFactory.propagateTimeout(e);
+    }
+
+    // Waits for the backup to be ready
+    int maxWaitRetries = 10;
+    int currentRetry = 0;
+    while (currentRetry < maxWaitRetries && !backup.isReady()) {
```
olavloite, line 97: Is this necessary, considering the call to `operation.get(1200, TimeUnit.SECONDS)` above? I would expect the backup to be ready once the operation has finished.

samples/snippets/src/main/java/com/example/spanner/CreateDatabaseWithEncryptionKey.java
```diff
@@ -0,0 +1,101 @@
...
+      Database createdDatabase = operation.get(120, TimeUnit.SECONDS);
+
+      System.out.printf(
+          "Database %s created with encryption key %s%n",
+          createdDatabase.getId(),
+          createdDatabase.getEncryptionConfig().getKmsKeyName()
+      );
+    } catch (ExecutionException e) {
+      // If the operation failed during execution, expose the cause.
+      throw (SpannerException) e.getCause();
```
olavloite, line 89: It could in theory be a different exception than a `SpannerException`, so this would be safer:

```suggestion
      throw SpannerExceptionFactory.asSpannerException(e.getCause());
```

samples/snippets/src/main/java/com/example/spanner/CreateBackupWithEncryptionKey.java
```diff
@@ -0,0 +1,119 @@
...
+    final OperationFuture<Backup, CreateBackupMetadata> operation = adminClient
+        .createBackup(backupToCreate);
+
+    Backup backup;
+    try {
+      System.out.println("Waiting for operation to complete...");
+      backup = operation.get(1200, TimeUnit.SECONDS);
+    } catch (ExecutionException e) {
+      // If the operation failed during execution, expose the cause.
+      throw (SpannerException) e.getCause();
```
olavloite, line 84: It could in theory be a different exception than a `SpannerException`, so this would be safer:

```suggestion
      throw SpannerExceptionFactory.asSpannerException(e.getCause());
```

samples/snippets/src/main/java/com/example/spanner/CreateBackupWithEncryptionKey.java
```diff
@@ -0,0 +1,119 @@
...
+          adminClient,
+          projectId,
+          instanceId,
+          databaseId,
+          backupId,
+          kmsKeyName);
+    }
+  }
+
+  static Void createBackupWithEncryptionKey(DatabaseAdminClient adminClient,
```
olavloite, line 63: Is there a specific reason that this method is returning `Void` instead of `void`?

samples/snippets/src/main/java/com/example/spanner/RestoreBackupWithEncryptionKey.java
```diff
@@ -0,0 +1,102 @@
...
+    final OperationFuture<Database, RestoreDatabaseMetadata> operation = adminClient
+        .restoreDatabase(restore);
+
+    Database database;
+    try {
+      System.out.println("Waiting for operation to complete...");
+      database = operation.get(1600, TimeUnit.SECONDS);
+    } catch (ExecutionException e) {
+      // If the operation failed during execution, expose the cause.
+      throw (SpannerException) e.getCause();
```
olavloite, line 78: It could in theory be a different exception than a `SpannerException`, so this would be safer:

```suggestion
      throw SpannerExceptionFactory.asSpannerException(e.getCause());
```

samples/snippets/src/test/java/com/example/spanner/DatabaseIdGenerator.java
```diff
@@ -0,0 +1,37 @@
...
+public class DatabaseIdGenerator {
+
+  private static final int DATABASE_NAME_MAX_SIZE = 30;
+  private static final String BASE_DATABASE_ID = System.getProperty(
+      "spanner.sample.database",
+      "sampletest"
+  );
+
+  static String generateDatabaseId() {
+    final String databaseId = (
```
olavloite, line 30: nit: just return the generated string, instead of first assigning it to `databaseId`.

samples/snippets/src/test/java/com/example/spanner/EncryptionKeyIT.java
```diff
@@ -0,0 +1,135 @@
...
+    String keyName = System.getProperty("spanner.test.key.name");
+    key = "projects/" + projectId + "/locations/" + keyLocation + "/keyRings/" + keyRing
+        + "/cryptoKeys/" + keyName;
+  }
+
+  @AfterClass
+  public static void tearDown() {
+    for (String databaseId : databasesToDrop) {
+      try {
+        databaseAdminClient.deleteBackup(instanceId, databaseId);
```
olavloite, line 70: This should probably be `databaseAdminClient.deleteDatabase(...)`

samples/snippets/src/test/java/com/example/spanner/EncryptionKeyIT.java
```diff
@@ -0,0 +1,135 @@
...
+  @BeforeClass
+  public static void setUp() {
+    final SpannerOptions options = SpannerOptions
+        .newBuilder()
+        .setAutoThrottleAdministrativeRequests()
+        .build();
+    projectId = options.getProjectId();
+    spanner = options.getService();
+    databaseAdminClient = spanner.getDatabaseAdminClient();
+    databasesToDrop = new ArrayList<>();
```
olavloite, line 56: nit: This can be made `final` and assigned on declaration.

samples/snippets/src/test/java/com/example/spanner/EncryptionKeyIT.java
```diff
@@ -0,0 +1,135 @@
...
+  public void testEncryptedDatabaseAndBackupAndRestore() throws Exception {
+    final String databaseId = DatabaseIdGenerator.generateDatabaseId();
+    final String backupId = DatabaseIdGenerator.generateDatabaseId();
+    final String restoreId = DatabaseIdGenerator.generateDatabaseId();
+
+    databasesToDrop.add(databaseId);
+    backupsToDrop.add(backupId);
+    databasesToDrop.add(restoreId);
+
+    String out = SampleRunner.runSample(() ->
```
olavloite, line 95: Have the samples already dropped support for Java 7? Otherwise, this unfortunately won't work....

samples/snippets/src/test/java/com/example/spanner/SampleRunner.java
```diff
@@ -0,0 +1,37 @@
...
+    final PrintStream stdOut = System.out;
+    final ByteArrayOutputStream bout = new ByteArrayOutputStream();
+    final PrintStream out = new PrintStream(bout);
+    System.setOut(out);
+    sample.call();
+    System.setOut(stdOut);
+    return bout.toString();
+  }
+
+  public interface Sample {
```
olavloite, line 34: AFAICT, this interface is not used anywhere

olavloite (commented):
More in general: This adds additional backup and restore operations to the total set of integration tests. Backup and restore operations can take a long time, so is this something that we really want?

samples/snippets/src/test/java/com/example/spanner/SampleRunner.java
thiagotnunes: Removed.

samples/snippets/src/test/java/com/example/spanner/EncryptionKeyIT.java
thiagotnunes: We have only supported java 8 from the start as far as I know (https://github.com/googleapis/java-spanner/blob/master/samples/snippets/pom.xml#L23-L24, https://github.com/googleapis/java-spanner/blob/master/samples/snapshot/pom.xml#L21-L22 and https://github.com/googleapis/java-spanner/blob/master/samples/install-without-bom/pom.xml#L21-L22)

samples/snippets/src/test/java/com/example/spanner/EncryptionKeyIT.java
thiagotnunes: Done.

samples/snippets/src/test/java/com/example/spanner/EncryptionKeyIT.java
thiagotnunes: Thanks for spotting this, it is fixed.

samples/snippets/src/test/java/com/example/spanner/DatabaseIdGenerator.java
thiagotnunes: Done.

samples/snippets/src/main/java/com/example/spanner/RestoreBackupWithEncryptionKey.java
thiagotnunes: Done.

samples/snippets/src/main/java/com/example/spanner/CreateBackupWithEncryptionKey.java
thiagotnunes: Yes, because `Callable` requires a return, so it is easier to the sample like this:

```
    String out = SampleRunner.runSample(() ->
        CreateDatabaseWithEncryptionKey.createDatabaseWithEncryptionKey(
            databaseAdminClient,
            projectId,
            instanceId,
            databaseId,
            key
        ));
```

Than like this:

```
    String out = SampleRunner.runSample(() -> {
        CreateDatabaseWithEncryptionKey.createDatabaseWithEncryptionKey(
            databaseAdminClient,
            projectId,
            instanceId,
            databaseId,
            key
        );
       return null;
    });
```

Personally I don't think it matters much on `void` vs `Void`, but if you have strong opinions I can change it.

samples/snippets/src/main/java/com/example/spanner/CreateBackupWithEncryptionKey.java
thiagotnunes: Fixed.

samples/snippets/src/main/java/com/example/spanner/CreateBackupWithEncryptionKey.java
thiagotnunes: Fixed.

thiagotnunes:
@olavloite that is a good question. I think that it is important to have these sample tests even if the backup / restore operations take a long time. However, I think we should move the samples tests to be executed in the nightly build only. This way we do not impact the day to day operation.

samples/snippets/src/main/java/com/example/spanner/CreateBackupWithEncryptionKey.java
olavloite: Cool, I didn't notice that one. Seems like a good reason.

samples/snippets/src/main/java/com/example/spanner/CreateBackupWithEncryptionKey.java
```diff
@@ -110,7 +102,7 @@ static Void createBackupWithEncryptionKey(DatabaseAdminClient adminClient,
             backup.getProto().getCreateTime().getSeconds(),
             backup.getProto().getCreateTime().getNanos(),
             OffsetDateTime.now().getOffset()),
-        ((CustomerManagedEncryption) backup.getEncryptionConfig()).getKmsKeyName()
+        backup.getEncryptionInfo().getKmsKeyVersion()
```
olavloite, line 105: According to [this](https://github.com/GoogleCloudPlatform/golang-samples/pull/2001#discussion_r596531313), this field will not necessarily be updated right away, and could return null, so it's probably not a good idea to use it here.

thiagotnunes: Changed this to be the user supplied `keyName` instead.

samples/snippets/src/main/java/com/example/spanner/CreateBackupWithEncryptionKey.java
```diff
@@ -0,0 +1,111 @@
...
+      // Throw when a thread is waiting, sleeping, or otherwise occupied,
+      // and the thread is interrupted, either before or during the activity.
+      throw SpannerExceptionFactory.propagateInterrupt(e);
+    } catch (TimeoutException e) {
+      // If the operation timed out propagates the timeout
+      throw SpannerExceptionFactory.propagateTimeout(e);
+    }
+
+    // Reload the metadata of the backup from the server.
+    backup = backup.reload();
```
olavloite, line 95: nit: This should not be necessary, as the most recent metadata should have been returned by the `operation.get()` call.

samples/snippets/src/main/java/com/example/spanner/RestoreBackupWithEncryptionKey.java
```diff
@@ -0,0 +1,102 @@
...
+      // Throw when a thread is waiting, sleeping, or otherwise occupied,
+      // and the thread is interrupted, either before or during the activity.
+      throw SpannerExceptionFactory.propagateInterrupt(e);
+    } catch (TimeoutException e) {
+      // If the operation timed out propagates the timeout
+      throw SpannerExceptionFactory.propagateTimeout(e);
+    }
+
+    // Reload the metadata of the database from the server.
+    database = database.reload();
```
olavloite, line 89: nit: same as above, this should not be necessary, as the most recent metadata should have been returned by the operation.

olavloite (approved):

samples/snippets/src/main/java/com/example/spanner/CreateBackupWithEncryptionKey.java
thiagotnunes: Thanks, fixed.

samples/snippets/src/main/java/com/example/spanner/RestoreBackupWithEncryptionKey.java
thiagotnunes: Thanks, fixed.

larkee (approved):

PR #1011, opened 2021-03-23
docs: improve error messages

olavloite:
Improve error messages when something is wrong in the connection string.

Fixes https://github.com/googleapis/java-spanner-jdbc/issues/399

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
@@ -775,7 +775,7 @@ public class ConnectionOptions {
         }
       } catch (UnavailableException e) {
         String msg;
-        if (emulatorHost == null) {
+        if (getHost() != null) {
           msg =
               String.format(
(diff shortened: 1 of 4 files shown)


olavloite:
cc @jimmyruska

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
```diff
@@ -731,6 +740,65 @@ static String checkValidProperties(String uri) {
     return res;
   }
 
+  /**
+   * Executes a quick check to see if this connection can actually connect to a local emulator host
+   * or other (mock) test server, if the options point to localhost instead of Cloud Spanner.
+   */
+  private void checkLocalConnection() {
```
thiagotnunes, line 747: Could we extract a class for this? I feel this class is very big already.

olavloite: Good point, I've moved it to a separate class.

thiagotnunes (approved):

PR #1012, opened 2021-03-23
Update cloudshell editor link and runtime directory

### LINKED ISSUE #1010: Open in Cloud Shell for sample uses the wrong repo
Repro:
  1. Click 'Open In Cloud Shell' from the samples readme: https://github.com/googleapis/java-spanner/tree/master/samples

Expected: it opens in cloud shell with the java-spanner repo(including samples)
Actual: it opens the GoogleCloudPlatform/java-docs-samples repo

elefeint:
This will get the correct repo to clone, and instruct the user to switch directories to run commands, but I am not sure the fix is sufficient.

The samples (`SpannerSample.java`, `QuickstartSample.java`, `BatchSample.java`) use the default project, which seems to be unset when cloud shell opens. Perhaps they should be modified to take project ID as a runtime argument, too?

Fixes #1010


thiagotnunes (approved):
Thanks for fixing this!

PR #1020, opened 2021-03-30
fix: plain text when testing emulator connection

thiagotnunes:
Uses plain text, instead of ssl when testing for local connections. This fixes connection checks when using the `MockSpannerServiceImpl` and it works with the emulator as well.

Without this fix we would get a transport error when testing the `getConnection` when using the `MockSpannerServiceImpl`, since the client would expect `TLS`, while the server would be serving `plainText`.


olavloite (approved):
Thanks for noticing that one!

PR #1030, opened 2021-03-31
feat: Support query hints for DML statements

### LINKED ISSUE #1029: spanner-jdbc: Support query hints for DML statements
New statement hint LOCK_SCANNED_RANGE is introduced to allow client transaction to request an exclusive lock on a set of ranges scanned by a transaction. This statement hint could be applied to both SELECT and UPDATE statements. 

However, the current StatementParser only allow statement hints to exist in SELECT statement but not in UPDATE statements. Doing so in UPDATE statements will result in the following exception:

`com.google.cloud.spanner.jdbc.JdbcSqlExceptionFactory$JdbcSqlExceptionImpl: INVALID_ARGUMENT: Unknown statement: @{LOCK_SCANNED_RANGES=exclusive} UPDATE ...`

Need to update the StatementParser to include support to this new hint in UPDATE statements.

ktchana:
Update StatementParser to include statement hints support for DML statements.

The Spanner JDBC driver would consider UPDATE statements that started with a statement hint as invalid statements. This change adds a check for statement hints at the beginning of a query, and accepts these as valid queries.

Fixes #1029 ☕️

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/StatementParser.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/StatementParser.java
@@ -24,6 +24,7 @@ import com.google.cloud.spanner.connection.ClientSideStatementImpl.CompileExcept
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Sets;
 import com.google.spanner.v1.ExecuteSqlRequest.QueryOptions;
 import java.util.Collections;
 import java.util.Objects;
@@ -457,17 +458,15 @@ public class StatementParser {
     // searching for the first occurrence of a keyword that should be preceded by a closing curly
     // brace at the end of the statement hint.
     int startStatementHintIndex = sql.indexOf('{');
-    // Statement hints are only allowed for queries.
+    // Statement hints are allowed for both queries and DML statements.
     int startQueryIndex = -1;
     String upperCaseSql = sql.toUpperCase();
-    for (String keyword : selectStatements) {
+    Set<String> selectAndDmlStatements =
+        Sets.union(selectStatements, dmlStatements).immutableCopy();
+    for (String keyword : selectAndDmlStatements) {
       startQueryIndex = upperCaseSql.indexOf(keyword);
-      if (startQueryIndex > -1) break;
-    }
-    if (startQueryIndex <= -1) {
-      for (String keyword : dmlStatements) {
-        startQueryIndex = upperCaseSql.indexOf(keyword);
-        if (startQueryIndex > -1) break;
+      if (startQueryIndex > -1) {
+        break;
       }
     }
     if (startQueryIndex > -1) {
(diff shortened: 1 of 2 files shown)


anantdamle:
@ktchana LGTM. Please fix the commit comment as per [conventionalcommits](https://github.com/googleapis/java-spanner/pull/1030/checks?check_run_id=2235767438) test.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/StatementParser.java
```diff
@@ -460,6 +464,12 @@ static String removeStatementHint(String sql) {
       startQueryIndex = upperCaseSql.indexOf(keyword);
       if (startQueryIndex > -1) break;
     }
+    if (startQueryIndex <= -1) {
```
anantdamle, line 467: instead of this `if` statement
you can probably make changes in the previous for-loop as follows:

```
Set<String> selectAndDmlStatements = Sets.union(selectStatements, dmlStatements).immutableCopy();
for (String keyword : selectAndDmlStatements) {
```
for (String keyword :

olavloite: Please apply this suggestion, as there is no need to do this in two separate loops. Also, please update the comment on line 460 to reflect the fact that statement hints are also supported for DML statements.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/StatementParser.java
```diff
@@ -460,6 +464,12 @@ static String removeStatementHint(String sql) {
       startQueryIndex = upperCaseSql.indexOf(keyword);
       if (startQueryIndex > -1) break;
```
olavloite, line 465: nit: I know this was not changed in this PR, but please apply the following as well, as the Google style guide dictates that all `if` statements should always use curly braces.

```suggestion
      if (startQueryIndex > -1) {
        break;
      }
```

olavloite (commented):
Thanks for the contribution and noticing this missing feature. This looks mostly good. PTAL at the comment given by @anantdamle .

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/StatementParser.java
ktchana: Updated both the for-loop and comment.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/StatementParser.java
ktchana: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/StatementParser.java
```diff
@@ -457,17 +458,15 @@ static String removeStatementHint(String sql) {
     // searching for the first occurrence of a keyword that should be preceded by a closing curly
     // brace at the end of the statement hint.
     int startStatementHintIndex = sql.indexOf('{');
-    // Statement hints are only allowed for queries.
+    // Statement hints are allowed for both queries and DML statements.
     int startQueryIndex = -1;
     String upperCaseSql = sql.toUpperCase();
-    for (String keyword : selectStatements) {
+    Set<String> selectAndDmlStatements =
+        Sets.union(selectStatements, dmlStatements).immutableCopy();
```
anantdamle, line 465: if style allows, consider moving the declaration and assignment on a single line.

ktchana: The line was split by the formatter so I guess this is what is needed to be style-conformant.

olavloite (approved):
LGTM

PR #1036, opened 2021-04-06
fix: local connection checker ignores exceptions

thiagotnunes:
The local connection checker should only check for unavailable exceptions when warning the user about the spanner emulator configuration. It should ignore all other kinds of errors.

### DIFF
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/LocalConnectionCheckerTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/LocalConnectionCheckerTest.java
@@ -17,39 +17,50 @@ package com.google.cloud.spanner.connection;
 
 import static com.google.cloud.spanner.MockSpannerTestUtil.SELECT1;
 import static com.google.cloud.spanner.MockSpannerTestUtil.SELECT1_RESULTSET;
+import static org.junit.Assert.assertEquals;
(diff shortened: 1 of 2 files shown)


google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/LocalConnectionCheckerTest.java
```diff
@@ -63,4 +74,23 @@ public void localConnectionCheckerWorksWithMockSpanner() {
       while (resultSet.next()) {}
     }
   }
+
+  @Test
+  public void testNoRunningEmulator() {
+    final String uri =
+        "cloudspanner://localhost:42424/projects/proj/instances/inst/databases/db?usePlainText=true";
```
olavloite, line 81: nit: It might be safer to use `server.getPort() - 1` instead of `42424`, as there is a 1/65000 chance that the port assigned to the mock server in this test happens to be `42424`.

olavloite (approved):

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/LocalConnectionCheckerTest.java
thiagotnunes: Good point, updated it.

PR #1042, opened 2021-04-07
samples: add samples for tagging feature

mayurkale22:
Adds a sample for assigning and querying tags. 

Follow-up to https://github.com/googleapis/java-spanner/pull/576

Note: We will add the test when the feature is fully available.

### DIFF
--- a/samples/snippets/src/main/java/com/example/spanner/TransactionWithTagSample.java
+++ b/samples/snippets/src/main/java/com/example/spanner/TransactionWithTagSample.java
@@ -30,32 +30,18 @@ public class TransactionWithTagSample {
 
   // [START spanner_set_transaction_and_request_tags]
   static void taggedTransaction(DatabaseClient databaseClient) {
-    // The request below would have transaction_tag set as "app=spanner,env=test".
+    // The request below would have transaction_tag set as "app=cart,env=dev".
     databaseClient
-        .readWriteTransaction(Options.tag("app=spanner,env=test"))
+        .readWriteTransaction(Options.tag("app=cart,env=dev"))
         .run(
             new TransactionCallable<Void>() {
               @Override
               public Void run(TransactionContext transaction) throws Exception {
-                // The request below would have request_tag set as
-                // "app=spanner,env=test,action=select".
-                ResultSet queryResultSet = transaction
-                    .executeQuery(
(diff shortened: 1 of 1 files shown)


samples/snippets/src/main/java/com/example/spanner/TransactionWithTagSample.java
```diff
@@ -0,0 +1,80 @@
...
+                        .set("FirstName").to("George")
+                        .set("LastName").to("Washington")
+                        .build());
+                return null;
+              }
+            });
+  }
+  // [END spanner_set_transaction_and_request_tags]
+
+  // [START spanner_query_tags]
```
thiagotnunes, line 65: This one looks like the client spec.

thiagotnunes (commented):
It seems that the first sample `spanner_set_transaction_and_request_tags` does not entirely match the one we have in the client spec.

Is that true, or did we decide to change directions @larkee ?

samples/snippets/src/main/java/com/example/spanner/TransactionWithTagSample.java
```diff
@@ -0,0 +1,80 @@
...
+  static void queryStats(DatabaseClient databaseClient) {
+    // Execute query on Query statistics
+    // see: https://cloud.google.com/spanner/docs/introspection/query-statistics, for more details.
+    String sql =
+        "SELECT t.REQUEST_TAG, t.AVG_LATENCY_SECONDS, t.AVG_CPU_SECONDS "
```
olavloite, line 70: As far as I can tell, `REQUEST_TAG` is not yet available in (public) Cloud Spanner databases. This should therefore not be merged until the column is available.

Furthermore: The `t.` prefix here does not work. There's no alias `t` in the query below.

mayurkale22:
> It seems that the first sample `spanner_set_transaction_and_request_tags` does not entirely match the one we have in the client spec.
> 
> Is that true, or did we decide to change directions @larkee ?

Updated the sample wrt the client spec. PTAL

thiagotnunes:
@mayurkale22 could you add some tests to this sample? We would like to make sure everything is working correctly before merging it.

mayurkale22:
> @mayurkale22 could you add some tests to this sample? We would like to make sure everything is working correctly before merging it.

Sure, I will add test soon. I will wait until the new column becomes visible for tests.

Issue #1043, opened 2021-04-07
Compile error found in "samples/snippets"


HailongWen:
I followed the [Getting started with Cloud Spanner in Java](https://cloud.google.com/spanner/docs/getting-started/java#prepare_your_local_environment) and did the following:
```
$ git clone https://github.com/googleapis/java-spanner.git
$ cd java-spanner/samples/snippets
$ mvn package
```

However, the compilation failed. Note that I did the same thing a while ago and it did work. So I guess this is caused by recent releases that include incompatible changes?

FYI, the errors are:
```
[INFO] -------------------------------------------------------------
[WARNING] COMPILATION WARNING : 
[INFO] -------------------------------------------------------------
[WARNING] /tmp/java-spanner/samples/snippets/src/main/java/com/example/spanner/TracingSample.java: /tmp/java-spanner/samples/snippets/src/main/java/com/example/spanner/TracingSample.java uses or overrides a deprecated API.
[WARNING] /tmp/java-spanner/samples/snippets/src/main/java/com/example/spanner/TracingSample.java: Recompile with -Xlint:deprecation for details.
[INFO] 2 warnings 
[INFO] -------------------------------------------------------------
[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /tmp/java-spanner/samples/snippets/src/main/java/com/example/spanner/CreateBackupWithEncryptionKey.java:[30,43] package com.google.cloud.spanner.encryption does not exist
[ERROR] /tmp/java-spanner/samples/snippets/src/main/java/com/example/spanner/RestoreBackupWithEncryptionKey.java:[26,32] cannot find symbol
  symbol:   class Restore
  location: package com.google.cloud.spanner
[ERROR] /tmp/java-spanner/samples/snippets/src/main/java/com/example/spanner/RestoreBackupWithEncryptionKey.java:[30,43] package com.google.cloud.spanner.encryption does not exist
[ERROR] /tmp/java-spanner/samples/snippets/src/main/java/com/example/spanner/CreateDatabaseWithEncryptionKey.java:[28,43] package com.google.cloud.spanner.encryption does not exist
[ERROR] /tmp/java-spanner/samples/snippets/src/main/java/com/example/spanner/CreateBackupWithEncryptionKey.java:[72,30] cannot find symbol
  symbol:   variable EncryptionConfigs
  location: class com.example.spanner.CreateBackupWithEncryptionKey
[...]

PR #1044, opened 2021-04-08
revert: adds samples for CMEK

olavloite:
See #1043

Temporarily reverts the addition of these samples, as encryption is not yet available in the Spanner client that is included in the BOM. Encryption requires Spanner client library version 6.x. The current BOM version is 5.1.0.

Reverts googleapis/java-spanner#989


thiagotnunes:
Hum that is a good catch, but we can not revert these, otherwise the documentation will break.

thiagotnunes:
We updated the bom version now: https://github.com/googleapis/java-spanner/commit/dcfe398e57eaa3c45184e2187a81a8b65c1995c5. This PR should be unnecessary.

PR #1066, opened 2021-04-15
feat: transaction callable as functional interface

thiagotnunes:
Makes the `TransactionCallable` a `@FunctionalInterface` so it can be used with lambdas.

This PR also modifies all of its usages to the lambda syntax. Only syntatic changes were made, no logic changes were made here.


olavloite (approved):
LGTM.

We should also update the examples in the Javadoc in the DatabaseClient interface (and possibly also other places), but I think we can do that in a separate PR.

PR #1067, opened 2021-04-15
chore: uses lambda syntax for Runnables and Callables

thiagotnunes:
Uses Java 8 lambda syntax for Runnables and Callables.

No logic change has been made, this is purely a syntatic change.


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -205,7 +202,7 @@ public void run() {
     private final boolean isSingleUse;
     private final AtomicInteger asyncOperationsCount = new AtomicInteger();
 
-    private Object lock = new Object();
+    private final Object lock = new Object();
```
thiagotnunes, line 205: The only change outside of syntax is this one, where we had to make this variable `final` to be used within a lambda.

olavloite (approved):

PR #1071, opened 2021-04-16
chore: fixes typos

thiagotnunes:
No logic changes, only fixes typos in non-public method names / comments (no breaking changes).


google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/StatementParser.java
```diff
@@ -360,7 +360,7 @@ public static String removeCommentsAndTrim(String sql) {
     final char HYPHEN = '-';
     final char DASH = '#';
     final char SLASH = '/';
-    final char ASTERIKS = '*';
+    final char ASTERISKS = '*';
```
olavloite, line 363: I think this should be `ASTERISK` (singular instead of plural)

google-cloud-spanner/src/test/java/com/google/cloud/spanner/OperationTest.java
```diff
@@ -174,7 +174,7 @@ public void waitForCompletes() {
   }
 
   @Test
-  public void waitForTimesout() {
+  public void waitForTimeOut() {
```
olavloite, line 177:
```suggestion
  public void waitForTimeout() {
```

olavloite (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/StatementParser.java
thiagotnunes: Thanks, fixed it.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/OperationTest.java
thiagotnunes: Thanks, fixed it.

PR #1072, opened 2021-04-16
refactor: cleans up code / uses java 8 features

thiagotnunes:
In this PR we have applied several cleanup tasks suggested by an automated code analysis. This were the tasks done:

1. Moves C style array declaration to Java style (`int name[] to int[] name`)
2. Uses `StandardCharset.UTF_8` instead of literal String `"UTF-8"`
3. Removes redundant `toString()` calls
4. Removes `static` modifiers from `enums`. Removes `private` modifier from `enum` constructors
5. Removes redundant modifiers from interfaces
6. Removes unnecessary semicolons
7. Removes redundant method throw clauses
8. Removes redundant calls to close methods
9. Optimises class imports
10. Removes dangling javadoc comment
11. Removes javadoc self-references
12. Refactors long literals (from `1l` to `1L`)
13. Removes octal integers
14. Simplify arithmetic expressions
15. Fixes malformed `String.format` calls
16. Removes ignored method call assignments
17. Compare Strings using `.equals` instead of `==`
18. Simplifies test assertions
19. Removes redundant calls to `String.format`
20. Removes explicit types in diamond operators
21. Uses try with resources where possible
22. Replaces if block with switch statement
23. Uses enhanced for loop
24. Removes unnecessary boxing
25. Removes unnecessary unboxing
26. Refactors to use `map.computeIfAbsent` where possible


thiagotnunes:
@olavloite sorry for the big PR, got carried away. This will be easier to review if we check each commit (they are self contained)

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITDirectPathFallback.java
```diff
@@ -110,7 +110,7 @@ public ITDirectPathFallback() {
   }
 
   @Before
-  public void setup() throws IOException, Throwable {
+  public void setup() throws Throwable {
```
olavloite, line 113: nit: Does this really need to be `throws Throwable`? Or could it throw something that is a little more specific?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITVPCNegativeTest.java
```diff
@@ -85,7 +85,7 @@ public static void setUpClass() {
     assertTrue(
         "GOOGLE_CLOUD_TESTS_VPCSC_OUTSIDE_PERIMETER_PROJECT environment variable needs "
             + "to be set to a GCP project that is outside the VPC perimeter",
-        OUTSIDE_VPC_PROJECT != null && OUTSIDE_VPC_PROJECT != "");
```
olavloite, line 89: nit: replace with assertFalse(..., `Strings.isNullOrEmpty(OUTSIDE_VPC_PROJECT))`

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
```diff
@@ -1824,17 +1824,14 @@ public void testBatchCreateSessionsFailure_shouldNotPropagateToCloseMethod() {
...
         while (rs.next()) {
           fail("Missing expected exception");
         }
       } catch (SpannerException e) {
         assertThat(e.getErrorCode()).isEqualTo(ErrorCode.RESOURCE_EXHAUSTED);
-      } finally {
-        // This should not cause any failures.
-        rs.close();
       }
+      // This should not cause any failures.
```
olavloite, line 1834: This comment now feels quite redundant. It applied to the `rs.close()` method call, but that is not clear from the context anymore. It might be better to just remove it.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/AsyncResultSetImplStressTest.java
```diff
@@ -153,7 +153,7 @@ private ResultSet createResultSetWithErrors(double errorFraction) {
   private List<Row> createExpectedRows() {
     List<Row> rows = new ArrayList<>(resultSetSize);
     for (int i = 0; i < resultSetSize; i++) {
-      rows.add(new Row(Long.valueOf(i + 1), String.format("Row %d", (i + 1))));
+      rows.add(new Row((long) (i + 1), String.format("Row %d", (i + 1))));
```
olavloite, line 156: nit: `i + 1L` is even simpler

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionClientTest.java
```diff
@@ -273,7 +273,7 @@ public void batchCreateSessionsWithExceptions() {
       final List<Long> errorOnChannels = new ArrayList<>();
       if (behavior == AddRemoveSetException.REMOVE) {
         for (int c = 0; c < spannerOptions.getNumChannels(); c++) {
-          errorOnChannels.add(Long.valueOf(c));
+          errorOnChannels.add((long) c);
```
olavloite, line 276: I'm actually not sure whether I agree with the (presumably automatically proposed) change here. The goal is to remove unnecessary boxing, but `List<Long>#add(Long e)` expects a `Long` and not a `long`, so I would say that the boxing is not unnecessary here. Or am I missing the point of this change?

olavloite (approved):
Thanks for this cleanup (and for the tip that the commits are self-contained, that made the review process a lot easier).

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionClientTest.java
thiagotnunes: That is true, I think it wants to avoid manual boxing of values. Let me know if you have strong opinions about this and I can change it back, otherwise I will leave it as is.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/AsyncResultSetImplStressTest.java
thiagotnunes: Thanks, fixed.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
thiagotnunes: Thanks, fixed.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITVPCNegativeTest.java
thiagotnunes: Thanks, fixed.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITDirectPathFallback.java
thiagotnunes: We did not need any `throws` actually, thanks for spotting this, fixed now.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionClientTest.java
olavloite: No, if this removes a warning from the static code checker, then I'm happy with it.

PR #1073, opened 2021-04-19
feat: adds getValue to ResultSet

thiagotnunes:
Adds the method `getValue` to the `ResultSet`. This will return an encapsulated value. It supports all the column types that are provided.

### DIFF
--- a/google-cloud-spanner/clirr-ignored-differences.xml
+++ b/google-cloud-spanner/clirr-ignored-differences.xml
@@ -592,4 +592,17 @@
     <className>com/google/cloud/spanner/AsyncTransactionManager$CommitTimestampFuture</className>
     <method>java.lang.Object get()</method>
   </difference>
+
+  <!-- Adds getValue to ResultSet -->
+  <!-- These are not breaking changes, since we provide default interface implementation -->
+  <difference>
+    <differenceType>7012</differenceType>
+    <className>com/google/cloud/spanner/StructReader</className>
+    <method>com.google.cloud.spanner.Value getValue(int)</method>
+  </difference>
+  <difference>
+    <differenceType>7012</differenceType>
+    <className>com/google/cloud/spanner/StructReader</className>
+    <method>com.google.cloud.spanner.Value getValue(java.lang.String)</method>
+  </difference>
 </differences>
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
@@ -362,59 +362,67 @@ abstract class AbstractResultSet<R> extends AbstractStructReader implements Resu
         Type fieldType = field.getType();
         switch (fieldType.getCode()) {
           case BOOL:
-            builder.set(fieldName).to(value.getBool());
+            builder.set(fieldName).to(value == null ? null : value.getBool());
             break;
           case INT64:
-            builder.set(fieldName).to(value.getInt64());
+            builder.set(fieldName).to(value == null ? null : value.getInt64());
             break;
           case FLOAT64:
-            builder.set(fieldName).to(value.getFloat64());
+            builder.set(fieldName).to(value == null ? null : value.getFloat64());
             break;
           case NUMERIC:
-            builder.set(fieldName).to(value.getNumeric());
+            builder.set(fieldName).to(value == null ? null : value.getNumeric());
             break;
           case STRING:
-            builder.set(fieldName).to(value.getString());
+            builder.set(fieldName).to(value == null ? null : value.getString());
             break;
           case BYTES:
-            builder.set(fieldName).to(value.getBytes());
+            builder.set(fieldName).to(value == null ? null : value.getBytes());
             break;
           case TIMESTAMP:
-            builder.set(fieldName).to(value.getTimestamp());
+            builder.set(fieldName).to(value == null ? null : value.getTimestamp());
             break;
           case DATE:
-            builder.set(fieldName).to(value.getDate());
+            builder.set(fieldName).to(value == null ? null : value.getDate());
             break;
           case ARRAY:
             switch (fieldType.getArrayElementType().getCode()) {
               case BOOL:
-                builder.set(fieldName).toBoolArray(value.getBoolArray());
+                builder.set(fieldName).toBoolArray(value == null ? null : value.getBoolArray());
                 break;
               case INT64:
-                builder.set(fieldName).toInt64Array(value.getInt64Array());
+                builder.set(fieldName).toInt64Array(value == null ? null : value.getInt64Array());
                 break;
               case FLOAT64:
-                builder.set(fieldName).toFloat64Array(value.getFloat64Array());
+                builder
+                    .set(fieldName)
+                    .toFloat64Array(value == null ? null : value.getFloat64Array());
                 break;
               case NUMERIC:
-                builder.set(fieldName).toNumericArray(value.getNumericArray());
+                builder
+                    .set(fieldName)
+                    .toNumericArray(value == null ? null : value.getNumericArray());
                 break;
               case STRING:
-                builder.set(fieldName).toStringArray(value.getStringArray());
+                builder.set(fieldName).toStringArray(value == null ? null : value.getStringArray());
                 break;
               case BYTES:
-                builder.set(fieldName).toBytesArray(value.getBytesArray());
+                builder.set(fieldName).toBytesArray(value == null ? null : value.getBytesArray());
                 break;
               case TIMESTAMP:
-                builder.set(fieldName).toTimestampArray(value.getTimestampArray());
+                builder
+                    .set(fieldName)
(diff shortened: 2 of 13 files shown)


google-cloud-spanner/clirr-ignored-differences.xml
```diff
@@ -592,4 +592,17 @@
     <className>com/google/cloud/spanner/AsyncTransactionManager$CommitTimestampFuture</className>
     <method>java.lang.Object get()</method>
   </difference>
+
```
thiagotnunes, line 595: Clirr does not support Java 8 features, so it thinks these are breaking changes, where in fact we have provided default implementations for the added interface methods. No need for a major release here.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
```diff
@@ -1122,117 +1081,6 @@ static NullPointerException throwNotNull(int columnIndex) {
         "Cannot call array getter for column " + columnIndex + " with null elements");
   }
 
-  /**
```
thiagotnunes, line 1125: @olavloite I removed this optimized Int64 and Float64. Do you think this will impact performance significantly?

In this constructor we were doing `O(n)` compute and using a `bitset` for `nulls`. The difference here will be basically be in memory usage, where here we were using `O(1)` and with the new implementation we copy the array into a list (so `O(n)`). However, I wonder how big is `n` usually and whether we should worry about this.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
```diff
@@ -699,65 +653,70 @@ protected Struct getStructInternal(int columnIndex) {
...
     }
 
     @Override
     protected double[] getDoubleArrayInternal(int columnIndex) {
-      return getDoubleListInternal(columnIndex).toPrimitiveArray(columnIndex);
+      final List<Double> values = rowData.get(columnIndex).getFloat64Array();
+      int i = 0;
+      double[] result = new double[values.size()];
+      for (Double value : values) {
+        result[i++] = value;
```
olavloite, line 682: I think this will break if `value` is null.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
olavloite: I do worry about the additional memory consumption here, because there are more changes in this PR that affect memory consumption than only the removal of the optimized array decoding of `Int64` and `Float64`. So I do think it is worth checking what effect this has. The other things that affect memory consumption in this PR:
1. Instead of keeping a reference to an `Object` for each column/row combination, we now create a `Value` object. The object itself is not big, but will need to be garbage collected afterwards. For large result sets this could be a significant number of additional objects.
2. Most array types used `Lists.transform` for decoding. That is virtually free as long as the column is not retrieved by the application. If a user executes a select that includes an array column that is not used by the application, it will now always be decoded for each row that is consumed, regardless of whether the column itself is consumed.
3. The optimized `Int64` and `Float64` array implementations are removed.

Also, I think that the current implementations for decoding `Array<Int64>` and `Array<Float64>` to primitive arrays will fail if one of the array elements is null.

I don't worry about the additional CPU cycles for the decoding itself, it's more the additional overhead that is created mainly for the garbage collection of the additional objects that we create.

thiagotnunes:
@olavloite got it, thanks for the feedback. I've re-implemented with a less intrusive manner, let me know if that makes sense.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
```diff
@@ -679,6 +680,60 @@ protected Date getDateInternal(int columnIndex) {
...
+            case BYTES:
+              return Value.bytesArray(getBytesListInternal(columnIndex));
+            case TIMESTAMP:
+              return Value.timestampArray(getTimestampListInternal(columnIndex));
+            case DATE:
+              return Value.dateArray(getDateListInternal(columnIndex));
+            case STRUCT:
+              return Value.structArray(
+                  this.type.getArrayElementType(), getStructListInternal(columnIndex));
+            default:
```
thiagotnunes, line 728: I assume you can't have arrays of arrays here.

olavloite: Correct, [arrays-of-arrays are not supported](https://cloud.google.com/spanner/docs/data-definition-language#arrays). Also, `ARRAY<STRUCT>` is not supported as the type of a table column, but can be [created in SQL statements](https://cloud.google.com/spanner/docs/arrays#querying_nested_arrays).

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITResultSetGetValue.java
```diff
@@ -0,0 +1,219 @@
...
+          resultSet.getValue("timestampArray"));
+      assertEquals(
+          Value.dateArray(
+              Arrays.asList(Date.fromYearMonthDay(2021, 2, 3), Date.fromYearMonthDay(2021, 3, 4))),
+          resultSet.getValue("dateArray"));
+    }
+  }
+
+  @Test
+  public void testReadNullInArrays() {
```
olavloite, line 178: I think it would be good to add null values for `INT64` and `FLOAT64` arrays as well in this test. Arrays of all types can always contain null values, regardless whether the column itself is nullable, and regardless of the array element type.

olavloite (approved):
This looks good to me, but I think for completeness it would be good to add null values to the integration tests for all array types. Now the primitive types are missing (Long and Double).

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITResultSetGetValue.java
thiagotnunes: These throw a `NPE` as they do if you use a `resultSet.getLongArray(...)` or `resultSet.getDoubleArray(...)`. I could test the exception if you feel like we should.

olavloite: No, in that case I actually think we should change the implementation. It is logical that `resultSet.getDoubleArray(..)` throws an `NPE`. I don't think it is logical that `resultSet.getValue("some-float64-array-col")` should ever throw an `NPE` because one of the elements is `null`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
thiagotnunes: Added tests for all possible literals I could think of (including `ARRAY<STRUCT>`)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Value.java
```diff
@@ -543,7 +543,7 @@ private Value() {}
    *
    * @throws IllegalStateException if {@code isNull()} or the value is not of the expected type
    */
-  abstract List<Struct> getStructArray();
+  public abstract List<Struct> getStructArray();
```
thiagotnunes, line 546: @olavloite I made this method public. It surprised me this was not public before, since users can `SELECT ARRAY(SELECT STRUCT ...)` and would not be able to get the results.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITResultSetGetValue.java
olavloite: Yes, I think we should change that as well now that we have a chance. `Value` also has an `isNull()` method, so it would make sense to let `resultSet.getValue(...)` return a `Value` instance when the column is null. The user can then check whether the value is actually null by calling `Value#isNull()`.

thiagotnunes: Awesome, changed the implementation to allow for that. If you could re-review, I'd appreciate it.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/StructReader.java
```diff
@@ -132,6 +132,16 @@
   /** Returns the value of a non-{@code NULL} column with type {@link Type#date()}. */
   Date getDate(String columnName);
 
+  /** Returns the value of a non-{@code NULL} column as a {@link Value}. */
+  default Value getValue(int columnIndex) {
+    throw new UnsupportedOperationException("method should be overwritten");
+  }
+
+  /** Returns the value of a non-{@code NULL} column as a {@link Value}. */
```
olavloite, line 140: nit: This should be updated to indicate that this method can be called for columns that contain a null value, and what the behavior is when the value is null.

olavloite (approved):

PR #1093, opened 2021-04-23
fix: do not keep references to invalidated clients

### LINKED ISSUE #1089: Reading from a non-existent database causes memory usage to grow
It seems like requests that fail with a NOT_FOUND error can cause the Spanner client to consume memory.  I suspect that it is caused by the invalidatedDbClients field in SpannerImpl:

https://github.com/googleapis/java-spanner/blob/71d7fd16aaa966b2a3cf0599fd25f4bd3c93060f/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java#L96

I ran this test program to reproduce the issue with Java 11 and google-cloud-spanner 6.3.2:

```java
import java.util.Arrays;
import org.junit.Test;

public class FailedRequestMemoryUsageTest {

  @Test
  public void testFailedRequestMemoryUsage() throws Exception {
    spanner
        .getInstanceAdminClient()
        .createInstance(
                .setNodeCount(1)
                .build())
        .get();

    for (int i = 0; i <= 10000; i++) {
      if (i % 500 == 0) {
        Runtime.getRuntime().gc();
        System.out.println(
            i + "\t" + (Runtime.getRuntime().totalMemory() - Runtime.getRuntime().freeMemory()));
      }
      try {
        spanner
            .singleUse()
      } catch (Exception e) {
        // Ignore NOT_FOUND error
      }
    }
  }
}
```

```
0	19573984
500	31193384
1000	35718312
1500	40337512
2000	44895408
2500	49444864
3000	54109824
3500	58690984
4000	63295664
4500	68078928
5000	72855264
5500	77463136
6000	82355544
6500	86576208
7000	91327472
7500	95821992
8000	100438752
8500	105055256
9000	109650504
9500	114148912
10000	118900040
```

java-spanner-memory-usage

olavloite:
SpannerImpl would keep a reference to all database clients that had been created and then invalidated because the database was deleted or did not exists in the first place. If the getDatabaseClient(String) method would be called multiple times for the same invalid database, the memory usage of the application would continue to increase.

Fixes #1089

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
@@ -93,9 +93,6 @@ class SpannerImpl extends BaseService<SpannerOptions> implements Spanner {
   private final CloseableExecutorProvider asyncExecutorProvider;
 
   @GuardedBy("this")
-  private final List<DatabaseClientImpl> invalidatedDbClients = new ArrayList<>();
-
-  @GuardedBy("this")
   private final Map<DatabaseId, SessionClient> sessionClients = new HashMap<>();
 
   private final DatabaseAdminClient dbAdminClient;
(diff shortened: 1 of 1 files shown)


thiagotnunes (approved):

sebright:
Thanks for the fix!

PR #1094, opened 2021-04-23
test: add additional logging to backup pagination test

olavloite:
Updates #1090


thiagotnunes (approved):

thiagotnunes:
Good catch, this seems to be happening from time to time.

PR #1110, opened 2021-04-28
fix: allow using case-insensitive user-agent key

olavloite:
Adding a custom user-agent was only possible by using an all lower-case key. A header key should however be case-insensitive. This PR changes the merging of a custom user-agent value with the default user-agent to use a case-insensitive comparison.


thiagotnunes (approved):
Thanks for the fix @olavloite

PR #1129, opened 2021-05-03
fix(3.3.x): allows setting custom user agent in any case

thiagotnunes:
Allows setting a custom user agent in any case, instead of only in lower case.


thiagotnunes:
Wait for release please to be configured for this branch before merging.

zoercai (approved):

PR #1139, opened 2021-05-05
chore: use assertThrows with lambda instead of try-catch

### LINKED ISSUE #1124: cleanup: Change tests for expected exceptions to use a lambda expression where possible
Many test cases currently use a try-catch block with a `fail` at the end of the try and a check for the expected exception in the catch block. With Java 8, that can in many cases be re-written as an `assertThrows(lambda expression)`.

olavloite:
Fixes #1124


thiagotnunes (approved):

thiagotnunes:
Thanks for the fix!

PR #1143, opened 2021-05-07
feat: support DATABASE statements

olavloite:
Adds support for the following DATABASE statements in the Connection API:
- `CREATE DATABASE <database_id>`: Creates a database (passes the statement through to the `CreateDatabase` RPC)
- `ALTER DATABASE <database_id>`: Alters an existing database (passes the statement through to the `UpdateDatabaseDdl` RPC)
- `DROP DATABASE <database_id>`: Drops an existing database (parsed and translated into a call to the `DropDatabase` RPC)
- `SHOW VARIABLE DATABASES`: Lists the databases on the current instance (parsed and translated into a call to the `ListDatabases` RPC)
- `USE DATABASE <database_id>`: Changes the database that is used by the connection to execute queries, updates, etc.

Also adds support for creating a `Connection` that is not connected to a specific database, but only to a Spanner instance. This can be used to connect to an empty instance, then create a database on the instance, and then call `USE DATABASE <database_id>` to use the database with the connection.

This feature allows users to execute scripts like this directly through the Connection API (and thereby also through JDBC):

```sql
CREATE DATABASE db;
ALTER DATABASE db SET OPTIONS (optimizer_version = 1);
USE DATABASE db;

CREATE TABLE Singers (...);
INSERT INTO Singers (...);
SELECT * FROM Singers;

DROP DATABASE db;
```

Needed for https://github.com/googleapis/java-spanner-jdbc/issues/457


thiagotnunes (commented):
I started an internal discussion in regards to the syntax of the following statements:

- `SHOW VARIABLE DATABASES`
- `USE DATABASE <database id>`

I will let you know once we have a response here, sorry for the delay.

PR #1145, opened 2021-05-07
feat: add bufferAsync methods

### LINKED ISSUE #1126: feat: Add bufferAsync method
The current `buffer(..)` methods are non-blocking, but do not return anything (return type `void`). With the addition of the async API and the switch to Java 8, having an async version of these methods that would return an `ApiFuture<Void>` would simplify writing chained async statements for a transaction. The semantics of the async versions would be exactly the same as the current methods, as those are already non-blocking, but it would make it possible to clean up client code.

olavloite:
Adds bufferAsync methods to TransactionContext. The existing buffer methods were already non-blocking, but the async versions also return an ApiFuture, which make them easier to use when chaining multiple async calls together.

Also changes some calls in the AsyncTransactionManagerTest to use lambdas instead of the test helper methods.

Fixes #1126

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
@@ -61,7 +61,6 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -152,7 +151,10 @@ class TransactionRunnerImpl implements SessionTransaction, TransactionRunner {
       }
     }
 
-    private final AtomicBoolean committing = new AtomicBoolean();
+    private final Object committingLock = new Object();
+
+    @GuardedBy("committingLock")
+    private volatile boolean committing;
 
     @GuardedBy("lock")
     private volatile SettableApiFuture<Void> finishedAsyncOperations = SettableApiFuture.create();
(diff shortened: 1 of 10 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
```diff
@@ -609,6 +609,15 @@ public void buffer(Mutation mutation) {
       }
     }
 
+    @Override
+    public ApiFuture<Void> bufferAsync(Mutation mutation) {
+      // Normally, we would call the async method from the sync method, but this is also safe as
+      // both are non-blocking anyways, and this prevents the creation of an ApiFuture that is not
+      // really used when the sync method is called.
+      buffer(mutation);
```
thiagotnunes, line 617: I do have a worry here: it seems we acquire a lock in the `buffer` method, so that could potentially have some waiting which is not expected by the user.

How big of a problem do you think this is?

olavloite: That's a good point. We actually take this lock in a couple of the other async methods as well already, so in that sense this does not deviate from the existing methods. The lock is only held for a short period of time in all cases, as it is only held while executing local operations. So in that sense I don't think it is a big problem. Still, we should preferably not take any locks in a method that is marked as non-blocking, as there could in theory be some waiting, albeit short.

I'll have a look and see if I can fix this for all the async methods in this class.

olavloite: I've not been able to come up with a solution without any locking, but I've reduced it to an absolute minimum by introducing a separate lock that is only for ensuring that mutations that are buffered are either included in the commit, or will throw an exception if commit has already been called. It is not perfect, but the locking should be absolutely minimal as the lock is only held for a couple of simple statements in all cases.

thiagotnunes (approved):
Thanks for the changes @olavloite

PR #1156, opened 2021-05-11
chore(3.3.3-sp): configure initial sp version

thiagotnunes:
Configures CI for branch

Automatically generated PR for LTS


olavloite (commented):
This seems reasonable to my (not very experienced) GitHub Actions eyes, but a couple of small questions:
1. Shouldn't this be a `chore` or `build` PR? It does not introduce any new features.
2. Should we worry about the linkage error?

PR #1185, opened 2021-05-17
test: verify that callback is not called while paused

olavloite:
Adds a test to verify that the callback function of an `AsyncResultSet` is not called while the result set is paused.

### DIFF
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/AsyncResultSetImplTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/AsyncResultSetImplTest.java
@@ -17,6 +17,8 @@
 package com.google.cloud.spanner;
 
 import static com.google.common.truth.Truth.assertThat;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertThrows;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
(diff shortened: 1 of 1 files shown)


thiagotnunes:
@olavloite I think I might be missing some context on this one. Do you mind providing some details?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/AsyncResultSetImplTest.java
```diff
@@ -301,6 +305,72 @@ public void pauseResume() throws InterruptedException {
...
+                  throw SpannerExceptionFactory.propagateInterrupt(e);
+                }
+              });
+      int rowCounter = 0;
+      while (!callbackResult.isDone()) {
+        Object o = queue.poll(1L, TimeUnit.MILLISECONDS);
+        if (o != null) {
+          rowCounter++;
+        }
+        Thread.yield();
```
thiagotnunes, line 362: If I understood this correctly, we could also sleep after this for a while (although I am not super happy with sleeps in tests) in order to guarantee that the callback does not run until resume is called.

thiagotnunes (approved):
Thanks for the context in the email!

google-cloud-spanner/src/test/java/com/google/cloud/spanner/AsyncResultSetImplTest.java
thiagotnunes: (minor correction, we don't need to yield if we sleep)

olavloite: Yes, that's correct. I tried it locally with a number of different sleep times, and it succeeded in all cases. I then changed it to `Thread.yield()` as I'm also not really a fan of `Thread.sleep()` in a test case if we can avoid it.

thiagotnunes: Ah great, that would be the approach I would take as well. LGTM

PR #1192, opened 2021-05-18
fix: stop invoking callback after pausing and cancelling result set

### LINKED ISSUE #1191: Cancelling and pausing an AsyncResultSet can cause the callback to be called repeatedly
The following test case shows that the callback will continue to be called after the callback has returned `PAUSE` and the result set has been cancelled. This will only happen if the callback method does not call `resultSet.tryNext()`.

```java
@Test
public void testCallbackIsNotCalledWhilePausedAndCanceled() throws InterruptedException, ExecutionException {
  Executor executor = Executors.newSingleThreadExecutor();
  ResultSet delegate = mock(ResultSet.class);

  final AtomicInteger callbackCounter = new AtomicInteger();

  try (AsyncResultSetImpl rs =
      new AsyncResultSetImpl(simpleProvider, delegate, AsyncResultSetImpl.DEFAULT_BUFFER_SIZE)) {
        rs.setCallback(executor,
            resultSet -> {
              callbackCounter.getAndIncrement();
              return CallbackResponse.PAUSE;
            });

    rs.cancel();

    while (callbackCounter.get() < 100) {
      Thread.yield();
    }

   assertThat(callbackCounter.get()).isGreaterThan(99);
  }
}
```

olavloite:
An AsyncResultSet that was PAUSED and then CANCELLED, would continue to invoke the callback until the callback would call tryNext(). If the callback never called tryNext(), the spinning would continue until the entire result set had been consumed.

Fixes #1191

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java
@@ -252,6 +252,13 @@ class AsyncResultSetImpl extends ForwardingStructReader implements ListenableAsy
             if (cursorReturnedDoneOrException) {
               break;
             }
+            if (state == State.CANCELLED) {
+              // The callback should always get at least one chance to catch the CANCELLED
+              // exception. It is however possible that the callback does not call tryNext(), and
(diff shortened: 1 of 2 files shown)


elefeint (approved):
I've verified that this works in R2DBC driver.

thiagotnunes (approved):

PR #1214, opened 2021-05-26
docs: document retry settings in sample

olavloite:
Fixes https://github.com/GoogleCloudPlatform/java-docs-samples/issues/5310

### DIFF
--- a/samples/snippets/src/main/java/com/example/spanner/CustomTimeoutAndRetrySettingsExample.java
+++ b/samples/snippets/src/main/java/com/example/spanner/CustomTimeoutAndRetrySettingsExample.java
@@ -57,18 +57,18 @@ class CustomTimeoutAndRetrySettingsExample {
                 .setInitialRetryDelay(Duration.ofMillis(500))
                 // The maximum amount of time to wait before retrying. I.e. after this value is
                 // reached, the wait time will not increase further by the multiplier.
-                .setMaxRetryDelay(Duration.ofSeconds(10))
+                .setMaxRetryDelay(Duration.ofSeconds(64))
                 // The previous wait time is multiplied by this multiplier to come up with the next
                 // wait time, until the max is reached.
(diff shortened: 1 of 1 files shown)


samples/snippets/src/main/java/com/example/spanner/CustomTimeoutAndRetrySettingsExample.java
```diff
@@ -53,14 +53,23 @@ static void executeSqlWithCustomTimeoutAndRetrySettings(
...
+                .setMaxRetryDelay(Duration.ofSeconds(10))
+                // The previous wait time is multiplied by this multiplier to come up with the next
+                // wait time, until the max is reached.
                 .setRetryDelayMultiplier(1.5)
 
                 // Configure RPC and total timeout settings.
-                .setInitialRpcTimeout(Duration.ofSeconds(60))
-                .setMaxRpcTimeout(Duration.ofSeconds(60))
-                .setRpcTimeoutMultiplier(1.0)
+                // Timeout for the first RPC call. Subsequent retries will be based off this value.
```
hengfengli, line 66: In the [doc](https://cloud.google.com/spanner/docs/custom-timeout-and-retry#java): 

> In the following sample, a timeout of 60 seconds is set for the given operation. If the operation takes longer than this timeout, a DEADLINE_EXCEEDED error is returned.

If we update the code sample to this, I don't it match this statement. We probably should update nodejs code sample to set all values to 60s.

olavloite: Oh, good catch. I'll change that back, and I'll also open a PR for the Node sample to be in line with the documentation page.

hengfengli (approved):
LGTM.

olavloite:
The Windows build error is addressed by #1218

PR #1224, opened 2021-05-31
chore(3.3.3-sp): dependencies check

thiagotnunes:
We had to do a couple of things in this PR:

1. We used the versions as specified 
https://github.com/GoogleCloudPlatform/cloud-opensource-java/blob/master/boms/cloud-lts-bom/pom.xml
2. We had to remove the dependency on google-cloud-shared-dependencies
3. There were some versions which were not specified in the above POM and we had to "guess" the best version taking in account maven build enforcer (which checks for transitive dependency version compatibilities).
4. We had compilation errors in the main and test code in regards to the google guava Predicate class. We had to implement the `test` method in addition to the `apply` method, since we are using Java 7 here and can not rely on `default` interface method definitions.

### DIFF
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerApiFuturesTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerApiFuturesTest.java
@@ -49,7 +49,7 @@ public class SpannerApiFuturesTest {
 
   @Test
   public void testGetOrNull() {
-    assertThat(SpannerApiFutures.getOrNull(null)).isNull();
+    assertThat(SpannerApiFutures.getOrNull((ApiFuture<Object>) null)).isNull();
   }
 
   @Test
--- a/pom.xml
+++ b/pom.xml
@@ -64,6 +64,43 @@
     <github.global.server>github</github.global.server>
     <site.installationModule>google-cloud-spanner-parent</site.installationModule>
     <google.cloud.shared-dependencies.version>0.18.0</google.cloud.shared-dependencies.version>
+
+    <!-- finalized versions -->
+    <!-- JRE for Java 8. Beam's dependency (gcsio) uses JRE flavor -->
+    <guava.version>30.1.1-jre</guava.version>
+    <google.autovalue.version>1.8.1</google.autovalue.version>
+    <protobuf.version>3.16.0</protobuf.version>
+    <http.version>1.39.2-sp.1</http.version>
+    <io.grpc.version>1.36.2</io.grpc.version>
+    <!-- We don't use gax-bom because it includes the artifacts with 'testlib' classifier.
+        When updating gax.version, update gax.httpjson.version too. -->
+    <gax.version>1.64.0-sp.1</gax.version>
+    <gax.httpjson.version>0.81.0-sp.1</gax.httpjson.version>
+    <google.api.client.version>1.31.3-sp.1</google.api.client.version>
+    <api.common.version>1.10.1-sp.1</api.common.version>
+    <google.auth.library.version>0.25.2-sp.1</google.auth.library.version>
+    <google.oauth.client.version>1.31.4-sp.1</google.oauth.client.version>
+    <google.cloud.spanner.version>3.3.3-sp.1</google.cloud.spanner.version>
+    <google.cloud.monitoring.version>2.1.0-sp.1</google.cloud.monitoring.version>
+    <google.cloud.trace.version>1.3.0-sp.1</google.cloud.trace.version>
+    <google.cloud.pubsub.version>1.111.0-sp.1</google.cloud.pubsub.version>
+    <google.cloud.bigquery.version>1.127.12-sp.1</google.cloud.bigquery.version>
+    <google.cloud.bigtable.version>1.22.0-sp.1</google.cloud.bigtable.version>
+    <bigtable-hbase-beam.version>1.20.0-sp.1</bigtable-hbase-beam.version>
+    <google.cloud.storage.version>1.113.14-sp.1</google.cloud.storage.version>
+    <datastore.v1.proto.client.version>1.6.4-sp.1</datastore.v1.proto.client.version>
+    <appengine.api.1.0.sdk.version>1.9.86</appengine.api.1.0.sdk.version>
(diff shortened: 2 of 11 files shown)


pom.xml
```diff
@@ -64,6 +64,43 @@
...
+    <google.cloud.trace.version>1.3.0-sp.1</google.cloud.trace.version>
+    <google.cloud.pubsub.version>1.111.0-sp.1</google.cloud.pubsub.version>
+    <google.cloud.bigquery.version>1.127.12-sp.1</google.cloud.bigquery.version>
+    <google.cloud.bigtable.version>1.22.0-sp.1</google.cloud.bigtable.version>
+    <bigtable-hbase-beam.version>1.20.0-sp.1</bigtable-hbase-beam.version>
+    <google.cloud.storage.version>1.113.14-sp.1</google.cloud.storage.version>
+    <datastore.v1.proto.client.version>1.6.4-sp.1</datastore.v1.proto.client.version>
+    <appengine.api.1.0.sdk.version>1.9.86</appengine.api.1.0.sdk.version>
+    <androidpublisher.version>v3-rev20210429-1.31.0</androidpublisher.version>
+
```
thiagotnunes, line 94: The dependencies below this line are the ones we had to "guess" and fix based on errors of the maven build enforcer checks (for transitive versions compatibilities). These would usually come from the google-cloud-shared-libraries, but we did not know which version of those to use.

@elharo Is this the approach we would like to take or will we release a version of the google-cloud-shared-libraries that is compatible with the lts versions of all the libraries we have?

pom.xml
```diff
@@ -104,20 +141,30 @@
         <version>3.3.3-sp.2-SNAPSHOT</version><!-- {x-version-update:google-cloud-spanner:current} -->
       </dependency>
 
+<!--      <dependency>-->
```
thiagotnunes, line 144: @elharo we had to remove the google-cloud-shared-dependencies, since some versions did not work with our build.

pom.xml
```diff
@@ -104,20 +141,30 @@
...
+<!--        <type>pom</type>-->
+<!--        <scope>import</scope>-->
+<!--      </dependency>-->
       <dependency>
-        <groupId>com.google.cloud</groupId>
-        <artifactId>google-cloud-shared-dependencies</artifactId>
-        <version>${google.cloud.shared-dependencies.version}</version>
-        <type>pom</type>
-        <scope>import</scope>
+        <groupId>com.google.errorprone</groupId>
```
thiagotnunes, line 152: We had to manually add `errorprone`

pom.xml
```diff
@@ -104,20 +141,30 @@
...
+        <artifactId>error_prone_annotations</artifactId>
+        <version>${errorprone.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>io.grpc</groupId>
```
thiagotnunes, line 157: We had to manually add `grpc-netty-shaded`

pom.xml
```diff
@@ -104,20 +141,30 @@
...
+        <artifactId>grpc-netty-shaded</artifactId>
+        <version>${io.grpc.version}</version>
       </dependency>
-
       <dependency>
         <groupId>junit</groupId>
         <artifactId>junit</artifactId>
-        <version>4.13.1</version>
+        <version>4.13.2</version>
```
thiagotnunes, line 164: We had to upgrade `junit`s version here.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerApiFuturesTest.java
```diff
@@ -49,7 +49,7 @@ public void testGetNull() {
 
   @Test
   public void testGetOrNull() {
-    assertThat(SpannerApiFutures.getOrNull(null)).isNull();
+    assertThat(SpannerApiFutures.getOrNull((ApiFuture<Object>) null)).isNull();
```
thiagotnunes, line 52: We had to add the explicit cast here, because of the Java 8 upgrade.

suztomo:
Memo:

Linkage Monitor complains about google-cloud-spanner-jdbc. This artifact is not in the LTS BOM.

```
SEVERE: Newly introduced problems:
(com.google.cloud:google-cloud-spanner:3.3.3-sp.2-SNAPSHOT) Interface com.google.cloud.spanner.connection.Connection's method getCommitResponse() is not found
  referenced from com.google.cloud.spanner.jdbc.JdbcConnection (com.google.cloud:google-cloud-spanner-jdbc:2.1.0)
(com.google.cloud:google-cloud-spanner:3.3.3-sp.2-SNAPSHOT) Interface com.google.cloud.spanner.connection.Connection's method setReturnCommitStats(boolean) is not found
  referenced from com.google.cloud.spanner.jdbc.JdbcConnection (com.google.cloud:google-cloud-spanner-jdbc:2.1.0)
(com.google.cloud:google-cloud-spanner:3.3.3-sp.2-SNAPSHOT) Interface com.google.cloud.spanner.connection.Connection's method isReturnCommitStats() is not found
  referenced from com.google.cloud.spanner.jdbc.JdbcConnection (com.google.cloud:google-cloud-spanner-jdbc:2.1.0)

com.google.cloud:google-cloud-spanner:3.3.3-sp.2-SNAPSHOT is at:
  com.google.cloud:google-cloud-spanner:3.3.3-sp.2-SNAPSHOT (compile)
com.google.cloud:google-cloud-spanner-jdbc:2.1.0 is at:
  com.google.cloud:google-cloud-spanner-jdbc:2.1.0 (compile)
```

This happens because the Linkage Monitor tries to get the latest Libraries BOM, which may not be compatible with the libraries in the LTS BOM.

The libraries BOM 20.0.0 (which will be used by the Beam release in LTS BOM) has `com.google.cloud:google-cloud-spanner-jdbc:2.0.0`

suztomo:
@thiagotnunes Thank you for the detailed notes.

> We had to remove the dependency on google-cloud-shared-dependencies

Was this also for the enforcer rule (upper-bound check)?

thiagotnunes:
@suztomo Yes, that was the reason for removing it

elharo (changes requested):
I don't think you need to merge this. This is all for testing, not shipping.

thiagotnunes:
@elharo ok, but I guess we will need to do some modifications once we have the final LTS BOM, which could be based on this PR.

PR #1227, opened 2021-06-01
feat: add gRPC-GCP channel pool as an option

nimf:
This enables a user to opt-in for using the gRPC-GCP channel pool and configure its options.

For example:

```java
    SpannerOptions.Builder optionsBuilder = SpannerOptions.newBuilder();

    // ...

    optionsBuilder.withGrpcGcpOptions(GcpManagedChannelOptions.newBuilder()
        .withResiliencyOptions(GcpResiliencyOptions.newBuilder()
            .setNotReadyFallback(true)
            .build())
        .build());

    Spanner spanner = optionsBuilder.build().getService();
```

The gRPC-GCP channel pool provides a couple resiliency improvements during some networking issues. It also provides additional channel pool metrics.


google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -491,6 +504,73 @@ public GapicSpannerRpc(final SpannerOptions options) {
...
+        MoreObjects.firstNonNull(options.getGrpcGcpOptions(), new GcpManagedChannelOptions());
+    GcpMetricsOptions metricsOptions = MoreObjects.firstNonNull(
+        grpcGcpOptions.getMetricsOptions(), GcpMetricsOptions.newBuilder().build());
+    GcpMetricsOptions.Builder metricsOptionsBuilder = GcpMetricsOptions.newBuilder(metricsOptions);
+    if (metricsOptions.getMetricRegistry() == null) {
+      metricsOptionsBuilder.withMetricRegistry(Metrics.getMetricRegistry());
+    }
+    List<LabelKey> labelKeys = metricsOptions.getLabelKeys();
+    List<LabelValue> labelValues = metricsOptions.getLabelValues();
+    // TODO: Add default labels with values: client_id, database, instance_id.
```
thiagotnunes, line 533: I see you would like to get label values for client_id, instance_id and database_id within a static method, which only has access SpannerOptions. SpannerOptions is not tied directly one database / instance, so you won't be able to do this here.

For the client id (which I imagine is the client language + version) you can do the following: https://github.com/googleapis/java-spanner/blob/6ce4c2da0e1378011793fd1fa136f669624234a1/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java#L250-L251

In terms of the resource we are using for a specific request, that is tied up to the session which is being used. We do add an extra header in every call context as specified [here](https://github.com/googleapis/java-spanner/blob/6ce4c2da0e1378011793fd1fa136f669624234a1/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java#L1630). The extra header key name will be [google-cloud-resource-prefix](https://github.com/googleapis/gax-java/blob/49efdc38f2d6ca3838dcecee63dc6cd07f6c2b23/gax/src/main/java/com/google/api/gax/rpc/ApiClientHeaderProvider.java#L99) and the value will be one of [these two](https://github.com/googleapis/java-spanner/blob/47f59655acaf0fcf5848039e6b4881b80cdceb5c/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerMetadataProvider.java#L32-L35) depending on whether you are performing an operation on an instance only resource or an instance + database resource.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -999,6 +1008,26 @@ public Builder setHost(String host) {
       return this;
     }
 
+    /**
+     * Sets the preference for gRPC-GCP extension (default: false). When enabled the gRPC-GCP
+     * channel pool will be used.
+     */
+    public Builder setUseGrpcGcpExtension(boolean useGrpcGcpExtension) {
```
thiagotnunes, line 1015: It seems that if the user wants to use the gRPC GCP extension they would have to set this as `true` AND configure the options below. Is that true?

If so, I would prefer if we could flatten the option setting by having something like `.withGrpcGcpOptions(GcpManagedChannelOptions)` that automatically sets any flag needed to `true` and simultaneously set the options.

google-cloud-spanner/src/main/resources/com/google/cloud/spanner/spi/v1/grpc-gcp-apiconfig.json
```diff
@@ -0,0 +1,106 @@
...
+  "method": [
+    {
+      "name": ["google.spanner.v1.Spanner/CreateSession"],
+      "affinity" : {
+        "command": "BIND",
+        "affinityKey": "name"
+      }
+    },
+    {
+      "name": ["google.spanner.v1.Spanner/BatchCreateSessions"],
```
olavloite, line 15: This will (if I understand it correctly) bind the channel that is used to call `BatchCreateSessions` to the sessions that are returned by call. Is that correct?
Does this function use the same `CHANNEL_HINT` [option](https://github.com/googleapis/java-spanner/blob/2023839cce80de0ff6451a4b6274f5da9b18416f/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerRpc.java#L81) as Gax?

The session pool in the Spanner client is initialized when a database client is created. By default it is filled with 100 sessions. These sessions are distributed over the available channels in the following way:
1. Divide the number of sessions that is needed by the number of available channels (i.e. 100 sessions / 4 channels = 25 sessions per channel)
2. [Iterate over each channel number](https://github.com/googleapis/java-spanner/blob/2023839cce80de0ff6451a4b6274f5da9b18416f/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionClient.java#L252) and execute `BatchCreateSessions(sessionCount=25)` for each channel. [Set the channel hint](https://github.com/googleapis/java-spanner/blob/2023839cce80de0ff6451a4b6274f5da9b18416f/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionClient.java#L288) in the request to the current channel number.

The above approach assumes that:
1. The available channels are handed out round-robin from the underlying channel pool.
2. Executing `BatchCreateSessions` on a specific channel and setting the channel hint will affiliate all the sessions that are returned by the invocation with the channel that was used for the invocation.
3. Subsequent RPCs that reference the affiliated session will automatically cause the RPC to be executed on the affiliated channel.

Will the above also work with the new channel pool?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
nimf: Agree. I'll add that.

google-cloud-spanner/src/main/resources/com/google/cloud/spanner/spi/v1/grpc-gcp-apiconfig.json
```diff
@@ -0,0 +1,106 @@
+{
+  "channelPool": {
+    "maxSize": 3,
```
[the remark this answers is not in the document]
nimf: No, it's the other way around -- the `maxSize` value in the json will be overwritten by the `setPoolSize`.

`setPoolSize` is a convenient way to set the maximum pool size.

google-cloud-spanner/src/main/resources/com/google/cloud/spanner/spi/v1/grpc-gcp-apiconfig.json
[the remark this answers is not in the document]
nimf: Correct, this will bind (create an affinity) the channel with the sessions created.
We do not use the `CHANNEL_HINT`

Yes, everything above will work with the new channel pool but in a slightly different manner.

When the new channel pool gets a request without a session (like BatchCreateSessions request) it picks the least busiest channel (i. e. the channel with the minimum number of in-flight RPCs).
Also if max number of channels is not reached and the least busiest channel has at least `maxConcurrentStreamsLowWatermark` in-flight RPCs then a new channel will be created and provided.

Thus having `maxConcurrentStreamsLowWatermark` set to 1 and having the first N requests to be `BatchCreateSessions` which are executed concurrently ensures that each of N channels will receive exactly one BatchCreateSessions request.

nimf: I'll resolve this one as it is the duplicate of above.

olavloite: Yeah, sorry. GitHub had some issues yesterday, and it appears that it added my comment twice.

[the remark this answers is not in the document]
olavloite: Thanks for the detailed explanation. I think this will work in most cases, but I have a couple of questions / worries for some corner cases.

I worry a little bit about the way that a channel is picked for the `BatchCreateSessions` RPC. In most cases for the initial creation of the session pool, it will probably work well, as the initial calls are initiated in parallel. There are still a couple of scenarios that could cause the behavior to change from what it is currently:
* If one of the `BatchCreateSessions` invocations should finish before the background executor has picked up all of the requests, it could be that one channel is assigned more calls than the others. This is not very probable when using the default settings and using only one `DatabaseClient`, as the executor uses more threads than the default number of channels, and each channel only needs one invocation (one invocation of `BatchCreateSessions` will never return more than 100 sessions, even if the requested session count is more than 100). However, the [default executor](https://github.com/googleapis/java-spanner/blob/0f56261857aa9cb516bf31f796e1badb0a792c86/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java#L181) that is used for the `SessionClient` is shared among all `DatabaseClient` instances that are created from a `Spanner` instance. That means that if an application creates multiple `DatabaseClient` instances (for different databases) during startup, the executor could be fully exhausted for a while, causing the initial `BatchCreateSessions` for at least some of the `DatabaseClient`s to not be executed in parallel.
* The default configuration sets `maxConcurrentStreamsLowWatermark` to 1 to force the creation of a new channel each time a (parallel) request is done. Is that only used to determine when to create a new channel in the pool, or will it also affect the number of parallel streams on a channel that can be used during the lifetime of the application?
* What happens if the user sets a higher value for `maxConcurrentStreamsLowWatermark` in their own configuration? (Can they do that?)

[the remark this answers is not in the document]
nimf: Thank you for pointing this out! Yeah, if there is a chance of some BatchCreateSession requests to finish before some others are started during the session pool creation, then we can go safer with configuring `maxConcurrentStreamsLowWatermark` to `0`. In this case a new channel will be created for each BatchCreateSession request regardless of if they are running concurrently or sequentially.

Answering two other questions:

- `maxConcurrentStreamsLowWatermark` is only used to determine if a new channel should be created. And it works up until the point when the pool reaches the maximum size.
- A user cannot change the `maxConcurrentStreamsLowWatermark`. `maxConcurrentStreamsLowWatermark` is configured via `ApiConfig` (together with affinity configuration for the Spanner RPCs) which is provided by the Spanner client and a user provides `GcpManagedChannelOptions` which configures other aspects like metrics and resiliency options.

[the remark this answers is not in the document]
olavloite: I think changing `maxConcurrentStreamsLowWatermark` to 0 will answer my last questions / worries. Other than that, this seems good to me.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
nimf: Changed `setGrpcGcpOptions` to `withGrpcGcpOptions`. But I also keep `setUseGrpcGcpExtension` to be able to disable the GCP extension.

google-cloud-spanner/src/main/resources/com/google/cloud/spanner/spi/v1/grpc-gcp-apiconfig.json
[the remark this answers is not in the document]
nimf: Changed low watermark to zero.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -491,6 +504,63 @@ public GapicSpannerRpc(final SpannerOptions options) {
...
+        MoreObjects.firstNonNull(options.getGrpcGcpOptions(), new GcpManagedChannelOptions());
+    GcpMetricsOptions metricsOptions = MoreObjects.firstNonNull(
+        grpcGcpOptions.getMetricsOptions(), GcpMetricsOptions.newBuilder().build());
+    GcpMetricsOptions.Builder metricsOptionsBuilder = GcpMetricsOptions.newBuilder(metricsOptions);
+    if (metricsOptions.getMetricRegistry() == null) {
+      metricsOptionsBuilder.withMetricRegistry(Metrics.getMetricRegistry());
+    }
+    // TODO: Add default labels with values: client_id, database, instance_id.
+    if (metricsOptions.getNamePrefix().equals("")) {
+      metricsOptionsBuilder.withNamePrefix("cloud.google.com/java/spanner/gcp-channel-pool/");
```
nimf, line 532: BTW, I could take `"cloud.google.com/java/spanner"` from the `MetricRegistryConstants` if it is ok to make that class public.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
olavloite: I think this still will be a little confusing for users, and that many will think that you need to call both to enable it. Do we really need this? Or would it maybe be better to change it to something like `disableGrpcGcpExtension()`?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -509,6 +520,63 @@ public GapicSpannerRpc(final SpannerOptions options) {
     }
   }
 
+  private static String parseGrpcGcpApiConfig() {
+    InputStream inputStream = GapicSpannerRpc.class.getResourceAsStream(API_FILE);
```
olavloite, line 524: I think you can simplify this method somewhat by using (something like) `com.google.common.io.Resources.toString(resourceClass.getResource(filename), Charset.forName("UTF8"));`

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
olavloite: My gut feeling is that we should try to keep that class package private, as it would otherwise become part of the public API and restrict us from making any changes in the future.

olavloite:
In order to fix the failed lint check you need to run the following command:

```
mvn com.coveo:fmt-maven-plugin:format
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
nimf: Another point for keeping this is that it is not necessary to provide options to enable the extension - in this case the extension will be enabled with the defaults. I can change this into two methods enable/disable GrpcGcpExtension. What's more common for SpannerOptions?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
nimf: Thanks! Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
thiagotnunes: Ah ok, so we have default options that can be applied. Is this case I would overload the `enable` method and have it as:

```java
// Uses the default options
enableGrpcGcpExtension();

// Uses the custom options
enableGrpcGcpExtension(GcpManagedChannelOptions);
```

google-cloud-spanner/pom.xml
```diff
@@ -117,6 +117,12 @@
   </build>
 
   <dependencies>
+
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>grpc-gcp</artifactId>
+      <version>1.0.0</version>
```
thiagotnunes, line 124: We tend to use the versions specified in [java-shared-dependencies](https://github.com/googleapis/java-shared-dependencies/blob/master/pom.xml). By that, I mean we usually do NOT specify version dependencies (like [here](https://github.com/googleapis/java-spanner/blob/master/google-cloud-spanner/pom.xml#L120-L123)), but instead rely on the versions provided by the [POM import](https://github.com/googleapis/java-spanner/blob/master/pom.xml#L107-L113) of that repository.

Could we do the same for this artifact (if it is part of any BOMs specified in the shared dependencies)?

olavloite:
Also, could we have a couple of tests for this feature that:
1. Verifies that the creation of sessions is actually spread across multiple channels. See `SessionClientTest` for the current tests for the non-gRPC-GCP channel pool solution.
2. An end-to-end test that enables this option and executes a couple of transactions to verify that enabling this option works (and continues to work in the future).

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
nimf: Sounds good! Done.

nimf:
> Also, could we have a couple of tests for this feature that:

> 2. An end-to-end test that enables this option and executes a couple of transactions to verify that enabling this option works (and continues to work in the future).

Added.

> 1. Verifies that the creation of sessions is actually spread across multiple channels. See `SessionClientTest` for the current tests for the non-gRPC-GCP channel pool solution.

For the GAX channel pool it is important which `CHANNEL_HINT` values the Spanner client uses so that every BatchCreateSession request uses different channel. And the `SessionClientTest` currently verifies that correct channel hints values are used. But the gRPC-GCP channel pool doesn't use channel hints. Instead these are the things required from the Spanner client to spread the sessions correctly:
- Issue N+ BatchCreateSession requests (where N is the number of channels) (this is probably already tested)
- Set up GAX with the pool size of 1
- Set up low watermark for gRPC-GCP to 0

Can you give me a hint on how it is easier to test the last two, please?

google-cloud-spanner/pom.xml
nimf: It is not a part of any BOMs. If I add it to the java-shared-dependencies will all the projects depending on it automatically get grpc-gcp as a dependency or they must explicitly add it to their dependencies?

thiagotnunes: They must explicitly add it to their dependencies, but they won't need to specify a version. This makes sure that we are always using compatible library versions.

olavloite:
@nimf I can't push any changes to your branch, so I'll attach a proposal for a test as an attachment here. Please feel free to make any changes you think would be appropriate to the test.
[ChannelUsageTest.java.txt](https://github.com/googleapis/java-spanner/files/6702163/ChannelUsageTest.java.txt)

nimf:
> @nimf I can't push any changes to your branch, so I'll attach a proposal for a test as an attachment here. Please feel free to make any changes you think would be appropriate to the test.
> [ChannelUsageTest.java.txt](https://github.com/googleapis/java-spanner/files/6702163/ChannelUsageTest.java.txt)

Wow, it's perfect! Thank you so much!

I've added you as the author, but now it requires a consent. Can you give your consent, please?

olavloite:
@googlebot I consent.

olavloite (approved):
LGTM

I assume the 1.0.0 version for the grpc-gcp in the pom.xml can be removed once the shared dependencies with this dependency included has been released (?)

nimf:
> I assume the 1.0.0 version for the grpc-gcp in the pom.xml can be removed once the shared dependencies with this dependency included has been released (?)

Yes, exactly. I've already added grpc-gcp to the java-shared-dependencies but it hasn't been released yet.

PR #1263, opened 2021-06-29
feat: support setting an async executor provider

### LINKED ISSUE #1262: support for custom ExecutorProvider in SpannerOptions.Builder
**Is your feature request related to a problem? Please describe.**
I would like to configure ExecutorProvider in SpannerOptions so can control the executor configs. Currently the Builder doesn't provide any setter for it. Due to this, the asyncExecutorProvider is null and hence the below code always falls back to the default provider:

```java
SpannerImpl(SpannerRpc gapicRpc, SpannerOptions options) {
...

    this.asyncExecutorProvider =
        MoreObjects.firstNonNull(
            options.getAsyncExecutorProvider(),
            SpannerOptions.createDefaultAsyncExecutorProvider());
...
}
```

**Describe the solution you'd like**
Add setter in SpannerOptions.Builder class.

**Describe alternatives you've considered**
So far found no alternative.

**Additional context**
Discussed about this issue at [discussion](https://github.com/googleapis/java-spanner/pull/81#discussion_r659277576)

olavloite:
The Spanner Async API uses an `ExecutorProvider` for producing rows for an `AsyncResultSet`. The default that is automatically created is fine for most use cases, but did not allow users to override the default settings that are:
- a pool containing at most 8 threads
- threads automatically time out after 1 minute if they are not used

This PR adds the possibility for users to supply their own `ExecutorProvider` to use with the async API.

Fixes #1262

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
@@ -471,13 +471,17 @@ public class SpannerOptions extends ServiceOptions<Spanner, SpannerOptions> {
   private static final AtomicInteger DEFAULT_POOL_COUNT = new AtomicInteger();
 
   /** {@link ExecutorProvider} that is used for {@link AsyncResultSet}. */
-  interface CloseableExecutorProvider extends ExecutorProvider, AutoCloseable {
+  public interface CloseableExecutorProvider extends ExecutorProvider, AutoCloseable {
     /** Overridden to suppress the throws declaration of the super interface. */
     @Override
     void close();
   }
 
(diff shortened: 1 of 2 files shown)


thiagotnunes (approved):
Do you know if there are any metrics available for the thread pool? I think we should expose thread pool usage in cloud monitoring for the default executor.

olavloite:
> Do you know if there are any metrics available for the thread pool? I think we should expose thread pool usage in cloud monitoring for the default executor.

The standard executor services in Java do not expose any metrics, so we would need to add that ourselves (or add some open source library, there are a couple that do this). I'll add a feature request for it to the repository.

pravsingh (commented):
Thank you!

PR #1266, opened 2021-06-30
samples: updates lci samples

thiagotnunes:
Updates the LCI samples to be accordingly to the spec. Instead of printing out the processing units directly from the instance that is created, another getInstance is done in order to retrieve the information.

### DIFF
--- a/samples/snippets/src/main/java/com/example/spanner/CreateInstanceWithProcessingUnitsExample.java
+++ b/samples/snippets/src/main/java/com/example/spanner/CreateInstanceWithProcessingUnitsExample.java
@@ -17,6 +17,7 @@
 package com.example.spanner;
 
 //[START spanner_create_instance_with_processing_units]
+
 import com.google.api.gax.longrunning.OperationFuture;
 import com.google.cloud.spanner.Instance;
 import com.google.cloud.spanner.InstanceAdminClient;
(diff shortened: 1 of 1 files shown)


samples/snippets/src/main/java/com/example/spanner/CreateInstanceWithProcessingUnitsExample.java
```diff
@@ -47,26 +47,28 @@ static void createInstance(String projectId, String instanceId) {
...
-          "Error: Creating instance %s failed with error message %s%n",
-          instanceInfo.getId(), e.getMessage());
-    } catch (InterruptedException e) {
-      System.out.println("Error: Waiting for createInstance operation to finish was interrupted");
+      System.out.printf("Waiting for operation on %s to complete...%n", instanceId);
+      Instance createdInstance = operation.get();
+
+      System.out.printf("Created instance %s.%n", createdInstance.getId().getInstance());
+
+      Instance instance = instanceAdminClient.getInstance(instanceId);
```
hengfengli, line 67: How do we use `FieldMask` here?

hengfengli: We need to show users how to use `FieldMask`.

hengfengli (approved):

PR #1283, opened 2021-07-06
feat: exposes default leader in database, and leader options / replicas in instance config

thiagotnunes:
This PR implements the functionality for CMMR phase 1. Here we do the following:

- Exposes a new field, default leader in the Database class. This will be populated when retrieving / listing the databases if the value was specified when creating or updating the database.
- Exposes a new field, replicas in the InstanceConfig class. This will be a list of replicas information available in the given instance (for more details see https://cloud.google.com/spanner/docs/replication#overview_of_replication_in_cloud_spanner).
- Exposes a new field, leader options in the InstanceConfig class. This will be a list of available leader locations for the databases using the given instance config. This will be populated when retrieving / listing the instance configs.

#### TODO
- [ ] Add Integration test for setting a database's default leader on CreateDatabase operation
- [ ] Add Integration test for setting a database's default leader using UpdateDatabaseDdl operation


thiagotnunes:
Backend has not enabled this functionality yet, so the integration tests should not succeed for the moment.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseTest.java
```diff
@@ -100,27 +102,13 @@ public void listDatabaseOperations() {
...
-    assertThat(db.getId().getName()).isEqualTo(NAME);
-    assertThat(db.getState()).isEqualTo(DatabaseInfo.State.CREATING);
-    assertThat(db.getVersionRetentionPeriod()).isEqualTo(VERSION_RETENTION_PERIOD);
-    assertThat(db.getEarliestVersionTime()).isEqualTo(EARLIEST_VERSION_TIME);
-    assertThat(db.getEncryptionConfig())
-        .isEqualTo(EncryptionConfigs.customerManagedEncryption(KMS_KEY_NAME));
-  }
-
-  @Test
-  public void testFromProtoWithEncryptionConfig() {
```
hengfengli, line 112: why do we remove this test?

hengfengli (approved):

hengfengli:
I don't the emulator would support `default_leader` and `leader_options`. You may want to skip these tests for the emulator.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseTest.java
thiagotnunes: We did not remove it, just changed the `assertThat` to `assertEquals`.

thiagotnunes:
@hengfengli thanks, will do

google-cloud-spanner/src/main/java/com/google/cloud/spanner/ReplicaInfo.java
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  public static class BuilderImpl extends Builder {
+
+    private String location;
+    private ReplicaType type;
+    private boolean defaultLeaderLocation;
+    private com.google.spanner.admin.instance.v1.ReplicaInfo proto;
+
+    /** The location of the serving resources. */
```
olavloite, line 44: Would it be possible (and make sense) to reference some online documentation that indicates what the possible values are, including the format that is expected?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseInfo.java
```diff
@@ -44,6 +45,11 @@
      */
     public abstract Builder setEncryptionConfig(CustomerManagedEncryption encryptionConfig);
 
+    /** The read-write region which will be used for the database's leader replicas. */
```
olavloite, line 48: Would it be possible (and make sense) to reference some online documentation that indicates what the possible values are, including the format that is expected?

hengfengli: I don't think we have published any doc yet. The possible value depends on the returned values in `leader_options`.

hengfengli: The format should like: https://cloud.google.com/spanner/docs/instances#available-configurations-multi-region

hengfengli: Just confirmed with the backend team. The option would not have the `regional-` prefix. It will be something like `us-east1` and `us-central1`.

hengfengli: @thiagotnunes I think the best we can do is point to [the multi-region instance configs page](https://cloud.google.com/spanner/docs/instances#available-configurations-multi-region) and say that all RW region names are valid. And that you can use `getInstanceConfig` to fetch the `leader options` available.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/ReplicaInfo.java
thiagotnunes: Updated the javadoc with the link to the possible values here.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseInfo.java
thiagotnunes: Updated the javadoc with the link to the possible values here.

olavloite (approved):

PR #1303, opened 2021-07-19
samples: add default leader options samples

thiagotnunes:
Adds the following samples regarding the default leader options:

1. GetInstanceConfig (for `nam6`) printing the available leader options
2. ListInstanceConfigs printing the available leader options
3. ListDatabases printing the default leader (if set) for each database
4. CreateDatabaseWithDefaultLeader
5. UpdataDatabaseDefaultLeader
6. GetDatabaseDdl printing all the DDL statements
7. QueryInformationSchema printing out the default leader set

- [x] TODO: Add integration tests for each sample


samples/snippets/src/main/java/com/example/spanner/QueryInformationSchemaDatabaseOptionsSample.java
```diff
@@ -0,0 +1,67 @@
...
+                  + " FROM INFORMATION_SCHEMA.DATABASE_OPTIONS"
+                  + " WHERE OPTION_NAME = 'default_leader'")
+          )) {
+        if (resultSet.next()) {
+          final String optionName = resultSet.getString("OPTION_NAME");
+          final String optionValue = resultSet.getString("OPTION_VALUE");
+
+          System.out.println("The " + optionName + " for " + id + " is " + optionValue);
+        } else {
+          System.out.println("Could not retrieve the default_leader option for " + id);
```
olavloite, line 61: nit: I think that a message like `Database id does not have a value for option 'default_leader'` would be more appropriate. There is no error occurring here, and no option value in this table means that no specific option has been set (at least, that is the case for other options, so I assume that this will be the case for leader options as well).

olavloite (approved):
LGTM, with a small nit/question on one of the log messages.

samples/snippets/src/main/java/com/example/spanner/QueryInformationSchemaDatabaseOptionsSample.java
thiagotnunes: Makes sense I will update the messages, thanks!

hengfengli:
@thiagotnunes The checkstyle fails due to a line length error: 

```
Error:  src/main/java/com/example/spanner/QueryInformationSchemaDatabaseOptionsSample.java:[61] (sizes) LineLength: Line is longer than 100 characters (found 102).
```

The 23.7 release has been rolled out. Can we try to rerun the integration test?

thiagotnunes:
Integration tests for samples will fail because we don't have a MR instance to run them against. We should fix this before merging.

PR #1316, opened 2021-07-26
refactor: refactor sample tests

thiagotnunes:
Extracts test base class to be re-used in sample tests.

This will be useful to write tests for new samples.


samples/snippets/src/test/java/com/example/spanner/SampleTestBase.java
```diff
@@ -0,0 +1,75 @@
...
+import com.google.cloud.spanner.DatabaseAdminClient;
+import com.google.cloud.spanner.Spanner;
+import com.google.cloud.spanner.SpannerOptions;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+
+/**
+ * Base class for sample integration tests.
+ */
+public class SampleTestBase {
```
thiagotnunes, line 28: Had to use inheritance here =(.

@olavloite let me know if that is ok or if you know of any better way to enforce `beforeClass` / `afterClass` in the tests.

samples/snippets/src/test/java/com/example/spanner/SampleRunner.java
```diff
@@ -19,21 +19,29 @@
...
 import java.util.function.Predicate;
 
-/** Runs a sample and captures the output as a String. */
+/**
+ * Runs a sample and captures the output as a String.
+ */
 public class SampleRunner {
-  public static String runSample(Callable<Void> sample) throws Exception {
+
+  @FunctionalInterface
```
thiagotnunes, line 29: Refactor so that the `runSample` does not have to return any values.

samples/snippets/src/test/java/com/example/spanner/SampleTestBase.java
olavloite: As far as I know, this is the best way to do that in Junit (4).

samples/snippets/src/test/java/com/example/spanner/SampleTestBase.java
```diff
@@ -0,0 +1,72 @@
...
+    idGenerator = new SampleIdGenerator(BASE_DATABASE_ID, BASE_BACKUP_ID);
+  }
+
+  @AfterClass
+  public static void tearDown() {
+    for (String databaseId : idGenerator.getDatabaseIds()) {
+      try {
+        databaseAdminClient.dropDatabase(instanceId, databaseId);
+      } catch (Exception e) {
+        System.out.println("Failed to drop database " + databaseId + ", skipping...");
```
olavloite, line 60: nit: I think it could be useful to print out the error message as well here to get an idea why it failed.

samples/snippets/src/test/java/com/example/spanner/SampleTestBase.java
```diff
@@ -0,0 +1,72 @@
...
+      }
+    }
+    for (String backupId : idGenerator.getBackupIds()) {
+      try {
+        databaseAdminClient.deleteBackup(instanceId, backupId);
+      } catch (Exception e) {
+        System.out.println("Failed to delete backup " + backupId + ", skipping...");
```
olavloite, line 67: nit: same as above

olavloite (approved):
Thanks for the cleanup, this looks good to me.

samples/snippets/src/test/java/com/example/spanner/SampleTestBase.java
thiagotnunes: Good point, added.

samples/snippets/src/test/java/com/example/spanner/SampleTestBase.java
thiagotnunes: Good point, added.

samples/snippets/src/test/java/com/example/spanner/SampleTestBase.java
thiagotnunes: Thanks for confirming.

PR #1319, opened 2021-07-26
chore: add ci task to compile samples

thiagotnunes:
Since the samples tests are not being run on every PR, we add a step to at least compile the samples to make sure we don't introduce any hard failures.

### DIFF
--- a/.github/workflows/samples.yaml
+++ b/.github/workflows/samples.yaml
@@ -14,10 +14,13 @@ jobs:
         working-directory: samples/snippets
   compile:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        java: [8, 11]
     steps:
     - uses: actions/checkout@v2
     - uses: actions/setup-java@v1
       with:
-        java-version: 8
+        java-version: ${{matrix.java}}
     - name: Compile samples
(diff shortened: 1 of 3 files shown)


.github/workflows/samples.yaml
```diff
@@ -12,3 +12,12 @@ jobs:
       - name: Run checkstyle
         run: mvn -P lint --quiet --batch-mode checkstyle:check
         working-directory: samples/snippets
+  compile:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-java@v1
+      with:
+        java-version: 8
```
olavloite, line 21: Could we do this for both Java 8 and Java 11? We had a failure in the past where a compilation error would only occur for Java 11 and not for Java 8.

olavloite (approved):
LGTM. If it would be possible to compile using both Java 8 and Java 11, it would be even better =)

.github/workflows/samples.yaml
thiagotnunes: Yup, changed it to compile against Java 8 and 11.

PR #1328, opened 2021-08-02
Autocommit

dzhang-bap:
This is to illustrate the implementation to inline commit into ExecuteSqlRequest, to improve the performance of executing a DML statement when the connection is in autocommit mode.


olavloite:
@dzhang-bap Please let us know when this is ready for review.

Note: You could also [convert this to a draft PR](https://github.blog/changelog/2020-04-08-convert-pull-request-to-draft/#:~:text=To%20convert%20a%20pull%20request,you%20to%20confirm%20the%20change.) while it is work-in-progress. Once it is ready for review, you can convert it back to a normal PR and we should automatically be notified.

dzhang-bap:
@thiagotnunes and @olavloite: this is ready for review now. Here's an explanation of the commits:

- feat: temporary synth'ed proto change to enable developing autocommit.
  - No need to review! This commit will be deleted before merging.
  - Until googleapis gets changed to include the new protobuf fields (ExecuteSqlRequest.autocommit and others), this temporary commit enables subsequent commits to use these fields.

- feat: support inline commit into an ExecuteSqlRequest.
  - If Connection.isAutocommit() is true, Connection.execute() or Connection.executeUpdate() will commit the transaction after executing a DML statement.
  - Under the hood, SingleUseTransaction.createWriteTransaction() will call DatabaseClient.readWriteTransaction() with a newly introduced option: Options.inlineCommitOption().
  - The users may directly call DatabaseClient.readWriteTransaction() with Options.inlineCommitOption(), and a DML statement will lead the transaction to be committed. (The transactionRunner should have a single DML statement.)
  - TransactionRunnerImpl.TransactionContextImpl is the main place where inline commit is implemented:
     - An InlineCommitState is introduced.
     - In runInternal(), if inlineCommit, do not inlineBeginTransaction. This is because we cannot inline both BeginTransaction and Commit into the same ExecuteSqlRequest, for replay protection purposes.
     - In executeUpdate(), if inline commit, getExecuteSqlRequestBuilder() will have autocommit set to true; and the returned resultSet should include a commitResponse which is recorded.
     - In Commit(), if the transaction was already committed, skip Commit.

- test: fix unit test failures.
  - There are a few unit tests that mock DatabaseClient.readWriteTransaction(). They should also mock the call if it uses Options.inlineCommitOption() as an argument.
 
- test: temporary commit that adds Donghui.java and changes integration...
  - This temporary commit will be deleted before merging.
  - One purpose is to temporarily change integration test to use donghuiz-instance in Cloud-Devel.
  - Another purpose is to introduce my temporary tests Donghui.java including performance tests. The result (24% speedup if using donghuiz-instance, or 2% speedup if using test-instance) is detailed in the comment at testPerformance().

- test: add an integration test for using inline autocommit in Database… 
   - Adds an integration test for exercising inlineCommit directly from DatabaseClient.
   - Note that the existing test ITReadWriteAutocommitSpannerTest already covers the case when Connection.isAutocommit()=true.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
```diff
@@ -462,11 +480,15 @@ TransactionSelector getTransactionSelector() {
...
             // BeginTransaction with the statement. The others will be waiting on the
             // transactionIdFuture until an actual transactionId is available.
+            // If inlineCommit is true, the first statement should wait on the transactionIdFuture.
             if (transactionIdFuture == null) {
               transactionIdFuture = SettableApiFuture.create();
               if (trackTransactionStarter) {
                 transactionStarter = new Exception("Requesting new transaction");
               }
+              if (options.inlineCommit()) {
+                tx = transactionIdFuture;
```
olavloite, line 490: I don't quite understand how this would work. the `transactionIdFuture` has just been created above as a `SettableApiFuture` with no value. On line 511 we will now be waiting for this future to return a value, but no one will be setting a value for it. Or am I missing something here?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
```diff
@@ -671,9 +701,18 @@ public long executeUpdate(Statement statement, UpdateOption... options) {
           throw new IllegalArgumentException(
               "DML response missing stats possibly due to non-DML statement as input");
         }
+        if (inlineCommitState == InlineCommitState.COMMITTING) {
+          if (resultSet.hasCommitResponse()) {
+            inlineCommitState = InlineCommitState.SUCCEEDED;
+            commitResponse = new CommitResponse(resultSet.getCommitResponse());
+          } else {
+            inlineCommitState = InlineCommitState.MISSING_COMMIT_RESPONSE;
```
olavloite, line 709: Shouldn't this also throw an exception? The user assumes that the statement was executed with autocommit enabled, and that it therefore also was committed. It now seems to silently fail.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
```diff
@@ -671,9 +701,18 @@ public long executeUpdate(Statement statement, UpdateOption... options) {
...
+          }
+        }
         // For standard DML, using the exact row count.
         return resultSet.getStats().getRowCountExact();
       } catch (Throwable t) {
+        inlineCommitState = InlineCommitState.FAILED;
```
olavloite, line 715: You should only set this if autocommit was enabled.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
```diff
@@ -123,6 +123,15 @@ public static ListOption pageToken(String pageToken) {
     return new PageTokenOption(pageToken);
   }
 
+  /**
+   * Specifying this will cause the transaction to be committed when executing the DML statement. A
+   * requirement is that the transaction should have a single action which is a DML statement. This
+   * option is automatically specified if Connection.isAutocommit() is true.
+   */
+  public static InlineCommitOption inlineCommitOption() {
```
olavloite, line 131: Would it in some way be possible to make this package private, or otherwise not accessible for normal users of the client library? It seems like an option that might be difficult for users to use in the correct way when using a normal transaction runner. If we want to support this in the client library, I think it would be better to add a separate method for it. Something like `DatabaseClient#executeUpdateAutocommit(..)`. That way the user does not have any way of accidentally executing multiple statements in a transaction with autocommit enabled.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITInlineCommitDatabaseClientTest.java
```diff
@@ -0,0 +1,166 @@
...
+        .readWriteTransaction(Options.inlineCommitOption())
+        .run(
+            transaction -> {
+              // Do a query.
+              ResultSet resultSet = transaction.executeQuery(Statement.of(COUNT));
+              assertTrue(resultSet.next());
+              assertEquals(0, resultSet.getLong(0));
+              assertFalse(resultSet.next());
+
+              // DML after query should succeed and should cause the transaction to be committed.
```
olavloite, line 115: I think that this will be a confusing API for users. The transaction is in `autocommit` mode, which means that you can execute any number of queries, but the first DML statement that is executed will end the transaction. I understand why it technically works like that, but I do worry that this might cause confusion.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITInlineCommitDatabaseClientTest.java
```diff
@@ -0,0 +1,166 @@
...
+        .readWriteTransaction(Options.inlineCommitOption())
+        .run(
+            transaction -> {
+              // The first DML should cause the transaction to be committed.
+              long rowCount = transaction.executeUpdate(Statement.of(insertDml()));
+              assertEquals(DML_COUNT, rowCount);
+
+              // Additional DML should error out.
+              try {
+                transaction.executeUpdate(Statement.of(insertDml()));
```
olavloite, line 135: nit: use `assertThrows(..)` instead.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
```diff
@@ -123,6 +123,15 @@ public static ListOption pageToken(String pageToken) {
     return new PageTokenOption(pageToken);
   }
 
+  /**
+   * Specifying this will cause the transaction to be committed when executing the DML statement. A
```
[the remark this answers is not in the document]
dzhang-bap: Executing other queries is fine but the first DML will cause the transaction to be committed, and any query or DML after that in the same transaction will fail.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
```diff
@@ -286,6 +300,10 @@ void commit() {
     volatile ApiFuture<CommitResponse> commitFuture;
 
     ApiFuture<CommitResponse> commitAsync() {
+      checkState(
```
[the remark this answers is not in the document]
dzhang-bap: Ah, good point. I didn't see any failure in "mvn test" but will modify the code accordingly.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
dzhang-bap: About 20 lines below, there is a call to tx.get() with comment "wait for the transaction to come available..." I believe that call will set the value for it. No?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
dzhang-bap: I thought about throwing an exception but I think it is not as user friendly, if for whatever reason the DML was executed successfully but the transaction is not committed.
I think the existing code's behavior is to silently fail only internally; since the transaction runner will call Commit() explicitly, the external behavior to the users may be that they will see the DML executed successfully and the transaction is committed.
But I can be persuaded either way.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
dzhang-bap: That was our initial plan. But I couldn't find a way to implement it.
What we need is for Connection to pass to DatabaseClient the info that Connection.isAutocommit() is true. But they are in different packages. Any function we introduce to DatabaseClient that enables Connection to call, the users can call too. Thiago and I discussed this too. Anything I missed?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
dzhang-bap: ok

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITInlineCommitDatabaseClientTest.java
dzhang-bap: One thing we may do is to ensure the comments at Options.inlineCommitOption is as clear as possible; e.g. we may say setting this requires the transaction to have nothing but a single DML statement. Are there better options?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITInlineCommitDatabaseClientTest.java
dzhang-bap: ok

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
olavloite: Ah, yes, that is true.

What are your thoughts on adding one of the following methods to `DatabaseClient` instead of allowing this option for `TransactionRunner`:
- `DatabaseClient#executeUpdate`: Allows the user to execute one single update statement without having to create a `TransactionRunner`. The method would use the inline commit option internally. The connection API could also call this method when it wants to execute a single update statement in autocommit mode.
- `DatabaseClient#singleUseTransaction()`: Returns a single-use read/write transaction that can be used to execute one update. Works otherwise in the same way as the above `executeUpdate` method. The idea behind this would be to keep it in line with how users execute single reads (`singleUse().executeQuery(..)`). It would also be a good entry point if we want to support inline commit for Batch DML: `singleUseTransaction().batchUpdate(statement1, statement2, ...)`

The disadvantage of the above is obviously that it does not allow any select-then-update semantics when using inline commit, but I'm not sure that is a big miss, considering that this option is mainly intended for supporting autocommit in the Connection API.

@thiagotnunes Do you have any thoughts on the above?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
olavloite: No. That line of code will wait for 60 seconds for a transaction to come available when __a different statement__ has already been executed and included an inlined `BeginTransaction` option. An application can use the async API of the `TransactionRunner` to execute multiple statements (partly) in parallel. Consider the following example:

```java
// The following query will include the BeginTransaction option in the request.
var rs1 = transaction.executeQueryAsync("SELECT * FROM Foo");

// The following query must wait until the first query has returned a transaction ID before it can start.
var rs2 = transaction.executeQueryAsync("SELECT * FROM Bar");
```

The first query in the above example will initiate the transaction by inlining the BeginTransaction option. The second query would normally have been executed in parallel with the first query, but in this case it has to wait for the first one to return a transaction id. That is the `wait for the transaction to come available...` part does in the code.

The reason that this (probably) works is because it is dead code. The if-branch will only be entered `if (transactionIdFuture == null)`. When inline commit is enabled, this will never be null, because it does not use inline begin. That means that the first action of the transaction is executing a `BeginTransaction` RPC which will initialize the `transactionIdFuture`.

Short version: I'm pretty sure that you can (and should) remove this whole if-block.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
olavloite: I would be leaning towards throwing an exception here for the following reasons:
1. If this is a situation that could happen sometimes (i.e. this is not a once in a trillion requests thing), we would have the strange situation that in some cases a transaction runner with inline commit does accept multiple DML statements. Or would the transaction still refuse any following DML statement even when the commit failed in such a case?
2. If this is a situation that doesn't really ever happen (i.e. this indicates some internal failure that should never happen), then I think it's better to fail as fast as possible and rollback the transaction. Otherwise, what are the chances that the following Commit call will succeed?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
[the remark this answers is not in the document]
olavloite: My bad, the `inlineCommitState` will obviously NOT be `SUCCEEDED` here if `AsyncTransactionManager` is being used, so this check will not fail. But (while looking into this):
1. Another problem with the current `InlineCommit` option is that it is defined as a `TransactionOption`. Those can also be specified when creating a `TransactionManager` or `AsyncTransactionManager`. What happens when a user does that? As far as I can tell from the code, it will be ignored, but that might also be a little confusing for users.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITInlineCommitDatabaseClientTest.java
olavloite: See my comments above. I think that if this is really only intended for executing a single DML statement, that we should implement it as a single method that only accepts one DML statement. That could be either something like:

```java
DatabaseClient client = spanner.getDatabaseClient(..);
client.executeUpdate("UPDATE Foo");
```

or something like

```java
DatabaseClient client = spanner.getDatabaseClient(..);
client.singleUseTransaction().executeUpdate("UPDATE Foo");
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
thiagotnunes: We raised this point before, but could not find any good enough solution. However, from reading your suggestions I quite like the introduction of a `DatabaseClient#singleUseTransaction` as it fits very nicely with the other methods that we have. This way we could remove the `Options#inlineCommit` and prevent users from having multiple statements within an auto-commit enabled transaction.

@dzhang-bap could you change the implementation to introduce this interface method instead? When you do, provide a default implementation that just throws an error like [here](https://github.com/googleapis/java-spanner/blob/7d6816f1fd14bcd2c7f91d814855b5d921ba970d/google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionContext.java#L107-L109), in order to prevent breaking changes.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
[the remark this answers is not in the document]
dzhang-bap: given the change of using singleDmlTransaction(), this is no longer relevant.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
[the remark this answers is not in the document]
dzhang-bap: changed the error message to be more generic.
Removed Options.inlineCommitOption() so users should not be able to set it.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
dzhang-bap: Indeed when using inlineCommit, transactionId cannot be empty. I've added a checkState() at the beginning of the "if (transactionId==null)" block.
This block cannot be removed because it is still in use when inlineCommit is not used.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
dzhang-bap: I tried to throw an exception here but encountered 6 unit test failures. I believe the root cause is that mockSpanner (defined in AbstractMockServerTest) does not return a commit response. I've added a TODO item there to change to throwing an exception after mockSpanner is extended.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
dzhang-bap: I love the idea!!! I've introduced a new class SingleDmlTransaction, and added DatabaseClient#singleDmlTransaction(). Note that there exists a class called SingleUseTransaction so I want to use a different name. Besides, "SingleDmlTransaction" seem more accurate because we only allow a DML here, not query.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITInlineCommitDatabaseClientTest.java
dzhang-bap: removed Options.inlineCommitOption.

google-cloud-spanner/pom.xml
```diff
@@ -69,8 +69,9 @@
         <configuration>
           <systemPropertyVariables>
             <spanner.testenv.config.class>com.google.cloud.spanner.GceTestEnvConfig</spanner.testenv.config.class>
-            <spanner.testenv.instance>projects/gcloud-devel/instances/spanner-testing</spanner.testenv.instance>
-            <spanner.gce.config.project_id>gcloud-devel</spanner.gce.config.project_id>
+            <spanner.testenv.instance>projects/span-cloud-testing/instances/donghuiz-instance</spanner.testenv.instance>
```
thiagotnunes, line 72: We will remove this before merging

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SingleDmlTransaction.java
```diff
@@ -0,0 +1,63 @@
...
+
+public class SingleDmlTransaction {
+  private static final String READ_WRITE_TRANSACTION = "CloudSpanner.ReadWriteTransaction";
+  private static final Tracer tracer = Tracing.getTracer();
+  private final TransactionRunner runner;
+
+  SingleDmlTransaction(PooledSessionFuture session, TransactionOption... options) {
+    Span span = tracer.spanBuilder(READ_WRITE_TRANSACTION).startSpan();
+    try (Scope s = tracer.withSpan(span)) {
+      TransactionOption[] new_options = Arrays.copyOf(options, options.length + 1);
```
thiagotnunes, line 37: nit: use camel case instead of snake case

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SingleDmlTransaction.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+  }
+
+  @VisibleForTesting
+  public SingleDmlTransaction(TransactionRunner runner) {
+    this.runner = runner;
+  }
+
+  public long executeUpdate(Statement statement, UpdateOption... options)
+      throws NullPointerException, SpannerException {
```
thiagotnunes, line 54: nit: no need to put the exceptions in the function signature, since they are unchecked exceptions

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SingleDmlTransaction.java
```diff
@@ -0,0 +1,63 @@
...
+    UpdateOption[] new_options = Arrays.copyOf(options, options.length + 1);
```
thiagotnunes, line 55: nit: use camel case instead of snake case

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
```diff
@@ -456,12 +473,14 @@ TransactionSelector getTransactionSelector() {
...
       // withInlineBegin and an earlier statement has already started a transaction.
       if (transactionId == null) {
+        checkState(!options.inlineCommit(), "InlineCommit requires a transactionId.");
         try {
           ApiFuture<ByteString> tx = null;
           synchronized (lock) {
             // The first statement of a transaction that gets here will be the one that includes
             // BeginTransaction with the statement. The others will be waiting on the
             // transactionIdFuture until an actual transactionId is available.
+            // If inlineCommit is true, the first statement should wait on the transactionIdFuture.
```
thiagotnunes, line 483: Remove this comment

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
thiagotnunes: We plan to do this work in a separate PR

google-cloud-spanner/src/test/java/com/google/cloud/spanner/Donghui.java
```diff
@@ -0,0 +1,392 @@
...
+//
+// For a quick run:
+// - One thing that prevents a run from finishing in seconds is that, if the table does not already
+//   include all keys in [0..NUM_PRE_INSERTED-1], these keys will be inserted. But this is a
+//   one-time thing.
+// - Another things that may cause the program to take a while to run is that
+//   testPerformance() uses multiples ways each updating NUM_TRANSACTIONS_IN_PERF_TEST records. For
+//   a quick run, it is recommended to temporarily reduce NUM_TRANSACTIONS_IN_PERF_TEST.
+//
+public class Donghui {
```
thiagotnunes, line 38: We will remove this file before merging

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/AutocommitDmlModeTest.java
```diff
@@ -56,6 +59,8 @@ private ConnectionImpl createConnection(ConnectionOptions options) {
     DdlClient ddlClient = mock(DdlClient.class);
     TransactionRunner txRunner = mock(TransactionRunner.class);
     when(dbClient.readWriteTransaction()).thenReturn(txRunner);
```
thiagotnunes, line 61: nit: could we remove this line?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/AutocommitDmlModeTest.java
```diff
@@ -84,7 +89,12 @@ public void testAutocommitDmlModeTransactional() {
       assertThat(connection.getAutocommitDmlMode(), is(AutocommitDmlMode.TRANSACTIONAL));
 
       connection.execute(Statement.of(UPDATE));
-      verify(txContext).executeUpdate(Statement.of(UPDATE));
+      if (connection.isAutocommit()) {
+        verify(txContext)
+            .executeUpdate(Matchers.eq(Statement.of(UPDATE)), any(Options.UpdateOption.class));
+      } else {
+        verify(txContext).executeUpdate(Statement.of(UPDATE));
```
thiagotnunes, line 96: nit: could we remove this?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
```diff
@@ -311,43 +310,41 @@ public static ConnectionImpl createConnection(final ConnectionOptions options) {
...
+            return commitResponse;
+          }
+
+          @Override
+          public TransactionRunner allowNestedTransaction() {
+            return this;
+          }
+        };
+
+    when(dbClient.readWriteTransaction()).thenAnswer(invocation -> transactionRunner);
```
thiagotnunes, line 345: nit: could we remove this one?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/SingleUseTransactionTest.java
```diff
@@ -387,56 +389,54 @@ private SingleUseTransaction createSubject(
...
+            return commitResponse;
+          }
+
+          @Override
+          public TransactionRunner allowNestedTransaction() {
+            return this;
+          }
+        };
+
+    when(dbClient.readWriteTransaction()).thenAnswer(invocation -> txnRunner);
```
thiagotnunes, line 437: nit: could we remove this one?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITInlineCommitDatabaseClientTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+/** Integration tests for InlineCommit used in DatabaseClient. */
+@Category(ParallelIntegrationTest.class)
+@RunWith(JUnit4.class)
+public final class ITInlineCommitDatabaseClientTest {
+  @ClassRule public static IntegrationTestEnv env = new IntegrationTestEnv();
+  private static DatabaseClient client;
+
+  /** Id prefix per test case. */
+  private static int id;
```
thiagotnunes, line 48: Could we revisit this, because there might be a race condition where two tests get the same `id`?

thiagotnunes (commented):
LGTM overall just a few nits

Before merging we need to remove the temporary files and the generated protos.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SingleDmlTransaction.java
dzhang-bap: done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SingleDmlTransaction.java
dzhang-bap: removed

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SingleDmlTransaction.java
dzhang-bap: done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
dzhang-bap: removed

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/AutocommitDmlModeTest.java
dzhang-bap: yes. removed

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/AutocommitDmlModeTest.java
dzhang-bap: yes, removed

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
dzhang-bap: removed

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/SingleUseTransactionTest.java
dzhang-bap: actually some tests still need it

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITInlineCommitDatabaseClientTest.java
dzhang-bap: I made it thread safe by introducing a static synchronized function that increments a private static counter.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SingleDmlTransaction.java
```diff
@@ -0,0 +1,73 @@
...
+import com.google.cloud.spanner.Options.UpdateOption;
+import com.google.cloud.spanner.SessionPool.PooledSessionFuture;
+import com.google.common.annotations.VisibleForTesting;
+import io.opencensus.common.Scope;
+import io.opencensus.trace.Span;
+import io.opencensus.trace.Tracer;
+import io.opencensus.trace.Tracing;
+import java.util.Arrays;
+
+public class SingleDmlTransaction {
```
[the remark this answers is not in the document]
dzhang-bap: added

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SingleDmlTransaction.java
```diff
@@ -0,0 +1,73 @@
...
+    } catch (RuntimeException e) {
+      TraceUtil.endSpanWithFailure(span, e);
+      throw e;
+    } finally {
+      span.end(TraceUtil.END_SPAN_OPTIONS);
+    }
+  }
+
+  @VisibleForTesting
+  public SingleDmlTransaction(TransactionRunner runner) {
```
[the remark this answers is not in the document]
dzhang-bap: I've made the constructor of SingleDmlTransaction (that takes runner as argument) visible only within the package. In order to call this version of the constructor from tests in other packages, I've added a SingleDmlTransactionTest class that bridged the gap.

I've also removed the other public method getRunner(). It was needed for one purpose: for SingleUseTransaction to be able to retrieve a commit response. To fill in the need, I've introduced a public method SingleDmlTransaction#getCommitResponse().

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java
```diff
@@ -346,6 +346,32 @@ CommitResponse writeAtLeastOnceWithOptions(
...
 
+  /**
+   * Returns a transaction for executing a single DML statement with retries. While a single DML
+   * statement can also be executed using {@link #readWriteTransaction()}, using
+   * SingleDmlTransaction tends to have better performance.
+   *
+   * <p>Example of using SingleDmlTransaction.
+   *
+   * <pre> <code>
+   * long rowCount = dbClient.singleDmlTransaction().execute(
```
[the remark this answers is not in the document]
dzhang-bap: nice catch!

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SingleDmlTransaction.java
```diff
@@ -0,0 +1,73 @@
...
+
+  private synchronized void ensureNotUsed() {
+    if (used) {
+      throw SpannerExceptionFactory.newSpannerException(
+          ErrorCode.FAILED_PRECONDITION, "A SingleDmlTransaction can only be used once.");
+    }
+    used = true;
+  }
+
+  public long executeUpdate(Statement statement, UpdateOption... options) {
```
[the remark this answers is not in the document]
dzhang-bap: done

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
```diff
@@ -311,43 +310,41 @@ public static ConnectionImpl createConnection(final ConnectionOptions options) {
...
+    when(dbClient.singleDmlTransaction())
+        .thenAnswer(invocation -> new SingleDmlTransaction(transactionRunner));
```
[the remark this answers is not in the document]
dzhang-bap: changed to call SingleDmlTransactionTest#newSingleDmlTransaction() instead

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITInlineCommitDatabaseClientTest.java
```diff
@@ -0,0 +1,143 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.spanner.it;
+
+import static org.junit.Assert.*;
```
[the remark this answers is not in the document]
dzhang-bap: done. Thanks!

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SingleDmlTransaction.java
```diff
@@ -0,0 +1,95 @@
...
+    ensureNotUsed();
+    UpdateOption[] newOptions = Arrays.copyOf(options, options.length + 1);
+    newOptions[options.length] = Options.INLINE_COMMIT_OPTION;
+    return runner.run(transaction -> transaction.executeUpdate(statement, newOptions));
+  }
+
+  /**
+   * Returns the {@link CommitResponse} of this transaction. It may only be called after a
+   * successfully call of executeUpdate.
```
olavloite, line 90: nit: 
```suggestion
   * successful call of executeUpdate.
```

olavloite (approved):
Thanks for the changes, this looks good to me.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SingleDmlTransaction.java
```diff
@@ -0,0 +1,95 @@
...
+ * <ul>
+ *   <li>When {@link Connection} is in autocommit mode, an execute() or executeUpdate() of a DML
+ *       statement will use {@link SingleDmlTransaction} under the hood.
+ *   <li>Application code may call {@link DatabaseClient#singleDmlTransaction} to get a {@link
+ *       SingleDmlTransaction} object, and call its executeUpdate() to execute a DML statement and
+ *       inline-commit the transaction.
+ * </ul>
+ */
+public class SingleDmlTransaction {
+  private static final String READ_WRITE_TRANSACTION = "CloudSpanner.ReadWriteTransaction";
```
thiagotnunes, line 44: Wondering if we could add an attribute to the trace to specify that this is from autocommit. WDYT?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
```diff
@@ -311,43 +310,41 @@ public static ConnectionImpl createConnection(final ConnectionOptions options) {
...
+            return commitResponse;
+          }
+
+          @Override
+          public TransactionRunner allowNestedTransaction() {
+            return this;
+          }
+        };
+
+    when(dbClient.readWriteTransaction()).thenAnswer(invocation -> txnRunner);
```
thiagotnunes, line 345: nit: do we still need this mock?

thiagotnunes:
In order to fix the clirr validations, add the following in the end of [this file](https://github.com/googleapis/java-spanner/blob/master/google-cloud-spanner/clirr-ignored-differences.xml):

```
<!-- Autocommit changes -->
<!-- These are not breaking changes, since we provide default interface implementation -->
<difference>
  <differenceType>7012</differenceType>
  <className>com/google/cloud/spanner/DatabaseClient</className>
  <method>com.google.cloud.spanner.SingleDmlTransaction singleDmlTransaction(com.google.cloud.spanner.Options$TransactionOption[])</method>
</difference>
```

(You added a default implementation, but clirr does not support this Java8 feature, that is why it is reporting as a breaking change).

In terms of unit / emulator failures, it seems that the following scenario is breaking: https://github.com/googleapis/java-spanner/blob/master/google-cloud-spanner/src/test/resources/com/google/cloud/spanner/connection/ConnectionImplGeneratedSqlScriptTest.sql#L2749-L2755

PR #1343, opened 2021-08-11
test: use unique name for threads in test

olavloite:
The tests that check that a Spanner instance closes all threads that it is
using was counting the total number of threads with a fixed thread name. That
also meant that if any other test would fail to (timely) close its own threads,
the thread tests could fail as a result. This is now mitigated by using unique
names for the threads for the tests, so the test can check only for its own
threads instead of all Spanner threads. The additional option that is added to
SpannerOptions is package-private, which means that the tests also need to
move to the same package. The thread tests have therefore been moved to a
separate file in the normal Spanner package.


thiagotnunes (approved):
Thanks for fixing this!

PR #1345, opened 2021-08-12
fix: set projectId to a default value in emulator case

dmivankov:
Spanner client wants to have a projectId option set, by default it also tries to pick it up from [lots of places](https://github.com/googleapis/java-core/blob/375983090b3700b3fb6a1953626db7efca49cc51/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java#L372)
For emulator case none of those may be present as it could be that neither gcloud nor default application credentials are configured, but emulator is accessible nevertheless.

Set default project id in SpannerOptions, value doesn't matter so much as emulator would be happy with any value it seems.
Note that alternatively it could be improved in base ServiceOptions, but not sure about assumptions there.

failure example in case where emulatorHost is set without projectId
```
  java.lang.IllegalArgumentException: A project ID is required for this service but could not be determined from the builder or the environment.  Please set a project ID using the builder.
  at com.google.common.base.Preconditions.checkArgument(Preconditions.java:142)
  at com.google.cloud.ServiceOptions.<init>(ServiceOptions.java:304)
  at com.google.cloud.spanner.SpannerOptions.<init>(SpannerOptions.java:533)
  at com.google.cloud.spanner.SpannerOptions.<init>(SpannerOptions.java:77)
  at com.google.cloud.spanner.SpannerOptions$Builder.build(SpannerOptions.java:1029)
  ```

Note that in non-emulator case one usually doesn't need to set projectId as it could be extracted from credentials

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
@@ -1137,7 +1137,7 @@ public class SpannerOptions extends ServiceOptions<Spanner, SpannerOptions> {
         // Default project id resolution might fail, but the emulator accepts any
         // project id, so we can safely use a dummy id.
         if (this.getProjectId() == null) {
-          this.setProjetcId("emulator-project-id");
+          this.setProjectId("emulator-project-id");
         }
       }
       return new SpannerOptions(this);
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
@@ -1137,7 +1137,7 @@ public class SpannerOptions extends ServiceOptions<Spanner, SpannerOptions> {
         // Default project id resolution might fail, but the emulator accepts any
         // project id, so we can safely use a dummy id.
         if (this.getProjectId() == null) {
-          this.setProjetcId("your-project-id");
+          this.setProjetcId("emulator-project-id");
         }
       }
       return new SpannerOptions(this);
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
@@ -1137,8 +1137,8 @@ public class SpannerOptions extends ServiceOptions<Spanner, SpannerOptions> {
         // Default project id resolution might fail, but in emulator case
         // any project id would work, one from credentials or filesystem isn't
         // strictly needed
-        if (super.getProjectId() == null) {
-          super.setProjetcId("your-project-id");
+        if (this.getProjectId() == null) {
+          this.setProjetcId("your-project-id");
         }
       }
       return new SpannerOptions(this);
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
@@ -1134,6 +1134,12 @@ public class SpannerOptions extends ServiceOptions<Spanner, SpannerOptions> {
         this.setChannelConfigurator(ManagedChannelBuilder::usePlaintext);
         // As we are using plain text, we should never send any credentials.
         this.setCredentials(NoCredentials.getInstance());
+        // Default project id resolution might fail, but in emulator case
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java [hunk 1]
```diff
@@ -1134,6 +1134,12 @@ public SpannerOptions build() {
         this.setChannelConfigurator(ManagedChannelBuilder::usePlaintext);
         // As we are using plain text, we should never send any credentials.
         this.setCredentials(NoCredentials.getInstance());
+        // Default project id resolution might fail, but in emulator case
+        // any project id would work, one from credentials or filesystem isn't
+        // strictly needed
+        if (super.getProjectId() == null) {
+          super.setProjetcId("your-project-id");
```
dmivankov, line 1141:
```suggestion
        if (this.getProjectId() == null) {
          this.setProjetcId("your-project-id");
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1134,6 +1134,12 @@ public SpannerOptions build() {
         this.setChannelConfigurator(ManagedChannelBuilder::usePlaintext);
         // As we are using plain text, we should never send any credentials.
         this.setCredentials(NoCredentials.getInstance());
+        // Default project id resolution might fail, but in emulator case
+        // any project id would work, one from credentials or filesystem isn't
+        // strictly needed
```
olavloite, line 1139:
```suggestion
        // Default project id resolution might fail, but the emulator accepts any
        // project id, so we can safely use a dummy id.
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1134,6 +1134,12 @@ public SpannerOptions build() {
...
+        if (this.getProjectId() == null) {
+          this.setProjetcId("your-project-id");
```
olavloite, line 1141:
```suggestion
          this.setProjetcId("emulator-project-id");
```

olavloite (commented):
Hi @dmivankov Thanks for your contribution. I think this is a good idea. I've left a couple of textual suggestions.

Also, would you be able to add a test that verifies that it will actually use this, and that it does not fail before it gets this far? If not, then that's OK as well. I can take a look at a later moment. It might be quite tricky as such a test would depend on a lot of different environment variables.

@thiagotnunes Do you have any additional comments on this?

thiagotnunes (approved):
There is a compilation error, but LGTM.

After fixing that we should be good to go.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1134,6 +1134,11 @@ public SpannerOptions build() {
         this.setChannelConfigurator(ManagedChannelBuilder::usePlaintext);
         // As we are using plain text, we should never send any credentials.
         this.setCredentials(NoCredentials.getInstance());
+        // Default project id resolution might fail, but the emulator accepts any
+        // project id, so we can safely use a dummy id.
+        if (this.getProjectId() == null) {
+          this.setProjetcId("emulator-project-id");
```
dmivankov, line 1140:
```suggestion
          this.setProjectId("emulator-project-id");
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1134,6 +1134,11 @@ public SpannerOptions build() {
         this.setChannelConfigurator(ManagedChannelBuilder::usePlaintext);
         // As we are using plain text, we should never send any credentials.
         this.setCredentials(NoCredentials.getInstance());
+        // Default project id resolution might fail, but the emulator accepts any
+        // project id, so we can safely use a dummy id.
+        if (this.getProjectId() == null) {
```
dmivankov, line 1139: oops, may need more work as there's no such getter in ServiceOptions builder and the field is private https://github.com/googleapis/java-core/blob/375983090b3700b3fb6a1953626db7efca49cc51/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java#L127

olavloite:
@dmivankov I've created a new PR that uses a slightly different approach to achieve the same goal: #1363

thiagotnunes:
CLosed in favour of #1363

PR #1356, opened 2021-08-17
build: migrate to owlbot

Neenu1995:


google-cloud-spanner/clirr-ignored-differences.xml
```diff
@@ -1,662 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
thiagotnunes, line 1: Don't we need this anymore?

thiagotnunes (commented):
Since we removed the `synth.py` now, how do we regenerate proto changes manually? We need this in order to test features before they are released.

What we used to do is the following:

1. git clone https://github.com/googleapis/googleapis.git
2. Make proto changes in googleapis
3. export SYNTHTOOL_GOOGLEAPIS=<path to googleapis>
4. go to java-spanner repository and run python3 -m synthtool

google-cloud-spanner/clirr-ignored-differences.xml
Neenu1995: Once the version with the breaking changes is published, the clirr check won't detect the same changes as breaking again. So the clirr-ignored-differences file is no longer required to ignore the same breaking changes.

samples/snippets/src/main/java/com/example/spanner/AsyncQueryExample.java
```diff
@@ -71,18 +71,18 @@ public CallbackResponse cursorReady(AsyncResultSet resultSet) {
                   try {
                     while (true) {
                       switch (resultSet.tryNext()) {
-                        // OK: There is a row ready.
+                          // OK: There is a row ready.
```
thiagotnunes, line 74: @Neenu1995 it seems like owlbot is wrongly formatting these comments (whenever it re-runs). We don't have any special checkstyle configuration, so even though we can manually fix this, owlbot will break it whenever it re-runs.

Is there anything we can do in owlbot to fix this?

google-cloud-spanner/clirr-ignored-differences.xml
thiagotnunes: I see, but it is nice to have this file with some examples if we have to add any exceptions. Could we leave a few entries here? Or is this automatically deleted?

lesv (approved):

Neenu1995:
Closing this in favour of #1569

PR #1357, opened 2021-08-17
prototype: autocommit v2

dzhang-bap:
@thiagotnunes and @olavloite , here is a crude prototype for autocommit v2.

- It is not thoroughly tested. We will do that after the server-side of autocommit v2 is implemented.
- It has System.out.println(), and test code Donghui.java, which are not meant to be merged.
- Given that the server does not yet support autocommit v2, this prototype fake it, by adding an explicit BeginTransaction inside commit().
- The point of providing this pull request is to illustrate how it may be done and whether it may be overly complicated to maintain on the client side. It looks to me that having a separate collection of write sessions (maintained as a PriorityQueue sorted by timeToExpire) does not add too much complexity. What do you think?

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClientImpl.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClientImpl.java
@@ -49,8 +49,8 @@ class DatabaseClientImpl implements DatabaseClient {
   }
 
   @VisibleForTesting
-  PooledSessionFuture getSession() {
-    return pool.getSession();
+  PooledSessionFuture getSession(boolean wantRwTxnId) {
+    return pool.getSession(wantRwTxnId);
   }
 
   @Override
@@ -98,7 +98,7 @@ class DatabaseClientImpl implements DatabaseClient {
   public ReadContext singleUse() {
     Span span = tracer.spanBuilder(READ_ONLY_TRANSACTION).startSpan();
     try (Scope s = tracer.withSpan(span)) {
-      return getSession().singleUse();
+      return getSession(false).singleUse();
     } catch (RuntimeException e) {
       TraceUtil.endSpanWithFailure(span, e);
       throw e;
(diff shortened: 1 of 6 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClientImpl.java
```diff
@@ -218,7 +218,7 @@ public long executePartitionedUpdate(final Statement stmt, final UpdateOption...
   }
 
   private <T> T runWithSessionRetry(Function<Session, T> callable) {
-    PooledSessionFuture session = getSession();
+    PooledSessionFuture session = getSession(true);
```
olavloite, line 221: This should not be statically true, but an input parameter for the method. This method is not a helper for retrying read/write transactions, but for retrying `SessionNotFound` errors that could occur for both read-only and read/write requests. (This happens when the backend has garbage collected or otherwise deleted/invalidated sessions without the client knowing about it)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2006,7 +2034,19 @@ PooledSessionFuture getSession() throws SpannerException {
                 resourceNotFoundException.getMessage()),
             resourceNotFoundException);
       }
-      sess = sessions.poll();
+      moveExpiredWriteSessionsToSessions();
+      if (wantRwTxnId) {
+        sess = writeSessions.poll();
```
olavloite, line 2039: I think that we would need to add some more logic here to prevent possible starvation in the following scenario (and similar scenarios):
1. Assume that `maxSessions=10`
2. The client executes 10 read/write transactions in parallel (or enough in parallel to use all 10 different sessions for a r/w transaction). All these sessions have now received a transaction id from the backend, and will be added to the queue of writeSessions and not to the list of sessions.
3. The client wants to execute a read-only transaction, but there are no sessions in the normal session list, and `maxSessions` has been reached. It will therefore wait indefinitely.

Also, this list of writeSessions does not seem to be considered when calculating the number of sessions currently in the pool, is not considered for sessions that should be removed from the pool because they have expired, and is not included in the logic for closing the session pool and reporting any leaked sessions. (I know it's prototype so it's logical that not all of this is implemented, but my point is that more complexity than this will still be needed to be added to the session pool.)

PR #1358, opened 2021-08-18
test: unskip cmmr tests

thiagotnunes:
We have configured a multi-regional instance for CMMR sample tests, so we can unskip the tests now.


hengfengli (approved):
LGTM.

PR #1366, opened 2021-08-23
feat: configure initial sp version

thiagotnunes:
Configures CI for branch

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,79 +1,86 @@
-on:
+'on':
   push:
     branches:
-    - master
-  pull_request:
+      - 6.4.4-sp
+  pull_request: null
 name: ci
 jobs:
   units:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        java: [8, 11]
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -1,79 +1,86 @@
...
-        java-version: ${{matrix.java}}
-    - run: java -version
-    - run: .kokoro/dependencies.sh
+      - uses: actions/checkout@v2
+      - uses: actions/setup-java@v1
+        with:
+          java-version: ${{matrix.java}}
+      - run: java -version
+      - run: .kokoro/dependencies.sh
   linkage-monitor:
```
suztomo, line 49: "ci / linkage-monitor (pull_request)" does not work for the sp branch, because it always tries to get the latest versions of other libraries.

suztomo (approved):

.github/workflows/ci.yaml
thiagotnunes: Thanks, I will removed it in a subsequent PR

PR #1367, opened 2021-08-23
test: run it tests with LTS dependencies

thiagotnunes:


thiagotnunes:
@suztomo The integration tests work fine, but the dependency checks fail (upper bound requirements). I am not sure this is a problem

suztomo:
Upper bound check is ok to fail. The unit tests and integration tests are what I need to confirm. Great to hear they succeed.

PR #1375, opened 2021-08-24
chore: add samples for JSON

zoercai:


samples/snippets/src/main/java/com/example/spanner/AddJsonColumnSample.java
```diff
@@ -0,0 +1,57 @@
+/*
+ * Copyright 2020 Google Inc.
```
olavloite, line 2: nit:

```suggestion
 * Copyright 2021 Google LLC
```

samples/snippets/src/main/java/com/example/spanner/QueryWithJsonParameterSample.java
```diff
@@ -0,0 +1,63 @@
+/*
+ * Copyright 2020 Google Inc.
```
olavloite, line 2: nit:
```suggestion
 * Copyright 2021 Google LLC
```

samples/snippets/src/main/java/com/example/spanner/SpannerSample.java
```diff
@@ -488,44 +498,45 @@ static void update(DatabaseClient dbClient) {
   static void writeWithTransaction(DatabaseClient dbClient) {
     dbClient
         .readWriteTransaction()
-        .run(transaction -> {
```
olavloite, line 491: Here and elsewhere in this file: There seem to be a number of unrelated formatting changes. I think those should be removed from this PR, and if there is something wrong with the formatting here, we should fix that in a separate PR.

samples/snippets/src/main/java/com/example/spanner/SpannerSample.java
```diff
@@ -1258,6 +1276,7 @@ static void createTableWithDatatypes(DatabaseAdminClient dbAdminClient, Database
                     + "  OutdoorVenue    BOOL, "
                     + "  PopularityScore FLOAT64, "
                     + "  Revenue         NUMERIC, "
+                    + "  VenueDetails  JSON, "
```
olavloite, line 1279: nit: try to align with the line above so the data type starts at the same column

samples/snippets/src/main/java/com/example/spanner/SpannerSample.java
```diff
@@ -1502,13 +1523,35 @@ static void queryWithNumeric(DatabaseClient dbClient) {
   }
   // [END spanner_query_with_numeric_parameter]
 
+  // [START spanner_query_with_json_parameter]
+  static void queryWithJson(DatabaseClient dbClient) {
```
olavloite, line 1527: It seems that this sample is added twice: Once as a standalone sample (the preferred way) and once in this monolith sample file (deprecated). I think this one should be removed.

samples/snippets/src/test/java/com/example/spanner/SpannerSampleIT.java
```diff
@@ -92,22 +92,26 @@ public static void setUp() throws Exception {
     dbId = DatabaseId.of(options.getProjectId(), instanceId, databaseId);
     // Delete stale test databases that have been created earlier by this test, but not deleted.
     deleteStaleTestDatabases(instanceId, baseDbId);
-    key = String.format("projects/%s/locations/%s/keyRings/%s/cryptoKeys/%s",
-        options.getProjectId(), keyLocation, keyRing, keyName);
+    key =
```
olavloite, line 95: These formatting changes also seem unrelated to this change.

samples/snippets/src/test/java/com/example/spanner/SpannerSampleIT.java
```diff
@@ -302,49 +306,50 @@ public void testSample() throws Exception {
     assertThat(out).contains("19 Venue 19 1200100");
     assertThat(out).contains("42 Venue 42 390650.99");
 
+    out = runSample("querywithjson");
```
olavloite, line 309: This should probably also be removed, if we remove the duplicate sample in the monolith sample file.

olavloite (commented):
The samples look good to me, but I noticed that there are now no integration tests at all for these samples. Is that intentional? Normally we add tests for such standalone samples like this: https://github.com/googleapis/java-spanner/blob/master/samples/snippets/src/test/java/com/example/spanner/GetCommitStatsSampleIT.java

PR #1376, opened 2021-08-24
test: add Spanner JSON test suite

zoercai:


google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITJsonWriteReadTest.java
```diff
@@ -0,0 +1,156 @@
...
+              .build();
+      mutations.add(mutation);
+    }
+
+    databaseClient.write(mutations);
+    ResultSet resultSet =
+        databaseClient.singleUse().executeQuery(Statement.of("SELECT * FROM " + TABLE_NAME));
+    int count = 0;
+    while (resultSet.next()) {
+      count++;
```
olavloite, line 109: I think we should either (in order of preference):
1. Do a `resultSet.getJson("json")` here to verify that the value is also gettable in the client.
2. Or: Just re-write the query to a `SELECT COUNT(*) ...`

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITJsonWriteReadTest.java
```diff
@@ -0,0 +1,156 @@
...
+      try {
+        databaseClient.write(
+            Collections.singletonList(
+                Mutation.newInsertBuilder(TABLE_NAME)
+                    .set("Id")
+                    .to(100L)
+                    .set("json")
+                    .to(Value.json(jsonStr))
+                    .build()));
+        fail(resource + " should be rejected.");
```
olavloite, line 135: nit: use `Assert.assertThrows(..)` instead of try-catch.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITJsonWriteReadTest.java
```diff
@@ -0,0 +1,156 @@
...
+      } catch (Exception e) {
+        System.out.println("Failed to insert JSON: " + jsonStr + ". From: " + resource);
+        e.printStackTrace();
```
olavloite, line 138: Remove this and consider replacing it with an assertion that you actually got the error that you expected.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITJsonWriteReadTest.java
```diff
@@ -0,0 +1,156 @@
...
+  private List<String> getJsonFilePaths(String folder) {
+    String fixturesRoot = Resources.getResource(folder).getPath();
+    final Path fixturesRootPath = Paths.get(fixturesRoot);
+    try {
+      return Files.walk(fixturesRootPath)
+          .filter(Files::isRegularFile)
+          .map(path -> fixturesRootPath.relativize(path).toString())
+          .filter(x -> x.endsWith(".json"))
+          .collect(Collectors.toList());
+    } catch (IOException e) {
```
olavloite, line 152: nit: I don't think you need the try-catch here. The test methods already declare `throws IOException`, so we could just as well just let that bubble up here.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITQueryTest.java
```diff
@@ -64,6 +64,7 @@
 @Category(ParallelIntegrationTest.class)
 @RunWith(JUnit4.class)
 public class ITQueryTest {
+
```
olavloite, line 67: nit: remove unrelated change

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITJsonWriteReadTest.java
zoercai: I ended up going with option 2, because the JSON that gets returned is often not the same as what's inserted (e.g. `[123e45]` gets returned as `[1.23e+47]`), and it's difficult to automatically determine whether the returned value is equivalent to the inserted value without the same library that the backend uses to parse JSON. Since the tests are only there to confirm that the backend is able to accept the valid values, I don't think I'll bother spending the extra time manually adding the expected results for each file. Hopefully this is okay!

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITJsonWriteReadTest.java
```diff
@@ -0,0 +1,150 @@
...
+              StandardCharsets.UTF_8);
+
+      assertThrows(
+          SpannerException.class,
+          () ->
+              databaseClient.write(
+                  Collections.singletonList(
+                      Mutation.newInsertBuilder(TABLE_NAME)
+                          .set("Id")
+                          .to(100L)
```
olavloite, line 134: nit: This should use the same kind of increasing id (or other unique id) as in the other test. Otherwise, if two or more invalid JSON values are accepted by the backend, you will get a wrong error message for the second one, as it will fail with a PRIMARY KEY constraint violation instead (in other words: if 10 values all of a sudden are accepted, each test run will only detect the first one)

olavloite (approved):

thiagotnunes (commented):
LGTM

thiagotnunes (approved):

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITJsonWriteReadTest.java
```diff
@@ -0,0 +1,150 @@
...
+      mutations.add(mutation);
+    }
+    databaseClient.write(mutations);
+
+    try (ResultSet resultSet =
+        databaseClient
+            .singleUse()
+            .executeQuery(Statement.of("SELECT COUNT(*) FROM " + TABLE_NAME))) {
+      resultSet.next();
+      assertEquals(resultSet.getLong(0), resources.size());
```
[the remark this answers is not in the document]
zoercai: We had a discussion about this above: https://github.com/googleapis/java-spanner/pull/1376#discussion_r698955405

PR #1409, opened 2021-09-06
chore: configure backups test to run in nightly build

thiagotnunes:
The main change of this PR is to move the `ITBackupTest` into a new JUnit category to be executed only on the nightly builds. This is because the Backup test is slow and flaky when executing several in parallel (due to RESOURCE_EXHAUSTED issues). We have done the following here:

1. Created a new JUnit category: `SlowTest`
2. Created a new JUnit category: `SerialIntegrationTest`
3. Removed unused categories `FlakyTest`
4. Made `SlowTest`, `SerialIntegrationTest` and `ParallelIntegrationTest` extend the `IntegrationTest` category (in order to make configuration easier)
5. Created a new profile in the `pom.xml` to run only the `SlowTest` tests
6. Created a kokoro configuration file to execute the slow tests

Addresses https://github.com/googleapis/java-spanner/issues/1336


google-cloud-spanner/pom.xml
```diff
@@ -47,9 +47,10 @@
         </configuration>
         <executions>
           <execution>
+            <!-- Executes unit tests -->
             <id>default-test</id>
             <configuration>
-              <excludedGroups>com.google.cloud.spanner.TracerTest,com.google.cloud.spanner.IntegrationTest</excludedGroups>
+              <excludedGroups>com.google.cloud.spanner.categories.TracerTest,com.google.cloud.spanner.categories.IntegrationTest</excludedGroups>
```
olavloite, line 53: You probably need to change the package name for `TracerTest` further below as well (line 62)

google-cloud-spanner/pom.xml
```diff
@@ -365,6 +366,46 @@
...
+              <artifactId>maven-failsafe-plugin</artifactId>
+              <executions>
+                <execution>
+                  <id>default</id>
+                  <configuration>
+                    <groups>com.google.cloud.spanner.categories.SlowTest</groups>
+                  </configuration>
+                </execution>
+                <execution>
+                  <id>parallel-integration-test</id>
```
olavloite, line 396: What is the intention with this execution? The name seems to indicate that it will execute parallel integration tests, but if I understand the configuration correctly, it will just skip all tests. Could we maybe add a comment to this section to explain why it's here?

olavloite (approved):

google-cloud-spanner/pom.xml
thiagotnunes: Ah yes, thanks, fixed it.

google-cloud-spanner/pom.xml
thiagotnunes: This is because the default configuration (line 88-98) defines an execution for the integration-test goal with such id. Thus, we need to override it in the slow tests profile to NOOP, otherwise the tests with the category `ParallelIntegrationTest` will be executed (as per the default one). I added a comment in this section.

PR #1440, opened 2021-09-23
chore: increase integration tests timeout

thiagotnunes:
Integration tests are timing out at the 54 minute mark. It correlates well with the current forkedProcess timeout set for failsafe plugin which is 3000 seconds (50 minutes).

Here we increase it to 6000 seconds to see if the issue is resolved.


olavloite (approved):
Do we have any idea why it has started to exceed the previous timeout? I don't think there were any new integration tests that were added recently that would take a long time. Or am I missing something?

thiagotnunes:
Not sure, it seems that some operations were a bit slower than usual in our development instance, especially instance admin operations. We will keep an eye out.

PR #1458, opened 2021-09-24
chore: fix failsafe plugin hanging

thiagotnunes:
Trying out solution from https://stackoverflow.com/questions/21661559/maven-failsafe-plugin-hangs-unexpectedly


thiagotnunes:
It seems that the failures are somehow related to parallel tests?
Running them all serially for now to unblock the build.

olavloite (approved):

PR #1486, opened 2021-10-05
fix: keep track of any BeginTransaction option for a Read

olavloite:
If a StreamingReadRequest that included a BeginTransaction option was retried as a result
of a transient error (UNAVAILABLE), the fact that the BeginTransaction option was included
would not be registered for the retried request. This could cause a transaction to fail if
the retried request returned an Aborted error, and that Aborted error was caught by the
application.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
@@ -790,7 +790,7 @@ abstract class AbstractReadContext
             SpannerRpc.StreamingCall call =
                 rpc.read(builder.build(), stream.consumer(), session.getOptions());
             call.request(prefetchChunks);
-            stream.setCall(call, selector != null && selector.hasBegin());
+            stream.setCall(call, /* withBeginTransaction = */ builder.getTransaction().hasBegin());
             return stream;
           }
         };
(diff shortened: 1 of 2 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
```diff
@@ -790,7 +790,7 @@ ResultSet readInternalWithOptions(
             SpannerRpc.StreamingCall call =
                 rpc.read(builder.build(), stream.consumer(), session.getOptions());
             call.request(prefetchChunks);
-            stream.setCall(call, selector != null && selector.hasBegin());
+            stream.setCall(call, /* withBeginTransaction = */ builder.getTransaction().hasBegin());
```
suztomo, line 793: Is this typo? Strange way to write an inline comment.

olavloite: No, it's not a typo. It's the closest we get to named parameters in Java. We sometimes use it for (boolean) arguments that otherwise are not obvious what they are for.

suztomo: Thank you for clarification

suztomo:
I see integration test failing.

olavloite:
> I see integration test failing.

Seems to be unrelated. I've created a [tracking issue](https://github.com/googleapis/java-spanner/issues/1487) and restarted them.

suztomo:
I see it failed again.

olavloite:
> I see it failed again.

I've increased the timeout in the specific test, and it now succeeds. The test is however completely unrelated to this change, so the test failure rather seems to be related to the backend needing more time than previously to delete a database.

suztomo (approved):

thiagotnunes (approved):

PR #1548, opened 2021-11-11
feat: support RPC priority for JDBC connections and statements

rahul2393:
[Issue](https://github.com/googleapis/java-spanner-jdbc/issues/656)
This PR will support setting RPC priority from connection URL and statements.

Fixes https://github.com/googleapis/java-spanner-jdbc/issues/656


google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
```diff
@@ -208,6 +208,7 @@ static UnitOfWorkType of(TransactionMode transactionMode) {
   private AutocommitDmlMode autocommitDmlMode = AutocommitDmlMode.TRANSACTIONAL;
   private TimestampBound readOnlyStaleness = TimestampBound.strong();
   private QueryOptions queryOptions = QueryOptions.getDefaultInstance();
+  private Options.RpcPriority rpcPriority = Options.RpcPriority.HIGH;
```
olavloite, line 211: Would it be an idea to use a different default value for this field, as there is currently no way to tell the difference between a connection that has been set to use `RpcPriority.HIGH` and one that uses the default value. This means that:
1. The connection must either always explicitly pass the priority as an option to the client library.
2. OR: Rely on the fact that HIGH is the default priority, which means that if that would ever change, this code would also need to change.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
```diff
@@ -1010,8 +1025,9 @@ public long executeUpdate(Statement update) {
       if (options == null || options.length == 0) {
         options = new QueryOption[] {Options.tag(statementTag)};
```
olavloite, line 1026: Why doesn't `rpcPriority` need to be added to this array here?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
```diff
@@ -1024,8 +1040,9 @@ public long executeUpdate(Statement update) {
       if (options == null || options.length == 0) {
         options = new UpdateOption[] {Options.tag(statementTag)};
```
olavloite, line 1041: Why doesn't `rpcPriority` need to be added to this array here?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
```diff
@@ -923,6 +938,12 @@ public boolean isAutoConfigEmulator() {
     return autoConfigEmulator;
   }
 
+  /** The initial rpcPriority value for connections created by this {@link ConnectionOptions} */
+  /** The {@link Options.RpcPriority} to use for the connection. */
```
olavloite, line 942: nit: Don't use two separate javadoc blocks for a single method, merge them into one.

google-cloud-spanner/src/main/resources/com/google/cloud/spanner/connection/ClientSideStatements.json
```diff
@@ -361,6 +369,23 @@
 	    "allowedValues": "'(([a-zA-Z][a-zA-Z0-9_\\-]{1,63})|())'",
 	    "converterName": "ClientSideStatementValueConverters$StringValueConverter"
 	  }
+	},
+	{
+		"name": "SET RPC_PRIORITY = 'HIGH'|'MEDIUM'|'LOW'",
```
olavloite, line 374: As you are adding a new statement to the client side statements set, you also need to run the following command in the root of the google-cloud-spanner project:

`mvn -P generate-test-sql-scripts compile`

See also https://github.com/googleapis/java-spanner/blob/b6db9b1524760f4fd158bbf5b2689211d2ed55e7/google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ClientSideStatementsTest.java#L39

This will generate test statements for the new statement(s) in the `ClientSideStatementsTest.sql` file. I had a quick look locally, and:
1. It seems that we forgot to do this also for the `statistics_package` statement that was added recently. That means that you will be getting some additional unrelated changes in the test script.
2. The generated tests will fail, as the current implementation does not allow the priority value to be specified in lower case (or anything other than all upper case).

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
rahul2393: I think we want #2 here since we want default to be HIGH https://github.com/googleapis/java-spanner/pull/676/files#diff-992d480774d874d97423da09934b0395ef460130558e2fd6f7b1461ba32535c0R29
cc: @olavloite

rahul2393:
Thanks for quick review @olavloite
Please review again, pushed all the changes as per suggestions

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/Connection.java
```diff
@@ -532,6 +533,26 @@ default String getOptimizerStatisticsPackage() {
   /** @return true if this connection requests commit statistics from Cloud Spanner */
   boolean isReturnCommitStats();
 
+  /**
+   * Sets the priority to use for RPCs executed by this connection..
+   *
+   * @param rpcPriority The RPC priority to use. Must be a string from (HIGH/MEDIUM/LOW). The empty
```
olavloite, line 539: nit: This comment needs to be updated, as the argument is no longer a string.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
```diff
@@ -208,6 +209,7 @@ static UnitOfWorkType of(TransactionMode transactionMode) {
   private AutocommitDmlMode autocommitDmlMode = AutocommitDmlMode.TRANSACTIONAL;
   private TimestampBound readOnlyStaleness = TimestampBound.strong();
   private QueryOptions queryOptions = QueryOptions.getDefaultInstance();
+  private RpcPriority rpcPriority = RpcPriority.HIGH;
```
olavloite, line 212: What I meant regarding this was that it might be good to consider using `null` as the default value (or even change the type of the internal value to the proto enum which also includes the `UNSPECIFIED` value), as the current implementation means that:
1. The client determines what the default value is. However, if no value is sent in the request, the backend can determine what the default value should be.
2. The client always sends a priority value, which makes it impossible to distinguish between clients deliberately setting HIGH, and clients that did not set a value at all.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
```diff
@@ -1008,10 +1024,11 @@ public long executeUpdate(Statement update) {
     if (this.statementTag != null) {
```
olavloite, line 1024: This `if` statement causes the priority value to only be set if there is also a statement tag. If the statement tag is `null` and the client has set a custom priority, the connection will still just use the default priority.

I also think that if you want to reuse this method for also setting the priority, that you should rename the method to better reflect what it is doing.

One thing that should also be considered here is that this method (and the one for update statements) are only invoked for queries and updates. They are not called for `Commit` invocations (and also not for `Rollback`, but that is probably less of an issue).

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
```diff
@@ -158,6 +159,7 @@ public String getDefaultValue() {
   private static final String DEFAULT_USER_AGENT = null;
   private static final String DEFAULT_OPTIMIZER_VERSION = "";
   private static final String DEFAULT_OPTIMIZER_STATISTICS_PACKAGE = "";
+  private static final RpcPriority DEFAULT_RPC_PRIORITY = RpcPriority.HIGH;
```
olavloite, line 162: Also here I would say that it might be better to use `null` (or some other 'undefined' value) as the default, and so prevent a priority from being included in every RPC invocation.

olavloite (commented):
I think that in general this PR could use some more testing to verify that the priority that is set by an application is actually sent to Spanner. As far as I can see, the current implementation will only include a priority in a request if there is also a statement tag.

Take a look at the `com.google.cloud.spanner.connection.TaggingTest` for an example for how you could set up such a test. The difficult thing with something like priority (and also tagging) is Spanner itself does not return anything that indicates whether the priority was actually included in the request or not, so you cannot really use a normal system test for that. Instead, you can use a mock server and then inspect the requests that the mock server receives.

rahul2393:
Added the test to assert RPCPriority and pushed the requested changes, please help review again @olavloite

olavloite (commented):
Thanks for the changes! This looks mostly good to me. I have a couple of questions regarding being able to set the priority back to null after it has been set to something else.

The rest of the comments is mainly small questions / nitpicks on some of the tests.

rahul2393:
@olavloite Added the requested changes, please help review.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
```diff
@@ -30,6 +30,7 @@
    * set a lower priority for a specific RPC invocation.
    */
   public enum RpcPriority {
+    UNSPECIFIED(Priority.PRIORITY_UNSPECIFIED),
```
olavloite, line 33: This worries me a little. Adding an `enum` constant __at the beginning__ of an existing enum, means that the existing enum constants will get new ordinal values (that is; LOW was 0 before this change and will be 1 after this change), which could break existing client application code. Although I don't think it is very probable that anyone is using/checking the ordinal values of these enum constants, it is the kind of change that we should not do unless there is a very good reason to do so.

Also; do we really need this value? It seems that it is only used in the `RpcPriorityConverter`. I think that it might be better to let the `RpcPriorityConverter` return a `com.google.spanner.v1.Priority` instance (the proto enum) instead of `RpcPriority`, and then translate from `com.google.spanner.v1.Priority` to `RpcPriority` in the `ConnectionStatementExecutorImpl#statementSetRPCPriority` method (instead of converting `RpcPriority.UNSPECIFIED` to `null`).

Another reason not to add a value to this enum is that it is also used by the normal client library, so this would mean that it will be possible for a user of the normal client library to also pass in `RpcPriority.UNSPECIFIED` as the priority of a call, and I'm not sure the client library will handle that correctly.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java
```diff
@@ -345,6 +345,9 @@ public StatementResult statementAbortBatch() {
 
   @Override
   public StatementResult statementSetRPCPriority(RpcPriority rpcPriority) {
+    if (rpcPriority == RpcPriority.UNSPECIFIED) {
```
olavloite, line 348: (see also above)
I think it would be better to let this method accept a `com.google.spanner.v1.Priority` instead of `RpcPriority`, as that would remove the need to add `UNSPECIFIED` to the `RpcPriority` enum.

google-cloud-spanner/src/main/resources/com/google/cloud/spanner/connection/ClientSideStatements.json
```diff
@@ -383,7 +383,7 @@
 		"setStatement": {
 			"propertyName": "RPC_PRIORITY",
 			"separator": "=",
-			"allowedValues": "'(HIGH|MEDIUM|LOW)'",
+			"allowedValues": "'(HIGH|MEDIUM|LOW|NULL)'",
```
olavloite, line 386: Adding a new value to `allowedValues` means that you also need to regenerate the tests so it will also include tests with the new value:

`mvn -P generate-test-sql-scripts compile`

rahul2393:
@olavloite Please help review again

olavloite (approved):
LGTM, thanks for your patience with me on this one.

rahul2393:
Thanks @olavloite for all the help and review, it was great learning for me, this will help me in future contributions

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
```diff
@@ -208,6 +209,7 @@ static UnitOfWorkType of(TransactionMode transactionMode) {
   private AutocommitDmlMode autocommitDmlMode = AutocommitDmlMode.TRANSACTIONAL;
   private TimestampBound readOnlyStaleness = TimestampBound.strong();
   private QueryOptions queryOptions = QueryOptions.getDefaultInstance();
+  private RpcPriority rpcPriority = null;
```
thiagotnunes, line 212: Should it be `private RpcPriority rpcPriority = Priority.PRIORITY_UNSPECIFIED`?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java
```diff
@@ -338,4 +343,24 @@ public StatementResult statementAbortBatch() {
     getConnection().abortBatch();
     return noResult(ABORT_BATCH);
   }
+
+  @Override
+  public StatementResult statementSetRPCPriority(Priority priority) {
+    RpcPriority value = null;
+    if (priority != null && priority != Priority.PRIORITY_UNSPECIFIED) {
+      value = RpcPriority.values()[priority.ordinal() - 1];
```
thiagotnunes, line 351: This worries me a bit. If the order changes / a new value is added in the priority that changes the ordinal this would break without us noticing it.

Could you use a `Map<Priority, RpcPriority>` instead?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionTest.java
```diff
@@ -281,4 +288,180 @@ public void testMaxSessions()
...
+      return super.getBaseUrl() + ";rpcPriority=MEDIUM";
+    }
+
+    @Test
+    public void testQuery_RPCPriority() {
+      try (Connection connection = createConnection()) {
+        for (boolean autocommit : new boolean[] {true, false}) {
+          connection.setAutocommit(autocommit);
+          try (ResultSet rs = connection.executeQuery(SELECT_COUNT_STATEMENT)) {}
+          assertEquals(1, mockSpanner.countRequestsOfType(ExecuteSqlRequest.class));
```
thiagotnunes, line 309: nit: newline before the assert

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/RpcPriorityConverterTest.java
```diff
@@ -0,0 +1,50 @@
...
+
+@RunWith(JUnit4.class)
+public class RpcPriorityConverterTest {
+
+  @Test
+  public void testConvert() throws CompileException {
+    String allowedValues = "'(HIGH|MEDIUM|LOW|NULL)'";
+    RpcPriorityConverter converter =
+        new ClientSideStatementValueConverters.RpcPriorityConverter(allowedValues);
+    assertEquals(converter.convert("high"), Priority.PRIORITY_HIGH);
```
thiagotnunes, line 37: The parameters are inverted here. It should be `assertEquals(expected, actual)`

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/RpcPriorityConverterTest.java
```diff
@@ -0,0 +1,50 @@
...
+    assertEquals(converter.convert("HIGH"), Priority.PRIORITY_HIGH);
+    assertEquals(converter.convert("High"), Priority.PRIORITY_HIGH);
+
+    assertEquals(converter.convert("medium"), Priority.PRIORITY_MEDIUM);
+    assertEquals(converter.convert("Low"), Priority.PRIORITY_LOW);
```
thiagotnunes, line 42: nit: could we move low below the medium tests

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
rahul2393: Cannot assign value of type Priority to RpcPriority, although I though of adding UNSPECIFIED in RpcPriority enum but after @olavloite suggestions I decided to not add the value https://github.com/googleapis/java-spanner/pull/1548#discussion_r748503248

thiagotnunes (approved):

PR #1554, opened 2021-11-15
feat: NaNs in Mutations are equal and have the same hashcode

thiagotnunes:


rahul2393 (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Value.java
```diff
@@ -1028,7 +1028,8 @@ void valueToString(StringBuilder b) {
 
     @Override
     boolean valueEquals(Value v) {
-      return ((Float64Impl) v).value == value;
+      final Float64Impl float64Value = (Float64Impl) v;
+      return Double.isNaN(value) && Double.isNaN(float64Value.value) || float64Value.value == value;
```
olavloite, line 1032: Why do we want to deviate from the IEEE 754 standard here? According to that (and also according to the Cloud Spanner backend) `NaN == NaN` should return false.

The following SQL statement for example returns `false`:

`select (IEEE_DIVIDE(0.0, 0.0) = IEEE_DIVIDE(0.0, 0.0)) as b`

thiagotnunes: That is a good point and I considered this.

The issue that we have is that on import / export in Apache Beam (see [here](https://github.com/apache/beam/blob/master/sdks/java/core/src/main/java/org/apache/beam/sdk/util/MutationDetectors.java#L37-L51)) we compare mutations before committing. Since we allow for `PGNumerics` that allow for such a value, import / export is failing and that is why we are loosening the restriction here.

If you feel strong that we should not make changes here, we could try and explore a change in Apache Beam itself, but I guess we would just be re-implementing equality checks over there.

olavloite: This is maybe a slightly hypothetical problem, but I don't think we should implement this in `Value`. However, it might make sense to implement it in `Mutation` (if doable). That is:

If someone executes the following query:

```sql
SELECT FloatValue1, FloatValue2
FROM SomeTable
WHERE FloatValue1 != FloatValue2
```

Then `resultSet.getValue("FloatValue1).equals(resultSet.getValue("FloatValue2"))` should return false for all rows.

On the other hand, for `Mutation`s the difference between two `NaN` values is void, as they will both set the value of a row to the same. So:

```java
      Mutation mutation1 = Mutation.newInsertBuilder("SomeTable")
          .set("FloatValue").to(Value.float64(Double.NaN))
          .build();
      Mutation mutation2 = Mutation.newInsertBuilder("SomeTable")
          .set("FloatValue").to(Value.float64(Double.NaN))
          .build();
      // The following could (should?) be true as both mutations will have the same effect.
      assertTrue(mutation1.equals(mutation2));
```

Would that solve the problem in Beam?

libingye816: Yes, this would solve the problem for us. In addition, is it possible to see Double.NaN in PgNumeric Value?

olavloite: > In addition, is it possible to see Double.NaN in PgNumeric Value?

Yes, that is possible.

thiagotnunes: Changed the implementation to check for NaNs only in mutations

libingye816: > > In addition, is it possible to see Double.NaN in PgNumeric Value?
> 
> Yes, that is possible.

Hi @thiagotnunes , is this case considered? I see we only try to cast the type to float64 but I am not an expert on client libraries. Ignore me if I am wrong.

thiagotnunes: Hey Michael, since this fix is on the public repository we don't have pg numeric here yet. I will make the necessary changes in the other repository as well.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Mutation.java
```diff
@@ -358,6 +358,29 @@ public int hashCode() {
...
+      return true;
+    } else if (values == null || otherValues == null) {
+      return false;
+    } else if (values.size() != otherValues.size()) {
+      return false;
+    } else {
+      for (int i = 0; i < values.size(); i++) {
+        final Value value = values.get(i);
+        final Value otherValue = otherValues.get(i);
+        if (!value.equals(otherValue) && (!isNaN(value) || !isNaN(otherValue))) {
```
olavloite, line 372: nit: maybe for future reference we should add a short comment on why we are doing this

olavloite (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Value.java
libingye816: oh, yes. Sorry, you are absolutely right.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Mutation.java
thiagotnunes: Good point, added a comment here

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Value.java
thiagotnunes: No worries!

PR #1565, opened 2021-11-29
samples: add Cloud Spanner Change Stream java sample

haikuo-google:
It does the following:
1. Set up the Cloud Spanner client.
2. Create a database with a table that's watched by a change stream.
3. Insert rows into the table.
4. Execute change stream initial query to get change stream partition tokens.
5. Execute a change stream partition query to get data change records of the inserted rows.

### DIFF
--- a/samples/snippets/src/main/java/com/example/spanner/ChangeStreamSample.java
+++ b/samples/snippets/src/main/java/com/example/spanner/ChangeStreamSample.java
@@ -1,0 +1,626 @@
+/*
+ * Copyright 2021 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.example.spanner;
+
+// [START spanner_change_stream]
+import com.google.api.gax.longrunning.OperationFuture;
+import com.google.cloud.Timestamp;
+import com.google.cloud.spanner.Database;
+import com.google.cloud.spanner.DatabaseAdminClient;
+import com.google.cloud.spanner.DatabaseClient;
+import com.google.cloud.spanner.DatabaseId;
+import com.google.cloud.spanner.ErrorCode;
+import com.google.cloud.spanner.Mutation;
+import com.google.cloud.spanner.ResultSet;
+import com.google.cloud.spanner.Spanner;
+import com.google.cloud.spanner.SpannerExceptionFactory;
+import com.google.cloud.spanner.SpannerOptions;
+import com.google.cloud.spanner.Statement;
+import com.google.cloud.spanner.Struct;
+import com.google.common.base.Stopwatch;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Sets;
+import com.google.spanner.admin.database.v1.CreateDatabaseMetadata;
+import java.io.Serializable;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.UUID;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import javax.annotation.Nullable;
+
+/**
+ * Sample code for Cloud Spanner Change Stream. It demonstrates how to:
+ * 1. Set up the Cloud Spanner client.
+ * 2. Create a database with a table that's watched by a change stream.
+ * 3. Insert rows into the table.
+ * 4. Execute change stream initial query to get change stream partition tokens.
+ * 5. Execute a change stream partition query to get data change records of the inserted rows.
+ */
+public class ChangeStreamSample {
+  public static void main(String... args) {
+    if (args.length != 2 && args.length != 3) {
+      System.err.println("Usage: ChangeStreamSample "
+          + "<project_id> "
+          + "<instance_id> "
+          + "<database_id> (optional)");
+      return;
+    }
+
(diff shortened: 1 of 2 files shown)


samples/snippets/src/main/java/com/example/spanner/ChangeStreamSample.java
```diff
@@ -0,0 +1,626 @@
...
+ * 4. Execute change stream initial query to get change stream partition tokens.
+ * 5. Execute a change stream partition query to get data change records of the inserted rows.
+ */
+public class ChangeStreamSample {
+  public static void main(String... args) {
+    if (args.length != 2 && args.length != 3) {
+      System.err.println("Usage: ChangeStreamSample "
+          + "<project_id> "
+          + "<instance_id> "
+          + "<database_id> (optional)");
```
thiagotnunes, line 63: To simplify the sample I think that the database id should not be optional

samples/snippets/pom.xml
```diff
@@ -113,6 +113,12 @@
       <version>1.1.3</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.beam</groupId>
+      <artifactId>beam-sdks-java-core</artifactId>
```
thiagotnunes, line 118: We should not depend on beam sdks here, since this would create a dependency cycle.

samples/snippets/src/main/java/com/example/spanner/ChangeStreamSample.java
```diff
@@ -0,0 +1,626 @@
...
+        }
+      }
+      res.addAll(records);
+    }
+
+    return res;
+  }
+
+  // Models of change stream result format.
+  interface ChangeStreamRecord extends Serializable {}
```
thiagotnunes, line 239: Do we need these models? Or could we just use the raw structs? If we could use the result set raw structs I think we could simplify the code / remove the mapper.

samples/snippets/src/test/java/com/example/spanner/ChangeStreamSampleIT.java
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  @After
+  public void tearDown() {
+    System.setOut(stdOut);
+  }
+
+  @Test
+  public void testQuickstart() {
+    assertThat(projectId).isNotNull();
```
thiagotnunes, line 56: Please avoid using `assertThat`, we are trying to eliminate the usage in preference of junit matchers, such as `assertEquals`.

samples/snippets/src/main/java/com/example/spanner/ChangeStreamSample.java
haikuo-google: Should we even keep database_id as an argument then? If we keep it, then that means user will need to have a database with Singers and ChangeStreamAll, otherwise we can't insert and query change stream. I lean forward to just remove database_id, and we always create a temp database, what do you think?

samples/snippets/src/main/java/com/example/spanner/ChangeStreamSample.java
haikuo-google: It does make the code a bit long, but if we just print out the struct, user still have to figure out how to parse the struct and extract the records in their app code, and this is non-trivial. 

On a separate note, It seems that these models are pretty common code for whoever uses raw change stream API, should we package these models somewhere so that user (and connector) can add them as dependencies if they want to use raw API?

samples/snippets/pom.xml
haikuo-google: Good point, didn't realize that

samples/snippets/src/test/java/com/example/spanner/ChangeStreamSampleIT.java
haikuo-google: Removed.

haikuo-google:
Sorry I forgot to add the changes that address comments as a separate commit, will keep that in mind next time.

samples/snippets/src/main/java/com/example/spanner/ChangeStreamSample.java
thiagotnunes: Sorry, I was thinking of keeping the database id, but creating the tables during the sample initialization (before creating the change streams).

samples/snippets/src/main/java/com/example/spanner/ChangeStreamSample.java
thiagotnunes: > It does make the code a bit long, but if we just print out the struct, user still have to figure out how to parse the struct and extract the records in their app code, and this is non-trivial.

Agreed, but the issue I see with this is that the samples need to be relatively simple so that we can add them to our documentation (like [here](https://cloud.google.com/spanner/docs/getting-started/java)). If we have these many lines of code it might be hard for us to add those there. I was thinking that we could provide a sample which only extracts the parts of the records from a struct that we care about reading for the sample. We can then refer the user to the Struct full specification which should be published somewhere else. WDYT?

> On a separate note, It seems that these models are pretty common code for whoever uses raw change stream API, should we package these models somewhere so that user (and connector) can add them as dependencies if they want to use raw API?

They are common, but I don't think they should live in the client libraries as they can evolve separately (client libraries should be a lower level layer them the models). For instance, there might be customers who are only interested in parsing some of the fields. If we were to have these in a library, IMHO they should leave in a higher abstraction library that uses the client libraries. To be honest, I don't think we should care about this at this moment. Open to discuss it further if you have a stronger opinion otherwise.

samples/snippets/src/test/java/com/example/spanner/ChangeStreamSampleIT.java
```diff
@@ -0,0 +1,72 @@
...
+    System.setOut(stdOut);
+  }
+
+  @Test
+  public void testQuickstart() {
+    assertNotNull(projectId);
+    assertNotNull(instanceId);
+    ChangeStreamSample.main(projectId, instanceId);
+    String got = bout.toString();
+    Assert.assertThat(got, containsString("Received a ChildPartitionsRecord"));
```
thiagotnunes, line 62: Sorry, I meant we should use the default matchers like `assertEquals`, `assertTrue`, ... instead of `assertThat`.

haikuo-google:
Sorry - I have to close this one since the same was submitted to java-docs-sample repo.

PR #1596, opened 2021-12-16
fix: re-adds test-jar to bom definition

thiagotnunes:
The test jar was wrongly removed during the owlbot migration.

`google-cloud-spanner-bom` has been added to an exclude list for owlbot. This means owlbot will not add new dependencies to the bom when a new version is introduced. Renovate bot will still update the versions and no other owlbot feature changes.


thiagotnunes:
It seems OwlBot is removing the dependency

Neenu1995 (approved):

PR #1600, opened 2021-12-17
deps(3.1.x): update OpenCensus API to 0.30.0

thiagotnunes:


olavloite:
Created tracking issue for the integration test failure here: https://github.com/googleapis/java-spanner/issues/1606

Seems to be caused by a test that will fail if two instances of the same test are running at the same time using the same project id.

olavloite (approved):

thiagotnunes:
Fixes #1606

Issue #1606, opened 2021-12-17
ITInstanceAdminTest.listInstances seems to fail if two instances of the test are running simultaneously


olavloite:
java.lang.IllegalArgumentException: expected one element but was: <Instance{name=projects/gcloud-devel/instances/spanner-testing, configName=projects/gcloud-devel/instanceConfigs/regional-us-central1, displayName=spanner-testing, nodeCount=2, state=READY, labels={}}, Instance{name=projects/gcloud-devel/instances/spanner-testing-east1, configName=projects/gcloud-devel/instanceConfigs/regional-us-east1, displayName=spanner-testing-east1, nodeCount=2, state=READY, labels={}}, Instance{name=projects/gcloud-devel/instances/spanner-testing-west1, configName=projects/gcloud-devel/instanceConfigs/regional-us-west1, displayName=spanner-testing-west1, nodeCount=2, state=READY, labels={}}>
	at com.google.common.collect.Iterators.getOnlyElement(Iterators.java:315)
	at com.google.cloud.spanner.it.ITInstanceAdminTest.listInstances(ITInstanceAdminTest.java:70)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
[...]

PR #1610, opened 2021-12-17
deps(3.3.x): update OpenCensus API to 0.30.0

thiagotnunes:


thiagotnunes:
Fixes #1606

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITInstanceAdminTest.java
```diff
@@ -69,9 +69,7 @@ public void listInstances() {
     Instance instance =
         Iterators.getOnlyElement(
             instanceClient
-                .listInstances(
-                    Options.filter(
-                        "name:instances/" + env.getTestHelper().getInstanceId().getInstance()))
+                .listInstances(Options.filter("name=" + env.getTestHelper().getInstanceId()))
```
olavloite, line 72: I think this should be `"name:" + env.getTestHelper().getInstanceId()`

See https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.admin.instance.v1#google.spanner.admin.instance.v1.ListInstancesRequest

thiagotnunes: The problem here is that `:` operator is a contains and we have 3 instances that start with `spanner-testing`. That is why the test is failing.

By using the `=` operator we do an exact match instead, having a single instance returned. More on operators here: https://cloud.google.com/sdk/gcloud/reference/topic/filters

PR #1673, opened 2022-02-08
feat: PostgreSQL dialect databases

olavloite:
Adds support for Cloud Spanner databases using the PostgreSQL dialect.


google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractStructReader.java
```diff
@@ -429,6 +437,18 @@ private void checkNonNullOfType(int columnIndex, Type expectedType, Object colum
...
   }
 
+  private void checkNonNullOfTypes(
+      int columnIndex, List<Type> expectedTypes, Object columnNameForError) {
+    Type actualType = getColumnType(columnIndex);
+    checkState(
+        expectedTypes.contains(actualType),
+        "Column %s is not of correct type: expected one of [%s] but was %s",
+        columnNameForError,
+        expectedTypes.stream().map(Type::toString).collect(Collectors.joining(", ")),
```
jin-jj, line 447: would this be executed everytime we call checkNonNullOfTypes? If so, seems too expensive to me.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/PostgreSQLStatementParser.java
```diff
@@ -0,0 +1,256 @@
...
+package com.google.cloud.spanner.connection;
+
+import com.google.api.core.InternalApi;
+import com.google.cloud.spanner.Dialect;
+import com.google.cloud.spanner.ErrorCode;
+import com.google.cloud.spanner.SpannerExceptionFactory;
+import com.google.common.base.Preconditions;
+
+@InternalApi
+public class PostgreSQLStatementParser extends AbstractStatementParser {
```
jin-jj, line 26: woa, hand wrote parser.  Nice job, we should also have this kind of parser at the backend.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Struct.java
```diff
@@ -371,6 +373,8 @@ private Object getAsObject(int columnIndex) {
             return getDoubleListInternal(columnIndex);
           case NUMERIC:
             return getBigDecimalListInternal(columnIndex);
+          case PG_NUMERIC:
```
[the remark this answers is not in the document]
olavloite: PG currently does not support structs in the backend, so in that sense we won't see this coming from the backend. The client library however allows users to convert a row to a struct, and in that case we could encounter it: https://github.com/googleapis/java-spanner/blob/8ea22205ea1361012b8f237af9150f320b41cc23/google-cloud-spanner/src/main/java/com/google/cloud/spanner/ResultSet.java#L52

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractStructReader.java
olavloite: Good point, I've updated it to just take a fixed string for now.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/PostgreSQLStatementParser.java
olavloite: You could re-write the backend to Java and then add this parser :-)

thiagotnunes (approved):
I think we did not parameterize some integration tests with the pg dialect. Could you add the parameterization for the ITBatchReadTest?

thiagotnunes:
Thanks for doing this!

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseInfo.java
```diff
@@ -239,6 +254,14 @@ public Timestamp getEarliestVersionTime() {
     return defaultLeader;
   }
 
+  /**
+   * The dialect that is used by the database. It can be one of the values as specified in {@link
+   * Dialect#values()}.
+   */
+  public @Nullable Dialect getDialect() {
+    return dialect;
```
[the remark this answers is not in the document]
olavloite: The `DatabaseInfo(DatabaseId id, State state)` constructor sets it to `null` (and also a lot of other fields that also could have had non-null values, like the encryption config), so I think we should keep it this way to keep it consistent with the other properties in the class.

google-cloud-spanner/clirr-ignored-differences.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- see http://www.mojohaus.org/clirr-maven-plugin/examples/ignored-differences.html -->
+<differences>
+  8001: com.google.cloud.spanner.connection.StatementParser: Class com.google.cloud.spanner.connection.StatementParser removed
```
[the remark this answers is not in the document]
olavloite: Nope, copy-paste error. Removed. Thanks for catching it (and strange that clirr does not complain about it...)

olavloite:
> I think we did not parameterize some integration tests with the pg dialect. Could you add the parameterization for the ITBatchReadTest?

I've parameterized the `ITBatchReadTest`, but I had to skip the two tests that use the Read API for the time being, as it returns an error. I'll reach out internally to see if anyone knows why that is happening.

PR #1680, opened 2022-02-11
fix: untyped null parameters would cause NPE

### LINKED ISSUE #1679: Statement.Builder#bind("some-param").to((Value) null) causes NullPointerException
Setting a statement parameter to an untyped null value will cause a `NullPointerException`. Example:

```java
Statement.newBuilder("INSERT INTO T (K, V) VALUES (@p1, @p2)")
    .bind("p1")
    .to("k1")
    .bind("p2")
    .to((Value) null)
    .build()))
```

Untyped null values are often used in ORMs and is commonly used in the JDBC driver. This will also be necessary for PostgreSQL connections, as the PostgreSQL JDBC driver allows applications to supply untyped null parameter values.

olavloite:
Adding an untyped null value as a parameter to a statement was not
possible, as:
1. The parameter collection would allow a null value to be added, but
   when the statement was built, it would throw a NullPointerException
   because it used an ImmutableMap internally, which does not support
   null values.
2. The translation from a hand-written Statement instance to a proto
   Statement instance would fail, as it did not take into account that
   the parameter could be null.

Fixes #1679


thiagotnunes (approved):
LGTM

PR #1688, opened 2022-02-14
fix: do not delete session in close method for BatchReadOnlyTransactionImpl

### LINKED ISSUE #1651: Spanner: BatchReadOnlyTransaction closes shared sessions on close()
[class `BatchReadOnlyTransactionImpl`](https://github.com/googleapis/java-spanner/blob/ae222be1e1689a15e4994a08e9cb8731f404b4a8/google-cloud-spanner/src/main/java/com/google/cloud/spanner/BatchClientImpl.java#L77) is an AutoClosable (via `MultiUseReadOnlyTransaction->AbstractReadContext->ReadContext->AutoClosable`)

The [`close()`](https://github.com/googleapis/java-spanner/blob/ae222be1e1689a15e4994a08e9cb8731f404b4a8/google-cloud-spanner/src/main/java/com/google/cloud/spanner/BatchClientImpl.java#L218) method always closes both the session and the transaction. 

This is fine if a session has been created for this transaction, using [`batchReadOnlyTransaction(TimestampBound)`](https://github.com/googleapis/java-spanner/blob/ae222be1e1689a15e4994a08e9cb8731f404b4a8/google-cloud-spanner/src/main/java/com/google/cloud/spanner/BatchClientImpl.java#L46)

But a BatchReadOnlyTransaction can also be created from an existing shared BatchTransaction: [`batchReadOnlyTransaction(BatchTransactionId)`](https://github.com/googleapis/java-spanner/blob/ae222be1e1689a15e4994a08e9cb8731f404b4a8/google-cloud-spanner/src/main/java/com/google/cloud/spanner/BatchClientImpl.java#L61). \
In this case, the Session is not _created_ for this transaction, but is taken from the `SessionID` from the `BatchTransactionId`, and so it is a _shared session_ so should not be closed when this transaction is closed.

Therefore the `close()` method should be fixed to _only_ close the session when the session has been created for this transaction. 

#### Environment details

2. OS type and version: all
3. Java version: all 
4.  version(s): latest (6.17.4)

#### Steps to reproduce

  1. Use the sample code at https://cloud.google.com/spanner/docs/samples/spanner-batch-client#spanner_batch_client-java
  2. Modify the code so that each thread creates a new separate batchReadOnlyTransaction, and uses that to perform the read (so that a different channel is used for the read)
      ```
      try (BatchReadOnlyTransaction readTxn = batchClient.batchReadOnlyTransaction(txn.getBatchTransactionId())) {
         try (ResultSet results = readTxn.execute(p)) {
            while (results.next()) {
            ...
            }
         }
      }
      ...
      ```
  3. Run the code on a large enough table to get a reasonably partitioned transaction in a single process
  4. Watch the code fail as one thread uses a session that has been closed by another

rajatbhatta:
-   remove session.close() from close() method implementation in
    BatchReadOnlyTransactionImpl class.

-   add a cleanup() method to BatchReadOnlyTransaction interface
    and give user the control to delete session when the session
    is no longer in use.

-   add tests for txn.cleanup() method.

Fixes #1651 ☕️

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/BatchClientImpl.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/BatchClientImpl.java
@@ -215,7 +215,7 @@ public class BatchClientImpl implements BatchClient {
     }
 
     /**
-     * Closes the session as part of the cleanup. It is the responsibility of the caller to make
+     * Closes the session as part of the cleanup. It is the responsibility of the caller to make a
      * call to this method once the transaction completes execution across all the channels (which
      * is understandably hard to identify). It is okay if the caller does not call the method
      * because the backend will anyways clean up the unused session.
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/BatchClientImpl.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/BatchClientImpl.java
@@ -214,9 +214,14 @@ public class BatchClientImpl implements BatchClient {
           partition.getPartitionToken());
     }
 
+    /**
+     * Closes the session as part of the cleanup. It is the responsibility of the caller to make
+     * call to this method once the transaction completes execution across all the channels (which
+     * is understandably hard to identify). It is okay if the caller does not call the method
+     * because the backend will anyways clean up the unused session.
+     */
     @Override
-    public void close() {
-      super.close();
+    public void cleanup() {
       session.close();
     }
   }
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/BatchReadOnlyTransaction.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/BatchReadOnlyTransaction.java
@@ -199,10 +199,10 @@ public interface BatchReadOnlyTransaction extends ReadOnlyTransaction {
   BatchTransactionId getBatchTransactionId();
 
   /**
-   * Closes the session as part of the cleanup. It is the responsibility of the caller to make
-   * call to this method once the transaction completes execution across all the channels (which
-   * is understandably hard to identify). It is okay if the caller does not call the method
-   * because the backend will anyways clean up the unused session.
+   * Closes the session as part of the cleanup. It is the responsibility of the caller to make call
+   * to this method once the transaction completes execution across all the channels (which is
+   * understandably hard to identify). It is okay if the caller does not call the method because the
+   * backend will anyways clean up the unused session.
    */
   default void cleanup() {}
 }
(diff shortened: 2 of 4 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/BatchClientImpl.java
```diff
@@ -214,9 +214,14 @@ public ResultSet execute(Partition partition) throws SpannerException {
...
 
+    /**
+     * Closes the session as part of the cleanup. It is the responsibility of the caller to make
+     * call to this method once the transaction completes execution across all the channels (which
+     * is understandably hard to identify). It is okay if the caller does not call the method
+     * because the backend will anyways clean up the unused session.
+     */
     @Override
-    public void close() {
-      super.close();
```
thiagotnunes, line 219: Wait, won't this break all the existing clients that are relying on the close method to return the session to the pool? (won't this possibly cause a session pool exhaustion for existing clients?)

cc @olavloite

rajatbhatta: As BatchReadOnlyTransactionImpl is an AutoCloseable (via MultiUseReadOnlyTransaction->AbstractReadContext->ReadContext->AutoClosable), having it within a try-with-resources construct allows automatic closing of the transaction resource, which leads the underlying session to be released to the session pool. Ideally, it should not lead to session pool exhaustion.

Having an overridden close method with just super.close() won't make sense, as the super's close method will anyways be called during automatic closing.

@olavloite : Please do share your viewpoint as well.

thiagotnunes: Oh got it, thanks for the explanation!

thiagotnunes (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BatchClientImpl.java
```diff
@@ -214,9 +214,14 @@ public ResultSet execute(Partition partition) throws SpannerException {
           partition.getPartitionToken());
     }
 
+    /**
+     * Closes the session as part of the cleanup. It is the responsibility of the caller to make
```
olavloite, line 218: nit:
```suggestion
     * Closes the session as part of the cleanup. It is the responsibility of the caller to make a
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BatchReadOnlyTransaction.java
```diff
@@ -197,4 +197,12 @@
    * BatchTransactionId guarantees the subsequent read/query to be executed at the same timestamp.
    */
   BatchTransactionId getBatchTransactionId();
+
+  /**
+   * Closes the session as part of the cleanup. It is the responsibility of the caller to make call
```
olavloite, line 202: nit:
```suggestion
   * Closes the session as part of the cleanup. It is the responsibility of the caller to make a call
```

olavloite (approved):

PR #1690, opened 2022-02-15
chore: make statement parsing public (marked as InternalApi)

olavloite:
Statement parsing is currently done internally in the Connection API do
determine what RPC to execute on the backend. The parsing is however
also very useful to some other applications, specifically PgAdapter,
which can use this to determine what response to send to a PG client.
This makes the parse method and related class and methods public, but
marks them as InternalApi, so they can be used in PgAdapter. This
follows the same pattern that was used to expose certain methods in the
Connection API for the JDBC driver.


thiagotnunes (approved):

ansh0l (approved):
lgtm

PR #1700, opened 2022-02-18
docs: add samples for PostgreSQL

olavloite:
Adds samples specifically for PG dialect databases.

cc @jin-jj


samples/snippets/src/main/java/com/example/spanner/PgCaseSensitivitySample.java
```diff
@@ -0,0 +1,151 @@
...
+                  // Column names in mutations are always case-insensitive, regardless whether the
+                  // columns were double-quoted or not during creation.
+                  .set("firstname")
+                  .to("Bruce")
+                  .set("lastname")
+                  .to("Allison")
+                  .build()));
+
+      try (ResultSet singers =
+          client.singleUse().executeQuery(Statement.of("SELECT * FROM Singers"))) {
```
jin-jj, line 92: Change "Select *" to explicit column list?

samples/snippets/src/main/java/com/example/spanner/PgInterleavedTableSample.java
```diff
@@ -0,0 +1,81 @@
...
+                      + "  SingerId  bigint NOT NULL PRIMARY KEY,"
+                      + "  FirstName varchar(1024) NOT NULL,"
+                      + "  LastName  varchar(1024) NOT NULL"
+                      + ")",
+                  "CREATE TABLE Albums ("
+                      + "  SingerId bigint NOT NULL,"
+                      + "  AlbumId  bigint NOT NULL,"
+                      + "  Title    varchar(1024) NOT NULL,"
+                      + "  PRIMARY KEY (SingerId, AlbumId)"
+                      + ") INTERLEAVE IN PARENT Singers ON DELETE CASCADE"),
```
jin-jj, line 67: Do we have sample for drop?  One of the benefit is that we only support cascade drop for interleaved table.

olavloite: What exactly do you mean? I don't think we support cascade drop, only cascade delete. Or is that what you meant?

jin-jj:
> LGTM.
> 
> @jin-jj : Can you have a look at the PR if it looks ok to you? Now that we have the PR for existing samples available, I think this can be merged together with that one. So marking it as approved.

LGTM. I have a comment about the INTERLEAVE IN on delete cascade, but let's worry about it later.

PR #1708, opened 2022-02-21
fix: use information_schema instead of pg_catalog for dialect detection

olavloite:
Automatic dialect detection should not rely on `pg_catalog` being present in the `information_schema` to determine that it is a PostgreSQL database. Instead, the presence of an upper/lower case `INFORMATION_SCHEMA` / `information_schema` should be used to determine which of the two dialects is being used.


ansh0l (approved):
LGTM

PR #1711, opened 2022-02-23
feat: Track PG Adapter usage from user-agent headers

gauravsnj:
Added the string "pg-adapter" in the list of allowed client library token, so that when PG Adapter connects via client library by passing the string "pg-adapter" in the user-agent header, it doesn't get discarded. Also, added the one test case to cover the same.

Fixes [b/213946108](http://b/213946108)

### DIFF
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerOptionsTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerOptionsTest.java
@@ -19,6 +19,7 @@ package com.google.cloud.spanner;
 import static com.google.common.truth.Truth.assertThat;
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertSame;
 import static org.junit.Assert.assertThrows;
 import static org.mockito.Mockito.mock;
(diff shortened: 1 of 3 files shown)


google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerOptionsTest.java
```diff
@@ -518,6 +519,14 @@ public void testSetClientLibToken() {
             .build();
     assertThat(options.getClientLibToken()).isEqualTo(hibernateToken);
 
+    options =
+        SpannerOptions.newBuilder()
+            .setProjectId("some-project")
+            .setCredentials(NoCredentials.getInstance())
+            .setClientLibToken(pgAdapterToken)
+            .build();
+    assertThat(options.getClientLibToken()).isEqualTo(pgAdapterToken);
```
olavloite, line 528: nit: we prefer the use of `assertEquals(expected, actual)` (add `import static org.junit.Assert.assertEquals` at the top of the file to import it)

I know that it is not consistent with the other tests here, but the general rule of thumb is to use this method for new code, even when it is different from the surrounding code.

olavloite (commented):
LGTM, with a small nit on the assertion method that is used. I'm going to hold off final approval to give @ansh0l a chance to put in a review as well.

ansh0l (approved):
LGTM

olavloite (approved):

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerOptionsTest.java
gauravsnj: Thanks @olavloite for mentioning this. I've changed it and will remember this in future

olavloite:
@ansh0l Do you know why the OwlBot check seems to be hanging? Is that because the PR has been created from a fork instead of a branch directly in the repository? And if so, is that something you can override for this PR?

thiagotnunes:
@olavloite it does not run by default from forks, we just need to add the label `owlbot:run`

PR #1726, opened 2022-03-03
fix: Correct recording values in opencensus measureMap in HeaderInterceptor

KiranmayiB:

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/HeaderInterceptor.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/HeaderInterceptor.java
@@ -95,8 +95,9 @@ class HeaderInterceptor implements ClientInterceptor {
       if (matcher.find()) {
         try {
           long latency = Long.parseLong(matcher.group("dur"));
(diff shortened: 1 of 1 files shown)


thiagotnunes:
@KiranmayiB can you explain how does this change solve the issue we are seeing?

KiranmayiB:
> @KiranmayiB can you explain how does this change solve the issue we are seeing?

Here, the code puts a value into the measureMap twice and says record twice. That is causing the crash as we cannot put anything into the measureMap after we say record. So i think the second time when it tried to put a value after the record, it could not internally find the bucket causing an NPE. This is what i gathered from the bug.

thiagotnunes:
@KiranmayiB is there any (manual) testing we can do for this (to simulate the bug and have the fix correct it)?

thiagotnunes:
@KiranmayiB @ansh0l any updates?

rajatbhatta:
@thiagotnunes: We shared the jar with the changes in this PR with the end user, and they confirmed that they're not facing the issue anymore. They're waiting for an official release to be available to use it. I think we can merge this PR.

thiagotnunes:
Sounds good, thanks for checking!

thiagotnunes (approved):

PR #1731, opened 2022-03-07
fix: PostgreSQL supports newline in quoted literals and identifiers

### LINKED ISSUE #1730: PostgreSQL statement parsing fails on string literals with newline characters
The PostgreSQL statement parser that removes comments and determines how many query parameters are in a query fails if the statement contains one or more string literals or quoted identifiers that contain a newline character.

So the following statements fail with an `INVALID_ARGUMENT: SQL statement contains an unclosed literal` error:

```sql
select 'test1
test2';

select "foo
bar" from some_table;
```

olavloite:
PostgreSQL supports newline characters in string literals and quoted
identifiers. Trying to execute a statement with a string literal or
quoted identifier that contained a newline character would cause an
'Unclosed string literal' error.

Fixes #1730

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/PostgreSQLStatementParser.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/PostgreSQLStatementParser.java
@@ -56,94 +56,43 @@ public class PostgreSQLStatementParser extends AbstractStatementParser {
   @Override
   String removeCommentsAndTrimInternal(String sql) {
     Preconditions.checkNotNull(sql);
-    String currentTag = null;
-    boolean isInQuoted = false;
     boolean isInSingleLineComment = false;
     int multiLineCommentLevel = 0;
-    char startQuote = 0;
-    boolean lastCharWasEscapeChar = false;
     StringBuilder res = new StringBuilder(sql.length());
     int index = 0;
     while (index < sql.length()) {
       char c = sql.charAt(index);
-      if (isInQuoted) {
-        if ((c == '\n' || c == '\r') && startQuote != DOLLAR) {
-          throw SpannerExceptionFactory.newSpannerException(
-              ErrorCode.INVALID_ARGUMENT, "SQL statement contains an unclosed literal: " + sql);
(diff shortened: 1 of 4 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/PostgreSQLStatementParser.java
```diff
@@ -56,95 +56,44 @@ protected boolean supportsExplain() {
   @Override
   String removeCommentsAndTrimInternal(String sql) {
     Preconditions.checkNotNull(sql);
-    String currentTag = null;
```
olavloite, line 59: It might be easier to open the entire file (online or in an IDE) to review it, than to try to do that based on this diff render from GitHub. The parser code has also been compacted a lot, as there was quite a lot of code duplication, and that makes this diff render a lot harder to read.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/PostgreSQLStatementParser.java
```diff
@@ -56,95 +56,44 @@ protected boolean supportsExplain() {
...
-            multiLineCommentLevel++;
-            index++;
-          }
+        // Check for -- which indicates the start of a single-line comment.
+        if (sql.length() > index + 1 && c == HYPHEN && sql.charAt(index + 1) == HYPHEN) {
+          // This is a single line comment.
+          isInSingleLineComment = true;
+          index += 2;
+          continue;
+        } else if (sql.length() > index + 1 && c == SLASH && sql.charAt(index + 1) == ASTERIKS) {
```
thiagotnunes, line 86: typo `ASTERIKS -> ASTERISKS`?

thiagotnunes (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/PostgreSQLStatementParser.java
olavloite: Fixed

PR #1732, opened 2022-03-07
feat: parse query parameters in PostgreSQL query

olavloite:
Adds a helper method to get the parameters from a PostgreSQL query. This
is needed for DESCRIBE statement messages in PGAdapter, as it must
return the data types of all query parameters in a query string. Even
though this parser is not able to determine the parameter types, it is
able to determine the number of parameters. This again makes it possible
to PGAdapter to return Oid.UNSPECIFIED for each parameter in the query
string, which is enough for most clients.

cc @pratickchokhani

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/PostgreSQLStatementParser.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/PostgreSQLStatementParser.java
@@ -21,6 +21,9 @@ import com.google.cloud.spanner.Dialect;
 import com.google.cloud.spanner.ErrorCode;
 import com.google.cloud.spanner.SpannerExceptionFactory;
 import com.google.common.base.Preconditions;
+import java.util.HashSet;
+import java.util.Set;
+import javax.annotation.Nullable;
 
 @InternalApi
 public class PostgreSQLStatementParser extends AbstractStatementParser {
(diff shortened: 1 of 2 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/PostgreSQLStatementParser.java
```diff
@@ -149,30 +152,76 @@ ParametersInfo convertPositionalParametersToNamedParametersInternal(char paramCh
...
+   * from some_table
+   * where col1=$1 and col2=$2
+   * and not col3=$1 and col4='$3'
+   * </code>
+   *
+   * @param sql the SQL-string to check for parameters. Must not contain comments.
+   * @return A set containing all the parameters in the SQL-string.
+   */
+  @InternalApi
+  public Set<String> getQueryParameters(String sql) {
```
thiagotnunes, line 174: I imagine we don't need to consider malformed queries here (`select col$4" from some_table ...`)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/PostgreSQLStatementParser.java
```diff
@@ -149,30 +152,76 @@ ParametersInfo convertPositionalParametersToNamedParametersInternal(char paramCh
...
+      }
+    }
+    return parameters;
+  }
+
+  private int skip(String sql, int currentIndex, @Nullable StringBuilder result) {
     char currentChar = sql.charAt(currentIndex);
     if (currentChar == SINGLE_QUOTE || currentChar == DOUBLE_QUOTE) {
-      result.append(currentChar);
+      appendIfNotNull(result, currentChar);
```
thiagotnunes, line 199: nit: will it ever be `null`?

thiagotnunes (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/PostgreSQLStatementParser.java
olavloite: Correct. That is caught by the preceding step (`removeCommentsAndTrim`) that should be taken by any well-informed users of the API.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/PostgreSQLStatementParser.java
olavloite: Yes, `result` will be null if the caller does not need the result. That's a (minor) performance optimization not to build a resulting query string when the call just wants to scan through the SQL string to extract specific information. So in this specific case we skip building a result string as we are only interested in extracting the query parameters, and not building a new SQL string with for example a different style query parameters (e.g. when converting JDBC style `?` parameters to PostgreSQL style parameters `$1`).

PR #1742, opened 2022-03-09
feat: [DO NOT MERGE] support customer managed instance configurations

rajatbhatta:


google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigInfo.java
```diff
@@ -69,6 +109,266 @@ public String getDisplayName() {
...
+    public InstanceConfigInfo build() {
+      return new InstanceConfigInfo(this);
+    }
+  }
+
+  public InstanceConfigInfo(InstanceConfigId id, String displayName) {
+    this((BuilderImpl) newBuilder(id).setDisplayName(displayName));
+  }
+
+  public static Builder newBuilder(InstanceConfigId id) {
```
olavloite, line 337: I would suggest removing this (or otherwise making it package-private if we need it for internal reasons). BaseConfig is always required for custom InstanceConfigs.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigInfo.java
```diff
@@ -69,6 +109,266 @@ public String getDisplayName() {
...
+      this.reconciling = false;
+      this.configType = Type.TYPE_UNSPECIFIED;
+      this.displayName = "";
+      this.etag = "";
+    }
+
+    BuilderImpl(InstanceConfigInfo instanceConfigInfo) {
+      this.id = instanceConfigInfo.id;
+      this.displayName = instanceConfigInfo.displayName;
+      this.replicas = instanceConfigInfo.replicas;
```
olavloite, line 237: Make a copy of the Lists / Maps etc.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -875,6 +880,48 @@ private void acquireAdministrativeRequestsRateLimiter() {
     return new Paginated<>(response.getInstanceConfigsList(), response.getNextPageToken());
   }
 
+  @Override
+  public OperationFuture<InstanceConfig, CreateInstanceConfigMetadata> createInstanceConfig(
+      String parent,
+      String instanceConfigId,
+      InstanceConfig instanceConfig,
+      @Nullable Boolean validateOnly)
```
olavloite, line 888: Is there a difference between null and false in this case? If not, I would prefer using a boolean instead of a Boolean.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerRpc.java
```diff
@@ -34,10 +34,7 @@
 import com.google.protobuf.Empty;
 import com.google.protobuf.FieldMask;
 import com.google.spanner.admin.database.v1.*;
-import com.google.spanner.admin.instance.v1.CreateInstanceMetadata;
-import com.google.spanner.admin.instance.v1.Instance;
-import com.google.spanner.admin.instance.v1.InstanceConfig;
-import com.google.spanner.admin.instance.v1.UpdateInstanceMetadata;
+import com.google.spanner.admin.instance.v1.*;
```
olavloite, line 37: Google style requires us to never use wildcard imports: https://google.github.io/styleguide/javaguide.html

google-cloud-spanner/src/test/java/com/google/cloud/spanner/InstanceConfigTest.java
```diff
@@ -63,26 +76,44 @@ public void testInstanceConfigFromProto() {
...
-                    .setDefaultLeaderLocation(false)
-                    .build(),
-                ReplicaInfo.newBuilder()
-                    .setLocation("Replica Location 3")
-                    .setType(ReplicaInfo.ReplicaType.WITNESS)
-                    .setDefaultLeaderLocation(false)
-                    .build()),
-            Arrays.asList("Leader Option 1", "Leader Option 2"),
-            client),
+            (InstanceConfig.Builder)
```
olavloite, line 79: I don't understand why we need a cast here?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigInfo.java
```diff
@@ -69,6 +109,266 @@ public String getDisplayName() {
...
+      this.reconciling = instanceConfigInfo.reconciling;
+      this.state = instanceConfigInfo.state;
+      this.labels = new HashMap<>(instanceConfigInfo.labels);
+    }
+
+    private void addAllReplicasFromBaseConfig() {
+      // Custom configurations contain all the replicas of the base config and at least one optional
+      // replica.
+      this.replicas = this.baseConfig.getReplicas();
+      this.replicas.add(baseConfig.getOptionalReplicas().get(0));
```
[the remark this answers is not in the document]
rajatbhatta: Oops, had to remove this after adding the addReadOnlyReplicas support. Will remove this method.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigInfo.java
```diff
@@ -69,6 +109,266 @@ public String getDisplayName() {
...
+      this.etag = instanceConfigInfo.etag;
+      this.reconciling = instanceConfigInfo.reconciling;
+      this.state = instanceConfigInfo.state;
+      this.labels = new HashMap<>(instanceConfigInfo.labels);
+    }
+
+    private void addAllReplicasFromBaseConfig() {
+      // Custom configurations contain all the replicas of the base config and at least one optional
+      // replica.
+      this.replicas = this.baseConfig.getReplicas();
```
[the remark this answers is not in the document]
rajatbhatta: This method is not needed now, after adding support for addReadOnlyReplicas method. Will remove it.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigInfo.java
```diff
@@ -69,6 +109,266 @@ public String getDisplayName() {
...
+    }
+
+    @Override
+    public Builder setDisplayName(String displayName) {
+      this.displayName = displayName;
+      return this;
+    }
+
+    @Override
+    public Builder addAllReplicas(List<ReplicaInfo> replicas) {
```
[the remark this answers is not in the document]
rajatbhatta: Agreed. Will change this to protected, as the user is setting replicas through newBuilder(InstanceConfigId id, InstanceConfigInfo baseConfig) and addReadOnlyReplicas method. We'll need this method only during conversion from com.google.spanner.admin.instance.v1.InstanceConfig object to com.google.cloud.spanner.InstanceConfig object.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigInfo.java
```diff
@@ -69,6 +109,266 @@ public String getDisplayName() {
...
+    }
+
+    @Override
+    public Builder setBaseConfig(InstanceConfigInfo baseConfig) {
+      this.baseConfig = baseConfig;
+      return this;
+    }
+
+    @Override
+    protected Builder setConfigType(Type configType) {
```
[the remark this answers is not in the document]
rajatbhatta: It has been marked as protected.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigInfo.java
rajatbhatta: Will make this newBuilder private.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigInfo.java
```diff
@@ -69,6 +109,266 @@ public String getDisplayName() {
...
+
+  public static Builder newBuilder(InstanceConfigId id) {
+    return new BuilderImpl(id);
+  }
+
+  public static Builder newBuilder(InstanceConfigId id, InstanceConfigInfo baseConfig) {
+    return new BuilderImpl(id, baseConfig);
+  }
+
+  public InstanceConfigInfo(
```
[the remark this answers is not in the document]
rajatbhatta: Won't removing this or making it private be a breaking change? Users may be using this constructor for creating objects currently.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
rajatbhatta: If the user does not pass in any validateOnlyOption, this value will be null. Hence, we have @Nullable here. If we have this value is null, we need not to populate validateOnly field to the backend.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -875,6 +880,48 @@ private void acquireAdministrativeRequestsRateLimiter() {
...
+    }
+    CreateInstanceConfigRequest request = builder.build();
+    GrpcCallContext context =
+        newCallContext(null, parent, request, InstanceAdminGrpc.getCreateInstanceConfigMethod());
+    return instanceAdminStub.createInstanceConfigOperationCallable().futureCall(request, context);
+  }
+
+  @Override
+  public OperationFuture<InstanceConfig, UpdateInstanceConfigMetadata> updateInstanceConfig(
+      InstanceConfig instanceConfig, @Nullable Boolean validateOnly, FieldMask fieldMask)
```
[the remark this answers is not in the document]
rajatbhatta: If the user does not pass in validateOnlyOption, this value will be null. Hence, we have Nullable here. If we have this value is null, we need not to populate validateOnly field to the backend (we need not even send false here, as that's the default behaviour if we do not pass in validateOnly).

google-cloud-spanner/src/test/java/com/google/cloud/spanner/InstanceConfigTest.java
rajatbhatta: All the setters exist in base InstanceConfigInfo.Builder/BuilderImpl class and return InstanceConfigInfo.Builder object. We just inherit these setters in InstanceConfig.Builder. When we use these setters, the object returned is of type InstanceConfigInfo.Builder and we cast it to more specific InstanceConfig.Builder.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigInfo.java
rajatbhatta: This newBuilder is used to create InstanceConfigInfo object to be passed during updateInstanceConfig. We do not require a baseConfig in that case.

google-cloud-spanner/clirr-ignored-differences.xml
```diff
@@ -55,6 +95,11 @@
     <className>com/google/cloud/spanner/DatabaseAdminClient</className>
     <method>com.google.api.gax.longrunning.OperationFuture copyBackup(com.google.cloud.spanner.BackupId, com.google.cloud.spanner.Backup)</method>
   </difference>
+  <difference>
```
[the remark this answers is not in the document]
rajatbhatta: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
```diff
@@ -151,6 +170,14 @@ public static ListOption filter(String filter) {
     return new FilterOption(filter);
   }
 
+  public static DeleteAdminAPIOption etag(String etag) {
```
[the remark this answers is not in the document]
rajatbhatta: Added.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerRpc.java
rajatbhatta: Made this change.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigInfo.java
```diff
@@ -69,6 +109,266 @@ public String getDisplayName() {
...
+   */
+  public Map<String, String> getLabels() {
+    return labels;
+  }
+
+  /** Builder for {@code InstanceConfigInfo}. */
+  public abstract static class Builder {
+    public abstract Builder setDisplayName(String displayName);
+
+    public abstract Builder addAllReplicas(List<ReplicaInfo> replicas);
```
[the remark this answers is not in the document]
rajatbhatta: Changed to protected.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigInfo.java
```diff
@@ -69,6 +109,266 @@ public String getDisplayName() {
...
+
+    @Override
+    public Builder addAllLeaderOptions(List<String> leaderOptions) {
+      this.leaderOptions = leaderOptions;
+      return this;
+    }
+
+    @Override
+    protected Builder addAllOptionalReplicas(List<ReplicaInfo> optionalReplicas) {
+      this.optionalReplicas = optionalReplicas;
```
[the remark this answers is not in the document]
rajatbhatta: Changed to `set`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigInfo.java
```diff
@@ -69,6 +109,266 @@ public String getDisplayName() {
...
+
+    @Override
+    public Builder addAllReplicas(List<ReplicaInfo> replicas) {
+      this.replicas = replicas;
+      return this;
+    }
+
+    @Override
+    public Builder addAllLeaderOptions(List<String> leaderOptions) {
+      this.leaderOptions = leaderOptions;
```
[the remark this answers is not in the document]
rajatbhatta: Changed to `set`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigInfo.java
```diff
@@ -69,6 +109,266 @@ public String getDisplayName() {
...
+      return this;
+    }
+
+    @Override
+    public InstanceConfigInfo build() {
+      return new InstanceConfigInfo(this);
+    }
+  }
+
+  public InstanceConfigInfo(InstanceConfigId id, String displayName) {
```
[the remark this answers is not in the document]
rajatbhatta: Won't making it private create a breaking change? Users may be using this constructor for creating objects currently.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigInfo.java
rajatbhatta: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigInfo.java
```diff
@@ -69,6 +109,266 @@ public String getDisplayName() {
...
+            newBuilder(id)
+                .setDisplayName(displayName)
+                .addAllReplicas(replicas)
+                .addAllLeaderOptions(leaderOptions));
+  }
+
+  InstanceConfigInfo(BuilderImpl builder) {
+    this.id = builder.id;
+    this.displayName = builder.displayName;
+    this.replicas = builder.replicas;
```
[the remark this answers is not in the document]
rajatbhatta: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigInfo.java
```diff
@@ -69,6 +109,266 @@ public String getDisplayName() {
...
+    }
+
+    @Override
+    protected Builder addAllOptionalReplicas(List<ReplicaInfo> optionalReplicas) {
+      this.optionalReplicas = optionalReplicas;
+      return this;
+    }
+
+    @Override
+    public Builder setBaseConfig(InstanceConfigInfo baseConfig) {
```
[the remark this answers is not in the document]
rajatbhatta: Changed it to protected.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigInfo.java
```diff
@@ -69,6 +109,266 @@ public String getDisplayName() {
...
+    }
+
+    @Override
+    public Builder addAllReplicas(List<ReplicaInfo> replicas) {
+      this.replicas = replicas;
+      return this;
+    }
+
+    @Override
+    public Builder addAllLeaderOptions(List<String> leaderOptions) {
```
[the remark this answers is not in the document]
rajatbhatta: Added.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigInfo.java
```diff
@@ -69,6 +109,266 @@ public String getDisplayName() {
...
+
+  /** Builder for {@code InstanceConfigInfo}. */
+  public abstract static class Builder {
+    public abstract Builder setDisplayName(String displayName);
+
+    public abstract Builder addAllReplicas(List<ReplicaInfo> replicas);
+
+    protected abstract Builder addAllOptionalReplicas(List<ReplicaInfo> optionalReplicas);
+
+    public abstract Builder setBaseConfig(InstanceConfigInfo baseConfig);
```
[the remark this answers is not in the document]
rajatbhatta: Changed to protected.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigInfo.java
```diff
@@ -69,6 +109,266 @@ public String getDisplayName() {
...
+
+    public abstract Builder addAllLeaderOptions(List<String> leaderOptions);
```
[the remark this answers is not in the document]
rajatbhatta: leader_options is not a read-only field. So, having it makes sense. Let me know if you think otherwise.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigInfo.java
```diff
@@ -69,6 +109,266 @@ public String getDisplayName() {
...
+
+    public abstract Builder setEtag(String etag);
+
+    protected abstract Builder setReconciling(boolean reconciling);
+
+    public abstract Builder addLabel(String key, String value);
+
+    public abstract Builder putAllLabels(Map<String, String> labels);
+
+    public abstract Builder addReadOnlyReplicas(List<ReplicaInfo> readOnlyReplicas);
```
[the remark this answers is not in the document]
rajatbhatta: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceAdminClient.java
```diff
@@ -19,14 +19,140 @@
...
+   *     InstanceConfig.newBuilder(InstanceConfigId.of(projectId, instanceConfigId))
+   *         .setDisplayName(displayName)
+   *         .build();
+   *
+   * // Only update display name.
+   * final OperationFuture<InstanceConfig, UpdateInstanceConfigMetadata> operation =
+   *     instanceAdminClient.updateInstanceConfig(
+   *         instanceConfigInfo, ImmutableList.of(InstanceConfigField.DISPLAY_NAME));
+   *
+   * InstanceConfig instanceConfig = operation.get(5, TimeUnit.MINUTES);
```
thiagotnunes, line 143: nit: should we use the same timeout in minutes for all operations? This one uses 5 minutes, while above we use 10 minutes. Does it mean this one is expected to be faster? If not, I would argue to use the same timeout.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
```diff
@@ -51,6 +51,22 @@
   public interface ReadQueryUpdateTransactionOption
       extends ReadOption, QueryOption, UpdateOption, TransactionOption {}
 
+  /**
+   * Marker interface to mark options applicable to Create, Update and Delete operations in admin
+   * API.
+   */
+  public interface CreateUpdateDeleteAdminAPIOption
```
thiagotnunes, line 58: Please use camel case for `Api` as well, per https://engdoc.corp.google.com/eng/doc/devguide/java/style/index.md?cl=head#s5.3-camel-case

Please change everywhere else too

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceAdminClientImpl.java
```diff
@@ -60,13 +65,81 @@ protected Policy fromPb(com.google.iam.v1.Policy policyPb) {
     this.dbClient = dbClient;
   }
 
+  @Override
+  public OperationFuture<InstanceConfig, CreateInstanceConfigMetadata> createInstanceConfig(
+      InstanceConfigInfo instanceConfig, CreateAdminAPIOption... options) throws SpannerException {
+    final Options createAdminAPIOptions = Options.fromAdminAPIOptions(options);
```
thiagotnunes, line 71: Please use camel case `Api`, per https://engdoc.corp.google.com/eng/doc/devguide/java/style/index.md?cl=head#s5.3-camel-case

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigInfo.java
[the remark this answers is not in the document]
thiagotnunes: +1, we can not change the visibility here. We could mark it as deprecated though (with docs pointing to the new method)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigInfo.java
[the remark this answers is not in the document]
thiagotnunes: +1, we can not change the visibility here. We could mark it as deprecated though (with docs pointing to the new method)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigInfo.java
```diff
@@ -81,17 +381,159 @@ public boolean equals(Object o) {
...
+            .setState(toProtoState(getState()));
+
+    if (getBaseConfig() != null) {
+      builder.setBaseConfig(getBaseConfig().getId().getName());
+    }
+
+    return builder.build();
+  }
+
+  static InstanceConfig.Type toProtoConfigType(Type type) {
```
thiagotnunes, line 456: Could this be private? If it is more open just for testing, could you add the `@VisibleForTesting` annotation?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigInfo.java
```diff
@@ -81,17 +381,159 @@ public boolean equals(Object o) {
...
+      case GOOGLE_MANAGED:
+        return com.google.spanner.admin.instance.v1.InstanceConfig.Type.GOOGLE_MANAGED;
+      case USER_MANAGED:
+        return InstanceConfig.Type.USER_MANAGED;
+      default:
+        throw new IllegalArgumentException("Unknown config type:" + type);
+    }
+  }
+
+  static InstanceConfig.State toProtoState(State state) {
```
thiagotnunes, line 469: Could this be private? If it is more open just for testing, could you add the `@VisibleForTesting` annotation?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigInfo.java
```diff
@@ -81,17 +381,159 @@ public boolean equals(Object o) {
...
+      case CREATING:
+        return com.google.spanner.admin.instance.v1.InstanceConfig.State.CREATING;
+      case READY:
+        return com.google.spanner.admin.instance.v1.InstanceConfig.State.READY;
+      default:
+        throw new IllegalArgumentException("Unknown state:" + state);
+    }
+  }
+
+  static com.google.cloud.spanner.InstanceConfig fromProto(
```
thiagotnunes, line 482: Could this be private? If it is more open just for testing, could you add the `@VisibleForTesting` annotation?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigInfo.java
```diff
@@ -81,17 +381,159 @@ public boolean equals(Object o) {
...
+      builder.setBaseConfig(
+          new com.google.cloud.spanner.InstanceConfigInfo.BuilderImpl(
+                  InstanceConfigId.of(proto.getBaseConfig()))
+              .build());
+    }
+
+    return builder.build();
+  }
+
+  static State fromProtoState(com.google.spanner.admin.instance.v1.InstanceConfig.State state) {
```
thiagotnunes, line 514: Could this be private? If it is more open just for testing, could you add the `@VisibleForTesting` annotation?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigInfo.java
```diff
@@ -81,17 +381,159 @@ public boolean equals(Object o) {
...
+      case CREATING:
+        return State.CREATING;
+      case READY:
+        return State.READY;
+      default:
+        throw new IllegalArgumentException("Unknown state:" + state);
+    }
+  }
+
+  static Type fromProtoConfigType(com.google.spanner.admin.instance.v1.InstanceConfig.Type type) {
```
thiagotnunes, line 527: Could this be private? If it is more open just for testing, could you add the `@VisibleForTesting` annotation?

pom.xml
```diff
@@ -4,7 +4,7 @@
   <groupId>com.google.cloud</groupId>
   <artifactId>google-cloud-spanner-parent</artifactId>
   <packaging>pom</packaging>
-  <version>6.23.2-SNAPSHOT</version><!-- {x-version-update:google-cloud-spanner:current} -->
+  <version>6.23.2-SNAPSHOT</version>
```
thiagotnunes, line 7: Why did we remove the metadata here?

google-cloud-spanner/clirr-ignored-differences.xml
```diff
@@ -35,6 +35,46 @@
...
+    <differenceType>7004</differenceType>
+    <className>com/google/cloud/spanner/InstanceConfig</className>
+    <method>InstanceConfig(com.google.cloud.spanner.InstanceConfigId, java.lang.String, java.util.List, java.util.List, com.google.cloud.spanner.InstanceAdminClient)</method>
+  </difference>
+  <difference>
+    <differenceType>7004</differenceType>
+    <className>com/google/cloud/spanner/InstanceConfigInfo</className>
+    <method>InstanceConfigInfo(com.google.cloud.spanner.InstanceConfigId, java.lang.String, java.util.List, java.util.List)</method>
```
thiagotnunes, line 46: I don't think we can have breaking changes in constructors, could you please overload it instead of changing it to avoid the breakage?

google-cloud-spanner/clirr-ignored-differences.xml
```diff
@@ -35,6 +35,46 @@
     <className>com/google/cloud/spanner/connection/ConnectionOptions</className>
     <method>com.google.cloud.spanner.Dialect getDialect()</method>
   </difference>
+  <difference>
+    <differenceType>7004</differenceType>
+    <className>com/google/cloud/spanner/InstanceConfig</className>
+    <method>InstanceConfig(com.google.cloud.spanner.InstanceConfigId, java.lang.String, java.util.List, java.util.List, com.google.cloud.spanner.InstanceAdminClient)</method>
```
thiagotnunes, line 41: I don't think we can have breaking changes in constructors, could you please overload it instead of changing it to avoid the breakage?

rajatbhatta: This constructor has not been changed, and these clirr differences are not required here. Will remove these. Thanks for pointing this out.

google-cloud-spanner/clirr-ignored-differences.xml
rajatbhatta: This constructor has not been changed, and these clirr differences are not required here. Will remove these. Thanks for pointing this out.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceAdminClient.java
rajatbhatta: Changing both the timeouts to 5 mins makes sense.

rajatbhatta: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
rajatbhatta: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceAdminClientImpl.java
rajatbhatta: Done.

google-cloud-spanner/clirr-ignored-differences.xml
rajatbhatta: Removed these differences.

google-cloud-spanner/clirr-ignored-differences.xml
rajatbhatta: Removed these differences.

pom.xml
rajatbhatta: We need not remove this. Re-added the same.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigInfo.java
rajatbhatta: Making it private makes sense. It is not directly used in unit tests, so I think we need not add `@VisibleForTesting` annotation.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigInfo.java
rajatbhatta: Making it private makes sense. It is not directly used in unit tests, so I think we need not add @VisibleForTesting annotation.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigInfo.java
rajatbhatta: Making it private makes sense. It is not directly used in unit tests, so I think we need not add @VisibleForTesting annotation.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigInfo.java
rajatbhatta: Making it private makes sense. It is not directly used in unit tests, so I think we need not add @VisibleForTesting annotation.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigInfo.java
rajatbhatta: This cannot be made private because the method is being used in InstanceAdminClientImpl.java in createInstanceConfig, updateInstanceConfig, etc.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigInfo.java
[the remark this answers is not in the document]
rajatbhatta: Marked this constructor as deprecated with docs to point to newBuilder method.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigInfo.java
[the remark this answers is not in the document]
rajatbhatta: Marked this constructor as deprecated with docs to point to newBuilder method.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigInfo.java
```diff
@@ -69,6 +109,266 @@ public String getDisplayName() {
...
+    private InstanceConfigInfo baseConfig;
+    private Type configType;
+    private String etag;
+    private boolean reconciling;
+    private State state;
+    private Map<String, String> labels;
+
+    BuilderImpl(InstanceConfigId id) {
+      this.id = id;
+      this.labels = new HashMap<>();
```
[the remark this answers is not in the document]
rajatbhatta: Thanks for a clear explanation. This is done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceConfigInfo.java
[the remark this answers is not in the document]
rajatbhatta: Done.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/InstanceConfigTest.java
olavloite: Ok, that makes sense, but I think that shows a small (potential?) problem with the current implementation: Users should not be forced to do any casts when working with the builder. So if the implementation in the abstract base class (`InstanceConfigInfo`) returns the abstract class, but all our public methods require the use of the concrete subclass (`InstanceConfig`), then we should make sure that the methods that are visible to users als return an instance of `InstanceConfig.Builder` and not `InstanceConfigInfo.Builder`.
Could you take a look at `InstanceConfig` and `InstanceConfigInfo` how that is handled there, and modify this test case (and potentially the implementation) to show that users do not need to do any casting when working with the builder(s)?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
olavloite: Makes sense to me.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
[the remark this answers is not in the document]
olavloite: Makes sense to me

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
```diff
@@ -151,6 +170,33 @@ public static ListOption filter(String filter) {
...
+   * Specifying this will not actually execute a request, and provide the same response. Operations
+   * that support this option are:
+   *
+   * <ul>
+   *   <li>{@link InstanceAdminClient#createInstanceConfig}
+   *   <li>{@link InstanceAdminClient#updateInstanceConfig}
+   *   <li>{@link InstanceAdminClient#deleteInstanceConfig}
+   * </ul>
+   */
+  public static CreateUpdateDeleteAdminApiOption validateOnly(Boolean validateOnly) {
```
thiagotnunes, line 196: nit: should this be `boolean` or can we explicitly set a `null`?

thiagotnunes (approved):
LGTM

google-cloud-spanner/src/test/java/com/google/cloud/spanner/InstanceConfigTest.java
rajatbhatta: Thanks for the explanation! 

The solution I've implemented now to handle this is to override the setters in InstanceConfig.Builder class and make them return InstanceConfig.Builder.

For example, overridden setDisplayName in InstanceConfig.Builder would look something like this:
```
@Override
public Builder setDisplayName(String displayName) {
  super.setDisplayName(displayName);
  return this;
}
```

olavloite (approved):

rajatbhatta:
Hi @thiagotnunes @olavloite : I've added support (both implementation and tests) for ListInstanceConfigOperations as well to this PR. Can you please review changes since previous code review? Thank you!

olavloite (approved):
The last changes also look good to me. I assume that we will add end-to-end tests either in a separate PR, or later in this PR once it is enabled on the backend?

rajatbhatta:
> The last changes also look good to me. I assume that we will add end-to-end tests either in a separate PR, or later in this PR once it is enabled on the backend?

Thanks for the review! Yes, we'll add end-to-end tests in a separate PR, along with the samples. We'll be writing integration tests using those samples (like in other clients).

PR #1760, opened 2022-03-19
fix: return errors from BatchCreateSession to dialect detection

### LINKED ISSUE #1759: Automatic dialect detection returns DEADLINE_EXCEEDED if session creation fails
The automatic dialect detection will return a `DEADLINE_EXCEEDED` error if the session pool failed to create any sessions. This for example happens if the user has requested a `DatabaseClient` for a database that does not exist, or that the user is not permitted to see. This is caused by the fact that the automatic dialect detection query is never triggered, as there is never any sessions being added to the pool.

olavloite:
Any error that was returned by BatchCreateSessions should also be set as
the result of the auto-detect dialect query, as the query will never be
able to return a result for the dialect when session creation fails.

Fixes #1759

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
@@ -2210,6 +2210,9 @@ class SessionPool {
           break;
         }
       }
+      if (!dialect.isDone()) {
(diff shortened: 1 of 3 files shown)


thiagotnunes (approved):

ansh0l (approved):
LGTM

PR #1777, opened 2022-03-23
feat: support PG show transaction isolation level

olavloite:
Adds support for the PostgreSQL statement SHOW TRANSACTION ISOLATION
LEVEL. This statement is used by the PG JDBC driver to get the isolation
level of a connection. The statement will always return the static
string 'serializable', as that is the only isolation level Spangres
supports.


thiagotnunes (approved):

olavloite:
@ansh0l Do you have any idea why so many of the checks seem to be hanging?

elefeint:
fwiw, github had issues yesterday around 10am EST. Try restarting the actions manually.

ansh0l (approved):
LGTM. 

Bulk of LOC is in the .sql file - where the new file looks largely good

PR #1778, opened 2022-03-24
feat: Copy Backup Support

asthamohta:


google-cloud-spanner/src/main/java/com/google/cloud/spanner/BackupInfo.java
```diff
@@ -84,6 +85,20 @@
 
     /** Builds the backup from this builder. */
     public abstract Backup build();
+
+    /**
+     * Output Only.
+     *
+     * <p>Returns the max allowed expiration time of the backup, with microseconds granularity.
+     */
+    protected abstract Builder setMaxExpireTime(Timestamp maxExpireTime);
```
thiagotnunes, line 94: This class is public, so you will have to remove the abstract and provide a default implementation. Otherwise, this would be a breaking change (which requires a major release).

thiagotnunes, line 94:
```suggestion
    protected Builder setMaxExpireTime(Timestamp maxExpireTime) {
      throw new UnsupportedOperationException("Unimplemented");
    }
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BackupInfo.java
```diff
@@ -84,6 +85,20 @@
...
+
+    /**
+     * Output Only.
+     *
+     * <p>Returns the names of the destination backups being created by copying this source backup.
+     */
+    protected abstract Builder setReferencingBackup(ProtocolStringList referencingBackup);
```
thiagotnunes, line 101: This class is public, so you will have to remove the abstract and provide a default implementation. Otherwise, this would be a breaking change (which requires a major release).

thiagotnunes, line 101:
```suggestion
    protected Builder setReferencingBackup(List<String> referencingBackup) {
      throw new UnsupportedOperationException("Unimplemented");
    }
```

thiagotnunes, line 101: We should not use the protobuf specific types here. Instead of `ProtocolStringList`, could you use `List<String>`?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BackupInfo.java
```diff
@@ -96,6 +111,8 @@
     private BackupEncryptionConfig encryptionConfig;
     private EncryptionInfo encryptionInfo;
     private com.google.spanner.admin.database.v1.Backup proto;
+    private Timestamp maxExpireTime;
+    private ProtocolStringList referencingBackup;
```
thiagotnunes, line 115: We should not use the protobuf specific types here. Instead of `ProtocolStringList`, could you use `List<String>`?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BackupInfo.java
```diff
@@ -184,6 +215,8 @@ Builder setProto(@Nullable com.google.spanner.admin.database.v1.Backup proto) {
   private final BackupEncryptionConfig encryptionConfig;
   private final EncryptionInfo encryptionInfo;
   private final com.google.spanner.admin.database.v1.Backup proto;
+  private final Timestamp maxExpireTime;
+  private final ProtocolStringList referencingBackup;
```
thiagotnunes, line 219: We should not use the protobuf specific types here. Instead of ProtocolStringList, could you use List<String>?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseAdminClient.java
```diff
@@ -178,6 +179,70 @@
...
+  /**
+   * Creates a copy of backup from an existing backup in a Cloud Spanner instance.
+   *
+   * <p>Example to copy a backup.
+   *
+   * <pre>{@code
+   * String instanceId                  = "my_instance_id";
+   * String sourceBackupId              = "source_backup_id";
+   * String destinationBackupId         = "destination_backup_id";
+   * Timestamp expireTime               = Timestamp.ofTimeMicroseconds(micros);
```
thiagotnunes, line 191: nit: we can remove the spacing between variable and `=` here

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseAdminClientImpl.java
```diff
@@ -165,6 +166,49 @@ private static String randomOperationId() {
...
+
+    return new OperationFutureImpl<>(
+        rawOperationFuture.getPollingFuture(),
+        rawOperationFuture.getInitialFuture(),
+        snapshot -> {
+          com.google.spanner.admin.database.v1.Backup proto =
+              ProtoOperationTransformers.ResponseTransformer.create(
+                      com.google.spanner.admin.database.v1.Backup.class)
+                  .apply(snapshot);
+          return Backup.fromProto(
```
thiagotnunes, line 198: Did we update the `Backup.fromProto` to populate the new fields?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/BackupTest.java
```diff
@@ -312,4 +314,19 @@ private Backup createBackup() {
...
+    com.google.spanner.admin.database.v1.Backup proto =
+        com.google.spanner.admin.database.v1.Backup.newBuilder()
+            .setName(NAME)
+            .setDatabase(DB)
+            .setExpireTime(
+                com.google.protobuf.Timestamp.newBuilder().setSeconds(1000L).setNanos(1000).build())
+            .setVersionTime(
+                com.google.protobuf.Timestamp.newBuilder().setSeconds(2000L).setNanos(2000).build())
+            .setEncryptionInfo(ENCRYPTION_INFO)
+            .setState(com.google.spanner.admin.database.v1.Backup.State.CREATING)
```
thiagotnunes, line 328: Could you add the new fields to the test as well?

samples/pom.xml
```diff
@@ -33,6 +33,24 @@
     <module>snippets</module>
   </modules>
 
+  <dependencies>
```
thiagotnunes, line 36: Just for completion, as we already discussed this, the samples changes should be removed from this PR

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseAdminClient.java
```diff
@@ -178,6 +179,70 @@
...
+   * }</pre>
+   *
+   * @param instanceId the id of the instance where the source backup is located and where the new
+   *     backup will be created.
+   * @param sourceBackupId the source backup id.
+   * @param destinationBackupId the id of the backup which will be created. It must conform to the
+   *     regular expression [a-z][a-z0-9_\-]*[a-z0-9] and be between 2 and 60 characters in length.
+   * @param expireTime the time that the new backup will automatically expire.
+   */
+  OperationFuture<Backup, CopyBackupMetadata> copyBackup(
```
thiagotnunes, line 208: This would be a breaking change. We need to add it with a default implementation

```
default OperationFuture<Backup, CopyBackupMetadata> copyBackup(String instanceId, String sourceBackupId, String destinationBackupId, Timestamp expireTime) throws SpannerException {
}

File: google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseAdminClient.java [hunk 10]
```diff
@@ -178,6 +179,70 @@
...
+   *     .build();
+   *
+   * OperationFuture<Backup, CopyBackupMetadata> op = dbAdminClient.copyBackUp(sourceBackupId, destinationBackup);
+   * Backup copiedBackup = op.get();
+   * }</pre>
+   *
+   * @param sourceBackupId the backup to be copied
+   * @param destinationBackup the new backup to create
+   */
+  OperationFuture<Backup, CopyBackupMetadata> copyBackup(
```
thiagotnunes, line 243: This would be a breaking change. We need to add it with a default implementation

```
default OperationFuture<Backup, CopyBackupMetadata> copyBackup(BackupId sourceBackupId, Backup destinationBackup) throws SpannerException {
}
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerRpc.java
```diff
@@ -230,6 +232,19 @@ public String getNextPageToken() {
...
+  /**
+   * Creates a copy backup from the source backup specified.
+   *
+   * @param destinationBackup the backup to create. The instance, database, and expireTime fields of
+   *     the backup must be filled. It may also optionally have an encryption config set. If no
+   *     encryption config has been set, the new backup will use the same encryption config as the
+   *     source backup.
+   * @return the operation that monitors the backup creation.
+   */
+  OperationFuture<Backup, CopyBackupMetadata> copyBackUp(
```
thiagotnunes, line 244: This would be a breaking change. Change to:

```
default OperationFuture<Backup, CopyBackupMetadata> copyBackup(BackupId sourceBackupId, com.google.cloud.spanner.Backup destinationBackup) throws SpannerException {
}
```

thiagotnunes (commented):
Be careful when accepting my code suggestions, as you need to change the commit title, otherwise the conventionalcommits check will fail. Feel free to, instead of accepting the suggestions, implementing the same in your local machine.

thiagotnunes (changes requested):
We need to resolve the breaking changes, populate the new fields when reading the proto and remove the samples

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseAdminClientImpl.java
asthamohta: @olavloite can you help answer this?

rajatbhatta: This is done. I see the following changes in Backup.fromProto in this PR.

```
+       .setMaxExpireTime(Timestamp.fromProto(proto.getMaxExpireTime()))
+       .setReferencingBackup(proto.getReferencingBackupsList())
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BackupInfo.java
```diff
@@ -163,6 +186,18 @@ Builder setProto(@Nullable com.google.spanner.admin.database.v1.Backup proto) {
       this.proto = proto;
       return this;
     }
+
+    @Override
+    public Builder setMaxExpireTime(Timestamp maxExpireTime) {
```
olavloite, line 191: Shouldn't this also be `protected`?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BackupInfo.java
```diff
@@ -163,6 +186,18 @@ Builder setProto(@Nullable com.google.spanner.admin.database.v1.Backup proto) {
...
+      this.maxExpireTime = Preconditions.checkNotNull(maxExpireTime);
+      return this;
+    }
+
+    @Override
+    public Builder addAllReferencingBackups(List<String> referencingBackup) {
```
olavloite, line 197: Shouldn't this also be protected?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseAdminClient.java
```diff
@@ -22,11 +22,7 @@
 import com.google.cloud.Timestamp;
 import com.google.cloud.spanner.Options.ListOption;
 import com.google.longrunning.Operation;
-import com.google.spanner.admin.database.v1.CreateBackupMetadata;
-import com.google.spanner.admin.database.v1.CreateDatabaseMetadata;
-import com.google.spanner.admin.database.v1.CreateDatabaseRequest;
-import com.google.spanner.admin.database.v1.RestoreDatabaseMetadata;
-import com.google.spanner.admin.database.v1.UpdateDatabaseDdlMetadata;
+import com.google.spanner.admin.database.v1.*;
```
olavloite, line 25: Wildcard imports are not allowed according to Google style: https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseAdminClient.java
```diff
@@ -178,6 +174,73 @@
...
+   * OperationFuture<Backup, CopyBackupMetadata> op = dbAdminClient
+   *     .copyBackup(
+   *         instanceId,
+   *         sourceBackupId,
+   *         destinationBackupId,
+   *         expireTime);
+   * Backup backup = op.get();
+   * }</pre>
+   *
+   * @param instanceId the id of the instance where the source backup is located and where the new
```
olavloite, line 196: This means that the method won't support copying a backup from one instance to another. Is that intentional? I thought that an important part of this feature was to allow copying a backup to a different instance.

olavloite: Ah, I see that the next method does allow copying from one instance to another. Maybe worth a small note in the comments that if a user wants to do that, then they need to use the method that takes a `BackupId` and `Backup` as arguments.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerRpc.java
```diff
@@ -32,31 +33,12 @@
 import com.google.longrunning.Operation;
 import com.google.protobuf.Empty;
 import com.google.protobuf.FieldMask;
-import com.google.spanner.admin.database.v1.Backup;
-import com.google.spanner.admin.database.v1.CreateBackupMetadata;
-import com.google.spanner.admin.database.v1.CreateDatabaseMetadata;
-import com.google.spanner.admin.database.v1.Database;
-import com.google.spanner.admin.database.v1.RestoreDatabaseMetadata;
-import com.google.spanner.admin.database.v1.UpdateDatabaseDdlMetadata;
+import com.google.spanner.admin.database.v1.*;
```
olavloite, line 36: Also here: No wildcard imports

(Strange that the formatter does not detect this)

olavloite (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BackupInfo.java
asthamohta: Yes. Changing

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseAdminClient.java
asthamohta: Okay sure I will add that. Thanks

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BackupInfo.java
```diff
@@ -84,6 +85,24 @@
...
+    protected Builder setMaxExpireTime(Timestamp maxExpireTime) {
+      throw new UnsupportedOperationException("Unimplemented");
+    }
+
+    /**
+     * Output Only.
+     *
+     * <p>Returns the names of the destination backups being created by copying this source backup.
+     */
+    protected Builder addAllReferencingBackups(List<String> referencingBackup) {
```
rajatbhatta, line 103: referencingBackup -> referencingBackups

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BackupInfo.java
```diff
@@ -96,6 +115,8 @@
     private BackupEncryptionConfig encryptionConfig;
     private EncryptionInfo encryptionInfo;
     private com.google.spanner.admin.database.v1.Backup proto;
+    private Timestamp maxExpireTime;
+    private List<String> referencingBackup;
```
rajatbhatta, line 119: referencingBackup -> referencingBackups

rajatbhatta: Please rename it at all other places as well.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerRpc.java
asthamohta: Actually the formatter changes it to this. In case of more than 3 imports from a library it imports the whole library

rajatbhatta (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerRpc.java
```diff
@@ -230,6 +212,20 @@ public String getNextPageToken() {
...
+  /**
+   * Creates a copy backup from the source backup specified.
+   *
+   * @param destinationBackup the backup to create. The instance, database, and expireTime fields of
+   *     the backup must be filled. It may also optionally have an encryption config set. If no
+   *     encryption config has been set, the new backup will use the same encryption config as the
+   *     source backup.
+   * @return the operation that monitors the backup creation.
+   */
+  default OperationFuture<Backup, CopyBackupMetadata> copyBackUp(
```
thiagotnunes, line 224: Please use `copyBackup` (lower case `up`)

google-cloud-spanner/src/test/java/com/google/cloud/spanner/BackupTest.java
```diff
@@ -309,6 +311,27 @@ private Backup createBackup() {
...
+            .setMaxExpireTime(
+                com.google.protobuf.Timestamp.newBuilder().setSeconds(3000L).setNanos(3000).build())
+            .addAllReferencingBackups(Arrays.asList(REFERENCING_BACKUP_NAME))
```
thiagotnunes, line 334: nit: Collections.singletonList(REFERENCING_BACKUP_NAME)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Backup.java
```diff
@@ -183,6 +183,8 @@ static Backup fromProto(
         .setDatabase(DatabaseId.of(proto.getDatabase()))
         .setEncryptionInfo(EncryptionInfo.fromProtoOrNull(proto.getEncryptionInfo()))
         .setProto(proto)
+        .setMaxExpireTime(Timestamp.fromProto(proto.getMaxExpireTime()))
```
thiagotnunes, line 186: Did we test with `null` max expire time and `null` referencingBackupsList?

google-cloud-spanner/clirr-ignored-differences.xml
```diff
@@ -35,4 +35,29 @@
...
+  </difference>
+  <difference>
+    <differenceType>7012</differenceType>
+    <className>com/google/cloud/spanner/DatabaseAdminClient</className>
+    <method>com.google.api.gax.longrunning.OperationFuture copyBackup(com.google.cloud.spanner.BackupId, com.google.cloud.spanner.Backup)</method>
+  </difference>
+  <difference>
+    <differenceType>7012</differenceType>
+    <className>com/google/cloud/spanner/spi/v1/SpannerRpc</className>
+    <method>com.google.api.gax.longrunning.OperationFuture copyBackUp(com.google.cloud.spanner.BackupId, com.google.cloud.spanner.Backup)</method>
```
thiagotnunes, line 61: You will have to change this one as well to `copyBackup` (lowercase `up`)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerRpc.java
olavloite: You should run the following formatter from the command line for the client library:

`mvn com.coveo:fmt-maven-plugin:format`

thiagotnunes (commented):
I will remove my request changes so you are not blocked by my review, but please address the method name change

thiagotnunes (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Backup.java
asthamohta: We won't get a null value over here. The addAll function checks for non null values-> https://github.com/protocolbuffers/protobuf/blob/master/java/core/src/main/java/com/google/protobuf/AbstractMessageLite.java#L408

PR #1840, opened 2022-04-20
feat: changes to support data, timestamp and arrays in IT tests

asthamohta:


google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITQueryTest.java
```diff
@@ -771,7 +767,7 @@ public void bindNumericArray_doesNotPreservePrecision() {
 
     Struct row =
         execute(
-            Statement.newBuilder("SELECT @v").bind("v").toNumericArray(asList(b1, b2, null)),
+            Statement.newBuilder(selectValueQuery).bind("p1").toNumericArray(asList(b1, b2, null)),
```
olavloite, line 770: This test case seem to be disabled for PostgreSQL still. Is that intentional?
The same also seems to be the case for other numeric array tests.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITResultSetGetValue.java
```diff
@@ -429,6 +522,59 @@ public void testReadNullValuesInArrays() {
     }
   }
 
+  @Test
+  public void testReadNullValuesInArraysPostgreSQL() {
```
olavloite, line 526: Why do we need separate tests for GoogleSQL and PostgreSQL in this case? They seem to be equal. Or am I missing something?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITResultSetGetValue.java
```diff
@@ -615,9 +793,16 @@ public void testReadFloat64LiteralsGoogleStandardSQL() {
   public void testReadFloat64LiteralsPostgreSQL() {
     Assume.assumeTrue(dialect.dialect == Dialect.POSTGRESQL);
     try (ResultSet resultSet =
-        databaseClient.singleUse().executeQuery(Statement.of("SELECT 10.0 AS float64"))) {
+        databaseClient
```
olavloite, line 796: Why does this need a separate test case?

olavloite (approved):
Looks good to me, with a couple of minor questions.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITResultSetGetValue.java
asthamohta: structs and json are not supported and column names are different as PG requires explicit quotes for case

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITResultSetGetValue.java
asthamohta: same as above

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITQueryTest.java
asthamohta: missed those, adding it

ansh0l (approved):
lgtm

PR #1845, opened 2022-04-21
feat: support CREATE DATABASE in Connection API

### LINKED ISSUE #1844: feat: Support CREATE DATABASE statements in the Connection API
Add support for the `CREATE DATABASE` statement in the Connection API. This will enable support for this statement in:
- JDBC connections (both PostgreSQL and GoogleSQL)
- PGAdapter

olavloite:
Adds support for the CREATE DATABASE statement in the Connection API.
The statement can only be used with a `SingleUseTransaction` (i.e. auto
commit mode). It is not supported in DDL batches. The database that is
created will have the same dialect as the current database that the user
is connected to.

Fixes #1844 

Towards b/229420558

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseAdminClient.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseAdminClient.java
@@ -71,6 +71,39 @@ public interface DatabaseAdminClient {
       String instanceId, String databaseId, Iterable<String> statements) throws SpannerException;
 
   /**
+   * Creates a new database in a Cloud Spanner instance with the given {@link Dialect}.
+   *
+   * <p>Example to create database.
+   *
+   * <pre>{@code
+   * String instanceId = "my_instance_id";
+   * String createDatabaseStatement = "CREATE DATABASE \"my-database\"";
+   * Operation<Database, CreateDatabaseMetadata> op = dbAdminClient
+   *     .createDatabase(
+   *         instanceId,
+   *         createDatabaseStatement,
+   *         Dialect.POSTGRESQL
+   *         Collections.emptyList());
+   * Database db = op.waitFor().getResult();
+   * }</pre>
+   *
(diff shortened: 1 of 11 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseAdminClient.java
```diff
@@ -70,6 +70,39 @@ public interface DatabaseAdminClient {
...
+   *     TABLE MyTable ( ... )}. This should not include {@code CREATE DATABASE} statement.
+   */
+  default OperationFuture<Database, CreateDatabaseMetadata> createDatabase(
+      String instanceId,
+      String createDatabaseStatement,
+      Dialect dialect,
+      Iterable<String> statements)
+      throws SpannerException {
+    throw new UnsupportedOperationException("Unimplemented");
+  }
```
ansh0l, line 104: We seem to be throwing `SpannerException` elsewhere with `createDatabase` in the interface. Why is it an `UnsupportedOperationException("Unimplemented");` for this one?

ansh0l (approved):
LGTM with a query, please take a look

rahul2393 (approved):

rajatbhatta (approved):
LGTM

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseAdminClient.java
olavloite: This is only to make this a non-breaking change. We add this when we add a new method to a public interface. This prevents compile errors for downstream users that might implement this interface in a test, as the method has a default implementation. See for example this example from the `DatabaseClient` where we did the same thing: https://github.com/googleapis/java-spanner/blob/049635d4bc3210bd9ce41444f17c8b9d67af969a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java#L36

PR #1847, opened 2022-04-22
chore: make DatabaseClient accessible in the Connection API

olavloite:
Makes the underlying `DatabaseClient` accessible in the Connection API so this can be used by drivers that need direct access to it.

See https://github.com/GoogleCloudPlatform/pgadapter/blob/8f0d47785e04dc9db6ee275074777849f006d797/src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationWriter.java#L439


ansh0l (approved):
LGTM

PR #1849, opened 2022-04-22
chore: include client statement type in parsed statement

olavloite:
Include the ClientSideStatementType in ParsedStatement. This enables
clients to directly check what type of statement it is, which removes
the need for textual checks.

This will for example remove these types of checks: https://github.com/GoogleCloudPlatform/pgadapter/blob/2a3f2eb7cd16cc3c71ffc673544ca9e2baaba21c/src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java#L354


google-cloud-spanner/src/main/resources/com/google/cloud/spanner/connection/ClientSideStatements.json
```diff
@@ -5,6 +5,7 @@
 	  "name": "SHOW VARIABLE AUTOCOMMIT",
 	  "executorName": "ClientSideStatementNoParamExecutor",
 	  "resultType": "RESULT_SET",
+		"statementType": "SHOW_AUTOCOMMIT",
```
ansh0l, line 8: nit: we seem to have added extra indentation to all modified lines in this file, please check.

ansh0l (approved):
LGTM with minor nit

google-cloud-spanner/src/main/resources/com/google/cloud/spanner/connection/ClientSideStatements.json
olavloite: Hmmm... That file needs a small cleanup. It uses both tabs AND spaces for indentation (first indentation is one tab, then second indentation is two spaces).
I've made this change consistent with the current format of the file. Let's do the cleanup in a separate PR.

PR #1850, opened 2022-04-22
chore: remove gsql parsing escape

olavloite:
Removes the `/*GSQL*/` escape hatch in the client side parsing, as it is no longer needed.

Fixes b/223655465


ansh0l (approved):
LGTM. 

@olavloite : IIRC, there was some gsql parsing removal in https://github.com/googleapis/java-spanner-jdbc/pull/759/files. Was it for a different purpose?

olavloite:
> @olavloite : IIRC, there was some gsql parsing removal in https://github.com/googleapis/java-spanner-jdbc/pull/759/files. Was it for a different purpose?

It was basically for the same purpose: The `GSQL` escape hatch in the backend would be removed, so we had to remove it from the JDBC driver where it was actively used. In the client library the option was also there, but it was (AFAICT) never used, so the urgency to remove it right away wasn't really there.

PR #1852, opened 2022-04-25
feat: Add support for Explain feature

gauravsnj:
Adds the support for the EXPLAIN, which will ultimately be utilised by PG Adapter


google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorTest.java
```diff
@@ -237,4 +251,290 @@ public void testStatementSetPgTransactionModeNoOp() {
     verify(connection, never()).setTransactionMode(TransactionMode.READ_ONLY_TRANSACTION);
     verify(connection, never()).setTransactionMode(TransactionMode.READ_WRITE_TRANSACTION);
   }
+
+  private ResultSet getMockResultSetForAnalyseQuery(){
```
olavloite, line 255: Use `Mockito` to create a mock for this instead. See here for an example how to do that: https://github.com/googleapis/java-spanner/blob/12a0acb02785987d8442c03a45b8b04a068375d7/google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java#L196

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorTest.java
```diff
@@ -237,4 +251,290 @@ public void testStatementSetPgTransactionModeNoOp() {
...
+      }
+    };
+    return rs;
+
+  }
+  @Test
+  public void testStatementExplain(){
+    String sql = "verbose select * from table";
+    try {
+      ResultSet rs = subject.statementExplain(sql);
```
olavloite, line 523: Use `Assert.assertThrows` to verify that a method invocation throws a specific exception. See here for an example: https://github.com/googleapis/java-spanner/blob/5f156f2efdb4726679766b385d500a030c24e477/google-cloud-spanner/src/test/java/com/google/cloud/spanner/DialectTest.java#L93

google-cloud-spanner/src/main/resources/com/google/cloud/spanner/connection/PG_ClientSideStatements.json
```diff
@@ -133,6 +133,14 @@
       "method": "statementShowTransactionIsolationLevel",
       "exampleStatements": ["show transaction isolation level","show variable transaction isolation level"]
     },
+    {
+      "name": "EXPLAIN '<sql>'",
+      "executorName": "ClientSideStatementNoParamExecutor",
```
olavloite, line 138: This won't work, as [the method that you have defined](https://github.com/googleapis/java-spanner/blob/242e35142357ad55d27444437320b37d4107a560/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java#L458) has a parameter; the sql string that is to be explained is the input parameter in this case. You have two options for this:
1. You could reuse the `ClientSideStatementSetExecutor` for this. That executor is normally used for statements like `SET <variable> = <value>` where the `value` is the input parameter. In this case, the input parameter would be the entire sql string that follows `EXPLAIN`.
2. You could create a custom executor specifically for this method. See `ClientSideStatementPgBeginExecutor` for an example for that.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java
```diff
@@ -442,4 +453,29 @@ public StatementResult statementShowRPCPriority() {
   public StatementResult statementShowTransactionIsolationLevel() {
     return resultSet("transaction_isolation", "serializable", SHOW_TRANSACTION_ISOLATION_LEVEL);
   }
+
+  @Override
+  public ResultSet statementExplain(String sql){
+
+    sql = sql.trim();
+    String firstString = sql.split(" ")[0].toLowerCase();
```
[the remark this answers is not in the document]
gauravsnj: Thanks for suggesting these improvements. I've added them in the code

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java
```diff
@@ -442,4 +453,29 @@ public StatementResult statementShowRPCPriority() {
...
+
+    if(explainOptions.contains(firstString)){
+      throw SpannerExceptionFactory.newSpannerException(
+          ErrorCode.UNIMPLEMENTED, String.format("%s is not implemented yet", firstString));
+    }
+
+    if(firstString.equals("analyze") || firstString.equals("analyse")) {
+      sql = sql.split(" ",2)[1];
```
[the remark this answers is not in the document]
gauravsnj: Done

google-cloud-spanner/src/main/resources/com/google/cloud/spanner/connection/PG_ClientSideStatements.json
gauravsnj: I think creating a new executor class will be better as ClientSideStatementSetExecutor is for specific type of commands and EXPLAIN clearly doesn't fit in that type. What do you say?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorTest.java
gauravsnj: Done

google-cloud-spanner/src/main/resources/com/google/cloud/spanner/connection/PG_ClientSideStatements.json
olavloite: Yes, that would be the neatest solution for this.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ClientSideStatementValueConverters.java
```diff
@@ -342,4 +342,17 @@ public Priority convert(String value) {
...
+  static class ExplainCommandConverter implements ClientSideStatementValueConverter<String> {
+
+    @Override
+    public Class<String> getParameterClass() {
+      return String.class;
+    }
+
+    @Override
+    public String convert(String value) {
+      return value.split(" +", 2)[1].toLowerCase();
```
olavloite, line 355: A couple of points regarding this:
1. I think it should be programmed a little more defensively, so check that the array actually contains at least 2 items before trying to get the element at index 1, so we know that we won't run into any `ArrayIndexOutOfBounds` exceptions.
2. It should probably be `value.split("\\s+")`. That ensures that the statement is also correctly handled if there is a line feed or a tab between `EXPLAIN` and the rest of the statement. Please also add tests for that.
3. Why the `.toLowerCase()`? I don't think we need (or should) call that on the entire sql string. If we need it to determine whether the first token in the SQL string is one of the `explain` options, then we should only call `toLowerCase()` for that part. Otherwise, we will be sending a modified SQL string to the backend.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java
```diff
@@ -83,6 +89,11 @@
  * calls are then forwarded into a {@link Connection}.
  */
 class ConnectionStatementExecutorImpl implements ConnectionStatementExecutor {
+
+  private final Set<String> explainOptions =
```
olavloite, line 93: nit:
```suggestion
  private static final Set<String> EXPLAIN_OPTIONS =
```

Also, we should consider moving this set to `ClientSideStatementExplainExecutor.java`, as it is specific for `explain`, and not something that is generically needed for all client side statements.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java
```diff
@@ -442,4 +453,26 @@ public StatementResult statementShowRPCPriority() {
   public StatementResult statementShowTransactionIsolationLevel() {
     return resultSet("transaction_isolation", "serializable", SHOW_TRANSACTION_ISOLATION_LEVEL);
   }
+
+  @Override
+  public ResultSet statementExplain(String sql) {
+
+    sql = sql.trim();
+
+    String[] arr = sql.split(" +", 2);
```
olavloite, line 462: Here also: 
```suggestion
    String[] arr = sql.split("\\s+", 2);
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java
```diff
@@ -442,4 +453,26 @@ public StatementResult statementShowRPCPriority() {
   public StatementResult statementShowTransactionIsolationLevel() {
     return resultSet("transaction_isolation", "serializable", SHOW_TRANSACTION_ISOLATION_LEVEL);
   }
+
+  @Override
+  public ResultSet statementExplain(String sql) {
+
+    sql = sql.trim();
+
```
olavloite, line 461: Remove

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java
```diff
@@ -442,4 +453,26 @@ public StatementResult statementShowRPCPriority() {
...
+    String[] arr = sql.split(" +", 2);
+
+    String option = arr[0];
+    String statementToBeExplained = arr[1];
```
olavloite, line 465: Check to ensure that we did not just receive something like `explain foo`, which would cause an `ArrayOutOfBoundsException` here. Also add tests for that kind of corner cases.

olavloite (commented):
This is looking good! Please add more tests for corner cases like `explain foo`, `explain`, `explain\nselect * from foo` etc.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ClientSideStatementExplainExecutor.java
```diff
@@ -0,0 +1,70 @@
...
+    Matcher matcher = statement.getPattern().matcher(sql);
+    if (matcher.find() && matcher.groupCount() >= 1) {
+      String value = matcher.group(0);
+      if (value != null) {
+        String res = converter.convert(value.trim());
+        if (res != null) {
+          return res;
+        }
+        throw SpannerExceptionFactory.newSpannerException(
+            ErrorCode.INVALID_ARGUMENT, String.format("Unknown transaction mode: %s", value));
```
ansh0l, line 65: Since this is the Explain method, is `Unknown transaction mode` in ErrorCode the relevant message?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ClientSideStatementExplainExecutor.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.regex.Matcher;
+
+/** Specific executor for the EXPLAIN statement for PostgreSQL. */
+class ClientSideStatementExplainExecutor implements ClientSideStatementExecutor {
+  private final ClientSideStatementImpl statement;
+  private final Method method;
+  private final ExplainCommandConverter converter;
+  public static final Set<String> EXPLAIN_OPTIONS =
+      ImmutableSet.of(
+          "verbose", "costs", "settings", "buffers", "wal", "timing", "summary", "format");
```
ansh0l, line 35: From https://www.postgresql.org/docs/current/sql-explain.html, it looks like `EXPLAIN (FORMAT JSON) SELECT * FROM foo;` is a possible query, but I don't think its getting covered in your implementation.

Please check

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ExplainCommandConverterTest.java
```diff
@@ -0,0 +1,58 @@
+/*
+ * Copyright 2019 Google LLC
```
ansh0l, line 2: `2022` since this is a new file.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ClientSideStatementValueConverters.java
```diff
@@ -342,4 +342,19 @@ public Priority convert(String value) {
...
+  static class ExplainCommandConverter implements ClientSideStatementValueConverter<String> {
+
+    @Override
+    public Class<String> getParameterClass() {
+      return String.class;
+    }
+
+    @Override
+    public String convert(String value) {
+      String arr[] = value.split("\\s+", 2);
```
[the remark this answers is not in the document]
ansh0l: Marking resolved based on https://www.geeksforgeeks.org/split-string-java-examples/

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ClientSideStatementExplainExecutor.java
gauravsnj: Changed the error message

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorTest.java
```diff
@@ -237,4 +241,101 @@ public void testStatementSetPgTransactionModeNoOp() {
...
+
+
+  }
+
+  @Test
+  public void testStatementExplain() {
+
+    ExplainCommandConverter converter = new ExplainCommandConverter();
+
+    /*Test for proper EXPLAIN*/
```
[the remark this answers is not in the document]
gauravsnj: Yes it means valid explain.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ExplainCommandConverterTest.java
gauravsnj: Changed

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ClientSideStatementExplainExecutor.java
gauravsnj: Thanks for pointing this out. I've added the code for handling this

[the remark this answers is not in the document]
gauravsnj: Yes the parsing is being done before checking. Please check [548](https://github.com/googleapis/java-spanner/blob/9f084bae78c99db01c3ef009cbd40325667405b2/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java#L548) and [512](https://github.com/googleapis/java-spanner/blob/9f084bae78c99db01c3ef009cbd40325667405b2/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java#L512)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java
```diff
@@ -442,4 +450,112 @@ public StatementResult statementShowRPCPriority() {
...
+    throw SpannerExceptionFactory.newSpannerException(ErrorCode.UNAVAILABLE, String.format("Couldn't fetch stats for %s",sql));
+
+  }
+
+  /*
+  * This method supports the EXPLAIN feature with ANALYSE option only.
+  * Any option other than ANALYSE is not supported currently.
+  * If the statement has combination of analyse with any other option
+  * like "explain (analyse, costs) select * from table"
+  * it will throw exception .
```
olavloite, line 488:
```suggestion
  * This method executes the given SQL string in either PLAN or PROFILE mode and returns
  * the query plan as a ResultSet containing a single String column with the query plan nodes.
  *
  * The only additional option that is supported is ANALYZE. The method will throw a SpannerException
  * if it is invoked with a statement that includes any other options.
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java
```diff
@@ -442,4 +450,112 @@ public StatementResult statementShowRPCPriority() {
...
+      throw SpannerExceptionFactory.newSpannerException(
+          ErrorCode.INVALID_ARGUMENT, String.format("Invalid String with Explain"));
+
+    }
+
+    if(sql.charAt(0) == '('){
+
+      int index = sql.indexOf(')');
+
+      String arr[] = sql.substring(1, index).split("\\s*,\\s*");
```
olavloite, line 504: Give `arr[]` a more logical name, maybe `options` or `optionExpression`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java
```diff
@@ -442,4 +450,112 @@ public StatementResult statementShowRPCPriority() {
...
+      boolean isAnalyse = false;
+      for(String probableOptions : arr){
```
olavloite, line 506:
```suggestion
      for(String option : arr){
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java
```diff
@@ -442,4 +450,112 @@ public StatementResult statementShowRPCPriority() {
...
+        String arr2[] = probableOptions.split("\\s+");
+        if(arr2.length >= 3){
+          isAnalyse = false;
+          break;
```
olavloite, line 510: I don't think this is correct. If we encounter a string that contains more than 2 words, we will assume that the options expression did not contain `ANALYZE` and that it for the rest was valid. From what I can see in the [PostgreSQL documentation](https://www.postgresql.org/docs/current/sql-explain.html), there should be no option expression that contains more than 2 words.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java
```diff
@@ -442,4 +450,112 @@ public StatementResult statementShowRPCPriority() {
...
+        else if(ClientSideStatementExplainExecutor.EXPLAIN_OPTIONS.contains(arr2[0].toLowerCase())){
+          throw SpannerExceptionFactory.newSpannerException(
+              ErrorCode.UNIMPLEMENTED, String.format("%s is not implemented yet", arr2[0]));
+        }
+        else if(arr2[0].equalsIgnoreCase("analyse") || arr2[0].equalsIgnoreCase("analyze")){
+          isAnalyse = true;
+        }
+        else{
+          isAnalyse = false;
+          break;
```
olavloite, line 521: Also here: I think that an unknown/invalid option expression should cause a failure, not be assumed to mean 'analyze is not in the options expression'.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java
```diff
@@ -442,4 +450,112 @@ public StatementResult statementShowRPCPriority() {
...
+        }
+
+        if(arr2.length == 2){
+          if(arr2[1].equalsIgnoreCase("false")){
```
olavloite, line 525: False may also be written as `0` and `off`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java
```diff
@@ -442,4 +450,112 @@ public StatementResult statementShowRPCPriority() {
...
+            isAnalyse = false;
+          }
+          else if(!arr2[1].equalsIgnoreCase("true")){
```
olavloite, line 528: True may also be written as `1` and `on`.

Any invalid boolean value should cause the method to throw an exception.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java
```diff
@@ -442,4 +450,112 @@ public StatementResult statementShowRPCPriority() {
...
+            isAnalyse = false;
+            break;
+          }
+        }
+
+      }
+
+      if(isAnalyse){
+        return executeStatement(sql.substring(index+1), QueryAnalyzeMode.PROFILE);
```
olavloite, line 537: An interesting question that we did not think of when designing this was: What do we do with the query results that are returned by `QueryAnalyzeMode.PROFILE`. Profile will return two things:
1. The query plan
2. The actual results of the query

I think that (for now) we should just discard the results.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java
```diff
@@ -442,4 +450,112 @@ public StatementResult statementShowRPCPriority() {
...
+  @Override
+  public StatementResult statementExplain(String sql) {
+
+    if(sql == null){
+      throw SpannerExceptionFactory.newSpannerException(
+          ErrorCode.INVALID_ARGUMENT, String.format("Invalid String with Explain"));
+
+    }
+
+    if(sql.charAt(0) == '('){
```
olavloite, line 500: A SQL statement could in theory start with a '('. If you try to execute `explain (select * from all_types limit 1);` on a PostgreSQL database, it will return the query plan of the query between the parens.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java
```diff
@@ -442,4 +450,112 @@ public StatementResult statementShowRPCPriority() {
...
+
+    }
+    else {
+      String[] arr = sql.split("\\s+", 2);
+
+      if (arr.length >= 2) {
+        String option = arr[0].toLowerCase();
+        String statementToBeExplained = arr[1];
+
+        if (ClientSideStatementExplainExecutor.EXPLAIN_OPTIONS.contains(option)) {
```
olavloite, line 551: Options can only be given inside parentheses, except for `analyze` and `verbose`, so:
* `explain select * from foo` is valid
* `explain analyze select * from foo` is valid
* `explain verbose select * from foo` is valid
* `explain costs select * from foo` is NOT valid
* `explain buffers select * from foo` is NOT valid

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorTest.java
```diff
@@ -237,4 +242,102 @@ public void testStatementSetPgTransactionModeNoOp() {
...
+    when(expectedRs.next()).thenReturn(true, false);
+    when(expectedRs.getColumnCount()).thenReturn(1);
+    when(expectedRs.getColumnIndex("query plan")).thenReturn(1);
+
+    when(connection.analyzeQuery(any(), any())).thenReturn(expectedRs);
+
+    ResultSet rs = subject.statementExplain(sql).getResultSet();
+
+    assertExplain(rs);
+    /*Test for EXPLAIN ANALYSE*/
```
olavloite, line 291: Split different test cases into different methods. That makes it easier to detect what fails if a test fails.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java
gauravsnj: resolved

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java
gauravsnj: If there will be an invalid boolean like "explain (analyse foo) select * from table", then it will send "(analyse foo) select * from table" as the query, which will eventually cause exception. Its just the error message will not be appropriate (it will be something like "(analyse foo) select * from table" is not a query).

If possible, can we take up the task of distinguishing the error message sometime after this initial PR is merged?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java
olavloite: Yes, that sounds good to me.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ExplainCommandConverterTest.java
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  @Test
+  public void testConvert() throws CompileException {
+    ExplainCommandConverter explainCommandConverter = new ExplainCommandConverter();
+    Assert.assertEquals(
+        "select * from table1", explainCommandConverter.convert("explain select * from table1"));
+    Assert.assertEquals(
+        "select    *   \t from table1",
+        explainCommandConverter.convert("explain \tselect    *   \t from table1"));
```
olavloite, line 45: I know it works, but for completeness: Could we have one test case with `explain` in upper or mixed case, instead of all lower case?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ExplainTest.java
```diff
@@ -0,0 +1,317 @@
...
+                .setStats(resultSetStats)
+                .build()));
+  }
+
+  @After
+  public void clearRequests() {
+    mockSpanner.clearRequests();
+  }
+
+  public void testExplain(String statement) {
```
olavloite, line 111: nit: I think this can be private

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ExplainTest.java
```diff
@@ -0,0 +1,317 @@
...
+    final Statement explainStatement = Statement.of(statement);
+
+    try (Connection connection = createConnection()) {
+
+      try (ResultSet resultSet = connection.execute(explainStatement).getResultSet()) {
+        int count = 0;
+
+        while (resultSet.next()) {
+          if (count == 1) {
+            Assert.assertEquals(true, false);
```
olavloite, line 122: nit: use `Assert.fail(String message)` instead.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ExplainTest.java
```diff
@@ -0,0 +1,317 @@
...
+    mockSpanner.clearRequests();
+    final Statement explainAnalyseStatement = Statement.of(statement);
+    try (Connection connection = createConnection()) {
+
+      try (ResultSet resultSet = connection.execute(explainAnalyseStatement).getResultSet()) {
+        int count = 0;
+
+        while (resultSet.next()) {
+          if (count == 1) {
+            Assert.assertEquals(true, false);
```
olavloite, line 154: nit: here also, use `Assert.fail`

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ExplainTest.java
```diff
@@ -0,0 +1,317 @@
...
+    statement = "    ExpLAin  (    " + EXPLAIN_STATEMENT_QUERY + "   ) ";
+    testExplain(statement);
+
+    statement = "    EXPLAIN  (    " + EXPLAIN_STATEMENT_QUERY + "   ) ";
+    testExplain(statement);
+  }
+
+  @Test
+  public void testValidExplainWithFalseAnalyse() {
+    String statement = "    explain (analyse false)      " + EXPLAIN_STATEMENT_QUERY;
```
olavloite, line 207: Here and elsewhere: PostgreSQL uses `analyze` instead of `analyse` in all documentation. I know both work, but I think it would be better if we also stuck to `analyze` (except for tests that verify that `analyse` also works).

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ExplainTest.java
```diff
@@ -0,0 +1,317 @@
...
+
+    statement =
+        "    EXPLAIN(analyse \n off  , analyse false , analyse  )(    "
+            + EXPLAIN_STATEMENT_QUERY
+            + " \t  ) ";
+    testExplainAnalyze(statement);
+  }
+
+  @Test
+  public void testInvalidExlain() {
```
olavloite, line 294: nit:
```suggestion
  public void testInvalidExplain() {
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/PartitionedDmlTransaction.java
```diff
@@ -95,7 +95,7 @@ long executeStreamingPartitionedUpdate(
               resumeToken = rs.getResumeToken();
             }
             if (rs.hasStats()) {
-              foundStats = true;
+              foundStats = rs.getStats().hasRowCountLowerBound();
```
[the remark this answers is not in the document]
gauravsnj: DatabaseClientImplTest.testExecutePartitionedDmlWithQuery test was failing without this line. I took this change from your PR (feat: support analyzeUpdate #1867 )

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ClientSideStatementValueConverters.java
```diff
@@ -342,4 +342,19 @@ public Priority convert(String value) {
...
+
+  static class ExplainCommandConverter implements ClientSideStatementValueConverter<String> {
+    @Override
+    public Class<String> getParameterClass() {
+      return String.class;
+    }
+
+    @Override
+    public String convert(String value) {
+      if (value.length() <= 7) {
```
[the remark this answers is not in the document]
gauravsnj: added

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java
```diff
@@ -442,4 +450,182 @@ public StatementResult statementShowRPCPriority() {
...
+      rs = ResultSets.forRows(Type.struct(StructField.of("QUERY PLAN", Type.string())), list);
+    }
+    return StatementResultImpl.of(rs);
+  }
+
+  private StatementResult executeStatement(String sql, QueryAnalyzeMode qam) {
+    Statement statement = Statement.newBuilder(sql).build();
+    ResultSet rs = getConnection().analyzeQuery(statement, qam);
+    while (rs.next()) {
+      continue;
```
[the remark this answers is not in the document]
gauravsnj: added

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java
```diff
@@ -442,4 +450,182 @@ public StatementResult statementShowRPCPriority() {
...
+                  StructField.of("QUERY PLAN", Type.string()),
+                  StructField.of("EXECUTION STATS", Type.string())),
+              list);
+    } else {
+      rs = ResultSets.forRows(Type.struct(StructField.of("QUERY PLAN", Type.string())), list);
+    }
+    return StatementResultImpl.of(rs);
+  }
+
+  private StatementResult executeStatement(String sql, QueryAnalyzeMode qam) {
```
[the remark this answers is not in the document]
gauravsnj: changed

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java
```diff
@@ -442,4 +450,182 @@ public StatementResult statementShowRPCPriority() {
...
+    Statement statement = Statement.newBuilder(sql).build();
+    ResultSet rs = getConnection().analyzeQuery(statement, qam);
```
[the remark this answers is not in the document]
gauravsnj: done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java
```diff
@@ -442,4 +450,182 @@ public StatementResult statementShowRPCPriority() {
...
+                .to(planNodeDescription)
+                .set("EXECUTION STATS")
+                .to(executionStats)
+                .build());
+      } else {
+        list.add(Struct.newBuilder().set("QUERY PLAN").to(planNodeDescription).build());
+      }
+    }
+
+    ResultSet rs;
```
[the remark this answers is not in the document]
gauravsnj: done

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ExplainCommandConverterTest.java
gauravsnj: added

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ExplainTest.java
gauravsnj: changed to private

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ExplainTest.java
gauravsnj: changed

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ExplainTest.java
gauravsnj: changed

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ExplainTest.java
gauravsnj: changed

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java
```diff
@@ -442,4 +450,182 @@ public StatementResult statementShowRPCPriority() {
...
+        list.add(
+            Struct.newBuilder()
+                .set("QUERY PLAN")
+                .to(planNodeDescription)
+                .set("EXECUTION STATS")
+                .to(executionStats)
+                .build());
+      } else {
+        list.add(Struct.newBuilder().set("QUERY PLAN").to(planNodeDescription).build());
+      }
```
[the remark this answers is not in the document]
gauravsnj: changed

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java
```diff
@@ -442,4 +450,182 @@ public StatementResult statementShowRPCPriority() {
...
+  private String removeParenthesisAndTrim(String sql) {
+    if (sql.charAt(0) == '(') {
+      sql = sql.substring(1, sql.length() - 1);
+    }
+    return sql.trim();
+  }
+
+  /*
+   * This method executes the given SQL string in either PLAN or PROFILE mode and returns
+   * the query plan as a ResultSet containing a single String column with the query plan nodes.
```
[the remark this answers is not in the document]
gauravsnj: changed

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java
```diff
@@ -442,4 +450,182 @@ public StatementResult statementShowRPCPriority() {
...
+
+    if (rs.getStats() != null && rs.getStats().getQueryPlan() != null) {
+      return getStatementResultFromQueryPlan(
+          rs.getStats().getQueryPlan(), qam.equals(QueryAnalyzeMode.PROFILE));
+    }
+    throw SpannerExceptionFactory.newSpannerException(
+        ErrorCode.FAILED_PRECONDITION, String.format("Couldn't fetch stats for %s", sql));
+  }
+
+  private String removeParenthesisAndTrim(String sql) {
```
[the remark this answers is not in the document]
gauravsnj: changed

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java
```diff
@@ -442,4 +450,182 @@ public StatementResult statementShowRPCPriority() {
...
+   */
+  @Override
+  public StatementResult statementExplain(String sql) {
+    if (sql == null) {
+      throw SpannerExceptionFactory.newSpannerException(
+          ErrorCode.INVALID_ARGUMENT, String.format("Invalid String with Explain"));
+    }
+
+    if (sql.charAt(0) == '(') {
+      int index = sql.indexOf(')');
```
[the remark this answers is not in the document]
gauravsnj: added

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ExplainCommandConverterTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+  @Parameters(name = "dialect = {0}")
+  public static Object[] data() {
+    return Dialect.values();
+  }
+
+  @Test
+  public void testConvert() throws CompileException {
+    ExplainCommandConverter explainCommandConverter = new ExplainCommandConverter();
+    Assert.assertEquals(
```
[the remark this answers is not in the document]
gauravsnj: done

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ExplainCommandConverterTest.java
```diff
@@ -0,0 +1,70 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameter;
+import org.junit.runners.Parameterized.Parameters;
+
+@RunWith(Parameterized.class)
+public class ExplainCommandConverterTest {
+  @Parameter public Dialect dialect;
```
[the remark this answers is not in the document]
gauravsnj: done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java
```diff
@@ -442,4 +450,182 @@ public StatementResult statementShowRPCPriority() {
   public StatementResult statementShowTransactionIsolationLevel() {
     return resultSet("transaction_isolation", "serializable", SHOW_TRANSACTION_ISOLATION_LEVEL);
   }
+
+  private String processQueryPlan(PlanNode planNode) {
+    String planNodeDescription = " : { ";
```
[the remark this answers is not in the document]
gauravsnj: done

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ExplainTest.java
gauravsnj: done

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ITAbstractSpannerTest.java
```diff
@@ -322,9 +322,9 @@ protected boolean tableExists(Connection connection, String table) {
     try (ResultSet rs =
         connection.executeQuery(
             Statement.newBuilder(
-                    "SELECT TABLE_NAME FROM INFORMATION_SCHEMA.TABLES WHERE UPPER(TABLE_NAME)=@table_name")
-                .bind("table_name")
-                .to(table.toUpperCase())
+                    String.format(
+                        "SELECT TABLE_NAME FROM INFORMATION_SCHEMA.TABLES WHERE UPPER(TABLE_NAME)=UPPER(\'%s\')",
```
olavloite, line 326: nit: We should never use string concatenation to add literals to a query, as it could be vulnerable to SQL injection attacks. It's not really relevant here, but it's good practice to always use statement parameters.

olavloite (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java
ansh0l: ack

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ClientSideStatementValueConverters.java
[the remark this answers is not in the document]
ansh0l: Consider a negative scenario like ```Explainselect 1```. Would it get converted to a successful query? If yes, we should add an additional check. I'm ok to have that as a follow up PR.

ansh0l (approved):
LGTM with minor nits.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ClientSideStatementValueConverters.java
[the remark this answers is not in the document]
gauravsnj: "Explainselect 1" will give error "Unknown statement: explainselect 1" because it will not pass the regex check. And even ideally, it should not be converted to "Explain select 1".

PR #1854, opened 2022-04-25
chore: support untyped values for statement parameters

olavloite:
Adds support for untyped values that can be used as statement
parameters. This is required for Spangres, as many native PG drivers
send (some) parameters without an explicit type, and expect the backend
to infer the type from the context. If we were to include a
(placeholder) type for such a parameter, the type inference on the
backend fails.

This is required for https://github.com/GoogleCloudPlatform/pgadapter/pull/125

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ If you are using Maven without BOM, add this to your dependencies:
 If you are using Gradle 5.x or later, add this to your dependencies
 
 ```Groovy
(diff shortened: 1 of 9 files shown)


olavloite:
> LGTM.
> 
> @olavloite : I'm just wondering if untyped is the right name, or should we call it a `genericPgType` or something different.

(See offline answer.)

PR #1869, opened 2022-05-03
feat: support CredentialsProvider in Connection API

olavloite:
Adds support for setting a `CredentialsProvider` instead of a
credentialsUrl in a connection string. The `CredentialsProvider` reference
must be a class name to a public class with a public no-arg constructor.
This option is available in the Connection API, which means that any
client that uses that API can directly benefit from it (this effectively
means the JDBC driver).

The option will only be available in the JDBC driver once the dependency on the client library in the JDBC driver has been updated to a version of the Java client library that contains this change.

Fixes b/231174409

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
@@ -588,7 +588,7 @@ public class ConnectionOptions {
     // Using credentials on a plain text connection is not allowed, so if the user has not specified
     // any credentials and is using a plain text connection, we should not try to get the
     // credentials from the environment, but default to NoCredentials.
-    if (builder.credentials == null
+    if (this.fixedCredentials == null
         && this.credentialsUrl == null
         && this.encodedCredentials == null
         && this.credentialsProvider == null
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
@@ -17,6 +17,7 @@
 package com.google.cloud.spanner.connection;
 
 import com.google.api.core.InternalApi;
+import com.google.api.gax.core.CredentialsProvider;
 import com.google.api.gax.rpc.TransportChannelProvider;
 import com.google.auth.Credentials;
 import com.google.auth.oauth2.AccessToken;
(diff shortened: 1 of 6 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
```diff
@@ -573,11 +591,21 @@ private ConnectionOptions(Builder builder) {
     if (builder.credentials == null
```
rajatbhatta, line 591: Shall we use `this.fixedCredentials` instead of `builder.credentials` here, to maintain consistency with the other credentials check?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/SpannerPool.java
```diff
@@ -155,10 +156,17 @@ static class SpannerPoolKey {
 
     @VisibleForTesting
     static SpannerPoolKey of(ConnectionOptions options) {
```
rajatbhatta, line 158:
```suggestion
    static SpannerPoolKey of(ConnectionOptions options) throws SpannerException {
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
```diff
@@ -537,22 +555,22 @@ private ConnectionOptions(Builder builder) {
...
-    // Check that not both credentials and an OAuth token have been specified.
+    // Check that at most one of credentials location, encoded credentials, credentials provider and
+    // OUAuth token has been specified in the connection URI.
     Preconditions.checkArgument(
-        (builder.credentials == null
-                && this.credentialsUrl == null
-                && this.encodedCredentials == null)
-            || this.oauthToken == null,
-        "Cannot specify both credentials and an OAuth token.");
+        Stream.of(
```
rajatbhatta, line 564: Why did we remove `builder.credentials` (or `this.fixedCredentials`) from the stream here, and move the fixedCredentials initializations to after this nonNull Preconditions check?

rajatbhatta (approved):
LGTM with a couple of small points/questions.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
olavloite: Good point, done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/SpannerPool.java
olavloite: I don't agree with that. This exception falls in the following category "__Runtime exceptions represent problems that are the result of a programming problem, and as such, the API client code cannot reasonably be expected to recover from them or to handle them in any way.__" There's nothing reasonable that we can do to catch and handle this exception in the client library, and it should naturally bubble up to the client application.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
olavloite: `builder.credentials` is not part of the public API and is not something users can set, so we don't need to check it here. (see https://github.com/googleapis/java-spanner/blob/2360e7e1bfd758fd773d52b43f31c20b41d77c97/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java#L495)

rajatbhatta: Thanks for clarifying!

rajatbhatta:
Merging manually, as automerge failed on this PR (even though all checks have passed).

PR #1877, opened 2022-05-14
fix: ignore errors during Connection.close()

olavloite:
A connection will automatically rollback any active transactions when
the connection is closed. If the rollback would throw an exception, the
close would fail and the connection would still be marked as open. This
fix adds a simple ignore error handler when rolling back a transaction
when the connection is closed.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
@@ -294,12 +294,16 @@ class ConnectionImpl implements Connection {
     if (!isClosed()) {
       List<ApiFuture<Void>> futures = new ArrayList<>();
       if (isTransactionStarted()) {
-        futures.add(rollbackAsync());
+        try {
+          futures.add(rollbackAsync());
+        } catch (Exception exception) {
+          // ignore and continue to close the connection.
+        }
       }
       // Try to wait for the current statement to finish (if any) before we actually close the
       // connection.
       this.closed = true;
-      // Add a no-op statement to the execute. Once this has been executed, we know that all
+      // Add a no-op statement to the executor. Once this has been executed, we know that all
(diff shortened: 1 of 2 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
```diff
@@ -294,12 +294,16 @@ public ApiFuture<Void> closeAsync() {
     if (!isClosed()) {
       List<ApiFuture<Void>> futures = new ArrayList<>();
       if (isTransactionStarted()) {
-        futures.add(rollbackAsync());
+        try {
+          futures.add(rollbackAsync());
+        } catch (Exception exception) {
```
rajatbhatta, line 299: What exception is specifically being thrown here by rollbackAsync(), in case rollback of transaction fails ? Does it make sense to catch that specific exception here ?

rajatbhatta (approved):
I've a small query. Apart from it, everything LGTM.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
olavloite: It intentionally catches wide exceptions, as a number of things could happen:
1. The connection could be in a state where rolling back is not allowed (resulting in something like an `IllegalStateException`)
2. A backend error could happen (`SpannerException`)
3. The statement executor could reject the execution (`RejectedExecutionException`)

In addition, there could in theory be other unexpected exceptions, but we don't really care about them here, as it is more important to close the connection and clean up the resources that it is using, than surfacing an unexpected exception occurring in a method the user did not actively call (the user called `close()` and not `rollback()`).

rajatbhatta: Sounds good!

PR #1887, opened 2022-05-24
fix: gracefully ignore RejectedExecutionException during Connection#close()

olavloite:
Gracefully catch and ignore `RejectedExecutionException`s that might happen while closing a connection. This prevents the `SpannerPool` from holding on to connections that failed to close because of such an error.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
@@ -57,6 +57,7 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Stack;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.ThreadFactory;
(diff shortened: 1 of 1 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
```diff
@@ -292,30 +293,38 @@ public void close() {
   }
 
   public ApiFuture<Void> closeAsync() {
-    if (!isClosed()) {
-      List<ApiFuture<Void>> futures = new ArrayList<>();
-      if (isBatchActive()) {
-        abortBatch();
-      }
-      if (isTransactionStarted()) {
+    synchronized (this) {
```
olavloite, line 296: This change looks bigger than it is. The actual change here is that:
1. The block is placed inside a `synchronized (this)` block.
2. The `futures.add(statementExecutor.submit(() -> null));` line is placed inside a try-catch block.

rajatbhatta (approved):
LGTM.

ansh0l (approved):
lgtm

PR #1914, opened 2022-06-10
feat: changes for supporting DML with Returning

rajatbhatta:
-   call executeQuery for both Query and Update
    statements.

-   allow executeQuery to execute all Update
    statements.


google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
```diff
@@ -943,23 +957,21 @@ private AsyncResultSet parseAndExecuteQueryAsync(
...
+        return ResultSets.toAsyncResultSet(
+            parsedStatement
+                .getClientSideStatement()
+                .execute(connectionStatementExecutor, parsedStatement.getSqlWithoutComments())
+                .getResultSet(),
+            spanner.getAsyncExecutorProvider(),
+            options);
+      case QUERY:
+      case UPDATE:
+        return internalExecuteQueryAsync(parsedStatement, analyzeMode, options);
```
olavloite, line 971: I think this also needs checks to verify that the result of the statement actually contains results and not just an update count.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
```diff
@@ -971,37 +983,46 @@ public long executeUpdate(Statement update) {
...
-      }
+        }
+        ResultSet rs = internalExecuteQuery(parsedStatement, AnalyzeMode.NONE);
+        if (rs.getColumnCount() > 0) {
+          throw SpannerExceptionFactory.newSpannerException(
+              ErrorCode.INVALID_ARGUMENT,
+              "Statement is not an normal DML statement: "
+                  + parsedStatement.getSqlWithoutComments());
+        }
+        while (rs.next()) ;
```
olavloite, line 998: We should add a comment to explain why we are doing this. Also, according to the Google style guide, a `while` statement must have curly braces.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
```diff
@@ -971,37 +983,46 @@ public long executeUpdate(Statement update) {
...
-        case DDL:
-        case UNKNOWN:
-        default:
-      }
+        }
+        ResultSet rs = internalExecuteQuery(parsedStatement, AnalyzeMode.NONE);
+        if (rs.getColumnCount() > 0) {
+          throw SpannerExceptionFactory.newSpannerException(
+              ErrorCode.INVALID_ARGUMENT,
+              "Statement is not an normal DML statement: "
```
olavloite, line 995: nit:
```suggestion
              "The statement returned a query result, for example because it is a query or a DML statement with a returning clause. Use executeQuery instead to execute these statements. Statement: "
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/SingleUseTransaction.java
```diff
@@ -417,6 +419,30 @@ private ApiFuture<ResultSetStats> executeTransactionalUpdateAsync(
...
+            return resultSet;
+          } catch (Throwable t) {
+            state = UnitOfWorkState.COMMIT_FAILED;
+            throw t;
+          }
+        };
+    return executeStatementAsync(
+        update,
+        callable,
+        ImmutableList.of(SpannerGrpc.getExecuteSqlMethod(), SpannerGrpc.getCommitMethod()));
```
olavloite, line 443: Hmm.... This is going to be a little bit of a difficult decision:
1. `executeUpdate` used to use the `executeSql` method, while `executeQuery` used the `executeStreamingSql`. Now all statements will use `executeStreamingSql`.
2. Users can set custom timeout and retry settings per RPC. This meant that they previously could in theory set a different timeout value for queries and DML statements. That is no longer possible, as they use the same RPC.
3. So users who currently set a timeout value for `ExecuteSql` will now have to set a timeout value for `ExecuteStreamingSql`.

It also means that we need to replace all references to `SpannerGrpc.getExecuteSqlMethod()` with `SpannerGrpc.getExecuteStreamingSqlMethod()`.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/AsyncResultSetImplTest.java
```diff
@@ -195,7 +195,7 @@ public void withCallback() throws InterruptedException {
     finishedLatch.await();
     // There should be between 1 and 4 callbacks, depending on the timing of the threads.
     // Normally, there should be just 1 callback.
-    assertThat(callbackCounter.get()).isIn(Range.closed(1, 4));
+    assertThat(callbackCounter.get()).isIn(Range.closed(1, 5));
```
olavloite, line 198: Why does this value change?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/GceTestEnvConfig.java
```diff
@@ -58,8 +58,11 @@ public class GceTestEnvConfig implements TestEnvConfig {
   private final SpannerOptions options;
 
   public GceTestEnvConfig() {
-    String projectId = System.getProperty(GCE_PROJECT_ID, "");
-    String serverUrl = System.getProperty(GCE_SERVER_URL, "");
+    // String projectId = System.getProperty(GCE_PROJECT_ID, "");
+    String projectId = "span-cloud-testing";
+    String serverUrl = "https://staging-wrenchworks.sandbox.googleapis.com";
+    // String serverUrl = System.getProperty(GCE_SERVER_URL,
+    // "https://staging-wrenchworks.sandbox.googleapis.com");
```
olavloite: nit: remove commented code etc.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/IntegrationTestEnv.java
```diff
@@ -82,7 +82,8 @@ protected void before() throws Throwable {
     this.config.setUp();
 
     SpannerOptions options = config.spannerOptions();
-    String instanceProperty = System.getProperty(TEST_INSTANCE_PROPERTY, "");
+    // String instanceProperty = System.getProperty(TEST_INSTANCE_PROPERTY, "");
+    String instanceProperty = "projects/span-cloud-testing/instances/rajatrb-test-instance";
```
olavloite: Here also: remove commented code

google-cloud-spanner/src/test/java/com/google/cloud/spanner/MockSpannerServiceImpl.java
```diff
@@ -1093,9 +1119,13 @@ public void executeBatchDml(
                       .build();
               break resultLoop;
             case RESULT_SET:
-              throw Status.INVALID_ARGUMENT
-                  .withDescription("Not a DML statement: " + statement.getSql())
-                  .asRuntimeException();
+              // Check if result containing UPDATE_COUNT exists. If not, throw an exception.
+              res = getResult(spannerStatement, StatementResult.StatementResultType.UPDATE_COUNT);
+              if (res == null) {
```
olavloite, line 1124: Does this mean that we don't support `RETURNING` for batch DML?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionAsyncApiTest.java
```diff
@@ -390,18 +397,6 @@ public void testExecuteClientSideQueryAsync() {
     }
   }
 
-  @Test
-  public void testExecuteInvalidQueryAsync() {
```
olavloite, line 394: Why is this test case removed? If I understand it correctly, this would execute a DML statement without a returning clause, which means that it should still fail.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
```diff
@@ -971,37 +983,46 @@ public long executeUpdate(Statement update) {
...
+        if (this.autocommitDmlMode == AutocommitDmlMode.PARTITIONED_NON_ATOMIC) {
           return get(internalExecuteUpdateAsync(parsedStatement));
-        case CLIENT_SIDE:
-        case QUERY:
-        case DDL:
-        case UNKNOWN:
-        default:
-      }
+        }
+        ResultSet rs = internalExecuteQuery(parsedStatement, AnalyzeMode.NONE);
```
olavloite, line 991: * What happens if this is called while the connection is in read-only mode?
* What happens if this is called while a read-only transaction is active?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
```diff
@@ -971,37 +983,46 @@ public long executeUpdate(Statement update) {
...
-          return internalExecuteUpdateAsync(parsedStatement);
-        case CLIENT_SIDE:
-        case QUERY:
-        case DDL:
-        case UNKNOWN:
-        default:
-      }
+    switch (parsedStatement.getType()) {
+      case UPDATE:
+        return internalExecuteUpdateAsync(parsedStatement);
```
olavloite, line 1020: What happens if this is called with a DML statement with a `RETURNING` clause?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
rajatbhatta: Have added tests for these scenarios.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
rajatbhatta: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
rajatbhatta: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/SingleUseTransaction.java
rajatbhatta: Done.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/AsyncResultSetImplTest.java
rajatbhatta: Tests pass without this change. Reverted this.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionAsyncApiTest.java
rajatbhatta: Reinstated the test with small change in when the exception will be raised.

rajatbhatta:
Closing this PR in favour of https://github.com/googleapis/java-spanner/pull/1978.

PR #1807, opened 2022-04-06
feat: Add support for db roles list

vi3k6i5:
feat: Add support for db roles list.

PR #1916, opened 2022-06-13
feat: add support for db roles list 

rahul2393:
* Added support of passing databaseRole in cloud-spanner sessions.

Old PR: https://github.com/googleapis/java-spanner/pull/1807


google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseRole.java
```diff
@@ -0,0 +1,78 @@
...
+
+package com.google.cloud.spanner;
+
+import static com.google.common.base.Preconditions.checkArgument;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
+import java.util.Objects;
+
+/** Represents a restore operation of a Cloud Spanner backup. */
```
olavloite, line 25: nit: copy-paste

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -830,6 +834,12 @@ public Builder setSessionPoolOption(SessionPoolOptions sessionPoolOptions) {
       return this;
     }
 
+    /** Sets the database role that should be used by this instance. */
```
olavloite, line 837: I think this needs a more elaborate Javadoc on what this means, as most users won't understand this automatically. I would suggest something like:

```suggestion
    /**
    * Sets the database role that should be used for connections that are created by this instance.
    * The database role that is used determines the access permissions that a connection has. This
    * can for example be used to create connections that are only permitted to access certain tables.
    */
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
```diff
@@ -213,8 +214,8 @@ public String[] getValidValues() {
   public static final String LENIENT_PROPERTY_NAME = "lenient";
   /** Name of the 'rpcPriority' connection property. */
   public static final String RPC_PRIORITY_NAME = "rpcPriority";
-  /** Dialect to use for a connection. */
-  private static final String DIALECT_PROPERTY_NAME = "dialect";
```
olavloite, line 217: I agree that we should remove this, as it is no longer needed, BUT:
1. We should consider doing so in a separate PR. It is not related to this change.
2. We should do it in a non-breaking way. Anyone who currently has `dialect=...` in their connection string will get an error after this change. We should therefore silently ignore it if we encounter `dialect=...` in a connection string.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
```diff
@@ -959,6 +969,11 @@ public TransportChannelProvider getChannelProvider() {
     }
   }
 
+  /** The creator role to use for the connection. */
```
olavloite, line 972: I don't think users will understand what 'creator role' means. This seems like an internal name (I (think I) understand that it means `the role that is used to create a session`, but that is not clear to a user.)

I would suggest: `The database role that is used for this connection. Assigning a role to a connection can be used to for example restrict the access of a connection to a specific set of tables.`

google-cloud-spanner/src/main/java/com/google/cloud/spanner/testing/RemoteSpannerHelper.java
```diff
@@ -50,6 +50,8 @@ public class RemoteSpannerHelper {
   private final InstanceId instanceId;
   private static final AtomicInteger dbSeq = new AtomicInteger();
   private static final int dbPrefix = new Random().nextInt(Integer.MAX_VALUE);
+  private static int dbRoleSeq;
```
olavloite, line 53: Although the probability is low that it goes wrong, it would be better to use an `AtomicInteger` for this, as our integration tests can run in parallel.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionClientTests.java
```diff
@@ -117,13 +117,15 @@ public void createAndCloseSession() {
     String sessionName = dbName + "/sessions/s1";
     Map<String, String> labels = new HashMap<>();
     labels.put("env", "dev");
+    String databaseRole = "role";
     when(spannerOptions.getSessionLabels()).thenReturn(labels);
+      when(spannerOptions.getDatabaseRole()).thenReturn(databaseRole);
```
olavloite, line 122: nit: indentation seems off here

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionClientTests.java
```diff
@@ -143,13 +145,16 @@ public void batchCreateAndCloseSessions() {
     final String sessionName = dbName + "/sessions/s%d";
     final Map<String, String> labels = new HashMap<>();
     labels.put("env", "dev");
+      String databaseRole = new String("role");
```
olavloite, line 148: nit: indentation seems to be off here (and also elsewhere in this class). Maybe run code formatter again?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerOptionsTest.java
```diff
@@ -76,7 +76,7 @@ public void defaultBuilder() {
     // id.
     SpannerOptions options = SpannerOptions.newBuilder().setProjectId("test-project").build();
     if (Strings.isNullOrEmpty(System.getenv("SPANNER_EMULATOR_HOST"))) {
-      assertThat(options.getHost()).isEqualTo("https://spanner.googleapis.com");
+      assertThat(options.getHost()).isEqualTo("https://staging-wrenchworks.sandbox.googleapis.com/");
```
olavloite, line 79: nit: Remove before merging

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
rahul2393: Reverted the change to remove Dialect, will be done in another PR

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Database.java
```diff
@@ -145,7 +145,7 @@ public Page<Operation> listDatabaseOperations() {
 
   /** Returns the IAM {@link Policy} for this database. */
   public Policy getIAMPolicy() {
-    return dbClient.getDatabaseIAMPolicy(instance(), database());
+    return dbClient.getDatabaseIAMPolicy(instance(), database(), 1);
```
olavloite, line 148: What is the `1` here?

EDIT: Apparently this is a version number. Should we add the version as an argument to the `getIAMPolicy` method?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseAdminClient.java
```diff
@@ -492,7 +495,7 @@ OperationFuture<Void, UpdateDatabaseDdlMetadata> updateDatabaseDdl(
   Operation getOperation(String name);
 
   /** Returns the IAM policy for the given database. */
-  Policy getDatabaseIAMPolicy(String instanceId, String databaseId);
+  Policy getDatabaseIAMPolicy(String instanceId, String databaseId, int version);
```
olavloite, line 498: The `version` argument needs some Javadoc. What does it do and when/why should I use it?

olavloite (approved):
Looks generally good to me, except we need some Javadoc for the version argument, as it is not self-explanatory.

PR #1919, opened 2022-06-22
test: test embedded comment in comment

olavloite:
Tests that block comments embedded in other block comments are parsed correctly by the PostgreSQL parser.


thiagotnunes (approved):

ansh0l:
@olavloite :  I think since the PR for this change is merged, should we close this PR?

olavloite:
> I think since the PR for this change is merged, should we close this PR?

No, this did not require a change, but I think we should merge it, as it adds a test for a scenario that was not covered (but did work as expected).

PR #1926, opened 2022-06-24
test: commit should retry unavailable error

olavloite:
Verify that an `UNAVAILABLE` error during a commit is retried by default in the client library.


ansh0l:
@olavloite : I think since the PR for this change is merged, should we close this PR?

olavloite:
> I think since the PR for this change is merged, should we close this PR?

Yes, this was only to verify that that `UNAVAILABLE` is retried as expected (there was a question regarding that).

PR #1929, opened 2022-06-28
Error Details Improvement

gauravpurohit06:
Adding following public methods under SpannerException to provide more info related to errors.
-getReason()
-getDomain()
-getMetaData
-getErrorDetails()

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerException.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerException.java
@@ -16,6 +16,8 @@
 
 package com.google.cloud.spanner;
 
+import com.google.api.gax.rpc.ApiException;
+import com.google.api.gax.rpc.ErrorDetails;
 import com.google.cloud.grpc.BaseGrpcServiceException;
 import com.google.common.base.Preconditions;
 import com.google.protobuf.util.Durations;
@@ -24,6 +26,7 @@ import com.google.rpc.RetryInfo;
 import io.grpc.Metadata;
 import io.grpc.Status;
 import io.grpc.protobuf.ProtoUtils;
+import java.util.Map;
 import javax.annotation.Nullable;
 
 /** Base exception type for all exceptions produced by the Cloud Spanner service. */
@@ -94,5 +97,33 @@ public class SpannerException extends BaseGrpcServiceException {
       }
     }
     return -1L;
+  }
+
+  public String getReason() {
+    if (this.getCause() instanceof ApiException) {
+      return ((ApiException) this.getCause()).getReason();
+    }
+    return null;
+  }
+
+  public String getDomain() {
+    if (this.getCause() instanceof ApiException) {
+      return ((ApiException) this.getCause()).getDomain();
+    }
+    return null;
+  }
+
+  public Map<String, String> getMetadata() {
+    if (this.getCause() instanceof ApiException) {
+      return ((ApiException) this.getCause()).getMetadata();
+    }
+    return null;
+  }
+
+  public ErrorDetails getErrorDetails() {
+    if (this.getCause() instanceof ApiException) {
+      return ((ApiException) this.getCause()).getErrorDetails();
+    }
+    return null;
   }
 }
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerExceptionFactory.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerExceptionFactory.java
@@ -295,12 +295,8 @@ public final class SpannerExceptionFactory {
       code = Status.Code.UNKNOWN;
     }
     ErrorCode errorCode = ErrorCode.fromGrpcStatus(Status.fromCode(code));
-    if (exception.getCause() != null) {
-      return SpannerExceptionFactory.newSpannerException(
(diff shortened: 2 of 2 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerException.java
```diff
@@ -95,4 +98,32 @@ static long extractRetryDelay(Throwable cause) {
     }
     return -1L;
   }
+
+  public String getReason() {
```
olavloite, line 102: nit: add javadoc for all public methods

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerExceptionFactory.java
```diff
@@ -295,12 +295,8 @@ private static SpannerException fromApiException(ApiException exception) {
       code = Status.Code.UNKNOWN;
     }
     ErrorCode errorCode = ErrorCode.fromGrpcStatus(Status.fromCode(code));
-    if (exception.getCause() != null) {
-      return SpannerExceptionFactory.newSpannerException(
-          errorCode, exception.getMessage(), exception.getCause());
-    } else {
-      return SpannerExceptionFactory.newSpannerException(errorCode, exception.getMessage());
-    }
+    return SpannerExceptionFactory.newSpannerException(
```
olavloite, line 298: (First of all: I think this in general seems like a good idea)

I wonder whether this modification makes this technically a breaking change. The exception that is returned here, and thereby also to users, will change slightly as the underlying cause will now be different than previously. Any code that inspects and makes any kind of choice based on the cause of an exception returned by this method could start behaving differently after this change has been merged.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerException.java
blakeli0: > nit: add javadoc for all public methods

+1

blakeli0:
> 🤖 I detect that the PR title and the commit message differ and there's only one commit. To use the PR title for the commit history, you can use Github's automerge feature with squashing, or use `automerge` label. Good luck human!
> 
> -- conventional-commit-lint bot https://conventionalcommits.org/

@gauravpurohit06 Please make sure the commit message and PR title follow the conventional commits in the future. In this case, I just renamed the title to `feat: Error Details Improvement`. Once this PR is ready, before you squash and merge, please make sure the squashed commit message is also `feat: Error Details Improvement`, we have a few automation bots rely on this convention.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerExceptionFactory.java
blakeli0: I don't have a strong opinion on this, but if you do decide it's a breaking change, please make sure the commit message is `feat!: Error Details Improvement` with an exclamation mark before merging in, so that our release bot will make a major version bump for this repo.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerException.java
```diff
@@ -59,10 +63,23 @@ public String getResourceName() {
       boolean retryable,
       @Nullable String message,
       @Nullable Throwable cause) {
-    super(message, cause, code.getCode(), retryable);
+    super(
+        message,
+        // If the cause is instance of APIException then using its cause to avoid the breaking
+        // change, because earlier we were passing APIException's cause to constructor.
+        cause instanceof ApiException ? cause.getCause() : cause,
```
olavloite, line 70: Hmm... Now that I see it a second time, I think we need to add a constructor overload that takes the `ApiException` as a separate argument, after all. I think that will make it a little more readable and understandable what is going on, but more importantly: The current implementation might also break existing code. Anyone that called this constructor directly with an instance of `ApiException` the cause (so not through the specific method in `SpannerExceptionFactory`) will now receive a different exception, as the `ApiException` is no longer set as the cause of this exception.

gauravpurohit06: Yeah, I think having the parameter based constructor will be more readable and understandable. SpannerException constructor is package-private and because we have factory written over it, in an ideal world everyone should use it to create the SpannerException, but as you correctly mentioned we can go ahead with parameterised constructor because on the top of readability and understandability we will also have more granular control to create SpannerException.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerExceptionFactory.java
```diff
@@ -295,8 +307,14 @@ private static SpannerException fromApiException(ApiException exception) {
...
     }
     ErrorCode errorCode = ErrorCode.fromGrpcStatus(Status.fromCode(code));
-    return SpannerExceptionFactory.newSpannerException(
-        errorCode, exception.getMessage(), exception);
+
+    Throwable exceptionCause = null;
+
+    if (exception.getCause() != null) {
+      exceptionCause = exception.getCause();
+    }
```
olavloite, line 315: nit: this block can be removed and replaced with `exception.getCause()` in the method call below

olavloite (approved):

ansh0l (approved):

PR #1949, opened 2022-07-16
feat: support multiple PostgreSQL transaction options

olavloite:
PostgreSQL allows BEGIN and other transaction statements to set multiple transaction
options at once (e.g. both 'read only' and 'isolation level serializable'). This was
not supported by the Connection API, which only allowed one option at a time. The
Python psycopg2 driver generates statements that include multiple transaction options
in one statement.

This PR adds support for:
* Begin transaction statements including both access mode and isolation level (these may be separated by both comma's and spaces, see https://www.postgresql.org/docs/current/sql-begin.html):
  * `begin read only isolation level serializable`
  * `begin isolation level serializable, read write`
* Set transaction statements using the same transaction modes as for begin:
  * `set transaction read only, isolation level serializable`
  * `set transaction isolation level serializable read write`
* Setting the `DEFAULT_TRANSACTION_ISOLATION` and `DEFAULT_TRANSACTION_READ_ONLY` properties


google-cloud-spanner/src/main/resources/com/google/cloud/spanner/connection/PG_ClientSideStatements.json
```diff
@@ -159,18 +159,21 @@
...
       "method": "statementBeginPgTransaction",
       "exampleStatements": [
         "begin", "start", "begin transaction", "start transaction", "begin work", "start work",
         "begin read only", "start read only", "begin transaction read only", "start transaction read only", "begin work read only", "start work read only",
         "begin read write", "start read write", "begin transaction read write", "start transaction read write", "begin work read write", "start work read write",
         "begin isolation level default", "start isolation level default", "begin transaction isolation level default", "start transaction isolation level default", "begin work isolation level default", "start work isolation level default",
-        "begin isolation level serializable", "start isolation level serializable", "begin transaction isolation level serializable", "start transaction isolation level serializable", "begin work isolation level serializable", "start work isolation level serializable"
```
ansh0l, line 173: We seem to be generating a bunch of these permutations on our own for exampleStatements, how do we track if we are not missing any specific permutation?

ansh0l (approved):
Have a couple of queries, but otherwise looks good.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/PgTransactionMode.java
```diff
@@ -16,36 +16,123 @@
...
+
+    /**
+     * Use this method to get the correct format for use in a SQL statement. The SQL statement for
+     * setting the mode to read-only should for example be without the underscore: <code>
+     * SET TRANSACTION READ ONLY</code>
+     *
+     * @return a string representation of this {@link TransactionMode} that can be used in a SQL
+     *     statement to set the transaction mode.
+     */
+    public String getStatementString() {
```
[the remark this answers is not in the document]
olavloite: It could be used to generate the SQL string that should be used (i.e. if there was a difference between `toString()` and this method), but it is not needed for these enums. Removed.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ClientSideStatementValueConverters.java
```diff
@@ -85,6 +88,43 @@ public Boolean convert(String value) {
...
+      }
+      if ("true".equalsIgnoreCase(value)
+          || "tru".equalsIgnoreCase(value)
+          || "tr".equalsIgnoreCase(value)
+          || "t".equalsIgnoreCase(value)) {
+        return Boolean.TRUE;
+      }
+      if ("false".equalsIgnoreCase(value)
+          || "fals".equalsIgnoreCase(value)
+          || "fal".equalsIgnoreCase(value)
```
gauravsnj, line 119: Can we add support for 'ON' and 'OFF' while converting the boolean value?

google-cloud-spanner/src/main/resources/com/google/cloud/spanner/connection/PG_ClientSideStatements.json
```diff
@@ -352,10 +355,53 @@
...
+      "statementType": "SET_READONLY",
+      "regex": "(?is)\\A\\s*set\\s+default_transaction_read_only(?:\\s*=\\s*|\\s+to\\s+)(.*)\\z",
+      "method": "statementSetReadOnly",
+      "exampleStatements": [
+        "set default_transaction_read_only = true",
+        "set default_transaction_read_only = false",
+        "set default_transaction_read_only = t",
+        "set default_transaction_read_only = f",
+        "set default_transaction_read_only to 't'",
+        "set default_transaction_read_only to \"f\""
```
gauravsnj, line 396: We can add a couple of example statements here related to ON and OFF

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ClientSideStatementValueConverters.java
```diff
@@ -85,6 +88,43 @@ public Boolean convert(String value) {
...
+        value = value.substring(1, value.length() - 1);
+      }
+      if ("true".equalsIgnoreCase(value)
+          || "tru".equalsIgnoreCase(value)
+          || "tr".equalsIgnoreCase(value)
+          || "t".equalsIgnoreCase(value)) {
+        return Boolean.TRUE;
+      }
+      if ("false".equalsIgnoreCase(value)
+          || "fals".equalsIgnoreCase(value)
```
gauravsnj, line 118: The values like `"tru", "fal", "tr"` are not supported without single quotes (`'`) but according to this code they will be supported without single quotes (`'`)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ClientSideStatementValueConverters.java
```diff
@@ -85,6 +88,43 @@ public Boolean convert(String value) {
...
+    public Boolean convert(String value) {
+      if (value == null) {
+        return null;
+      }
+      if (value.length() > 1
+          && ((value.startsWith("'") && value.endsWith("'"))
+              || (value.startsWith("\"") && value.endsWith("\"")))) {
+        value = value.substring(1, value.length() - 1);
+      }
+      if ("true".equalsIgnoreCase(value)
```
gauravsnj, line 111: I think we should also trim the string to consider the cases like `'  true  '`

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ClientSideStatementValueConverters.java
```diff
@@ -85,6 +88,43 @@ public Boolean convert(String value) {
...
+      return Boolean.class;
+    }
+
+    @Override
+    public Boolean convert(String value) {
+      if (value == null) {
+        return null;
+      }
+      if (value.length() > 1
+          && ((value.startsWith("'") && value.endsWith("'"))
```
gauravsnj, line 107: I think Double Quotes(`""`) are used to identify objects inside the database like column names, table names etc. So, `"true"` shouldn't work

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ClientSideStatementValueConverters.java
```diff
@@ -85,6 +88,43 @@ public Boolean convert(String value) {
...
+              || (value.startsWith("\"") && value.endsWith("\"")))) {
```
gauravsnj, line 108: I think Double Quotes("") are used to identify objects inside the database like column names, table names etc. So, `"true"` shouldn't work

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ClientSideStatementValueConverters.java
```diff
@@ -286,16 +326,39 @@ public TransactionMode convert(String value) {
...
+    public PgTransactionIsolationConverter(String allowedValues) {}
+
+    @Override
+    public Class<IsolationLevel> getParameterClass() {
+      return IsolationLevel.class;
+    }
+
+    @Override
+    public IsolationLevel convert(String value) {
+      // Isolation level may contain multiple spaces.
```
gauravsnj, line 343: Is checking the value string won't be null done somewhere else?

google-cloud-spanner/src/main/resources/com/google/cloud/spanner/connection/PG_ClientSideStatements.json
olavloite: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ClientSideStatementValueConverters.java
olavloite: Strangely enough, they are. I've tried a multitude of different statements with a real PG instance. See below for the results:

```sql
knut-test-db=# set default_transaction_read_only to "on";
SET

knut-test-db=# set default_transaction_read_only to tr;
SET

knut-test-db=# set default_transaction_read_only to ' true ';
ERROR:  parameter "default_transaction_read_only" requires a Boolean value

knut-test-db=# set default_transaction_read_only to "true";
SET

knut-test-db=# set default_transaction_read_only to 'true';
SET

knut-test-db=# set default_transaction_read_only to ye;
SET

knut-test-db=# set default_transaction_read_only to 'ye';
SET
```

So TLDR:
* Both single and double quotes are allowed (which is strange, as double quotes would normally be used to quote identifiers)
* Unique prefixes like `tr` and `n` are allowed both with and without quotes.
* Spaces inside quotes are __not__ allowed, so `' true '` is not a valid value.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ClientSideStatementValueConverters.java
olavloite: (See also above). PostgreSQL does not consider that to be a valid boolean value. When the value is quoted with either single or double quotes, it may not contain any spaces.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ClientSideStatementValueConverters.java
olavloite: I agree that it strange, but it actually allowed (see above)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ClientSideStatementValueConverters.java
gauravsnj: That's strange because once I tested inserting boolean values in some table and the results were totally opposite :) Btw thanks for clarifying

gauravsnj (approved):
LGTM

rajatbhatta (approved):

PR #1957, opened 2022-07-21
feat: add support for Optimistic Concurrency Control

ko3a4ok:


ko3a4ok:
@ansh0l @rajatbhatta could you please help with review?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionImpl.java
```diff
@@ -67,6 +67,14 @@ static void throwIfTransactionsPending() {
     }
   }
 
+  static TransactionOptions createReadWriteOptions(Options options) {
```
rajatbhatta, line 70:
```suggestion
  static TransactionOptions createReadWriteTransactionOptions(Options options) {
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionImpl.java
```diff
@@ -299,14 +307,16 @@ ByteString beginTransaction() {
...
+  ApiFuture<ByteString> beginTransactionAsync(Options transactionOptions) {
     final SettableApiFuture<ByteString> res = SettableApiFuture.create();
     final Span span = tracer.spanBuilder(SpannerImpl.BEGIN_TRANSACTION).startSpan();
     final BeginTransactionRequest request =
         BeginTransactionRequest.newBuilder()
             .setSession(name)
-            .setOptions(
-                TransactionOptions.newBuilder()
-                    .setReadWrite(TransactionOptions.ReadWrite.getDefaultInstance()))
+            .setOptions(createReadWriteOptions(transactionOptions))
```
rajatbhatta, line 319:
```suggestion
            .setOptions(createReadWriteTransactionOptions(transactionOptions))
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
```diff
@@ -492,9 +492,7 @@ TransactionSelector getTransactionSelector() {
           }
           if (tx == null) {
             return TransactionSelector.newBuilder()
-                .setBegin(
-                    TransactionOptions.newBuilder()
-                        .setReadWrite(TransactionOptions.ReadWrite.getDefaultInstance()))
+                .setBegin(SessionImpl.createReadWriteOptions(options))
```
rajatbhatta, line 495:
```suggestion
                .setBegin(SessionImpl.createReadWriteTransactionOptions(options))
```

rajatbhatta (changes requested):
Overall, this looks good to me, apart from a couple of suggestions. Please take a look.

rajatbhatta (commented):
Proto changes should be merged as part of autogenerated PRs (once the proto file changes are made public). Are the corresponding proto file changes public yet?

ko3a4ok:
> @rajatbhatta, Proto changes are not in public and  behind visibility restriction: TRUSTED_TESTER_OCC. The feature isn't ready on the server side yet, but API has been defined. I wanted to update a client so I can test the feature end-end. So what should I do with protos?

rajatbhatta:
> @rajatbhatta, Proto changes are not in public and  behind visibility restriction: TRUSTED_TESTER_OCC. The feature isn't ready on the server side yet, but API has been defined. I wanted to update a client so I can test the feature end-end. So what should I do with protos?

While merging this PR, we'll have to remove the protos. Which means that we should merge this PR only after visibility restriction is lifted, and proto changes are merged and made public (as part of autogenerated PR).

google-cloud-spanner/src/test/java/com/google/cloud/spanner/ReadWriteTransactionWithInlineBeginTest.java
```diff
@@ -389,6 +404,71 @@ public void failedBatchUpdateAndThenUpdate() {
...
 
+  @Test
+  public void executeSqlWithOptimisticConcurrencyControl() {
+    client
+        .readWriteTransaction(Options.optimisticLock())
+        .run(
+            transaction -> {
+              try (ResultSet rs = transaction.executeQuery(SELECT1)) {
+                while (rs.next()) {
+                  assertThat(rs.getLong(0)).isEqualTo(1);
```
olavloite, line 415: nit (here and elsewhere): prefer the standard org.junit.Assert `assertEquals(expected, actual)` over `assertThat(..)`.

(I know that there's a lot of Google Truth `assertThat(..)` in this and also other files for historical reasons, but the preference for the client libraries is to use `assertEquals` for new code)

google-cloud-spanner/src/test/java/com/google/cloud/spanner/ReadWriteTransactionWithInlineBeginTest.java
```diff
@@ -389,6 +404,71 @@ public void failedBatchUpdateAndThenUpdate() {
...
+  }
+
+  @Test
+  public void beginTransactionWithOptimisticConcurrencyControl() {
+    client
+        .readWriteTransaction(Options.optimisticLock())
+        .run(
+            transaction -> {
+              // This query carries the BeginTransaction, but fails. The BeginTransaction will
+              // then be carried by the subsequent statement.
```
olavloite, line 457: This comment is not accurate. Instead of adding the `BeginTransaction` option to the next statement, the client library will force a complete retry of the entire transaction, and use an explicit `BeginTransaction` RPC invocation for that transaction in order to include the failed statement in the transaction as well.

olavloite (approved):
Looks generally good to me (with some nits on assertions).

Also; would you mind adding a test to this file: https://github.com/googleapis/java-spanner/blob/f01d263bb890679c1591ffb2232781d68e8303a8/google-cloud-spanner/src/test/java/com/google/cloud/spanner/RetryOnInvalidatedSessionTest.java#L519

Those tests verify that everything works as expected if a transaction fails in case of a `SessionNotFoundException`. That can happen if a session has been garbage collected by the backend with the client knowing/expecting it.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/ReadWriteTransactionWithInlineBeginTest.java
ko3a4ok: Done

google-cloud-spanner/src/test/java/com/google/cloud/spanner/ReadWriteTransactionWithInlineBeginTest.java
ko3a4ok: Updated the comment, thanks!

ansh0l:
cc @surbhigarg92

PR #1965, opened 2022-08-04
fix: Use the key instead of the value to verify the number of channels created in ChannelUsageTest

blakeli0:
This is to fix an issue where the test would fail if the value of the key changes in gRPC. See https://github.com/googleapis/java-spanner/pull/1960 for details.


suztomo (approved):

suztomo:
@ansh0l Can you take a look. This pull request (which fixes https://github.com/googleapis/java-spanner/pull/1960) is a blocker for our BOM release.

ansh0l (approved):

PR #1977, opened 2022-08-17
test: Creating dummy spanner instance for ITInstanceAdminTest

### LINKED ISSUE #1626: ITInstanceAdminTest #update / #updateViaEntity fail when two tests are running simultaneously
The update and updateViaEntity tests update the current instance display name to a random one. If there are two tests running simultaneously the assertions might fail, because the updated name can match the wrong display name.

We need to fix these tests so that they are self-contained.

gauravpurohit06:
Fixes #1626

### DIFF
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/IntegrationTestEnv.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/IntegrationTestEnv.java
@@ -58,14 +58,16 @@ public class IntegrationTestEnv extends ExternalResource {
   private InstanceAdminClient instanceAdminClient;
   private DatabaseAdminClient databaseAdminClient;
   private boolean isOwnedInstance;
-  private boolean useNewInstance;
+  private final boolean createNewInstance;
   private RemoteSpannerHelper testHelper;
 
-  public IntegrationTestEnv(final boolean useNewInstance) {
-    this.useNewInstance = useNewInstance;
+  public IntegrationTestEnv() {
+    this(false);
   }
 
-  public IntegrationTestEnv() {}
+  public IntegrationTestEnv(final boolean createNewInstance) {
+    this.createNewInstance = createNewInstance;
+  }
 
   public RemoteSpannerHelper getTestHelper() {
     checkInitialized();
@@ -91,7 +93,7 @@ public class IntegrationTestEnv extends ExternalResource {
     SpannerOptions options = config.spannerOptions();
     String instanceProperty = System.getProperty(TEST_INSTANCE_PROPERTY, "");
     InstanceId instanceId;
-    if (!instanceProperty.isEmpty() && !useNewInstance) {
+    if (!instanceProperty.isEmpty() && !createNewInstance) {
       instanceId = InstanceId.of(instanceProperty);
       isOwnedInstance = false;
       logger.log(Level.INFO, "Using existing test instance: {0}", instanceId);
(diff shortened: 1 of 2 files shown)


google-cloud-spanner/src/test/java/com/google/cloud/spanner/IntegrationTestEnv.java
```diff
@@ -58,8 +58,15 @@ public class IntegrationTestEnv extends ExternalResource {
...
   private DatabaseAdminClient databaseAdminClient;
   private boolean isOwnedInstance;
+  private boolean useNewInstance;
   private RemoteSpannerHelper testHelper;
 
+  public IntegrationTestEnv(final boolean useNewInstance) {
+    this.useNewInstance = useNewInstance;
+  }
+
+  public IntegrationTestEnv() {}
```
rajatbhatta, line 68: Should we set a default value for useNewInstance in this constructor?
```suggestion
  public IntegrationTestEnv() {
    this.useNewInstance = false;
  }
```

rajatbhatta (changes requested):
Overall it looks good to me, except a comment. PTAL.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/IntegrationTestEnv.java
olavloite: A couple of nits regarding this:
1. It is common to have constructors with less parameters before constructors with more parameters (i.e. move this one to line 63)
2. It is good practice to reuse code in constructors, so I would suggest calling `this(false)` from the parameter-less constructor.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/IntegrationTestEnv.java
```diff
@@ -58,8 +58,15 @@ public class IntegrationTestEnv extends ExternalResource {
   private InstanceAdminClient instanceAdminClient;
   private DatabaseAdminClient databaseAdminClient;
   private boolean isOwnedInstance;
+  private boolean useNewInstance;
```
olavloite, line 61: nit: Maybe rename to `alwaysUseOwnedInstance`?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITInstanceAdminTest.java
```diff
@@ -45,11 +45,12 @@
 @Category(SerialIntegrationTest.class)
```
olavloite, line 45: As this is now always using an owned instance, you can probably change this to a `ParallelIntegrationTest`.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/IntegrationTestEnv.java
gauravpurohit06: Thanks Knut. It seems like a bit confusing with `isOwnedInstance`, can we use `createNewInstance`?

olavloite: Could we in that case use `alwaysCreateNewInstance`. My point is that any of these tests __could__ create a new instance. I think it will be confusing if the test creates a new instance while the property `createNewInstance` is `false`. The `always` prefix indicates that it forces the specific test to always create a new instance, regardless whether the test environment has defined an existing instance name that should be used.

olavloite (approved):
LGTM, with a suggestion regarding the naming of the `createNewInstance` / `alwaysCreateNewInstance` property.

rajatbhatta (approved):
LGTM.

PR #1978, opened 2022-08-17
feat: support DML with Returning clause in Connection API

rajatbhatta:
This PR adds support for running DML statements with Returning clause in the Connection API, and incorporates the following:
- API documentation changes to document the behaviour of `execute`, `executeAsync`, `executeQuery`, `executeQueryAsync`, `executeUpdate`, `executeUpdateAsync`, `executeBatchUpdate`, and `executeBatchUpdateAsync` APIs when executing DML statements with Returning clause.
- Parses the SQL string to check if the DML statement has a Returning clause. Parsing logic has been added for both GoogleStandardSql and PostgreSql dialects.
- Changes to MockSpannerServiceImpl to put results for DML statements with Returning clause, and to add support in `executeBatchDml`.
- Integration tests for running DML statements with Returning clause using each of the above APIs, for both GoogleStandardSql and PostgreSql dialects.
- Unit tests for abort scenarios while running DML statements with Returning clause.
- Parser tests for both GoogleStandardSql and PostgreSql dialects.


google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ReadWriteTransaction.java
```diff
@@ -337,7 +338,10 @@ public ApiFuture<ResultSet> executeQueryAsync(
       final ParsedStatement statement,
       final AnalyzeMode analyzeMode,
       final QueryOption... options) {
-    Preconditions.checkArgument(statement.isQuery(), "Statement is not a query");
+    Preconditions.checkArgument(
+        (statement.getType() == StatementType.QUERY)
+            || (statement.getType() == StatementType.UPDATE && statement.hasReturningClause()),
+        "Statement must be a query or DML with returning clause");
     checkValidTransaction();
```
olavloite, line 345: Can we add a couple of tests to https://github.com/googleapis/java-spanner/blob/main/google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/AbortedTest.java for DML statements with `RETURNING` clauses to ensure that it works correctly with the internal retry mechanism of the Connection API?

rajatbhatta: Added the tests.

rajatbhatta:
Note: Integration tests are not expected to pass until DML with Returning clause support is launched.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/SpannerStatementParser.java
```diff
@@ -250,4 +269,70 @@ ParametersInfo convertPositionalParametersToNamedParametersInternal(char paramCh
     }
     return new ParametersInfo(paramIndex - 1, named.toString());
   }
+
+  private boolean isReturning(String sql, int index) {
+    return Pattern.compile("[\\s)]THEN(\\s+)RETURN[\\s(][\\s\\S]*", Pattern.CASE_INSENSITIVE)
+        .matcher(sql.substring(index))
+        .matches();
```
olavloite, line 276: Also here: I think this could be made more efficient by at least removing the `compile` statement out of this method, but I also think that you don't need a regex at all. You can just do a simple string check if you first replace all double spaces with single spaces, and make the SQL string all lower or upper case.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/GceTestEnvConfig.java
```diff
@@ -58,8 +58,10 @@ public class GceTestEnvConfig implements TestEnvConfig {
   private final SpannerOptions options;
 
   public GceTestEnvConfig() {
-    String projectId = System.getProperty(GCE_PROJECT_ID, "");
-    String serverUrl = System.getProperty(GCE_SERVER_URL, "");
+    // String projectId = System.getProperty(GCE_PROJECT_ID, "");
```
olavloite, line 61: nit: add a TODO here + a Do not merge label in the PR itself to make sure this is modified before the PR is merged.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/MockSpannerServiceImpl.java
```diff
@@ -1120,10 +1125,20 @@ public void executeBatchDml(
...
+        switch (res.getType()) {
+          case UPDATE_COUNT:
+            updateCount = res.getUpdateCount();
+            break;
+          case RESULT_SET:
+            updateCount = res.getResultSet().getStats().getRowCountExact();
+            break;
+          default:
+            throw new IllegalStateException("Invalid result type: " + res.getType());
+        }
```
olavloite, line 1138: Note that while this works fine here and is semantically correct, it might be a little confusing, because (I think) the backend does not allow DML statements with a RETURNING clause to be executed using `BatchDml`.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/AbortedTest.java
```diff
@@ -115,53 +155,63 @@ public void testAbortedDuringRetryOfFailedQuery() {
 
   @Test
   public void testAbortedDuringRetryOfFailedUpdate() {
-    final Statement invalidStatement = Statement.of("INSERT INTO FOO");
+    final Iterable<Statement> invalidStatements =
+        ImmutableList.of(
+            Statement.of("INSERT INTO FOO"), Statement.of("INSERT INTO FOO THEN RETURN *"));
```
olavloite, line 160: I think it would be better to leave these tests as they were and add separate tests for DML statements with returning clauses. This potentially removes existing test coverage for DML statements without a returning clause, and those are handled differently than DML statements with a returning clause, so we should have separate tests for each.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/AbortedTest.java
```diff
@@ -201,63 +251,77 @@ public void testAbortedDuringRetryOfFailedQueryAsFirstStatement() {
...
     mockSpanner.putStatementResult(StatementResult.update(INSERT_STATEMENT, UPDATE_COUNT));
+    mockSpanner.putStatementResult(
+        StatementResult.updateReturning(INSERT_RETURNING_STATEMENT, UPDATE_RETURNING_RESULTSET));
     try (ITConnection connection = createConnection()) {
       connection.setTransactionTag("transaction-tag");
       connection.setStatementTag("statement-tag");
       connection.executeUpdate(INSERT_STATEMENT);
       connection.setStatementTag("statement-tag");
-      connection.executeBatchUpdate(Collections.singleton(INSERT_STATEMENT));
+      connection.executeBatchUpdate(ImmutableList.of(INSERT_STATEMENT, INSERT_RETURNING_STATEMENT));
```
olavloite, line 320: An interesting question here is whether the client should already throw an exception when you try to execute it, or whether we should rely on the backend doing so (at least, as far as I can remember, `BatchDml` does not support returning clauses).

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/AbstractMockServerTest.java
```diff
@@ -96,8 +97,18 @@ public abstract class AbstractMockServerTest {
                   .build())
           .setMetadata(SELECT_COUNT_METADATA)
           .build();
+  public static final com.google.spanner.v1.ResultSet UPDATE_RETURNING_RESULTSET =
+      com.google.spanner.v1.ResultSet.newBuilder()
+          .setStats(ResultSetStats.newBuilder().setRowCountExact(1))
+          .setMetadata(
+              ResultSetMetadata.getDefaultInstance()
+                  .toBuilder()
+                  .setRowType(StructType.getDefaultInstance()))
```
olavloite, line 106: This doesn't seem right. A DML statement with a `RETURNING` clause would return a result set with at least one column.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/StatementParserTest.java
```diff
@@ -1266,6 +1266,156 @@ public void testPostgreSQLGetQueryParameters() {
...
+        parser
+            .parse(Statement.of("insert into x (a,b) values (1,2)then/*comment*/return *"))
+            .hasReturningClause());
+    assertTrue(
+        parser
+            .parse(Statement.of("insert into x (a,b) values (1,2)then/*comment*/return(a)"))
+            .hasReturningClause());
+    assertTrue(
+        parser
+            .parse(Statement.of("insert into x (a,b) values (1,2)then /*comment*/return(a)"))
```
olavloite, line 1316: Could we also add some tests with (string) literals surrounding the `then return` clause?

So for example `delete from x where y='then return'` and `delete from x where y='z'then return *`

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/StatementParserTest.java
```diff
@@ -1266,6 +1266,156 @@ public void testPostgreSQLGetQueryParameters() {
...
+        parser
+            .parse(
+                Statement.of(
+                    "insert into x (a,b) values (1,2)/*comment" + "*/returning/*" + "comment*/(a)"))
+            .hasReturningClause());
+    assertTrue(
+        parser
+            .parse(Statement.of("insert into t1 select 1 as returning returning *"))
+            .hasReturningClause());
+    assertTrue(
```
olavloite, line 1395: Same as above: Can we add some tests with string literals?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/StatementParserTest.java
```diff
@@ -1266,6 +1266,156 @@ public void testPostgreSQLGetQueryParameters() {
...
+            .parse(Statement.of("insert into x (a,b) values (1,2)returning(a)"))
+            .hasReturningClause());
+    assertTrue(
+        parser
+            .parse(Statement.of("insert into x (a,b) values (1,2)/*comment*/returning(a)"))
+            .hasReturningClause());
+    assertTrue(
+        parser
+            .parse(
+                Statement.of("insert into x (a,b) values (1,2)/*comment*/returning/*comment*/(a)"))
```
olavloite, line 1383: We also need tests with nested comments (see also my comment on the parser itself)

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/it/ITDmlReturningTest.java
```diff
@@ -0,0 +1,343 @@
...
+    }
+  }
+
+  @Test
+  public void testDmlReturningExecuteUpdate() {
+    try (Connection connection = createConnection()) {
+      connection.setAutocommit(false);
+      try {
+        connection.executeUpdate(UPDATE_RETURNING_MAP.get(dialect));
+        fail("missing exception");
```
olavloite, line 187: nit: use `assertThrows` instead of a `try-catch-fail` construct

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/it/ITDmlReturningTest.java
```diff
@@ -0,0 +1,343 @@
...
+                    assertEquals(resultSet.getStats().getRowCountExact(), 3);
+                    return CallbackResponse.DONE;
+                  case NOT_READY:
+                    return CallbackResponse.CONTINUE;
+                  default:
+                    throw new IllegalStateException();
+                }
+              }
+            } catch (SpannerException e) {
+              System.out.printf("Error in callback: %s%n", e.getMessage());
```
olavloite, line 174: Remove or replace with an actual logger

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/it/ITDmlReturningTest.java
```diff
@@ -0,0 +1,343 @@
...
+    }
+  }
+
+  @Test
+  public void testDmlReturningExecuteBatchUpdate() {
+    try (Connection connection = createConnection()) {
+      connection.setAutocommit(false);
+      final Statement UPDATE_STMT = UPDATE_RETURNING_MAP.get(dialect);
+      long[] counts =
+          connection.executeBatchUpdate(ImmutableList.of(UPDATE_STMT, UPDATE_STMT, UPDATE_STMT));
```
olavloite, line 213: If this actually works, then my previous comments regarding `BatchDml` and returning are void. But I also think you should check with the developers of this feature whether this is intentional, as the result of the `RETURNING` clause is lost in these cases.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/Connection.java
```diff
@@ -984,8 +987,9 @@ default ResultSetStats analyzeUpdate(Statement update, QueryAnalyzeMode analyzeM
   ApiFuture<Long> executeUpdateAsync(Statement update);
 
   /**
-   * Executes a list of DML statements in a single request. The statements will be executed in order
-   * and the semantics is the same as if each statement is executed by {@link
+   * Executes a list of DML statements (can be simple DML statements or DML statements with
+   * returning clause) in a single request. The statements will be executed in order and the
```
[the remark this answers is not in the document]
rajatbhatta: DML with Returning clause over `BatchDML` is allowed. We can find corresponding integration tests for `BatchDml` and `BatchDmlAsync` available in ITDmlReturningTest in this PR.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/Connection.java
```diff
@@ -1006,8 +1010,9 @@ default ResultSetStats analyzeUpdate(Statement update, QueryAnalyzeMode analyzeM
   long[] executeBatchUpdate(Iterable<Statement> updates);
 
   /**
-   * Executes a list of DML statements in a single request. The statements will be executed in order
-   * and the semantics is the same as if each statement is executed by {@link
+   * Executes a list of DML statements (can be simple DML statements or DML statements with
+   * returning clause) in a single request. The statements will be executed in order and the
```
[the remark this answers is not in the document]
rajatbhatta: DML with Returning clause over BatchDML is allowed. We can find corresponding integration tests for BatchDml and BatchDmlAsync available in ITDmlReturningTest in this PR.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/it/ITDmlReturningTest.java
rajatbhatta: Yes, this feature is intentional, as discussed with the developers of this feature.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/SpannerStatementParser.java
```diff
@@ -250,4 +269,70 @@ ParametersInfo convertPositionalParametersToNamedParametersInternal(char paramCh
...
+            startQuote = 0;
+          }
+        } else if (c == '\\') {
+          lastCharWasEscapeChar = true;
+        } else {
+          lastCharWasEscapeChar = false;
+        }
+      } else {
+        if (!hasReturningClause && isReturning(sql, index)) {
+          hasReturningClause = true;
```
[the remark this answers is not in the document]
rajatbhatta: I had the exception at the end of this function for unclosed literal and I realise that it is redundant as we already have this handling when we remove comments from string. I'll remove that handling, and return true directly from here.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/PostgreSQLStatementParser.java
```diff
@@ -92,6 +108,10 @@ String removeCommentsAndTrimInternal(String sql) {
           continue;
         } else if (sql.length() > index + 1 && c == SLASH && sql.charAt(index + 1) == ASTERISK) {
           multiLineCommentLevel++;
+          if (index >= 1) {
+            whitespaceBeforeOrAfterMultiLineComment = Character.isWhitespace(sql.charAt(index - 1));
+          }
+          multiLineCommentStartIdx = index;
```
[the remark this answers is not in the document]
rajatbhatta: I think this would not go wrong. We are checking for space before/after only for the outermost comment. I've added a test case having nested comments too in that regard.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/StatementParserTest.java
rajatbhatta: I've already added test with nested comments. Responded on the above comment.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/PostgreSQLStatementParser.java
```diff
@@ -267,4 +287,36 @@ private void appendIfNotNull(
       result.append(prefix).append(tag).append(suffix);
     }
   }
+
+  private boolean isReturning(String sql, int index) {
+    // RETURNING is a reserved keyword in PG, but requires a
+    // leading AS to be used as column label, to avoid ambiguity.
+    // We thus check for cases which do not have a leading AS.
+    // (https://www.postgresql.org/docs/current/sql-keywords-appendix.html)
+    return Pattern.compile("[\\s)]RETURNING[\\s(][\\s\\S]*", Pattern.CASE_INSENSITIVE)
```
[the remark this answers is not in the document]
rajatbhatta: Thanks for the points!

- I have now optimised the isReturning method to only look for a particular length of substring (for example - in GoogleStandardSql, we would need to only check for substrings of length 13 and check for any of the characters out of ``)`"'`` before `then return` and any of the characters out of ``(`"'`` after it). Regex check would then have fairly same time complexity as normal comparison.
- I've now Incorporated point 1 as well.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/SpannerStatementParser.java
rajatbhatta: Added answer to this in https://github.com/googleapis/java-spanner/pull/1978#discussion_r952676785.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/MockSpannerServiceImpl.java
rajatbhatta: DML with Returning clause over BatchDML is allowed. Have added corresponding integration tests for `executeBatchDml` and `executeBatchDmlAsync` in ITDmlReturningTest in this PR.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/it/ITDmlReturningTest.java
rajatbhatta: Removed it.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/AbortedTest.java
rajatbhatta: DML with Returning clause over `BatchDML` is allowed. Have added corresponding integration tests for `executeBatchDml` and `executeBatchDmlAsync` in ITDmlReturningTest in this PR.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/StatementParserTest.java
rajatbhatta: Added.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/AbstractMockServerTest.java
rajatbhatta: Modified it to set row type to a StructType with 1 field (so that result set has 1 column).

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/AbortedTest.java
rajatbhatta: Added separate tests for DML statements with returning clause.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/PostgreSQLStatementParser.java
```diff
@@ -267,4 +290,32 @@ private void appendIfNotNull(
...
+    Preconditions.checkNotNull(rawSql);
+    int index = 0;
+    String sql = rawSql.replaceAll("\\s+", " ").toLowerCase();
+    while (index < sql.length()) {
+      if (isReturning(sql, index)) {
+        return true;
+      } else {
+        index = skip(sql, index, null);
+      }
+    }
```
olavloite, line 318: This is still a very expensive check, as we always iterate through the entire SQL string and repeatedly check for a regex match. Can we improve this a little by doing the following:
1. Do a pre-check to see if the (lower)-case string contains `"returning"` at all. If it does not, we know for sure that the regex will also never match. See https://gist.github.com/olavloite/a857ff860c4f12c17d51712bed57b9d3 for a small benchmark program. Some quick tests indicate that having such a pre-check improves performance with a factor 10x (400ms for 100,000 checks as opposed to 4,000ms on my local machine).
2. Do the while loop in the opposite order (i.e. start at `index=sql.length()`). If we do a pre-check and that check returns true, the probability that we will find a match is a lot bigger, as `returning` is not something that is very common in any given SQL string. And a `returning` clause will be somewhere at the end of the SQL string.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/SpannerStatementParser.java
```diff
@@ -250,4 +271,65 @@ ParametersInfo convertPositionalParametersToNamedParametersInternal(char paramCh
...
+
+  private boolean isReturning(String sql, int index) {
+    return (index >= 1)
+        && (index + 12 <= sql.length())
+        && THEN_RETURN_PATTERN.matcher(sql.substring(index - 1, index + 12)).matches();
+  }
+
+  @InternalApi
+  @Override
+  boolean checkReturningClauseInternal(String rawSql) {
```
olavloite, line 283: Could we also do a couple of optimizations here:
1. Do a `sql.contains("return") && sql.contains("then")` check before anything else, as that would allow us to return false quickly in most cases.
2. (Maybe in a separate PR, as this is more complicated here because of the checking for escaping etc.) Loop in the opposite direction if the quick check succeeds?

rajatbhatta: Good point! I'll work on enhancement #2 in a separate PR, and will incorporate enhancement #1 in this PR for now.

rajatbhatta: The check can be `sql.contains("then return")` as we must have already replaced all consecutive whitespaces with a single space.

olavloite: No, I would recommend that we do not do the last. So the order of operations would roughly be:
1. sql.toLowerCase()
2. sql.contains("return") && sql.contains("then")
3. sql.replaceAll("\\s+", " ")

The reason for this is that the `replaceAll` is also quite expensive, as it is also a regular expression check for each element in the string.

rajatbhatta: Makes sense.

rajatbhatta: Made this change.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/StatementParserTest.java
```diff
@@ -1266,6 +1266,178 @@ public void testPostgreSQLGetQueryParameters() {
...
+    assertFalse(
+        parser.parse(Statement.of("insert into x (a,b) values (1,2)")).hasReturningClause());
+    assertFalse(
+        parser
+            .parse(Statement.of("insert into x (a,b) values (1,2)thenreturn*"))
+            .hasReturningClause());
+  }
+
+  @Test
+  public void testPostgreSQLReturningClause() {
```
c2nes, line 1366: PostgreSQL supports some additional syntaxes for strings, [Dollar-Quoted String Constants](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-DOLLAR-QUOTING) in particular, that you may want to add test cases for. For instance,

```sql
UPDATE foo SET bar = $$RETURNING a, b, c$$ WHERE baz = 123
```

does not have a returning clause. Nor does,


```sql
UPDATE foo SET bar = $foobar$RETURNING a, b, c$foobar$ WHERE baz = 123
```

nor,

```sql
UPDATE foo SET bar = $returning$Hello, world!$returning$ WHERE baz = 123
```

rajatbhatta: @c2nes: Thanks for the test cases! Added these as well in https://github.com/googleapis/java-spanner/pull/1978/commits/9f59715e76511dceb43825a66c06f290978ed02e.

olavloite (approved):
@rajatbhatta Thanks for your patience with me for this PR. LGTM.

rajatbhatta:
@olavloite: Thank you for a very detailed review of this PR. :)

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/StatementParserTest.java
```diff
@@ -1266,6 +1266,197 @@ public void testPostgreSQLGetQueryParameters() {
...
```
bvandiver, line 1366: bvandiver=# insert into t1 select 2returning *;
 a 
---
 2
(1 row)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/AbstractStatementParser.java
```diff
@@ -522,4 +542,19 @@ static int countOccurrencesOf(char c, String string) {
...
+
+  /**
+   * Checks if the given SQL string contains a Returning clause. This method is used only in case of
+   * a DML statement.
+   *
+   * @param sql The sql string without comments that has to be evaluated.
+   * @return A boolean indicating whether the sql string has a Returning clause or not.
+   */
+  @InternalApi
+  abstract boolean checkReturningClauseInternal(String sql);
```
thiagotnunes, line 554: Can this be `protected`?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/PostgreSQLStatementParser.java
```diff
@@ -267,4 +290,38 @@ private void appendIfNotNull(
...
+  }
+
+  @InternalApi
+  @Override
+  boolean checkReturningClauseInternal(String rawSql) {
+    Preconditions.checkNotNull(rawSql);
+    String sql = rawSql.toLowerCase();
+    // Do a pre-check to check if the SQL string definitely does not have a returning clause.
+    // If this check fails, do a more involved check to check for a returning clause.
+    if (!sql.contains("returning")) {
```
thiagotnunes, line 313: nit: create a constant for "returning"

google-cloud-spanner/src/test/java/com/google/cloud/spanner/GceTestEnvConfig.java
```diff
@@ -58,8 +58,11 @@ public class GceTestEnvConfig implements TestEnvConfig {
   private final SpannerOptions options;
 
   public GceTestEnvConfig() {
-    String projectId = System.getProperty(GCE_PROJECT_ID, "");
-    String serverUrl = System.getProperty(GCE_SERVER_URL, "");
+    // TODO: Revert back any changes to this file before merging.
+    // String projectId = System.getProperty(GCE_PROJECT_ID, "");
+    String projectId = "span-cloud-testing";
+    // String serverUrl = System.getProperty(GCE_SERVER_URL, "");
```
thiagotnunes, line 64: It seems we forgot to clean this file?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/IntegrationTestEnv.java
```diff
@@ -91,7 +91,9 @@ protected void before() throws Throwable {
     this.config.setUp();
 
     SpannerOptions options = config.spannerOptions();
-    String instanceProperty = System.getProperty(TEST_INSTANCE_PROPERTY, "");
+    // TODO: Revert back any changes to this file before merging.
```
thiagotnunes, line 94: It seems we forgot to clean this file?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/AbstractStatementParser.java
rajatbhatta: Makes sense. Done.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/GceTestEnvConfig.java
rajatbhatta: I'll clean it before merging. Have added a TODO here.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/IntegrationTestEnv.java
rajatbhatta: I'll clean it before merging. Have added a TODO here.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/StatementParserTest.java
rajatbhatta: Thanks for this test case. I've handled this scenario by checking if the substring containing returning doesn't start with a valid first character of an identifier. In that case, the substring can be divided further into substrings. Have added more tests around that as well.

rajatbhatta:
@olavloite: Can you take a quick look at the changes in https://github.com/googleapis/java-spanner/pull/1978/commits/e2a463e4f91c602cd4a237634f557bcf4793de21? Thanks!

thiagotnunes (approved):

olavloite:
> @olavloite: Can you take a quick look at the changes in [e2a463e](https://github.com/googleapis/java-spanner/commit/e2a463e4f91c602cd4a237634f557bcf4793de21)? Thanks!

Can we also add a test for `insert into t1 select 10.returning*` (this is equivalent with `insert into t1 select 10.0 returning *`)

rajatbhatta:
> > @olavloite: Can you take a quick look at the changes in [e2a463e](https://github.com/googleapis/java-spanner/commit/e2a463e4f91c602cd4a237634f557bcf4793de21)? Thanks!
> 
> Can we also add a test for `insert into t1 select 10.returning*` (this is equivalent with `insert into t1 select 10.0 returning *`)

Added in https://github.com/googleapis/java-spanner/pull/1978/commits/60fcb4cdfb47fdc64924bff914e35d5cd454aa3d.

olavloite:
> 60fcb4c

Thanks. LGTM (once CI errors have been cleared)

PR #1986, opened 2022-08-29
test: Deleting stale encrypted instances

### LINKED ISSUE #1966: Encrypted test instance is not always deleted after test run
[This line](https://github.com/googleapis/java-spanner/blob/787ccadcba01193d541bfd1b80b055fb5d4c2bb3/samples/snippets/src/test/java/com/example/spanner/SpannerSampleIT.java#L458) will not be executed if the test is killed, for example if the test times out. That will leave the encrypted test instance lingering around. We should therefore add a cleanup task at the start of the test run to remove any old instances still hanging around.

gauravpurohit06:
Fixes #1966

### DIFF
--- a/samples/snippets/src/test/java/com/example/spanner/SpannerSampleIT.java
+++ b/samples/snippets/src/test/java/com/example/spanner/SpannerSampleIT.java
@@ -70,7 +70,7 @@ public class SpannerSampleIT {
   private static final String encryptedDatabaseId = formatForTest(baseDbId);
   private static final String encryptedBackupId = formatForTest(baseDbId);
   private static final String encryptedRestoreId = formatForTest(baseDbId);
-  private static final long SECONDS_IN_ONE_DAY = 86400;
+  private static final long SECONDS_IN_ONE_DAY = TimeUnit.SECONDS.convert(24L, TimeUnit.HOURS);
   static Spanner spanner;
   static DatabaseId dbId;
   static DatabaseAdminClient dbClient;
--- a/samples/snippets/src/test/java/com/example/spanner/SpannerSampleIT.java
+++ b/samples/snippets/src/test/java/com/example/spanner/SpannerSampleIT.java
@@ -26,18 +26,22 @@ import com.google.cloud.spanner.Database;
 import com.google.cloud.spanner.DatabaseAdminClient;
 import com.google.cloud.spanner.DatabaseId;
 import com.google.cloud.spanner.ErrorCode;
+import com.google.cloud.spanner.Instance;
 import com.google.cloud.spanner.InstanceAdminClient;
 import com.google.cloud.spanner.InstanceConfigId;
 import com.google.cloud.spanner.InstanceId;
 import com.google.cloud.spanner.InstanceInfo;
+import com.google.cloud.spanner.Options;
 import com.google.cloud.spanner.Spanner;
 import com.google.cloud.spanner.SpannerException;
 import com.google.cloud.spanner.SpannerExceptionFactory;
 import com.google.cloud.spanner.SpannerOptions;
 import com.google.common.base.Preconditions;
+import com.google.common.collect.Lists;
 import com.google.common.util.concurrent.Uninterruptibles;
 import java.io.ByteArrayOutputStream;
 import java.io.PrintStream;
+import java.util.List;
 import java.util.UUID;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Predicate;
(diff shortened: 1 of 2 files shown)


samples/snippets/src/test/java/com/example/spanner/SpannerSampleIT.java
```diff
@@ -93,10 +97,45 @@ public static void setUp() throws Exception {
...
+                .getInstanceAdminClient()
+                .listInstances(Options.filter("name:encrypted-test-"))
+                .iterateAll());
+
+    Timestamp now = Timestamp.now();
+
+    for (Instance instance : instances) {
+      if (TimeUnit.HOURS.convert(
+              now.getSeconds() - instance.getCreateTime().getSeconds(), TimeUnit.SECONDS)
+          > 24) {
```
rajatbhatta, line 132: Converting from a smaller unit(SECONDS) to a larger unit(HOUR), will lead to loss of precision. For example, 3600 to 7199 SECONDS will all return 1 HOUR.
All instances that got created between -24th hour to -25th hour from now will not get cleaned up here.

We can do either of the below:
- Change `> 24` to `>= 24`.
- 
```suggestion
      if (now.getSeconds() - instance.getCreateTime().getSeconds()
          > TimeUnit.SECONDS.convert(24L, TimeUnit.HOURS)) {
```
I think the second option would be better as we are not losing any precision whatsoever. We could have the threshold time `TimeUnit.SECONDS.convert(24L, TimeUnit.HOURS))` as a constant at the start of the function.

samples/snippets/src/test/java/com/example/spanner/SpannerSampleIT.java
```diff
@@ -93,10 +97,45 @@ public static void setUp() throws Exception {
...
+
+  /**
+   * Deleting all the test instances with name beginning from 'encrypted-test-' and were created
+   * before 24 hours.
+   *
+   * @throws InterruptedException If Thread.sleep() interrupted
+   */
+  static void deleteStaleTestEncryptedInstances() throws InterruptedException {
+    List<Instance> instances =
+        Lists.newArrayList(
```
rajatbhatta, line 121: Instead of creating a new list here using Lists.newArrayList(), we could directly iterate over spanner.getInstanceAdminClient().listInstances(Options.filter("name:encrypted-test-")).iterateAll() as follows:

```
for (Instance instance :
    spanner
        .getInstanceAdminClient()
        .listInstances(Options.filter("name:encrypted-test-"))
        .iterateAll()) {
```

rajatbhatta (changes requested):

samples/snippets/src/test/java/com/example/spanner/SpannerSampleIT.java
```diff
@@ -93,10 +97,45 @@ public static void setUp() throws Exception {
...
+    deleteStaleTestEncryptedInstances();
+  }
+
+  /**
+   * Deleting all the test instances with name beginning from 'encrypted-test-' and were created
+   * before 24 hours.
+   *
+   * @throws InterruptedException If Thread.sleep() interrupted
+   */
+  static void deleteStaleTestEncryptedInstances() throws InterruptedException {
```
rajatbhatta, line 119: Can we change this method to `private`, and change the method name to `deleteStaleEncryptedTestInstances` (in line with the name of the instances) ?
```suggestion
  private static void deleteStaleEncryptedTestInstances() throws InterruptedException {
```

samples/snippets/src/test/java/com/example/spanner/SpannerSampleIT.java
gauravpurohit06: We can better define a long final constant at class level for SECONDS_IN_ONE_DAY and use it.

` private static final long SECONDS_IN_ONE_DAY = 86400;`

rajatbhatta: IMO, having `TimeUnit.SECONDS.convert(24L, TimeUnit.HOURS))` as a constant will be better as it is more flexible to make changes to it in the longer term. Say in the future, we had to change the number of hours (to say, 12 hours), we would have to make only a small change (`24L` -> `12L`).

Also, I think converting from a larger unit(HOURS) to a smaller unit(SECONDS) won't be a problem as we won't be at a risk of overflow here (as the number of seconds in our use case won't exceed the range of long).

samples/snippets/src/test/java/com/example/spanner/SpannerSampleIT.java
```diff
@@ -93,10 +96,40 @@ public static void setUp() throws Exception {
...
+     * Delete stale instances that have been created earlier by this test but not deleted.
+     * Backups needed to be deleted from the instance first, as the instance can only be
+     * deleted once all backups have been deleted.
+     * */
+    deleteStaleEncryptedTestInstances();
   }
-  
+
+  /**
+   * Deleting all the test instances with name beginning from 'encrypted-test-' and were created
```
rajatbhatta, line 113:
```suggestion
   * Deleting all the test instances with name starting with 'encrypted-test-' and were created
```
Minor nit

samples/snippets/src/test/java/com/example/spanner/SpannerSampleIT.java
gauravpurohit06: Converting from larger unit to smaller unit won't be a problem for sure, but I think defining primitive constants will avoid unnecessary calculation on the fly but as it's only limited to Integration tests and if readability is getting affected we can also define it using `TimeUnit`.

rajatbhatta: I agree that using `TimeUnit.convert` can lead to some unnecessary calculation on the fly, but as these are integration tests and this threshold for cleanup is prone to changes with time, having `TimeUnit.convert` will give us more flexibility in terms of future changes.

Thanks for having this discussion offline as well.

samples/snippets/src/test/java/com/example/spanner/SpannerSampleIT.java
```diff
@@ -68,6 +70,7 @@ public class SpannerSampleIT {
   private static final String encryptedDatabaseId = formatForTest(baseDbId);
   private static final String encryptedBackupId = formatForTest(baseDbId);
   private static final String encryptedRestoreId = formatForTest(baseDbId);
+  private static final long SECONDS_IN_ONE_DAY = TimeUnit.SECONDS.convert(24L, TimeUnit.HOURS);
```
rajatbhatta, line 73: Does it make sense to rename `SECONDS_IN_ONE_DAY` to `STALE_INSTANCE_THRESHOLD_SECS`? In that case, if we need to change the number of hours to some other value in the future, we need not have to change this variable name.

rajatbhatta (approved):
LGTM, except just one comment https://github.com/googleapis/java-spanner/pull/1986/files#r960443381. Thanks for working on this.

PR #2007, opened 2022-09-13
samples: DML with Returning clause support

rajatbhatta:
This PR showcases how users can execute DML statements with returning clause using the Java Spanner client library, for both GoogleStandardSql and PostgreSql dialects.


samples/snippets/src/main/java/com/example/spanner/PgSpannerSample.java
```diff
@@ -577,6 +577,86 @@ static void writeUsingDml(DatabaseClient dbClient) {
   }
   // [END spanner_postgresql_dml_getting_started_insert]
 
+  // [START spanner_postgresql_insert_dml_returning]
```
olavloite, line 580: The preferred format for new samples is to put these in separate files, and not to add them to the large monolithic sample file. See for example `PgBatchDmlSample`.

samples/snippets/src/main/java/com/example/spanner/PgSpannerSample.java
```diff
@@ -577,6 +577,86 @@ static void writeUsingDml(DatabaseClient dbClient) {
...
+    dbClient
+        .readWriteTransaction()
+        .run(transaction -> {
+          String sql =
+              "INSERT INTO Singers (SingerId, FirstName, LastName) VALUES "
+                  + "(12, 'Melissa', 'Garcia'), "
+                  + "(13, 'Russell', 'Morales'), "
+                  + "(14, 'Jacqueline', 'Long'), "
+                  + "(15, 'Dylan', 'Shaw') RETURNING FullName";
+          try (ResultSet resultSet = transaction.executeQuery(Statement.of(sql))) {
```
olavloite, line 596: nit: I think it's worth adding a small comment in the code both here and in the other samples that users must use `transaction.executeQuery(..)` for DML with `RETURNING` clauses.

olavloite (commented):
Looks generally good to me. But aren't new samples supposed to go into separate files?

Also, I assume that we will be adding integration tests for these samples?

samples/snippets/src/main/java/com/example/spanner/PgSpannerSample.java
rajatbhatta: Done.

samples/snippets/src/main/java/com/example/spanner/PgSpannerSample.java
rajatbhatta: Done.

rajatbhatta:
@olavloite: I've made the changes based on the comments. Can you please take a look?

samples/snippets/src/test/java/com/example/spanner/DmlReturningSampleIT.java
```diff
@@ -0,0 +1,137 @@
...
+                .set("MarketingBudget")
+                .to(20000L)
+                .build()));
+  }
+
+  @After
+  public void removeTestData() {
+    final DatabaseClient client = spanner.getDatabaseClient(databaseId);
+    client.write(Collections.singletonList(Mutation.delete("Singers", KeySet.all())));
+    client.write(Collections.singleton(Mutation.delete("Albums", KeySet.all())));
```
olavloite, line 99: super-nit: This is not necessary, as Albums is interleaved in Singers with `ON DELETE CASCADE`.

samples/snippets/src/test/java/com/example/spanner/PgDmlReturningSampleIT.java
```diff
@@ -0,0 +1,150 @@
...
+                .set("MarketingBudget")
+                .to(20000L)
+                .build()));
+  }
+
+  @After
+  public void removeTestData() {
+    final DatabaseClient client = spanner.getDatabaseClient(databaseId);
+    client.write(Collections.singletonList(Mutation.delete("Singers", KeySet.all())));
+    client.write(Collections.singleton(Mutation.delete("Albums", KeySet.all())));
```
olavloite, line 112: nit: can also be removed

olavloite (approved):

samples/snippets/src/test/java/com/example/spanner/DmlReturningSampleIT.java
rajatbhatta: Thanks for pointing this out! Removed this line of code.

rajatbhatta:
@olavloite: Thank you for reviewing this PR :)

PR #2010, opened 2022-09-14
fix: retries of updates in the Connection API ignored analyze mode

### LINKED ISSUE #2009: analyzeUpdate is executed as executeUpdate during retry in the Connection API
If `analyzeUpdate(Statement)` is executed in a read/write transaction in the Connection API, and the read/write transaction is aborted, the retry will retry the statement as `executeUpdate(Statement)` instead of `analyzeUpdate(Statement)`. This will fail, either because the update count that is returned is different, or because the statement contained parameters and these were not bound for the analyze.

olavloite:
Retries of read/write transactions in the Connection API did not respect the analyze mode of update statements. This would cause updates to be executed using AnalyzeMode.NORMAL during retries, regardless of what was used during the initial attempt.

Fixes #2009

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/RetriableUpdate.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/RetriableUpdate.java
@@ -41,11 +41,9 @@ final class RetriableUpdate implements RetriableStatement {
       AnalyzeMode analyzeMode,
       long updateCount,
       UpdateOption... options) {
-    Preconditions.checkNotNull(transaction);
-    Preconditions.checkNotNull(statement);
-    this.transaction = transaction;
-    this.statement = statement;
-    this.analyzeMode = analyzeMode;
+    this.transaction = Preconditions.checkNotNull(transaction);
+    this.statement = Preconditions.checkNotNull(statement);
+    this.analyzeMode = Preconditions.checkNotNull(analyzeMode);
     this.updateCount = updateCount;
     this.options = options;
   }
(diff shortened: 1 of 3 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/RetriableUpdate.java
```diff
@@ -54,7 +57,15 @@ public void retry(AbortedException aborted) throws AbortedException {
       transaction
           .getStatementExecutor()
           .invokeInterceptors(statement, StatementExecutionStep.RETRY_STATEMENT, transaction);
-      newCount = transaction.getReadContext().executeUpdate(statement.getStatement(), options);
+      if (analyzeMode == null || analyzeMode == AnalyzeMode.NONE) {
```
rajatbhatta, line 60: I think we can check only for `AnalyzeMode.NONE` here, as we never pass in `analyzeMode` as `null`. We could also do a `Preconditions.checkNotNull(analyzeMode)` in the constructor.
```suggestion
      if (analyzeMode == AnalyzeMode.NONE) {
```

We should either do the above or add the `null` handling to below as well (as in the current state of code, it would not call `executeUpdate` for `null` analyzeMode case): https://github.com/googleapis/java-spanner/blob/1974f924eb2c4c9aea2638100d71fa1cace9dc17/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ReadWriteTransaction.java#L431.

rajatbhatta (approved):
LGTM.

PR #2095, opened 2022-10-05
test: create prototype for values with unrecognized types

olavloite:
Prototype for accepting values of unrecognized type and returning these as strings.


olavloite:
Closing, this was implemented in #2219

PR #2107, opened 2022-10-12
refactor: Removing unused method

gauravpurohit06:


gauravpurohit06:
Method is required for serialisation. 
Link: https://docs.oracle.com/en/java/javase/11/docs/specs/serialization/output.html#the-writereplace-method

PR #2117, opened 2022-10-19
feat: enable client to server compression

olavloite:
Enable compression of network traffic from client to server.

See the below screenshots from Wireshark for tests that show that the compression works both ways. Note: The rate of compression will depend on the type of data that is being sent. The below example for client->server communication is for sending 1,000 very similar mutations to Spanner. That compresses very well with gzip. More random data will compress less.

### Server -> Client - No Compression
server-_client identity

### Server -> Client - Gzip Compression
server-_client gzip

### Client -> Server - No Compression
client-_server identity

### Client -> Server - Gzip Compression
client-_server gzip


rajatbhatta (approved):
I don't think we have any unit tests to check if server -> client compression is set. Does it make sense to add tests for that case?

Apart from that, the changes looks good to me.

olavloite:
> I don't think we have any unit tests to check if server -> client compression is set. Does it make sense to add tests for that case?
> 
> Apart from that, the changes looks good to me.

Thanks for the quick review. I've added a unit test to verify that the header for server -> client compression is included in the request.

PR #2131, opened 2022-10-28
feat: support PostgreSQL END statement

### LINKED ISSUE #2130: Support the PostgreSQL END transaction statement
PostgreSQL supports both `COMMIT` and `END` as valid commands to successfully end a transaction. The Connection API currently only supports `COMMIT`.

See also https://www.postgresql.org/docs/current/sql-end.html

olavloite:
Adds support for the PostgreSQL END statement in the Connection API.

See also https://www.postgresql.org/docs/current/sql-end.html

Fixes #2130


rajatbhatta:
@olavloite: Generally looks good to me except that I have one question. How are we updating the `.sql` files under main/resources, and test/resources? Do we use any automated way to generate/change these?

olavloite:
> @olavloite: Generally looks good to me except that I have one question. How are we updating the `.sql` files under main/resources, and test/resources? Do we use any automated way to generate/change these?

Yes, those files are automatically generated from the `ClientSideStatements.json` file (specifically from the `exampleStatements` elements in that JSON file). So whenever we add a new (simple) statement to one of these files, we should:
1. Add a number examples for the statement to the JSON file.
2. Run the [command listed here](https://github.com/googleapis/java-spanner/blob/d0195b45423b73969636bc911980613a46dffa97/google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplGeneratedSqlScriptTest.java#L39). That will generate both valid and invalid statements for the command and add these to the generated SQL test scripts.

rajatbhatta (approved):

PR #2143, opened 2022-11-03
refactor: Removing duplicated code to handle array values

gauravpurohit06:
Business logic to decode data types enclosed in an array  is duplicated at multiple places which is prone to error and bugs.


google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
```diff
@@ -540,88 +539,25 @@ private static Struct decodeStructValue(Type structType, ListValue structValue)
...
-          return Lists.transform(
-              listValue.getValuesList(),
-              input -> input.getKindCase() == KindCase.NULL_VALUE ? null : input.getBoolValue());
         case INT64:
           // For int64/float64 types, use custom containers.  These avoid wrapper object
           // creation for non-null arrays.
           return new Int64Array(listValue);
         case FLOAT64:
           return new Float64Array(listValue);
+        case BOOL:
```
olavloite, line 548: nit: I think it would be better to keep BOOL conversion as is. That conversion also does not really convert the array, but creates a 'view' of the current list and only looks at the value if someone tries to actually get one of the values in the list.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
```diff
@@ -540,88 +539,25 @@ private static Struct decodeStructValue(Type structType, ListValue structValue)
...
-              if (value.getKindCase() == KindCase.NULL_VALUE) {
-                list.add(null);
-              } else {
-                ListValue structValue = value.getListValue();
-                list.add(decodeStructValue(elementType, structValue));
-              }
-            }
-            return list;
-          }
+          return listValue.getValuesList().stream()
```
olavloite, line 558: Would it be possible to make this even more efficient by either:
1. Not actually converting the values, but creating a view of the underlying list, and do the conversion on the fly when an element in the list is requested (basically the same as what we are currently doing for boolean values)?
2. OR (even better): Create some kind of a view of the list that converts all the elements in the list the first time an element in the list is requested, and after that reuses the converted list.

The above would remove the current inefficiency that array values are __always__ converted when a row is consumed, even if the client application does not really get the value of the array column.

olavloite (commented):
Brilliant find. I have a couple of questions whether we could make this even more efficient, now that we are looking into this. Would you mind taking a look at that?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
gauravpurohit06: Option#1
It s definitely possible and we can follow the pattern similar to `boolean` conversion of creating a view of the list and then do the lazy conversion whenever required but downside with this approach is there is no persistence so transformation function will run every time a user is accessing any value from list doesn't matter if it has been accessed earlier. I don't think it's a problem for now as our transform function to `decodeValue` is not much expensive. 

Option#2
I don't think it's possible using some library, we need to write our own implementation of list and use it. We can extend [TransformingRandomAccessList](https://github.com/google/guava/blob/master/guava/src/com/google/common/collect/Lists.java#L593) to add desire functionality of transforming complete list when any element is accessed and persist the transform values.

olavloite: I'm leaning towards option 1 then in that case. Our `decodeValue` methods are generally speaking not very expensive, and the only time that a view solution would be more expensive than what we have now is if the application accesses all the elements in the list once + at least one element twice or more.

gauravpurohit06: @rajatbhatta and @ansh0l, Please chime in and provide your opinions on the same.

rajatbhatta: Hi, my preference would be **Option#1** because user experience would be better in that case. In a general case, an application would not be accessing the same array column value for a row repeatedly. Most likely, application would access each array column value only once.

ansh0l: Writing a separate library for the use case doesn't sound optimal, so I'm not in favor of option 2.

For option 1, I think, it depends on what we want to optimize for:

For optimizing the worst case scenario performance, it's better to let the current proposed approach stay and not make any changes.
For optimizing the average case scenario performance, option 1 makes sense.

In general, I've a bias here that we should worry more about the worst case scenario performance more than the average case scenario, since Spanner is used for high load/throughput applications. That argument can be flipped around to say that customers win the most time when we optimize the average case.

So, is there a way to let the user override the behavior for lazy conversion? If yes, we should evaluate that as well. If there is no such way, then let's go with the consensus of option 1.

olavloite: I think that (advanced) users have the option to override the default lazy conversion if we go for option 1 by just creating a copy of the list that we return. I expect that most users won't be able to see any difference in any case, and that those who do need optimizations for the non-average case would be able to look into the code and decide that they should just create a hard copy of the lazy list that we are providing. That would effectively be the same as providing an additional option for overriding the default, but without adding any additional fields or options to the API.

Maybe we should add a comment about the latter to the list methods. So something like: The list that is returned by this method is lazily constructed. Create a copy of it if you intend to access each element in the list multiple times.

ansh0l: Sounds Good Knut, @gauravpurohit06 if it looks good to you, please incorporate, and I think then we are good to go ahead.

gauravpurohit06: Thanks @olavloite, @ansh0l, @rajatbhatta  for review. We can go ahead with the lazy conversion and also update the function doc for list APIs to mention about lazy conversions.

rajatbhatta (approved):
LGTM. Thanks for working on this enhancement!

PR #2147, opened 2022-11-05
feat: adding samples for Jsonb data type

rajatbhatta:


samples/snippets/src/main/java/com/example/spanner/UpdateJsonbDataSample.java
```diff
@@ -0,0 +1,75 @@
...
+    String databaseId = "my-database";
+
+    try (Spanner spanner =
+        SpannerOptions.newBuilder().setProjectId(projectId).build().getService()) {
+      DatabaseClient client =
+          spanner.getDatabaseClient(DatabaseId.of(projectId, instanceId, databaseId));
+      updateJsonbData(client);
+    }
+  }
+
```
asthamohta, line 43: Can you add the following comment:
```
PG JSONB takes the last value in the case of duplicate keys.
PG JSONB sorts first by key length and then lexicographically with equivalent key length.
```
And can we duplicate any one key to show this?

samples/snippets/src/test/java/com/example/spanner/PgSpannerStandaloneExamplesIT.java
```diff
@@ -99,6 +100,7 @@ public static void createTestDatabase() throws Exception {
             "CREATE TABLE Venues ("
                 + "VenueId      bigint NOT NULL,"
                 + "Revenue      NUMERIC,"
+                + "VenueDetails JSONB,"
```
asthamohta, line 103: This column should be added later using alter and not before

rajatbhatta: VenueDetails column added here, is initially being dropped in `addJsonbColumn_shouldSuccessfullyAddColumn` integration test using `ALTER TABLE DROP COLUMN`, and then re-added using the sample. This is done so that all the JSONB integration tests can run independent of each other (for example, `queryWithJsonbParameter_shouldReturnResults` would not have `addJsonbColumn_shouldSuccessfullyAddColumn` as a pre-requisite in that case).

asthamohta: Ohh okay. Thanks

samples/snippets/src/main/java/com/example/spanner/UpdateJsonbDataSample.java
rajatbhatta: This is done.

asthamohta (approved):

PR #2156, opened 2022-11-09
feat: analyze update returns param types

olavloite:
Add the method analyzeUpdateStatement that returns the undeclared
parameters in an update statement. This allows connection based APIs
to return more metadata for a statement than is currently possible:
1. JDBC should return the parameter data types when PreparedStatement#getMetaData()
   is called.
2. PGAdapter should return the parameter types when a DescribeStatement
   message is received.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/ResultSets.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/ResultSets.java
@@ -29,6 +29,7 @@ import com.google.common.base.Preconditions;
 import com.google.common.base.Supplier;
 import com.google.common.collect.Lists;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
+import com.google.spanner.v1.ResultSetMetadata;
 import com.google.spanner.v1.ResultSetStats;
 import java.math.BigDecimal;
 import java.util.List;
@@ -156,6 +157,11 @@ public final class ResultSets {
     public ResultSetStats getStats() {
       throw new UnsupportedOperationException(
           "ResultSetStats are available only for results returned from analyzeQuery() calls");
+    }
+
+    @Override
+    public ResultSetMetadata getMetadata() {
+      throw new UnsupportedOperationException();
     }
 
     @Override
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/ResultSets.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/ResultSets.java
@@ -161,7 +161,8 @@ public final class ResultSets {
 
     @Override
     public ResultSetMetadata getMetadata() {
-      throw new UnsupportedOperationException();
+      throw new UnsupportedOperationException(
+          "ResultSetMetadata are available only for results that were returned from Cloud Spanner");
     }
 
     @Override
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/Connection.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/Connection.java
@@ -29,6 +29,7 @@ import com.google.cloud.spanner.ErrorCode;
 import com.google.cloud.spanner.Mutation;
 import com.google.cloud.spanner.Options.QueryOption;
 import com.google.cloud.spanner.Options.RpcPriority;
+import com.google.cloud.spanner.Options.UpdateOption;
 import com.google.cloud.spanner.ReadContext.QueryAnalyzeMode;
 import com.google.cloud.spanner.ResultSet;
 import com.google.cloud.spanner.SpannerBatchUpdateException;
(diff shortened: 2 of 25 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/Connection.java
```diff
@@ -968,11 +969,29 @@ default RpcPriority getRPCPriority() {
    * the statement. {@link com.google.cloud.spanner.ReadContext.QueryAnalyzeMode#PROFILE} executes
    * the DML statement, returns the modified row count and execution statistics, and the effects of
    * the DML statement will be visible to subsequent operations in the transaction.
+   *
+   * @deprecated Use {@link #analyzeUpdateStatement(Statement, QueryAnalyzeMode, UpdateOption...)}
```
rajatbhatta, line 973:
```suggestion
   * @deprecated Use {@link #analyzeUpdateStatement(Statement, QueryAnalyzeMode, UpdateOption...)}
                instead.
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ReadWriteTransaction.java
```diff
@@ -431,25 +434,27 @@ private ApiFuture<ResultSetStats> internalExecuteUpdateAsync(
...
-                          updateCount =
+                          ResultSet resultSet =
                               get(txContextFuture)
-                                  .analyzeUpdate(
+                                  .analyzeUpdateStatement(
                                       update.getStatement(),
                                       analyzeMode.getQueryAnalyzeMode(),
                                       options);
+                          updateCount =
+                              resultSet.getStats() == null
```
rajatbhatta, line 451: When will `resultSet.getStats()` be `null`? Should we assume `updateCount` as `0` in that case?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ReadWriteTransaction.java
```diff
@@ -396,26 +396,29 @@ public void onSuccess(ResultSet result) {}
...
       final ParsedStatement update, final UpdateOption... options) {
     return ApiFutures.transform(
         internalExecuteUpdateAsync(update, AnalyzeMode.NONE, options),
-        ResultSetStats::getRowCountExact,
+        Tuple::x,
         MoreExecutors.directExecutor());
   }
 
-  private ApiFuture<ResultSetStats> internalExecuteUpdateAsync(
+  private ApiFuture<Tuple<Long, ResultSet>> internalExecuteUpdateAsync(
```
rajatbhatta, line 416: I understand that this is internal, but can we add a comment somewhere to make it clear that we are setting only one of `Tuple::x` and `Tuple::y` and not both (even though both are available).

rajatbhatta (changes requested):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/ResultSets.java
```diff
@@ -158,6 +159,11 @@ public ResultSetStats getStats() {
           "ResultSetStats are available only for results returned from analyzeQuery() calls");
     }
 
+    @Override
+    public ResultSetMetadata getMetadata() {
+      throw new UnsupportedOperationException();
```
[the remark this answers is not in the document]
olavloite: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/Connection.java
olavloite: Good point, done. (note, Javadoc block elements that only contain one sentence should not end with a punctuation mark)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ReadWriteTransaction.java
olavloite: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ReadWriteTransaction.java
olavloite: In this case it should never be null, but IntelliJ complains because `getStats()` is annotated as `@Nullable`. I've changed it to use a `requireNonNull` method call instead, which also silences the warning in IntelliJ.

rajatbhatta: Got it, thanks!

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ReadWriteTransaction.java
rajatbhatta: That's a very detailed comment, thank you for it!

rajatbhatta (approved):
LGTM.

PR #2186, opened 2022-12-05
fix: use a proper endpoint for DirectPath tests

ko3a4ok:


rajatbhatta:
Hi @ko3a4ok: Just had one question. Why are we setting the endpoint to wrenchworks-nonprod? Is it because this endpoint receives SpanFE releases early, and we would want to catch issues early on in the release cycle?

ko3a4ok:
> Hi @ko3a4ok: Just had one question. Why are we setting the endpoint to wrenchworks-nonprod? Is it because this endpoint receives SpanFE releases early, and we would want to catch issues early on in the release cycle?

@rajatbhatta this is a canary endpoint for prod SpanFE servers. It receives the binary release earlier than the rest of prod Spanner frontend servers. We're using a different set of probers to detect any regressions for DirectPath on the server side. This endpoint(wrenchworks-nonprod) will help us to detect regressions on the client side(Spanner lib, or grpc, rls, xds etc). So, technically, we can use any Spanner endpoint. But spanner.googleapis.com isn't available publicly with DirectPath, but wrenchworks-nonprod.googleapis.com has been setup and tested with DirectPath for a while

rajatbhatta:
> > Hi @ko3a4ok: Just had one question. Why are we setting the endpoint to wrenchworks-nonprod? Is it because this endpoint receives SpanFE releases early, and we would want to catch issues early on in the release cycle?
> 
> @rajatbhatta this is a canary endpoint for prod SpanFE servers. It receives the binary release earlier than the rest of prod Spanner frontend servers. We're using a different set of probers to detect any regressions for DirectPath on the server side. This endpoint(wrenchworks-nonprod) will help us to detect regressions on the client side(Spanner lib, or grpc, rls, xds etc). So, technically, we can use any Spanner endpoint. But spanner.googleapis.com isn't available publicly with DirectPath, but wrenchworks-nonprod.googleapis.com has been setup and tested with DirectPath for a while

Thanks for clarifying! LGTM.

rajatbhatta (approved):

PR #2195, opened 2022-12-13
chore: make Partition class getters public

rajatbhatta:
Make getters in Partition class public so that they can be accessed outside com.google.cloud.spanner package.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/Partition.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/Partition.java
@@ -16,7 +16,6 @@
 
 package com.google.cloud.spanner;
 
-import com.google.common.collect.ImmutableList;
 import com.google.protobuf.ByteString;
 import java.io.Serializable;
 import java.util.Objects;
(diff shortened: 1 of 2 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/Partition.java
```diff
@@ -92,39 +92,39 @@ static Partition createQueryPartition(
...
+  public String getTable() {
     return table;
   }
 
-  KeySet getKeys() {
+  public KeySet getKeys() {
     return keys;
   }
 
-  Iterable<String> getColumns() {
```
olavloite, line 107: We should make this return a copy or something immutable (or otherwise make sure that the `Iterable` that we are returning is always immutable). Otherwise an external caller could by accident change the partition by changing this iterable.

olavloite (approved):
LGTM, with a small nit on the `getColumns()` method.

Why do we need to make this public?

rajatbhatta:
> Why do we need to make this public?

I only need getPartitionToken to be made public for some Google-internal tests to be run. There are some asserts which need to check for the partitionToken, and with default access specifier, it was not accessible outside com.google.cloud.spanner package.

I'll probably have the rest of the getters as default and will make them public in the future if need be.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Partition.java
rajatbhatta: Makes sense. I'll skip changing it to public for now. Will make the changes on a need basis at a later point.

PR #2212, opened 2022-12-28
test: Proto Column Integration tests

gauravpurohit06:
This PR includes Integration tests for following scenarios: 
-  Write & Read Proto Messages & Enums and array of it. 
- Check compatibility between ProtoMessage/Bytes & ProtoEnun/Int64. 
- Negative Case when incorrect Proto object is provided while reading the data.
-   Parameterised queries, Primary Keys and Indexes.

### DIFF
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITProtoColumnTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITProtoColumnTest.java
@@ -80,7 +80,7 @@ public class ITProtoColumnTest {
 
   @After
   public void after() throws Exception {
-    databaseClient.write(ImmutableList.of(Mutation.delete( "Types", KeySet.all())));
+    databaseClient.write(ImmutableList.of(Mutation.delete("Types", KeySet.all())));
     databaseClient.write(ImmutableList.of(Mutation.delete("Singers", KeySet.all())));
   }
 
@@ -130,7 +130,7 @@ public class ITProtoColumnTest {
 
     databaseClient.write(
         ImmutableList.of(
-            Mutation.newInsertOrUpdateBuilder( "Types")
+            Mutation.newInsertOrUpdateBuilder("Types")
                 .set("RowID")
                 .to(11)
                 .set("Int64a")
@@ -152,7 +152,7 @@ public class ITProtoColumnTest {
                 .build()));
 
     try (ResultSet resultSet =
-        databaseClient.singleUse().executeQuery(Statement.of("SELECT * FROM " +  "Types"))) {
+        databaseClient.singleUse().executeQuery(Statement.of("SELECT * FROM " + "Types"))) {
 
       resultSet.next();
       assertEquals(11, resultSet.getLong("RowID"));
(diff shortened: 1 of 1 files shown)


google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITProtoColumnTest.java
```diff
@@ -0,0 +1,331 @@
...
+public class ITProtoColumnTest {
+  private static String projectId;
+  private static String instanceId;
+  private static String databaseId;
+  private static DatabaseClient databaseClient;
+
+  @BeforeClass
+  public static void beforeClass() throws Exception {
+    assumeFalse(
+        "Proto Column is not supported in the emulator", EmulatorSpannerHelper.isUsingEmulator());
```
harshachinta, line 65: Do we need this here as it is also mentioned inside `@Test` methods below?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITProtoColumnTest.java
```diff
@@ -0,0 +1,331 @@
...
+        Arrays.asList(
+            singerInfoBytes,
+            null,
+            ByteArray.copyFrom(SingerInfo.getDefaultInstance().toByteArray()));
+
+    List<ProtocolMessageEnum> enumList = Arrays.asList(Genre.FOLK, null, Genre.ROCK);
+    List<Long> enumConstList =
+        Arrays.asList((long) Genre.FOLK_VALUE, null, (long) Genre.ROCK_VALUE);
+
+    databaseClient.write(
```
harshachinta, line 122: Check compatibility between Proto Messages & Bytes, Proto Enum & Int64 during write operation also

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITProtoColumnTest.java
```diff
@@ -0,0 +1,331 @@
...
+                          "INSERT INTO Singers (SingerId, FirstName, LastName, SingerInfo, SingerGenre) VALUES (1, \"FirstName1\", \"LastName1\", @singerInfo, @singerGenre)")
+                      .bind("singerInfo")
+                      .to(singerInfo1)
+                      .bind("singerGenre")
+                      .to(genre1)
+                      .build();
+
+              Statement statement2 =
+                  Statement.newBuilder(
+                          "INSERT INTO Singers (SingerId, FirstName, LastName, SingerInfo, SingerGenre) VALUES (2, \"FirstName2\", \"LastName2\", @singerInfo, @singerGenre)")
```
harshachinta, line 222: Same as above SingerId value

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITProtoColumnTest.java
gauravpurohit06: It's good to have it at both the places as `beforeClass` will be executed prior to test.  Going forward if we have some test which are supported we can remove it from here.

harshachinta (approved):
LGTM

PR #2214, opened 2022-12-29
feat: Leader Aware Routing

yifanzyifanz:
Add `x-goog-spanner-route-to-leader` header to Spanner RPC contexts for RW/PDML transactions.

The header is added to support leader-aware-routing feature, which aims at reducing cross-regional latency for RW/PDML transactions in a multi-region instance.

This feature is off by default and can be enabled by enabling Leader Aware Routing while creating Spanner object.

`Spanner spanner = SpannerOptions.newBuilder().enableLeaderAwareRouting().build().getService();`


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1155,6 +1158,15 @@ public Builder setEmulatorHost(String emulatorHost) {
       return this;
     }
 
+    /**
+     * Enable or disable leader aware routing. Leader aware routing would route all requests in
+     * RW/PDML transactions to the leader region.
+     */
+    public Builder setLeaderAwareRouting(boolean leaderAwareRoutingEnabled) {
```
rajatbhatta, line 1165: I think we should have a method for disabling leaderAwareRouting, and one for enabling leaderAwareRouting. As the default is true and we need a process for users to opt out of it, it would make more sense to have `disableLeaderAwareRouting()` method (which sets the flag to false), and a method enableLeaderAwareRouting(to set the flag back to true). This is to improve user experience with using `SpannerOptions`.

CC: @olavloite for any inputs on this.

google-cloud-spanner/clirr-ignored-differences.xml
```diff
@@ -222,4 +222,103 @@
...
+    <differenceType>7004</differenceType>
+    <className>com/google/cloud/spanner/spi/v1/GapicSpannerRpc</className>
+    <method>com.google.api.core.ApiFuture beginTransactionAsync(com.google.spanner.v1.BeginTransactionRequest, java.util.Map)</method>
+  </difference>
+  <difference>
+    <differenceType>7004</differenceType>
+    <className>com/google/cloud/spanner/spi/v1/GapicSpannerRpc</className>
+    <method>com.google.api.core.ApiFuture beginTransactionAsync(com.google.spanner.v1.BeginTransactionRequest, java.util.Map)</method>
+  </difference>
+  <difference>
```
rajatbhatta, line 240: Were these methods not already present (prior to this PR)? I felt only the ones which pass in an extra boolean for `routeToLeader` (thus updating the existing definitions) would need additions to this file?

rajatbhatta (changes requested):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerMetadataProvider.java
```diff
@@ -28,7 +28,7 @@
 class SpannerMetadataProvider {
   private final Map<Metadata.Key<String>, String> headers;
   private final String resourceHeaderKey;
-
+  private static final String routeToLeaderHeaderKey = "x-goog-spanner-route-to-leader";
```
olavloite, line 31: nit: Java `static final` fields are considered constants, and should be named using `UPPER_SNAKE_CASE`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerMetadataProvider.java
```diff
@@ -66,6 +66,12 @@ Map<String, List<String>> newExtraHeaders(
         .build();
   }
 
+  Map<String, List<String>> newRouteToLeaderHeader() {
+    return ImmutableMap.<String, List<String>>builder()
+        .put(routeToLeaderHeaderKey, Collections.singletonList("true"))
+        .build();
+  }
```
olavloite, line 73: You don't need to create a method for this. Instead, you can just create a `static final` field for this, as the contents of the map is always the same (and it is immutable). That saves the creation of a new map every time the option is used.

olavloite (approved):
LGTM, with a couple of nits.

Tip: You could add an extra verification to this test to verify that the header is really sent to Cloud Spanner: https://github.com/googleapis/java-spanner/blob/d44e468b7f29cb48bc3e12d12e00ff27a64713a7/google-cloud-spanner/src/test/java/com/google/cloud/spanner/ChannelUsageTest.java#L133

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionImpl.java
```diff
@@ -275,7 +275,7 @@ public AsyncTransactionManagerImpl transactionManagerAsync(TransactionOption...
   @Override
   public void prepareReadWriteTransaction() {
```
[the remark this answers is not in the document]
rajatbhatta: Sure, I'll take a look.

rajatbhatta (approved):
LGTM (apart from one of my pending comments and nits from @olavloite)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerMetadataProvider.java
yifanzyifanz: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
```diff
@@ -380,6 +383,7 @@ void initTransaction() {
   Span span;
   private final int defaultPrefetchChunks;
   private final QueryOptions defaultQueryOptions;
+  protected boolean routeToLeader = false;
```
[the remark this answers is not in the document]
yifanzyifanz: We actually allow subclasses to change the value depending on the txn type. i.e. If the subclass represents a RW txn context, then overwrite the value with true, otherwise keep it false.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerMetadataProvider.java
yifanzyifanz: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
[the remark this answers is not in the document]
olavloite: In that case I would say that it would be better to define a protected method `protected boolean isRouteToLeader()` that returns `false` in this base class, and override the method in the specific class(es) that should return `true`. That represents the actual situation better than a variable, as it forces the value to be fixed to either true or false depending on the class.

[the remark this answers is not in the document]
yifanzyifanz: Yes that is a good option. Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1155,6 +1158,24 @@ public Builder setEmulatorHost(String emulatorHost) {
...
+     * transactions to the leader region.
+     */
+    public Builder enableLeaderAwareRouting() {
+      this.leaderAwareRoutingEnabled = true;
+      return this;
+    }
+
+    /**
+     * Enable leader aware routing. Leader aware routing would route all requests in RW/PDML
+     * transactions to the leader region.
```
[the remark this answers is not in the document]
yifanzyifanz: Done.

rajatbhatta (changes requested):
Adding "Request changes" to prevent accidental merge of this PR. Note that the changes still look good to me.

rajatbhatta (approved):

rajatbhatta:
Changing the Title to make it consistent across clients.

PR #2219, opened 2023-01-05
feat: support UNRECOGNIZED types + decode BYTES columns lazily

olavloite:
### Unrecognized Types
Adds support for types that do not have a known type code in the client library (yet). This can happen if a new type is added to Cloud Spanner before this client is updated, or if someone is using an old version of the client library with a version of Cloud Spanner that contains a type that was not available at the time that the specific version of the client library was built.

Unrecognized types are assigned a type with type code `UNRECOGNIZED`. The values of these types (and all other types) can be retrieved as follows:
1. Calling `ResultSet#getValue(..)` will return the column value as a `Value` instance.
2. Calling `Value#getAsString()` will return a string representation of the given value. The string value will be a valid and complete representation of the underlying value. This method is guaranteed to work for all known and unknown types. Array types will return a comma-separated string enclosed in square brackets (e.g. `[true, NULL, false]` for a boolean array).
3. Calling `Value#getAsStringList()` will return a list of strings. The list of strings will contain:
   1. In case the value is an array: One string for each element in the array. Each element is a complete and valid string representation of the underlying array element value.
   2. In case the value is a non-array type: A singleton list containing only the string representation of the specific value.
   3. A user can always call `Value#getType()#getCode() == Code.ARRAY` to check whether a value is an array with one element, or whether it is a non-array type to determine what the reason is that `getAsStringList()` returns a list with one element.

### Decoding BYTES
BYTES columns are encoded as Base64 strings. Decoding these are relatively CPU-heavy, especially for large values. Decoding them is not always necessary if the user only needs the Base64 string. Also, the internally used Guava decoder is less efficient than JDK implementations that are available from Java 8 and onwards. This change therefore delays the decoding of BYTES columns until it is actually necessary, and then uses the JDK implementation instead of the Guava version. The JDK implementation in OpenJDK 17 uses approx 1/3 of the CPU cycles of the Guava version.

This feature is combined with unrecognized types so it can make use of the new `getAsString()` methods. This prevents the introduction of what is (theoretically) a breaking change. Currently, `Value#getString(..)` throws an exception if you try to call this on a `BYTES` value. This remains the case after this change, and instead users should call `Value#getAsString()` to get the underlying Base64 value of a `BYTES` column.


google-cloud-spanner/src/main/java/com/google/cloud/spanner/Type.java
```diff
@@ -378,8 +393,14 @@ public int getFieldIndex(String fieldName) {
...
 
   void toString(StringBuilder b) {
-    if (code == Code.ARRAY) {
-      b.append("ARRAY<");
+    if (code == Code.ARRAY || (proto != null && proto.hasArrayElementType())) {
+      if (code == Code.ARRAY) {
+        b.append("ARRAY<");
+      } else {
+        // This is very unlikely to happen. It would mean that we have introduced a type that
+        // is not an ARRAY, but does have an array element type.
```
aseering, line 401: Can the new PROTO type do this?  @charvisingla

aseering (commented):
Just did a quick scan and left a comment or two.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractStructReader.java
```diff
@@ -251,7 +251,6 @@ public Date getDate(String columnName) {
 
   @Override
   public Value getValue(int columnIndex) {
-    checkNonNull(columnIndex, columnIndex);
```
thiagotnunes, line 254: Is this a breaking change? We are modifying the contract here, I wonder if any customer is relying on this? I would imagine not, but double checking.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Type.java
```diff
@@ -204,13 +210,23 @@ private Type(
       Code code,
       @Nullable Type arrayElementType,
       @Nullable ImmutableList<StructField> structFields) {
+    this.proto = null;
```
thiagotnunes, line 213: Could we always populate the corresponding proto type?

thiagotnunes (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Type.java
olavloite: That's a little difficult. This constructor is used to create the fixed type constants for all known types in the client library. They do not have a proto definition available at construction time. See for example here: https://github.com/googleapis/java-spanner/blob/f637f1e5b595c97f4355ac7ab147aa0d261d20a5/google-cloud-spanner/src/main/java/com/google/cloud/spanner/Type.java#L49

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractStructReader.java
olavloite: Hmm... That's a good point. I don't think anyone would be relying on it, but you're right that it is a change of the contract. In this case we will always be returning a non-null value, even if the database value is `NULL` (we're returning a `Value` instance whose internal value is null and whose type is set to the type of the column), so it also a safe value to return. I think there are two (three) options here:
1. Accept the small change in contract.
2. Add an extra method named something like 'getValueOrNull`
3. Do a major version bump (but I don't feel like this feature is worth that)

@thiagotnunes @rajatbhatta @ansh0l Any specific thoughts on this?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/RandomResultSetGenerator.java
```diff
@@ -39,7 +39,7 @@
  * of Cloud Spanner filled with random data.
  */
 public class RandomResultSetGenerator {
-  private static Type[] generateTypes(Dialect dialect) {
+  public static Type[] generateAllTypes(Dialect dialect) {
```
rajatbhatta, line 42: Is this method meant to be used publicly? If not, can we have it annotated as `VisibleForTesting`, if it's meant to be used only in tests?

rajatbhatta (approved):
LGTM, apart from a few nits or questions. Thank you @olavloite for working on this!

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Type.java
```diff
@@ -204,13 +210,23 @@ private Type(
...
     this.arrayElementType = arrayElementType;
     this.structFields = structFields;
   }
 
+  private Type(@Nonnull com.google.spanner.v1.Type proto) {
+    this.proto = proto;
+    this.code = Code.UNRECOGNIZED;
+    this.arrayElementType =
+        proto.hasArrayElementType() ? new Type(proto.getArrayElementType()) : null;
+    this.structFields = null;
```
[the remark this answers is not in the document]
olavloite: Good point, that cleans it up a bit. Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
```diff
@@ -353,6 +361,79 @@ private boolean isMergeable(KindCase kind) {
...
 
+  static final class LazyByteArray implements Serializable {
+    private static final Base64.Encoder ENCODER = Base64.getEncoder();
+    private static final Base64.Decoder DECODER = Base64.getDecoder();
+    private final String base64String;
+    private transient AbstractLazyInitializer<ByteArray> byteArray;
+
+    LazyByteArray(@Nonnull String base64String) {
+      this.base64String = Preconditions.checkNotNull(base64String);
+      byteArray = defaultInitializer();
```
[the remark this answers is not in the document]
olavloite: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
```diff
@@ -353,6 +361,79 @@ private boolean isMergeable(KindCase kind) {
...
+    private AbstractLazyInitializer<ByteArray> defaultInitializer() {
+      return new AbstractLazyInitializer<ByteArray>() {
+        @Override
+        protected ByteArray initialize() {
+          return ByteArray.copyFrom(DECODER.decode(base64String));
+        }
+      };
+    }
+
+    private void readObject(java.io.ObjectInputStream in)
```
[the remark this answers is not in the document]
olavloite: This is part of the custom serializable implementation for this class. Java allows a class to control its own (de)serialization through this method. See https://docs.oracle.com/javase/7/docs/platform/serialization/spec/input.html#2971

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
```diff
@@ -671,13 +772,17 @@ protected Value getValueInternal(int columnIndex) {
...
+          return Value.internalBytes(isNull ? null : getLazyBytesInternal(columnIndex));
         case TIMESTAMP:
           return Value.timestamp(isNull ? null : getTimestampInternal(columnIndex));
         case DATE:
           return Value.date(isNull ? null : getDateInternal(columnIndex));
         case STRUCT:
           return Value.struct(isNull ? null : getStructInternal(columnIndex));
+        case UNRECOGNIZED:
+          return Value.unrecognized(
+              isNull ? NULL_VALUE : (com.google.protobuf.Value) rowData.get(columnIndex),
```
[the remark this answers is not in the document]
olavloite: Good point, that makes this more consistent. Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Value.java
```diff
@@ -1661,6 +1830,16 @@ com.google.protobuf.Value valueToProto() {
       return com.google.protobuf.Value.newBuilder().setListValue(list).build();
     }
 
+    @Nonnull
+    @Override
+    public ImmutableList<String> getAsStringList() {
+      ImmutableList.Builder<String> builder = ImmutableList.builder();
+      for (int i = 0; i < value.size(); i++) {
+        builder.add(value.get(i) == null ? NULL_STRING : elementToString(value.get(i)));
```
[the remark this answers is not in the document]
olavloite: Ah, thanks! I didn't notice that this class was an iterable. Fixed.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/RandomResultSetGenerator.java
olavloite: It is indeed only intended for testing, but I don't think we need to add the annotation in this case, as it is only defined in `src/test/java`, meaning that it will also only be available to test code.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/RandomResultSetGenerator.java
```diff
@@ -109,7 +109,7 @@ private static Type[] generateTypes(Dialect dialect) {
     };
   }
 
-  private static ResultSetMetadata generateMetadata(Type[] types) {
+  public static ResultSetMetadata generateAllTypesMetadata(Type[] types) {
```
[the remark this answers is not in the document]
olavloite: Same as above.

olavloite:
CI failure for Windows is a setup failure for that CI run and not related to this change.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
```diff
@@ -2381,4 +2393,565 @@ public void testAnalyzeUpdateStatement() {
...
+                .setMetadata(
+                    ResultSetMetadata.newBuilder()
+                        .setRowType(
+                            StructType.newBuilder()
+                                .addFields(
+                                    Field.newBuilder()
+                                        .setName("c")
+                                        .setType(
+                                            Type.newBuilder()
+                                                .setCodeValue(Integer.MAX_VALUE)
```
olavloite, line 2731: This line sets a type code value that does not exist, and will be seen as an unrecognized type in the client library

PR #2247, opened 2023-01-21
test: fix restore database unit test in google-cloud-spanner

harshachinta:


harshachinta:
EncryptionConfig need not be explicitly set when making a call to restoreDatabase rpc. If it is not specified the restored database will use the same encryption configuration as the backup by default.

PR #2277, opened 2023-02-13
feat: add support for Proto Columns DDL

harshachinta:
This PR adds feature support for Proto Columns DDL with supporting integration tests.


google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseAdminClient.java
```diff
@@ -407,6 +409,10 @@ OperationFuture<Void, UpdateDatabaseDdlMetadata> updateDatabaseDdl(
       @Nullable String operationId)
       throws SpannerException;
 
+  OperationFuture<Void, UpdateDatabaseDdlMetadata> updateDatabaseDdl(
```
olavloite, line 412: nit: add Javadoc

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseAdminClient.java
```diff
@@ -434,6 +440,8 @@ OperationFuture<Void, UpdateDatabaseDdlMetadata> updateDatabaseDdl(
    */
   List<String> getDatabaseDdl(String instanceId, String databaseId);
 
+  GetDatabaseDdlResponse getDatabaseDdlResponse(String instanceId, String databaseId);
```
olavloite, line 443: nit: add Javadoc

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Database.java
```diff
@@ -88,6 +88,11 @@ public OperationFuture<Void, UpdateDatabaseDdlMetadata> updateDdl(
     return dbClient.updateDatabaseDdl(instance(), database(), statements, operationId);
   }
 
+  public OperationFuture<Void, UpdateDatabaseDdlMetadata> updateDdl(
```
olavloite, line 91: nit: add Javadoc

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Type.java
```diff
@@ -46,6 +46,7 @@
  */
 @Immutable
 public final class Type implements Serializable {
+
```
olavloite, line 49: nit: (here and elsewhere) remove unrelated formatting changes

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseInfo.java
```diff
@@ -141,6 +155,29 @@ public Builder setDialect(Dialect dialect) {
...
+      this.protoDescriptors = ByteString.copyFrom(byteArray);
+      return this;
+    }
+
+    @Override
+    public Builder setProtoDescriptors(String filePath) throws IOException {
+      Preconditions.checkState(filePath.length() != 0, "Input File Path cannot be empty.");
+      InputStream inputStream = getClass().getClassLoader().getResourceAsStream(filePath);
+      Preconditions.checkNotNull(inputStream, "Input File path is invalid.");
+      return setProtoDescriptors(inputStream);
```
rajatbhatta, line 178:
```suggestion
      return setProtoDescriptors(inputStream);
```
`public Builder setProtoDescriptors(@Nonnull InputStream inputStream)` already does the NotNull check, so we need not have a redundant check here.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseTest.java
```diff
@@ -79,6 +87,18 @@ public void setUp() {
         .thenAnswer(
             invocation ->
                 new Database.Builder(dbClient, (DatabaseId) invocation.getArguments()[0]));
+    try {
+      protoDescriptorsInputStream =
+          getClass()
+              .getClassLoader()
+              .getResourceAsStream("com/google/cloud/spanner/descriptors.pb");
```
rajatbhatta, line 94: We should have this resource path as a constant `PROTO_DESCRIPTORS_RESOURCE_PATH`, and use it wherever needed.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseTest.java
```diff
@@ -178,6 +198,41 @@ public void testBuildWithDatabaseDialect() {
...
+  public void testBuildWithProtoDescriptorsFromByteArray() throws IOException {
+    final Database database =
+        dbClient
+            .newDatabaseBuilder(DatabaseId.of("my-project", "my-instance", "my-database"))
+            .setProtoDescriptors(protoDescriptorsByteArray)
+            .build();
+
+    assertEquals(protoDescriptors, database.getProtoDescriptors());
+  }
+
```
rajatbhatta, line 235: We should add tests for exception cases (invalid inputstream, invalid file path, etc) as well.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITProtoColumnTest.java
```diff
@@ -21,20 +21,11 @@
...
 
 import com.google.cloud.ByteArray;
-import com.google.cloud.spanner.DatabaseClient;
-import com.google.cloud.spanner.DatabaseId;
-import com.google.cloud.spanner.Key;
-import com.google.cloud.spanner.KeySet;
-import com.google.cloud.spanner.Mutation;
-import com.google.cloud.spanner.ParallelIntegrationTest;
-import com.google.cloud.spanner.ResultSet;
+import com.google.cloud.spanner.*;
```
rajatbhatta, line 24: Wildcard imports are not allowed according to Google style: https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITProtoColumnTest.java
```diff
@@ -43,39 +34,104 @@
...
-            .build()
-            .getService();
-
-    databaseClient = spanner.getDatabaseClient(DatabaseId.of(projectId, instanceId, databaseId));
+  public static void setUpDatabase() throws Exception {
+    // Get default spanner options for an integration test.
+    // SpannerOptions.Builder builder = env.getTestHelper().getOptions().toBuilder();
+    // Create a new testHelper with the cloud-devel host.
+    // testHelper = RemoteSpannerHelper.create(builder.build(),
+    // env.getTestHelper().getInstanceId());
```
rajatbhatta, line 66: I think we do not need these comments.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITProtoColumnTest.java
```diff
@@ -43,39 +34,104 @@
...
+                        + "  STORING (SingerId, FirstName, LastName)"))
+            .get(5, TimeUnit.MINUTES);
+
+    assertEquals(databaseID.getDatabase(), createdDatabase.getId().getDatabase());
+
+    // TODO(harsha): Check with backend team as this is not working for generated columns yet.
+    // GetDatabaseDdlResponse response =
+    // dbAdminClient.getDatabaseDdlResponse(databaseID.getInstanceId().getInstance(),
+    // databaseID.getDatabase());
+    // assertNotNull(response.getProtoDescriptors());
```
rajatbhatta, line 118: We should make sure we remove these comments.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITProtoColumnTest.java
```diff
@@ -85,23 +141,16 @@ public void after() throws Exception {
...
-   * checks for compatability between following types: 1. Proto Messages & Bytes 2. Proto Enums &
-   * Int64
-   *
-   * <p>Table `Types` was created through gcloud using following DDL:
-   * ************************************** CREATE TABLE Types ( RowID INT64 NOT NULL, Int64a INT64,
-   * Bytes BYTES(MAX), Int64Array ARRAY<INT64>, BytesArray ARRAY<BYTES(MAX)>, ProtoMessage
-   * spanner.examples.music.SingerInfo, ProtoEnum spanner.examples.music.Genre, ProtoMessageArray
-   * ARRAY<spanner.examples.music.SingerInfo>, ProtoEnumArray ARRAY<spanner.examples.music.Genre>, )
-   * PRIMARY KEY (RowID); **************************************
+   * Test to check data update and read queries on Proto columns and Enums and their arrays. Test
```
rajatbhatta, line 144:
```suggestion
   * Test to check data update and read queries on Proto Messages, Proto Enums and their arrays. Test
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseAdminClient.java
harshachinta: Done. Can you check?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Database.java
harshachinta: Done. Please check.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseAdminClient.java
harshachinta: Done. Please check

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITProtoColumnTest.java
harshachinta: This is needed to verify if the `proto_descriptor` field is populated while making a call to `getDatabaseDdlResponse` method. Currently the changes are not yet available and will uncomment it after that.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseTest.java
harshachinta: Done.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITProtoColumnTest.java
harshachinta: Done. Thanks for the reference.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITProtoColumnTest.java
harshachinta: Done

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITProtoColumnTest.java
harshachinta: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseInfo.java
harshachinta: @rajatbhatta 
When the input file path is not valid, then `getResourceAsStream` does not throw any exception instead `InputStream` becomes null. If we don't make a check here then it calls `public Builder setProtoDescriptors(@Nonnull InputStream inputStream)` which has a `NotNull` check, but the user gets an error that `InputStream cannot be null` which is not expected as the user is passing a file path as String.
Instead the user should know that the Input File Path is invalid and hence we might need this check here.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseTest.java
harshachinta: Done. Can you check?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseInfo.java
```diff
@@ -141,6 +155,29 @@ public Builder setDialect(Dialect dialect) {
...
+      Preconditions.checkNotNull(inputStream);
+      byte[] byteArray = ByteStreams.toByteArray(inputStream);
+      this.protoDescriptors = ByteString.copyFrom(byteArray);
+      return this;
+    }
+
+    @Override
+    public Builder setProtoDescriptors(String filePath) throws IOException {
+      Preconditions.checkState(filePath.length() != 0, "Input File Path cannot be empty.");
+      InputStream inputStream = getClass().getClassLoader().getResourceAsStream(filePath);
```
gauravpurohit06, line 176: It looks like the method only support reading files from `Project Resources`. If yes please clarify it in java doc.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -1246,19 +1249,26 @@ public OperationFuture<Database, CreateDatabaseMetadata> createDatabase(
...
+      com.google.cloud.spanner.Database databaseInfo,
       final Iterable<String> updateDatabaseStatements,
       @Nullable final String updateId)
       throws SpannerException {
     acquireAdministrativeRequestsRateLimiter();
-    final UpdateDatabaseDdlRequest request =
+    UpdateDatabaseDdlRequest.Builder requestBuilder =
         UpdateDatabaseDdlRequest.newBuilder()
-            .setDatabase(databaseName)
+            .setDatabase(databaseInfo.getId().getName())
```
gauravpurohit06, line 1259: Add null check on `databaseInfo.getId()`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -1316,17 +1326,15 @@ public Database getDatabase(String databaseName) throws SpannerException {
   }
 
   @Override
-  public List<String> getDatabaseDdl(String databaseName) throws SpannerException {
+  public GetDatabaseDdlResponse getDatabaseDdl(String databaseName) throws SpannerException {
```
gauravpurohit06, line 1329: I don't expect any customer to be using this API directly but if someone is using it, it will be breaking change.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseTest.java
```diff
@@ -178,6 +198,72 @@ public void testBuildWithDatabaseDialect() {
...
+    in.close();
+    // case1: Test empty file path
+    assertThrows(
+        IllegalStateException.class,
+        () ->
+            dbClient
+                .newDatabaseBuilder(DatabaseId.of("my-project", "my-instance", "my-database"))
+                .setProtoDescriptors("")
+                .build());
+    // case2: Test invalid file path
```
gauravpurohit06, line 249: If the file path is invalid then we should more meaningful exception then `NullPointerException `

gauravpurohit06 (changes requested):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
harshachinta: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseInfo.java
```diff
@@ -141,6 +155,29 @@ public Builder setDialect(Dialect dialect) {
...
+      Preconditions.checkNotNull(protoDescriptors);
+      this.protoDescriptors = ByteString.copyFrom(protoDescriptors);
+      return this;
+    }
+
+    @Override
+    public Builder setProtoDescriptors(@Nonnull InputStream inputStream) throws IOException {
+      Preconditions.checkNotNull(inputStream);
+      byte[] byteArray = ByteStreams.toByteArray(inputStream);
+      this.protoDescriptors = ByteString.copyFrom(byteArray);
```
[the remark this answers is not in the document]
harshachinta: Thanks for the suggestion. Didn't notice the method

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
harshachinta: I guess `GapicSpannerRpc.java` class is marked as `@InternalApi`. This means the code is subject to change and customers can use it at their own risk. I have added necessary changes to clirr file to mark this change.
@rajatbhatta @gauravpurohit06  correct me if I am missing something here.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseTest.java
harshachinta: Here `assertThrows` checks the class of exception that is returned. Thus, we check only for `NullPointerException`. 
However, the below is the actual exception that is thrown to user.
```
java.lang.NullPointerException: Input Proto Descriptors File path is invalid.
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
rajatbhatta: Makes sense to me.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseInfo.java
```diff
@@ -58,6 +63,13 @@ public Builder setDialect(Dialect dialect) {
       throw new UnsupportedOperationException("Unimplemented");
     }
 
+    public abstract Builder setProtoDescriptors(@Nonnull byte[] protoDescriptors);
```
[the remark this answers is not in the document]
harshachinta: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseInfo.java
```diff
@@ -58,6 +63,13 @@ public Builder setDialect(Dialect dialect) {
...
+
+    public abstract Builder setProtoDescriptors(@Nonnull InputStream inputStream)
```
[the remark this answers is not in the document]
harshachinta: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseInfo.java
```diff
@@ -58,6 +63,13 @@ public Builder setDialect(Dialect dialect) {
...
+        throws IOException;
+
+    public abstract Builder setProtoDescriptors(String filePath) throws IOException;
```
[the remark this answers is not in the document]
harshachinta: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseInfo.java
harshachinta: Added in java doc

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseInfo.java
```diff
@@ -63,11 +62,39 @@ public Builder setDialect(Dialect dialect) {
...
+     * <p>It is used by CREATE/ALTER PROTO BUNDLE statements which are part of DDL statements.
+     * Contains a protobuf-serialized [google.protobuf.FileDescriptorSet]. To generate a proto
+     * descriptors file run `protoc` with --include_imports and --descriptor_set_out.
+     */
     public abstract Builder setProtoDescriptors(@Nonnull byte[] protoDescriptors);
 
+    /**
+     * Optional for creating a new database.
+     *
+     * <p>The proto descriptors input as InputStream to be used for the database.
```
gauravpurohit06, line 79: nit: Please use javadoc tags while writing javadoc. Use `@param` for the parameters description.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseInfo.java
```diff
@@ -63,11 +62,39 @@ public Builder setDialect(Dialect dialect) {
...
+     *
+     * <p>It is used by CREATE/ALTER PROTO BUNDLE statements which are part of DDL statements.
+     * Contains a protobuf-serialized [google.protobuf.FileDescriptorSet]. To generate a proto
+     * descriptors file run `protoc` with --include_imports and --descriptor_set_out.
```
gauravpurohit06, line 83: You can use `@code` for defining some commands or for other text which you don't want to get interpreted as HTML.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseInfo.java
```diff
@@ -63,11 +62,39 @@ public Builder setDialect(Dialect dialect) {
       throw new UnsupportedOperationException("Unimplemented");
     }
 
+    /**
+     * Optional for creating a new database.
+     *
+     * <p>The proto descriptors input as byte[] to be used for the database.
```
gauravpurohit06, line 68: nit: Please use javadoc tags while writing javadoc. Use @param for the parameters description.)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseInfo.java
```diff
@@ -63,11 +62,39 @@ public Builder setDialect(Dialect dialect) {
...
+     *
+     * <p>It is used by CREATE/ALTER PROTO BUNDLE statements which are part of DDL statements.
+     * Contains a protobuf-serialized [google.protobuf.FileDescriptorSet]. To generate a proto
+     * descriptors file run `protoc` with --include_imports and --descriptor_set_out.
```
gauravpurohit06, line 72: You can use @code for defining some commands or for other text which you don't want to get interpreted as HTML.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseInfo.java
```diff
@@ -63,11 +62,39 @@ public Builder setDialect(Dialect dialect) {
...
+     */
     public abstract Builder setProtoDescriptors(@Nonnull InputStream inputStream)
         throws IOException;
 
+    /**
+     * Optional for creating a new database.
+     *
+     * <p>The proto descriptors file path input as String to be used for the database. The proto
```
gauravpurohit06, line 91: same as above.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseInfo.java
```diff
@@ -63,11 +62,39 @@ public Builder setDialect(Dialect dialect) {
...
+     * descriptors file must be present with in the project resources directory.
+     *
+     * <p>It is used by CREATE/ALTER PROTO BUNDLE statements which are part of DDL statements.
+     * Contains a protobuf-serialized [google.protobuf.FileDescriptorSet]. To generate a proto
+     * descriptors file run `protoc` with --include_imports and --descriptor_set_out.
```
gauravpurohit06, line 96: same as above.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseInfo.java
```diff
@@ -172,7 +199,8 @@ public Builder setProtoDescriptors(@Nonnull InputStream inputStream) throws IOEx
     @Override
     public Builder setProtoDescriptors(@Nonnull String filePath) throws IOException {
       Preconditions.checkNotNull(filePath);
-      Preconditions.checkState(filePath.length() != 0, "Input Proto Descriptors File Path cannot be empty.");
+      Preconditions.checkState(
+          filePath.length() != 0, "Input Proto Descriptors File Path cannot be empty.");
       InputStream inputStream = getClass().getClassLoader().getResourceAsStream(filePath);
```
gauravpurohit06, line 204: What will be the behaviour of this function if we give any random path? Can we mention it in function doc for clarity?

gauravpurohit06 (changes requested):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseInfo.java
harshachinta: Done. Can you check it?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseInfo.java
harshachinta: Done. Can you check it?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseInfo.java
harshachinta: Done. Can you check it?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseInfo.java
harshachinta: Done. Can you check it?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseInfo.java
harshachinta: Done. Can you check it?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseInfo.java
harshachinta: The current functionality is to read file with in the classpath under `resources` directory. If the file path is random other than this it throws an `IOException: Input Proto Descriptors File path is invalid`.
I have added this behaviour in the `@throws` parameter of javadoc.

gauravpurohit06 (approved):
LGTM

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseInfo.java
Shua1: Is Project Resources a good default?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseInfo.java
```diff
@@ -141,6 +155,29 @@ public Builder setDialect(Dialect dialect) {
...
+    @Override
+    public Builder setProtoDescriptors(@Nonnull InputStream inputStream) throws IOException {
+      Preconditions.checkNotNull(inputStream);
+      byte[] byteArray = ByteStreams.toByteArray(inputStream);
+      this.protoDescriptors = ByteString.copyFrom(byteArray);
+      return this;
+    }
+
+    @Override
+    public Builder setProtoDescriptors(String filePath) throws IOException {
```
Shua1: Do we have a precedence of reading file on user's behalf? What about other Google Cloud client libraries. Are we reading files in the recommended way in case there is one?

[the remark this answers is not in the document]
harshachinta: Having different method names for the same functionality would not be a good user experience. Overloading `setProtoDescriptors` to accept different types of parameters (byte[], Filepath, InputStream) as input is the standard followed in Java.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseInfo.java
harshachinta: Yes. Having the resources with in the class path can help users provide a relative file path as input and Project resources directory is a good place to maintain all these files.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseInfo.java
harshachinta: User's can read the file on their end and pass in the byte[] or InputStream as input to the `setProtoDescriptors` method. Alternatively, we also add another overloaded method that accepts the file path as input.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Database.java
```diff
@@ -88,6 +88,20 @@ public OperationFuture<Void, UpdateDatabaseDdlMetadata> updateDdl(
     return dbClient.updateDatabaseDdl(instance(), database(), statements, operationId);
   }
 
+  /**
+   * Updates a database in a Cloud Spanner instance
```
olavloite, line 92: nit:
```suggestion
   * Updates the schema of a database in a Cloud Spanner instance
```

olavloite: Just to be sure: This method only updates the schema of the database, right? Or does it do anything additional?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseAdminClient.java
```diff
@@ -407,6 +409,39 @@ OperationFuture<Void, UpdateDatabaseDdlMetadata> updateDatabaseDdl(
       @Nullable String operationId)
       throws SpannerException;
 
+  /**
+   * Updates a database in a Cloud Spanner instance. Few configuration options in the {@link
+   * Database} instance will be included in the {@link UpdateDatabaseDdlRequest}.
```
olavloite, line 414:
```suggestion
   * Updates a database in a Cloud Spanner instance. Any proto descriptors that have been set for the {@link
   * Database} instance will be included in the {@link UpdateDatabaseDdlRequest}.
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Database.java
```diff
@@ -88,6 +88,20 @@ public OperationFuture<Void, UpdateDatabaseDdlMetadata> updateDdl(
...
+   *
+   * @param database Database object to set configuration options such as proto_descriptors.
+   * @param statements DDL statements to run while updating the database.
+   * @param operationId Operation id assigned to this operation. If null, system will autogenerate
+   *     one. This must be unique within a database abd must be a valid identifier
+   *     [a-zA-Z][a-zA-Z0-9_]*.
+   */
+  public OperationFuture<Void, UpdateDatabaseDdlMetadata> updateDdl(
```
olavloite, line 100: Sorry, I didn't notice this during the last review, but this seems a little strange. It is a method in the `Database` class and takes a `Database` instance as an argument. Why does it not just operate on this database?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Database.java
harshachinta: Updated the comment as per the suggestion.
Yes @olavloite. This method updates the schema of the database which also includes updating schema of the Proto Bundles which lie in that database.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseAdminClient.java
harshachinta: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Database.java
harshachinta: The reason it cannot operate on this database is because a user cannot set `ProtoDescriptor` directly to Database Class as `setProtoDescriptors` method belongs to `Database.Builder`. Below is a small example to illustrate this,
```
Database database = client.getDatabase(INSTANCE_ID, DB_ID);

// If a user wants to update the ProtoDescriptor field then, 
// the below line cannot be written as setProtoDescriptors method does not belong to Database class.
database.setProtoDescriptors("descriptors.pb"); 

// This does not update proto descriptor in the database because the field was not updated.
database.updateDdl(Collections.singletonList("DROP TABLE Singer"), null); 
```
So a new Database instance has to be passed to update the ProtoDescriptor field. However, I agree that this is incorrect and confusing at same time.

**Another Approach**
@olavloite  Please validate if below is more meaningful to do.

Expose a new method `toBuilder()` in `Database.java` that return a Builder of that `Database` instance. This allows users to `setProtoDescriptor` by calling the builder.
The change in `Database.java` is
```
public class Database extends DatabaseInfo {
  public Builder toBuilder() {
      return new Builder(this);
    }


  public OperationFuture<Void, UpdateDatabaseDdlMetadata> updateDdl(
      Iterable<String> statements, String operationId) throws SpannerException {
     // The updateDatabaseDdl is called with 'this' as argument instead of instance(), database()
    return dbClient.updateDatabaseDdl(this, statements, operationId);
  }
}
```
The user can use it like below
```
Database database = client.getDatabase(INSTANCE_ID, DB_ID);
// The below database instance has the updated ProtoDescriptor field
database = database.toBuilder().setProtoDescriptors("descriptors.pb").build();
// Update works fine
database.updateDdl(Collections.singletonList("DROP TABLE Singer"), null);
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseInfo.java
harshachinta: @olavloite 
My understanding here is class loader can find a resource if it is there in the classpath and resources directory would always be a part of the classpath. I tried this with a small demo application and it works fine.
However, I am not aware of the behaviour if there are any custom/multiple class loaders. I would need some time to analyse the behaviour in this case.
In the interest of time, I am planning to remove this overload that accepts a file path as argument and take it separately in another PR.
Can you please let me know your views here?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Database.java
olavloite: Hmmm.... If the `protoDescriptors` field is not something that is actually part of the `Database` class and thereby also not stored with the database, and is only needed for the update method, then we need to add it as an argument to the method afte rall. Other approaches would be non-idiomatic Java.

So the signature of the method would then become `Database#updateDatabaseDdl(Iterable<String> statements, String protoDescriptors, String operationId)`. Another option would be to create a `DatabaseUpdateOptions` class that contains all possible update options, the first one being `protoDescriptors` (that would make it more easily reusable for additional options in the future).

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseInfo.java
olavloite: I think that removing this for now is a good choice. People can easily use resources from their class path using the overload that takes an `InputStream` by calling `getResourceAsStream()` themselves.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Database.java
harshachinta: I agree with you @olavloite. Creating a new `DatabaseUpdateOptions` class that can update options will help us reuse in future. I will take this effort as part of the next PR. I will tag this comment in the other PR later to understand the need of that change.

olavloite (approved):

rajatbhatta (approved):

PR #2304, opened 2023-02-24
test: Configuring ITDatabaseRolePermissionTest to use SpannerOptions …

gauravpurohit06:
Using SpannerOptions from helper class while creating a new SpannerOptions to override specific properties.


rajatbhatta (approved):
Nice catch!

PR #2308, opened 2023-02-27
feat: Leader Aware Routing in Connection API

rajatbhatta:
By default, RW/PDML requests are preferred to route to leader after LAR is enabled from backend.

This PR adds `routeToLeader` Boolean connection option for implementing opt-out feature for LAR.
Example URL: `cloudspanner:/projects/test-project-123/instances/test-instance-123/databases/test-database-123?routeToLeader=false`

**Note:** You can also pass in `routeToLeader=true`, but that is the default behaviour (even if you do not pass in this option).

BEGIN_COMMIT_OVERRIDE
feat: Enable leader aware routing by default in Connection API. This enables its use in the JDBC driver and PGAdapter. The update contains performance optimisations that will reduce the latency of read/write transactions that originate from a region other than the default leader region.
END_COMMIT_OVERRIDE


google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/SpannerPoolTest.java
```diff
@@ -145,6 +151,9 @@ public void testGetSpanner() {
     spanner1 = pool.getSpanner(options3, connection1);
     spanner2 = pool.getSpanner(options4, connection2);
     assertThat(spanner1).isNotEqualTo(spanner2);
+    spanner1 = pool.getSpanner(options7, connection1);
```
surbhigarg92, line 154: We are defining options4 and options5 below but using options7 and options8 here. Is it expected ?

surbhigarg92 (approved):

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/SpannerPoolTest.java
rajatbhatta: That is a different test (`testSpannerPoolKeyEquality`). This test is using the data members of the `SpannerPoolTest` class (`private ConnectionOptions options7 = mock(ConnectionOptions.class);`)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
```diff
@@ -231,6 +234,10 @@ public String[] getValidValues() {
                       READONLY_PROPERTY_NAME,
                       "Should the connection start in read-only mode (true/false)",
                       DEFAULT_READONLY),
+                  ConnectionProperty.createBooleanProperty(
+                      ROUTE_TO_LEADER_PROPERTY_NAME,
+                      "Should the RW/PDML requests be routed to leader region (true/false)",
```
olavloite, line 239: nit: This string is allowed to be long, so I would suggest the following additions:

```suggestion
                      "Should read/write transactions and partitioned DML be routed to leader region (true/false)",
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
```diff
@@ -1025,6 +1041,10 @@ public boolean isAutocommit() {
   public boolean isReadOnly() {
     return readOnly;
   }
+  /** Whether RW/PDML requests are preferred to be routed to the leader region. */
```
olavloite, line 1044: nit: add empty line

olavloite (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
```diff
@@ -426,6 +433,8 @@ private boolean isValidUri(String uri) {
      *       created on the emulator if any of them do not yet exist. Any existing instance or
      *       database on the emulator will remain untouched. No other configuration is needed in
      *       order to connect to the emulator than setting this property.
+     *   <li>routeToLeader (boolean): Sets the routeToLeader flag to route requests to leader (true)
+     *       or any region (false) in RW/PDML transactions. Default is true.
```
arpan14, line 437: Abbreviations RW/PDML can be expanded in the documentation.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
```diff
@@ -719,6 +730,11 @@ static boolean parseReadOnly(String uri) {
     return value != null ? Boolean.parseBoolean(value) : DEFAULT_READONLY;
   }
 
+  static boolean parseRouteToLeader(String uri) {
```
arpan14, line 733: Are we using this method within unit tests? If yes, for consistency we can add the @VisibleForTesting annotation like other methods.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/SpannerPoolTest.java
```diff
@@ -65,6 +65,8 @@ public class SpannerPoolTest {
```
arpan14, line 65: Query : Do we have a test which fails when the new property is not added to the equals() method?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/SpannerPoolTest.java
```diff
@@ -145,6 +151,9 @@ public void testGetSpanner() {
     spanner1 = pool.getSpanner(options3, connection1);
     spanner2 = pool.getSpanner(options4, connection2);
     assertThat(spanner1).isNotEqualTo(spanner2);
+    spanner1 = pool.getSpanner(options7, connection1);
+    spanner2 = pool.getSpanner(options8, connection2);
+    assertNotEquals(spanner1, spanner2);
```
arpan14, line 156: Prefer `assertThat` method over `assertNotEquals` method. It provides readability and type check benefits and is also consistent with what is used in existing test.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionOptionsTest.java
```diff
@@ -148,6 +149,27 @@ public void testBuildWithAutoConfigEmulator() {
     assertTrue(options.isUsePlainText());
   }
 
+  @Test
+  public void testBuildWithRouteToLeader() {
+    ConnectionOptions.Builder builder = ConnectionOptions.newBuilder();
+    builder.setUri(
+        "cloudspanner:/projects/test-project-123/instances/test-instance-123/databases/test-database-123?routeToLeader=false");
+    ConnectionOptions options = builder.build();
+    assertEquals(options.getHost(), DEFAULT_HOST);
```
arpan14, line 158: While the asserts here are consistent with what is used in the test class, we should try having a single assert statement. For ex - `assertEquals(expectedConnectionOptions, actualConnectionOptions)`

This provides benefits where you asserting the entire object and not picking few of the members and can help to pick code side-effects where any member got modified by mistake.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionOptionsTest.java
```diff
@@ -148,6 +149,27 @@ public void testBuildWithAutoConfigEmulator() {
...
+    assertEquals(options.getProjectId(), "test-project-123");
+    assertEquals(options.getInstanceId(), "test-instance-123");
+    assertEquals(options.getDatabaseName(), "test-database-123");
+    assertFalse(options.isRouteToLeader());
```
arpan14, line 162: Can we define `private static final` variables for all such constants instead of defining them inline? We can either have it within this class or a new class in case we see these constants to be generic enough to be reused across different test classes.

This may be ok in case the project ID, instance ID and database name changed with every unit test but I don't see that to be the case here.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/SpannerPoolTest.java
```diff
@@ -460,14 +469,29 @@ public void testSpannerPoolKeyEquality() {
...
+    SpannerPoolKey key4 = SpannerPoolKey.of(options4);
+    SpannerPoolKey key5 = SpannerPoolKey.of(options5);
 
     assertNotEquals(key1, key2);
     assertEquals(key2, key3);
     assertNotEquals(key1, key3);
     assertNotEquals(key1, new Object());
+    assertEquals(key3, key4);
+    assertNotEquals(key4, key5);
+    assertEquals(key2, key4);
```
arpan14, line 495: From a readability point of view, its a little hard to understand why these assertions work. Do you think adding some comments would help? Let's discuss offline. 

I understand the key4 and key5 won't be equal since LAR boolean property is part of the equals() method but I didn't get how this is equal to key3 and key2.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
rajatbhatta: This is **not** used in unit tests. So, I think it would be cleaner to remove the annotation from other methods, rather than adding it here. I'll take it up as a follow-up PR, as that would be unrelated to the changes here.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/SpannerPoolTest.java
rajatbhatta: Will add comments in code. key3 will be equal to key4 because not passing in `routeToLeader` in Connection URI is equivalent to passing it as `true`, as routeToLeader is `true` by default.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
rajatbhatta: Done.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/SpannerPoolTest.java
rajatbhatta: We can take it as a separate exercise, as we'll need this across classes. Keeping the thread open for the time being.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/SpannerPoolTest.java
rajatbhatta: Done. Added comments while creating ConnectionOptions.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionOptionsTest.java
rajatbhatta: This is done. And I've used them in this method for now. Refactoring other methods would be unrelated to this PR, and can be taken up later.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionOptionsTest.java
```diff
@@ -151,21 +154,19 @@ public void testBuildWithAutoConfigEmulator() {
 
   @Test
   public void testBuildWithRouteToLeader() {
+    final String BASE_URI =
```
arpan14, line 157: Nit: Can be defined as private static.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionOptionsTest.java
```diff
@@ -151,21 +154,19 @@ public void testBuildWithAutoConfigEmulator() {
...
+        "cloudspanner:/projects/test-project-123/instances/test-instance-123/databases/test-database-123";
     ConnectionOptions.Builder builder = ConnectionOptions.newBuilder();
-    builder.setUri(
-        "cloudspanner:/projects/test-project-123/instances/test-instance-123/databases/test-database-123?routeToLeader=false");
+    builder.setUri(BASE_URI + "?routeToLeader=false");
```
arpan14, line 160: Nit: Prefer StringBuilder() . `+` is a more expensive way of concatenating strings.

arpan14 (approved):

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionOptionsTest.java
rajatbhatta: For single statement, the compiler would use `StringBuilder` automatically: https://stackoverflow.com/a/4645155.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionOptionsTest.java
rajatbhatta: There are several places in this file where this can be done and replace the hard-codings with constants. We can take it as a follow-up task in a separate PR.

rajatbhatta:
Pls do not merge this PR until LAR is enabled by default in the Java client library.

PR #2324, opened 2023-03-07
fix: support DDL/ClientSideStatement through Connection#executeUpdate API

rajatbhatta:
DDL statements and ClientSideStatements which return no result are now supported through executeUpdate API in ConnectionImpl and should return 0L in that case.

We add this change to the corresponding async API (executeUpdateAsync) too.

This change will be propagated to executeUpdate APIs through Statement and PreparedStatement in JDBC.


google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/it/ITClientSideStatementTest.java
```diff
@@ -0,0 +1,164 @@
...
+              SpannerException.class,
+              () -> connection.executeUpdateAsync(Statement.of(SHOW_VARIABLE_AUTOCOMMIT)));
+      assertEquals(e.getErrorCode(), ErrorCode.FAILED_PRECONDITION);
+      try (ResultSet rs = connection.executeQuery(Statement.of(SHOW_VARIABLE_AUTOCOMMIT))) {
+        while (rs.next()) {
+          assertFalse(rs.getBoolean(0));
+        }
+      }
+    } catch (ExecutionException | InterruptedException e) {
+      // pass
```
olavloite, line 90: This should cause a test failure and not be silently ignored.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/AbstractConnectionImplTest.java
```diff
@@ -993,6 +993,18 @@ private void testExecuteUpdate(final StatementType type) {
             },
             connection,
             ErrorCode.FAILED_PRECONDITION);
+      } else if (type == StatementType.DDL) {
```
olavloite, line 996: This change also requires that we re-generate the generated tests. You can do that with the following Maven command:

```
mvn -Ddo_log_statements=true exec:java -Dexec.mainClass=com.google.cloud.spanner.connection.SqlTestScriptsGenerator -Dexec.classpathScope="test"
```

See https://github.com/googleapis/java-spanner/blob/27ef53c8447bd51a56fdfe6b2b206afe234fad80/google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplGeneratedSqlScriptTest.java#L39

rajatbhatta: This is done. Thanks for pointing it out.

arpan14:
Closing due to inactivity.

PR #2340, opened 2023-03-20
feat: foreign key on delete cascade

arpan14:
* Adding integration tests for foreign key on delete cascade feature testing both GoogleSQL + PostGres dialects.
* Adding documentation samples along with integration tests for the samples.
  *  Making a minor enhancement in the samples integration test setup framework to allow accepting a host URL in Spanner options to allow overriding and testing in staging env.


samples/snippets/src/main/java/com/example/spanner/AlterTableWithForeignKeyDeleteCascadeSample.java
```diff
@@ -0,0 +1,66 @@
...
+            ImmutableList.of("ALTER TABLE ShoppingCarts\n"
+                + "              ADD CONSTRAINT FKShoppingCartsCustomerName\n"
+                + "              FOREIGN KEY (CustomerName)\n"
+                + "              REFERENCES Customers(CustomerName)\n"
+                + "              ON DELETE CASCADE\n"),
+            null);
+
+    System.out.printf(String.format("Altered ShoppingCarts table with FKShoppingCartsCustomerName\n"
+        + "foreign key constraint on database %s on instance %s", databaseId, instanceId));
+
```
olavloite, line 63: nit: remove empty line

samples/snippets/src/main/java/com/example/spanner/AlterTableWithForeignKeyDeleteCascadeSample.java
```diff
@@ -0,0 +1,66 @@
...
+  static void alterForeignKeyDeleteCascadeConstraint(
+      DatabaseAdminClient adminClient, String instanceId, String databaseId) {
+    adminClient.updateDatabaseDdl(
+            instanceId,
+            databaseId,
+            ImmutableList.of("ALTER TABLE ShoppingCarts\n"
+                + "              ADD CONSTRAINT FKShoppingCartsCustomerName\n"
+                + "              FOREIGN KEY (CustomerName)\n"
+                + "              REFERENCES Customers(CustomerName)\n"
```
olavloite, line 57: Is this something that has been standardized across multiple client libraries? I am a little surprised that our sample uses `CustomerName` as the referenced column, as it requires the column to be unique. `CustomerName` does not sound like a column that would normally be unique.

samples/snippets/src/main/java/com/example/spanner/DropForeignKeyConstraintDeleteCascadeSample.java
```diff
@@ -0,0 +1,62 @@
...
+    adminClient.updateDatabaseDdl(
+        instanceId,
+        databaseId,
+        ImmutableList.of("ALTER TABLE ShoppingCarts\n"
+            + "              DROP CONSTRAINT FKShoppingCartsCustomerName\n"),
+        null);
+
+    System.out.printf(String.format("Altered ShoppingCarts table to drop FKShoppingCartsCustomerName\n"
+        + "foreign key constraint on database %s on instance %s\n", databaseId, instanceId));
+
```
olavloite, line 59: remove empty line

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITForeignKeyDeleteCascadeTest.java
```diff
@@ -0,0 +1,540 @@
...
+    return params;
+  }
+
+  private static final String TABLE_NAME_SINGER = "Singer";
+  private static final String TABLE_NAME_CONCERTS = "Concerts";
+  private static final String CONCERTS_SINGER_FOREIGN_KEY_CONSTRAINT_NAME = "Fk_Concerts_Singer";
+  private static final String CONCERTS_SINGER_FOREIGN_KEY_CONSTRAINT_NAME_V2 = "Fk_Concerts_Singer_V2";
+  private static final String DELETE_RULE_CASCADE = "CASCADE";
+  private static final String DELETE_RULE_DEFAULT = "NO ACTION";
+  private static final String DELETE_RULE_COLUMN_NAME = "DELETE_RULE";
```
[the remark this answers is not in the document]
arpan14: Sure I will modify this to enhance the readability as suggested.

samples/snippets/src/main/java/com/example/spanner/AlterTableWithForeignKeyDeleteCascadeSample.java
arpan14: Yes I agree on this. And unfortunately this has been standardised across all client libraries.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITForeignKeyDeleteCascadeTest.java
```diff
@@ -0,0 +1,540 @@
...
+              transaction.executeUpdate(singerInsertStatementWithValues);
+              return null;
+            });
+    databaseClient
+        .readWriteTransaction()
+        .run(
+            transaction -> {
+              transaction.executeUpdate(concertInsertStatementWithValues);
+              return null;
+            });
```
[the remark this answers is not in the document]
arpan14: Earlier I wasn't sure if the order would be maintained in batchUpdate since due to FK relationship we'd want singer inserts to execute before concert inserts.

After reading documentation of `batchUpdate` method, I think we can have a single transaction.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITForeignKeyDeleteCascadeTest.java
```diff
@@ -0,0 +1,540 @@
...
+  @Test
+  public void testForeignKeyDeleteCascadeConstraints_withCreateDDLStatements() {
+    final DatabaseClient databaseClient = getCreatedDatabaseClient();
+    final String referentialConstraintQuery = getReferentialConstraintsQueryStatement();
+    try (final ResultSet rs =
+        databaseClient
+            .singleUse()
+            .executeQuery(Statement.of(referentialConstraintQuery))) {
+      while (rs.next()) {
+        assertThat(rs.getString(DELETE_RULE_COLUMN_NAME)).isEqualTo(DELETE_RULE_CASCADE);
```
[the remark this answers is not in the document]
arpan14: Shouldn't `assertThat` be preferred over `assertEquals`. Even the internal style docs suggest to prefer this over `assertEquals`?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITForeignKeyDeleteCascadeTest.java
```diff
@@ -0,0 +1,540 @@
...
+                        Mutation.newInsertBuilder("Singer")
+                            .set(getSingerIdColumnName()).to(4L)
+                            .set(getSingerFirstNameColumnName()).to("singerName").build(),
+                        Mutation.delete("Singer", Key.of(4L))));
+                return null;
+              });
+      fail("Expected exception");
+    } catch (SpannerException ex) {
+      assertThat(ex.getErrorCode()).isEqualTo(ErrorCode.FAILED_PRECONDITION);
+    }
```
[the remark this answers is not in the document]
arpan14: Yes, I had this same query since it looked unrelated to this feature but we got a guidance from backend team to include it in the set of tests. @asthamohta Are we including this (#7 test case in spec document) in other libraries?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITForeignKeyDeleteCascadeTest.java
```diff
@@ -0,0 +1,540 @@
...
+  }
+
+  private Database createDatabase(final List<String> statements) {
+    final Database database;
+    if (dialect.dialect == Dialect.POSTGRESQL) {
```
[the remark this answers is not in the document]
arpan14: Yes, done

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITForeignKeyDeleteCascadeTest.java
```diff
@@ -0,0 +1,540 @@
...
+    }
+  }
+
+  private String getDeleteStatementForSingerTable() {
+    if (dialect.dialect == Dialect.POSTGRESQL) {
+      return "DELETE FROM Singer WHERE singer_id = $1";
+    } else {
+      return "DELETE FROM Singer WHERE SingerId = @p1";
+    }
+  }
```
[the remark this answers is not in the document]
arpan14: @olavloite I like the suggestion to simplify this but `VALUES ($1, $2)` is not the only thing that we need to construct. We will also require to differentiate the column name based on the dialect. If we follow the above suggestion the test code will have the additional if/else block to generate the column names. This won't be the most readable code in my opinion.

Hence I went with the option of having a private method per table's insert statement. 

Let me know if you had any other suggestions.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITForeignKeyDeleteCascadeTest.java
```diff
@@ -0,0 +1,540 @@
...
+    final String referentialConstraintQuery = getReferentialConstraintsQueryStatement();
+    try (final ResultSet rs =
+        databaseClient
+            .singleUse()
+            .executeQuery(
+                Statement.of(referentialConstraintQuery))) {
+      while (rs.next()) {
+        assertThat(rs.getString(DELETE_RULE_COLUMN_NAME)).isEqualTo(DELETE_RULE_CASCADE);
+      }
+    }
```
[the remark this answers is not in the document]
arpan14: For all the tests apart from `testForeignKeyDeleteCascadeConstraints_withAlterDDLStatements` the DDL statements are within the `setUpDatabase` method which is executed once per class. Hence, for majority of the tests we won't be able to make the SQL statement inline with the test method.

For `testForeignKeyDeleteCascadeConstraints_withAlterDDLStatements` because we had to execute a different set of DDL statements those were put in private methods `getCreateAndAlterTableStatementsWithForeignKey` which is just a replacement of `setUpDatabase` method.

I can still think to eliminate the private method `getCreateAndAlterTableStatementsWithForeignKey` and make statements inline with the test method but that will just eliminate the concern with 1 test method. For the remaining tests we will still need to scroll to `setUpDatabase` method in order to understand the DDL statement. I'll also eliminate the `getCreateAndAlterTableStatementsWithForeignKey` method and eliminate a few of the constants but the core readability concern of having to scroll different methods to understand test context will still remain.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITForeignKeyDeleteCascadeTest.java
```diff
@@ -0,0 +1,540 @@
...
+    if (dialect.dialect == Dialect.POSTGRESQL) {
+      return POSTGRES_QUERY_REFERENTIAL_CONSTRAINTS;
+    } else {
+      return QUERY_REFERENTIAL_CONSTRAINTS;
+    }
+  }
+
+  private String getReferentialConstraintsQueryStatementV2() {
+    if (dialect.dialect == Dialect.POSTGRESQL) {
+      return POSTGRES_QUERY_REFERENTIAL_CONSTRAINTS_V2;
```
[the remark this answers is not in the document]
arpan14: I had to create a different set of tables for one of the test cases since all other test cases operate on a common set of tables created in `@BeforeClass` method.

The V2 constants don't really have any different functionality. They just operate on the `V2` tables.

The only alternate I can think of is using a different set of entities for naming the tables apart from Singers/Concerts. In my opinion that could be an overkill. Scrolling to the constant value does make it clear that the V2 constants are used for the different tables.

Let me know what you think?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITForeignKeyDeleteCascadeTest.java
```diff
@@ -0,0 +1,582 @@
...
+
+  @Parameterized.Parameters(name = "Dialect = {0}")
+  public static List<DialectTestParameter> data() {
+    List<DialectTestParameter> params = new ArrayList<>();
+    params.add(new DialectTestParameter(Dialect.GOOGLE_STANDARD_SQL));
+    params.add(new DialectTestParameter(Dialect.POSTGRESQL));
+    return params;
+  }
+
+  private static final String TABLE_NAME_SINGER = "Singer";
```
olavloite, line 62: super nit: We're mixing plural and singular table names in this test (`Singer` vs `Concerts`)

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITForeignKeyDeleteCascadeTest.java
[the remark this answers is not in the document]
olavloite: Can we take that discussion in a separate doc / forum? I'm not necessarily against changing the default. I am against implicitly changing the default in a random PR.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITForeignKeyDeleteCascadeTest.java
```diff
@@ -0,0 +1,582 @@
...
+  }
+
+  @Test
+  public void testForeignKeyDeleteCascadeConstraints_withCreateDDLStatements() {
+    assumeFalse(
+        "Emulator does not yet support foreign key delete cascade",
+        EmulatorSpannerHelper.isUsingEmulator());
+
+    final DatabaseClient databaseClient = getCreatedDatabaseClient();
+    final String referentialConstraintQuery = getReferentialConstraintsQueryStatement();
```
olavloite, line 171: Here and elsewhere:

I think you should:
1. Remove the `getReferentialConstraintsQueryStatement()` method and directly have the SQL statement here. That makes it easier to read what this test case does. AFAICT, it is only used twice, so the code duplication is minimal.
2. The query itself does not need to be dialect specific. Both dialects support using single quotes for string literals. (That is; both dialects can use `select col1, col2, ... from my_table where col1='some value'`).
3. It might be even better to change the query into using a query parameter. That would require the query to be slightly different per dialect, but only for the name of the query parameter (that is; `$1` vs `@p1`). Generating a query parameter could be put in a separate method, so you don't need that part in the test method itself.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITForeignKeyDeleteCascadeTest.java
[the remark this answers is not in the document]
arpan14: Sure, changing it to `assertEquals` based on the reasoning that it's more widely adopted externally.

samples/snippets/src/main/java/com/example/spanner/AlterTableWithForeignKeyDeleteCascadeSample.java
olavloite: Ack, thanks for checking.

samples/snippets/src/main/java/com/example/spanner/DropForeignKeyConstraintDeleteCascadeSample.java
```diff
@@ -0,0 +1,57 @@
...
+package com.example.spanner;
+
+// [START spanner_drop_foreign_key_constraint_delete_cascade]
+import com.google.cloud.spanner.DatabaseAdminClient;
+import com.google.cloud.spanner.Spanner;
+import com.google.cloud.spanner.SpannerOptions;
+import com.google.common.collect.ImmutableList;
+
+class DropForeignKeyConstraintDeleteCascadeSample {
```
olavloite, line 25: Should we just name this 
```suggestion
class DropForeignKeyConstraintSample {
```
? Or would that conflict with some other sample?

olavloite (approved):

samples/snippets/src/test/java/com/example/spanner/AlterTableWithForeignKeyDeleteCascadeSampleIT.java
```diff
@@ -0,0 +1,65 @@
...
+
+    // Runs sample
+    final String out =
+        SampleRunner.runSample(
+            () ->
+                AlterTableWithForeignKeyDeleteCascadeSample.alterForeignKeyDeleteCascadeConstraint(
+                    databaseAdminClient, instanceId, databaseId));
+
+    assertTrue(
+        "Expected to have created database "
```
anuragsrivstv, line 59: May be instead of checking the output message we should verify that the table was actually altered by the sample or not ?

samples/snippets/src/test/java/com/example/spanner/DropForeignKeyConstraintDeleteCascadeSampleIT.java
```diff
@@ -0,0 +1,63 @@
...
+
+    // Runs sample
+    final String out =
+        SampleRunner.runSample(
+            () ->
+                DropForeignKeyConstraintDeleteCascadeSample.deleteForeignKeyDeleteCascadeConstraint(
+                    databaseAdminClient, instanceId, databaseId));
+
+    assertTrue(
+        "Expected to have dropped foreign-key constraints from tables in created database "
```
anuragsrivstv, line 59: Same comment as above. If possible we should test changes in actual table.

samples/snippets/src/test/java/com/example/spanner/AlterTableWithForeignKeyDeleteCascadeSampleIT.java
```diff
@@ -0,0 +1,65 @@
...
+                "CREATE TABLE Customers (\n"
+                    + "              CustomerId INT64 NOT NULL,\n"
+                    + "              CustomerName STRING(62) NOT NULL,\n"
+                    + "              ) PRIMARY KEY (CustomerId)",
+                "CREATE TABLE ShoppingCarts (\n"
+                    + "              CartId INT64 NOT NULL,\n"
+                    + "              CustomerId INT64 NOT NULL,\n"
+                    + "              CustomerName STRING(62) NOT NULL,\n"
+                    + "              CONSTRAINT FKShoppingCartsCustomerId"
+                    + " FOREIGN KEY (CustomerId)\n"
```
anuragsrivstv, line 46: Hi, As we are adding `Foreign Key Delete Cascade` constraint , this should **Not** depend on sample which has already added the constraints while creating the tables, isn't it ? cc @olavloite

arpan14: Yes, thanks for pointing out. This shouldn't contain `ON DELETE CASCADE`. I'll remove the same and let the sample take the action.

asthamohta (approved):

PR #2367, opened 2023-04-07
Revert "build(deps): update dependency org.jacoco:jacoco-maven-plugin to v0.8.9"

olavloite:
Reverts googleapis/java-spanner#2363

@rajatbhatta It seems that we need to revert this, as the integration tests have started failing after this update with an error related to this version bump.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ If you are using Maven without the BOM, add this to your dependencies:
 If you are using Gradle 5.x or later, add this to your dependencies:
 
 ```Groovy
-implementation platform('com.google.cloud:libraries-bom:26.11.0')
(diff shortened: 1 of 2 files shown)


rajatbhatta:
@olavloite: thanks but it’s weird that all github checks passed before we merged that version upgrade PR. We should analyze why that PR merge wasn’t blocked.

rajatbhatta (approved):

olavloite:
> @olavloite: thanks but it’s weird that all github checks passed before we merged that version upgrade PR. We should analyze (maybe after this is fixed) why that PR merge wasn’t blocked.

Yes, I noticed that too, and that is weird. As far as I can see, the PR did run all tests, including the integration tests that started failing after it was merged.

PR #2373, opened 2023-04-10
feat: add support for Directed Read options

rajatbhatta:
This PR contains implementation, unit tests and integration tests for supporting DirectedReads in Cloud Spanner Java Client.

DirectedReads feature enables customers to load-balance their workload across multiple regions and to achieve better/uniform CPU utilization for the multi-region instance config.


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerUtil.java
```diff
@@ -0,0 +1,65 @@
...
+        throw SpannerExceptionFactory.newSpannerException(
+            ErrorCode.FAILED_PRECONDITION,
+            "DirectedReadOptions can't be set for Read-Write or Partitioned DML transactions");
+      }
+    }
+    // If DirectedReadOptions is not set at request-level, the request object won't be
+    // having DirectedReadOptions field set. Though, if DirectedReadOptions is set at client-level
+    // (through SpannerOptions), we must modify the request object to set the DirectedReadOptions
+    // proto field to this value.
+    if ((directedReadOptionsForRequest == null) && (directedReadOptionsForClient != null)) {
```
olavloite, line 60: nit: remove redundant parentheses

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
```diff
@@ -227,6 +228,11 @@ public static CreateUpdateDeleteAdminApiOption validateOnly(Boolean validateOnly
     return new ValidateOnlyOption(validateOnly);
   }
 
+  /** Option to request DirectedRead for ReadOnlyTransaction and SingleUseTransaction. */
```
olavloite, line 231: Can we elaborate on what this option does? I have no idea when I should use directed read and when I should not.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
```diff
@@ -387,6 +387,8 @@ void initTransaction() {
   @GuardedBy("lock")
   protected boolean isClosed = false;
 
+  protected boolean readOnly;
```
olavloite, line 390: Can this be final?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1081,6 +1085,25 @@ public Builder setCompressorName(@Nullable String compressorName) {
       return this;
     }
 
+    /**
+     * Sets the {@link DirectedReadOptions} that specify which replicas or regions should be used
+     * for non-transactional reads or queries.
+     *
+     * <p>Note that DirectedReadOptions can only be set for ReadOnlyTransaction and
+     * SingleUseTransaction; it cannot be set for ReadWriteTransaction and
+     * PartitionedDmlTransaction.
```
olavloite, line 1094: Is this relevant here? This method is not running a transaction.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1081,6 +1085,25 @@ public Builder setCompressorName(@Nullable String compressorName) {
...
+     *
+     * <p>DirectedReadOptions set at the request level will take precedence over the options set
+     * using this method. If DirectedReadOptions are set at the request level, they will override
+     * the options set here when the RPC call is made.
```
olavloite, line 1098: This seems to be saying the same twice

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1081,6 +1085,25 @@ public Builder setCompressorName(@Nullable String compressorName) {
...
+     */
+    public Builder setDirectedReadOptions(DirectedReadOptions directedReadOptions) {
+      Preconditions.checkNotNull(directedReadOptions, "DirectedReadOptions cannot be null");
+      SpannerUtil.verifyDirectedReadOptions(directedReadOptions);
```
olavloite, line 1102: nit: Could we let the `SpannerUtil#verifyDirectedReadOptions` method return the options that it was given if they are valid? That way the assignment can be on the same line. (I.e. the same as `Preconditions.checkNotNull(..)`)

olavloite (commented):
Do we want to support this in the JDBC driver and other related libraries as well? If so, we would need to add support for it to the Connection API.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerUtil.java
```diff
@@ -0,0 +1,65 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.spanner;
+
+import com.google.spanner.v1.DirectedReadOptions;
+
+/** Utility methods for Spanner. */
+public class SpannerUtil {
```
arpan14, line 22: Can we call this class something specific like `DirectedReadsUtil`? One single class `SpannerUtil` for all utility method would eventually reduce the readability of this class since it will hold all kind of utilities. 

Keeping the class specific would be more maintainable. If you agree with this it will be good to have a sub-folder as well
`google-cloud-spanner/src/main/java/com/google/cloud/spanner/util/`

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerUtil.java
```diff
@@ -0,0 +1,65 @@
...
+  static final int MAX_REPLICA_SELECTIONS_COUNT = 10;
+
+  public static void verifyDirectedReadOptions(DirectedReadOptions directedReadOptions) {
```
arpan14, line 25: IMO `validateDirectedReadOptions` would be a better name.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerUtil.java
arpan14, line 60: It's more readable to tweak this condition to 
```
if (directedReadOptionsForRequest != null) {
return directedReadOptionsForRequest;
}
return directedReadOptionsForClient;
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerUtil.java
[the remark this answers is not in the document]
rajatbhatta: This is being used in other packages (`com.google.cloud.spanner.spi.v1`) and thus need to be public. I can mark it as an InternalApi.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
```diff
@@ -161,7 +162,7 @@ public static ListOption pageSize(int pageSize) {
    * is not generally available now).
    */
   @BetaApi
-  public static DataBoostQueryOption dataBoostEnabled(Boolean dataBoostEnabled) {
+  public static ReadAndQueryOption dataBoostEnabled(Boolean dataBoostEnabled) {
```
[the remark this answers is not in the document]
rajatbhatta: I've removed this change.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
rajatbhatta: We'll not be able to change its value if it's final. We need to change its value in its subclasses to return whether it's read-only or not.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
rajatbhatta: Removed.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
rajatbhatta: Removed the latter line.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerUtil.java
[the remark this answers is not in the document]
rajatbhatta: This is being used in other packages (`com.google.cloud.spanner.spi.v1`) and thus need to be public. I can mark it as an `InternalApi`.

rajatbhatta: I've renamed it to DirectedReadsUtil. Moved it to util/ as well.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
rajatbhatta: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerUtil.java
rajatbhatta: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerUtil.java
rajatbhatta: Makes sense. Done!

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
rajatbhatta: Added some more documentation.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/util/DirectedReadsUtil.java
```diff
@@ -0,0 +1,71 @@
...
+              MAX_REPLICA_SELECTIONS_COUNT));
+    }
+    return directedReadOptions;
+  }
+
+  public static DirectedReadOptions validateAndGetPreferredDirectedReadOptions(
+      DirectedReadOptions directedReadOptionsForClient,
+      DirectedReadOptions directedReadOptionsForRequest,
+      boolean readOnly) {
+    if (!readOnly) {
```
arpan14, line 55: This will end up failing all the RW transactions if the directed read options is set for the client. IMO that shouldn't be the right behaviour. Instead this setting should be silently ignored for RW transactions and considered only for RO transactions.

@olavloite WDYT? 

If we avoid adding this validation (and avoid exception) apart from improving the customer experience, we will also simplify a lot of things done in this PR (for ex - introducing `readOnly` across multiple method interfaces) .

olavloite: I don't remember exactly how this feature works, but in general I would say:
1. If the configuration is set at the client level (so configure once for all transactions executed by this client): This check should be removed.
2. If the option is set at the request level, which for a read/write transaction would mean something like:
   1. Start read/write transaction
   2. Execute a query with a `DirectedReadOption` set.
   3. Then we should fail at this point instead of silently ignoring, as the user is explicitly trying to use a feature with a read/write transaction that is not supported in that case. Silently ignoring might cause the user to believe that the feature is working.

arpan14: The configuration can be set at both the client as well as request level. The flexibility is provided at client level so that it can be treated as a global option and avoid the need to set it per RPC. If custom sets the config both at client as well as request, then the option at client is ignored in favour of the option set at the request.

arpan14:
Closing this PR in favor of https://github.com/googleapis/java-spanner/pull/2766 . I tried re-using this PR but there are too many merge conflicts and now we are using a new approach to resolve the default options (client + RPC level option provided) instead of doing at at `GapicSpannerRPC` file.

PR #2396, opened 2023-04-20
feat: add TransactionExecutionOptions support to executor.

ko3a4ok:
This PR adds support for TransactionExecutionOptions to Cloud Client Executor Framework.


google-cloud-spanner-executor/src/main/java/com/google/cloud/executor/spanner/CloudClientExecutor.java
```diff
@@ -227,13 +228,16 @@ private static class ReadWriteTransaction {
...
     private SpannerException error;
     private final String transactionSeed;
+    private final boolean optimistic;
     // Set to true when the transaction runner completed, one of these three could happen: runner
     // committed, abandoned or threw an error.
     private boolean runnerCompleted;
 
-    public ReadWriteTransaction(DatabaseClient dbClient, String transactionSeed) {
+    public ReadWriteTransaction(
+        DatabaseClient dbClient, String transactionSeed, boolean optimistic) {
```
rajatbhatta, line 237: Do we have plans to add more such transaction options in the future (for example, opaque)? If that's the case, adding TransactionExecutionOptions as a parameter in the constructor would allow it to scale better when more fields are added to the proto.

rajatbhatta (approved):

PR #2419, opened 2023-05-09
feat: long running transaction clean up background task. Adding configuration options for closing inactive transactions.

arpan14:
Adding a couple of option in `SessionPoolOptions` for auto-detecting long running transactions, logging the stack-trace and removing such transactions from consuming more resources. There can be scenarios where transactions are unexpectedly taking more time and hence can consume the `session` for longer durations. With this change, customers get two new options which can be set while providing `SessionPoolOptions`
1. `setWarnIfInactiveTransactions()` - Calling this method will generate warning logs and help in identifying faulty code which is leading to incorrect consumption of sessions.
2. `setWarnAndCloseIfInactiveTransactions()` - Calling this method will close the faulty transactions that are consuming the resources. We will also be generating the warning logs which can be later referred to know the transactions that were closed through this option.

Note : The feature is currently under development. The final usage pattern will be updated in a subsequent commit.


google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
```diff
@@ -195,6 +198,160 @@ public void tearDown() {
...
+          assertNotNull(manager.getCommitTimestamp());
+          break;
+        } catch (AbortedException e) {
+          transaction = manager.resetForRetry();
+        }
+      }
+    }
+    Instant endExecutionTime = client.pool.poolMaintainer.lastExecutionTime;
+
+    assertThat(client.pool.getNumberOfSessionsInPool()).isEqualTo(1);
```
olavloite, line 251: nit: use `assertEquals`

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
```diff
@@ -195,6 +198,160 @@ public void tearDown() {
...
+      while (true) {
+        try {
+          // Simulate a delay of 20s to ensure that the below transaction is a long-running one.
+          // We require to wait for 20s so that the main thread executes at-least once every 10s
+          // As per this test, anything which takes more than 5s is long-running
+          mockSpanner.setExecuteSqlExecutionTime(
+              SimulatedExecutionTime.ofMinimumAndRandomTime(
+                  (int) Duration.ofSeconds(20).toMillis(), 0));
+          transaction.executeUpdate(UPDATE_STATEMENT);
+          manager.commit();
```
olavloite, line 241: This should fail, as the session has been returned to the pool. We should make sure we take measures that once we have forced the session back into the pool and someone is still using it, they get a reasonable error message telling them what went wrong. This will randomly succeed or fail in real life, depending on whether the session has been garbage collected by the backend or not.

I also wonder how hard it will fail if the session actually has been garbage collected, because the client also contains a retry-loop for `SessionNotFoundException`. So in that case the following will happen:
1. The session has been removed from the pool and garbage collected by the backend.
2. Trying to use it will lead to a `SessionNotFoundException`.
3. The client library will try to replace the session that was used by the transaction with a new one. But the session that was used has already been removed. What happens in that case?

olavloite (commented):
The log of the integration tests for the emulator (and possibly also for the other test runs) are spammed with a very large number of warnings for closed long-running transactions.
1. Is that expected? Or are we accidentally removing a lot more sessions from the pool than we should?
2. If it is expected, can we somehow silence that log? It now makes the log too large to be really useful.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1659,16 +1679,24 @@ final class PoolMaintainer {
...
+    /**
+     * Variable maintaining the last execution time of the long-running transaction cleanup task.
+     *
+     * <p>The long-running transaction cleanup needs to be performed every X minutes. The X minutes
+     * recurs multiple times within the invocation of the pool maintainer thread. For ex - If the
+     * main thread runs every 10s and the long-running transaction clean-up needs to be performed
+     * every 2 minutes, then we need to keep a track of when was the last time that this task
+     * executed and makes sure we only execute it every 2 minutes and not every 10 seconds.
+     */
+    @VisibleForTesting public volatile Instant lastExecutionTime;
```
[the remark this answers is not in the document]
arpan14: Changed to package-private. That's sufficient for tests.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
arpan14: `assertThat` is already used with a lot of the existing tests in the class. Hence, I am continuing to retain this for the new tests.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -180,6 +190,18 @@ public boolean isAutoDetectDialect() {
     return autoDetectDialect;
   }
 
+  public boolean closeInactiveTransactions() {
```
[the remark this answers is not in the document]
arpan14: There are two conventions being followed across languages. My guess is, the external options (within `SpannerOptions` in case of Java) has references to `InactiveTransaction` since that's how we are explaining the feature to to an external customer. Any other files can use 'long-running sessions' terminology.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1792,6 +1828,60 @@ private void replenishPool() {
...
+        }
+      } catch (final Throwable t) {
+        logger.log(Level.WARNING, "Failed removing long running transactions", t);
+      }
+    }
+
+    private void removeLongRunningSessions(
+        final Instant currentTime,
+        final InactiveTransactionRemovalOptions inactiveTransactionRemovalOptions) {
+      synchronized (lock) {
```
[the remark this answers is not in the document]
arpan14: Yes, it should be safe to remove lock from here since it internally invokes another method `removeLongRunningSessions` which has the necessary locks.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -50,6 +50,9 @@ public class SessionPoolOptions {
   private final ActionOnSessionNotFound actionOnSessionNotFound;
   private final ActionOnSessionLeak actionOnSessionLeak;
   private final boolean trackStackTraceOfSessionCheckout;
+  private final ActionOnInactiveTransaction actionOnInactiveTransaction;
```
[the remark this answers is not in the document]
arpan14: Yes, that can be done. But I felt it was a better semantic to have all our session leak options at one place. For ex - `ActionOnSessionLeak` and `ActionOnSessionNotFound` are just above this. `InactiveTransactionRemovalOptions` ideally wouldn't exist apart from being used in the unit tests to tweak other-wise private properties. Hence, it will be nice to have this separation between the two properties - `ActionOnInactiveTransaction` and `InactiveTransactionRemovalOptions` .

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -274,6 +409,7 @@ private Builder(SessionPoolOptions options) {
       this.actionOnSessionNotFound = options.actionOnSessionNotFound;
       this.actionOnSessionLeak = options.actionOnSessionLeak;
       this.trackStackTraceOfSessionCheckout = options.trackStackTraceOfSessionCheckout;
+      this.actionOnInactiveTransaction = options.actionOnInactiveTransaction;
```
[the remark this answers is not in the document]
arpan14: This is the method called from public `toBuilder` method. We are not setting `inactiveTransactionRemovalOptions` here since its only used in our tests and hence is not needed to be passed through public methods. Customers only need to set `actionOnInactiveTransaction` by calling one of the two public method `setCloseIfInactiveTransactions` or `setWarnIfInactiveTransactions`.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
```diff
@@ -195,6 +198,160 @@ public void tearDown() {
...
+    try (TransactionManager manager = client.transactionManager()) {
+      TransactionContext transaction = manager.begin();
+      while (true) {
+        try {
+          // Simulate a delay of 20s to ensure that the below transaction is a long-running one.
+          // We require to wait for 20s so that the main thread executes at-least once every 10s
+          // As per this test, anything which takes more than 5s is long-running
+          mockSpanner.setExecuteSqlExecutionTime(
+              SimulatedExecutionTime.ofMinimumAndRandomTime(
+                  (int) Duration.ofSeconds(20).toMillis(), 0));
```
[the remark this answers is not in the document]
arpan14: Yes, 20s is a large value. The reason I couldn't tweak it to be lower is since the main thread (for session maintenance) executes every 10s. I'll see if I can configure this to be a lower value, say every 2s and probably tweak all the `InactiveTransactionRemovalOptions` in order of milli-seconds and not seconds.

arpan14:
> The log of the integration tests for the emulator (and possibly also for the other test runs) are spammed with a very large number of warnings for closed long-running transactions.

Much of the spam is originating from `SessionPoolStressTest` class. This internally uses a fake clock which is initialising the last used time of the session to be 0 and this translates to epoch value `1970-01-21T16:53:00Z` . As a result all of the sessions were last used > threshold limit of 60 minutes and hence are getting classified as long running. 

> 1. Is that expected? Or are we accidentally removing a lot more sessions from the pool than we should?

I think it's expected. All sessions are getting classified as long running and then getting logged. Ideally, `lastUseTime` field should get initialised with current epoch value. Since we are using a fake clock instance, it's incorrectly initialised to 0 value. 
* Reference - https://github.com/googleapis/java-spanner/blob/977b5851b9cc1a538dedf57d006483dc58ec90b9/google-cloud-spanner/src/test/java/com/google/cloud/spanner/BaseSessionPoolTest.java#L87

> 2. If it is expected, can we somehow silence that log? It now makes the log too large to be really useful.

I agree, I think the right step would be to fix the initialisation of session `lastUseTime` value to reflect the current time. I'll see what's the best way to fix this. One way could be to to add below in `BaseSessionPoolTest`
```
volatile long currentTimeMillis = Instant.now().toEpochMilli();
```

The class is using `FakeClock` via `BaseSessionPoolTest`, so I need to see if changing this would affect other tests or not.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
arpan14: > This should fail, as the session has been returned to the pool.

This particular test did not fail since the session is not garbage collected in the backend. I can add another test where I throw a `SessionNotFoundException` and assert the behaviour. This particular test will continue to test the scenario where the session gets cleaned up on the front end. Is there any other behaviour that should be asserted for this case when the front-end returns the session while the backend hasn't garbage collected?

> We should make sure we take measures that once we have forced the session back into the pool and someone is still using it, they get a reasonable error message telling them what went wrong.

I will verify this through a new test, but I guess bubbling up a general `ResourceNotFound` / `SessionNotFound` error would be good enough? We originally thought of adding some messaging like `Transaction has been closed as it was running for more than 60 minutes. For long running transactions use batch or partitioned transactions` . But I think this is one of the reasons why a session is closed and we shouldn't be bubbling this to clients in all cases.

> The client library will try to replace the session that was used by the transaction with a new one. But the session that was used has already been removed. What happens in that case?

In the `replaceSession` flow we invalidate the previously used session. There is a check to do nothing in case the session is already in the closed state, hence I don't think we need anymore change here. I'll check through a test whether the flow here is expected or not. Overall, for the customer the transaction will still pass since we will replace with a new session.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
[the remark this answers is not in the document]
olavloite: There are many other fields as well that are not set by users, but we still set them in this constructor. Imagine the confusion if someone adds a test that actually uses this field, and then calls this method to create a copy of their SessionPoolOptions and the value for this field not being copied (while everything else is).

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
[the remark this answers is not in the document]
olavloite: I would follow the logic of `ActionOnInactiveTransaction` is part of the `InactiveTransactionRemovalOptions` (put another way: It is always used in combination with those settings). The fact that the other fields of `InactiveTransactionRemovalOptions` are (currently) package-private does not change that.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/MockSpannerServiceImpl.java
```diff
@@ -596,7 +596,7 @@ private static void checkStreamException(
   private int maxNumSessionsInOneBatch = 100;
   private int maxTotalSessions = Integer.MAX_VALUE;
   private AtomicInteger numSessionsCreated = new AtomicInteger();
-
+  private boolean onCommitThrowSessionNotFoundException = false;
```
olavloite, line 599: Could we do the following instead of adding yet another field to this class:


1. Extract the creation of a `SessionNotFound` error from `setSessionNotFound` into a public method, so something like this:

```java
  public StatusRuntimeException createSessionNotFoundException(String name) {
    ResourceInfo resourceInfo =
        ResourceInfo.newBuilder()
            .setResourceType(SpannerExceptionFactory.SESSION_RESOURCE_TYPE)
            .setResourceName(name)
            .build();
    Metadata.Key<ResourceInfo> key =
        Metadata.Key.of(
            resourceInfo.getDescriptorForType().getFullName() + Metadata.BINARY_HEADER_SUFFIX,
            ProtoLiteUtils.metadataMarshaller(resourceInfo));
    Metadata trailers = new Metadata();
    trailers.put(key, resourceInfo);
    return
        Status.NOT_FOUND
            .withDescription(String.format("Session not found: Session with id %s not found", name))
            .asRuntimeException(trailers);
  }
```

2. Re-use the above method in the `setSessionNotFound` method.
3. Use the following in the test method(s) that use this:

```java    mockSpanner.setCommitExecutionTime(SimulatedExecutionTime.ofException(mockSpanner.createSessionNotFoundException("test-session-name")));
```

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
olavloite: I think this is good. My initial mistake was thinking that the session was forced back to the pool, but we are not doing that. We are removing it from the pool alltogether.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
```diff
@@ -195,6 +198,177 @@ public void tearDown() {
...
+            .build()
+            .getService();
+    DatabaseClientImpl client =
+        (DatabaseClientImpl)
+            spanner.getDatabaseClient(DatabaseId.of(TEST_PROJECT, TEST_INSTANCE, TEST_DATABASE));
+    Instant initialExecutionTime = client.pool.poolMaintainer.lastExecutionTime;
+
+    try (TransactionManager manager = client.transactionManager()) {
+      TransactionContext transaction = manager.begin();
+      mockSpanner.setOnCommitThrowSessionNotFoundException(true);
```
olavloite, line 234: See also my comment further below; This should use `mockSpanner.setExecutionTime(SimulatedExcecutionTime.ofException(..))`.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
```diff
@@ -195,6 +198,177 @@ public void tearDown() {
...
+      while (true) {
+        try {
+          // Simulate a delay of 4s to ensure that the below transaction is a long-running one.
+          // We require to wait for 4s so that the main thread executes at-least once every 1s
+          // As per this test, anything which takes more than 2s is long-running
+          mockSpanner.setExecuteSqlExecutionTime(
+              SimulatedExecutionTime.ofMinimumAndRandomTime(
+                  (int) Duration.ofSeconds(4).toMillis(), 0));
+          transaction.executeUpdate(UPDATE_STATEMENT);
```
olavloite, line 243: Could we try to solve this using the following strategy instead:
1. Add `Clock` as a package-private option to `SessionPoolOptions`, so we can set that in a test like this as well.
2. Then instead of adding a long wait to an RPC, just manually advance the clock after executing the statement, and then force a maintenance loop. Note: The long-running transaction handler is in real-life not very likely to kick in while an RPC is running, it is much more likely to kick in while there is nothing more going on, because the application has leaked the session.

Regarding point 2 above: Ideally, the solution that we are adding here should not even kick in at all in this scenario, as the transaction is actually still running and has an active request. The fact that this request takes a 'long' time is something that should not be seen as inactivity. Instead, if the statement runs for too long, it is the deadline for that specific statement that should let it fail. After that, the transaction could be eligible for being regarded as inactive.

olavloite: Hmmm.... Coming back on my comment on point 2 above: It is probably not really a problem if we allow sessions with an RPC to be cleaned up by this task. Although it's not the optimal solution, it's probably the easiest, and as far as I can see it won't cause any problems, because:
1. The fact that the session is removed from the pool while still active does not make the session unusable for the transaction that is already using it. It only means that it won't be used by any other transactions in the future.
2. We're apparently also able to handle the fact that the transaction tries to return a session to the pool that the pool no longer knows.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
olavloite: We need to think this through a bit more. Currently, if the `commit()` succeeds without any errors, the following happens:
1. The session is removed from the pool by the cleanup task.
2. The transaction continues, as it is not really inactive. It was just executing an RPC that was slow.
3. The `commit` succeeds and the `TransactionManager` is closed. This releases the session back into the pool. This again leaves the session pool in the following inconsistent state:
4.1. `SessionPool#allSessions` does not contain the session anymore.
4.2. `SessionPool#sessions` (the list of sessions currently checked in to the pool) does contain the session.

So we need to get this consistent, by either:
1. Making sure that a session that has been cleaned up by the background task cannot in any way be returned to the pool, even if the transaction that had it succeeded.
2. OR: Make sure that a session that had been cleaned up but that turns out to be alive after all can be fully returned to the pool (that is; the 'remove' action is completely undone)

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
```diff
@@ -195,6 +198,179 @@ public void tearDown() {
...
+    // On retry, session again found to be long-running and cleaned up.
+    // During commit, there was no exception from backend.
+    final int numSessionsInPool = client.pool.getNumberOfSessionsInPool();
+    final long numLeakedSessionsRemoved = client.pool.numLeakedSessionsRemoved();
+
+    assertNotEquals(
+        endExecutionTime,
+        initialExecutionTime); // if session clean up task runs then these timings won't match
+    assertTrue(numSessionsInPool > 0 && numSessionsInPool < 3);
+    assertThat(numLeakedSessionsRemoved > 0 && numLeakedSessionsRemoved <= 2);
```
olavloite, line 269:
```suggestion
    assertTrue(numLeakedSessionsRemoved > 0 && numLeakedSessionsRemoved <= 2);
```

This is currently causing this test to falsely report that everything succeeded, as it is not really verifying anything. The expression in the `assertThat` call is actually false at the moment.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
arpan14: Yes, I have added this strategy now. And have removed the need to introduce delays in mock RPC call.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
arpan14: We concluded to favour the approach #1 where an exception never surfaces to the customer in case the session is recycled on client side. This is also consistent with what's done today when session is garbage collected on backend and we have a retry loop (after replacing client side session). 

I have added the necessary handling in the `Session#close()` method to ensure a session that is removed from the pool is not added back to the pool during close. Similar handling is already present in case of `SessionNotFound` exception.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1659,16 +1697,24 @@ final class PoolMaintainer {
...
+    /**
+     * Variable maintaining the last execution time of the long-running transaction cleanup task.
+     *
+     * <p>The long-running transaction cleanup needs to be performed every X minutes. The X minutes
+     * recurs multiple times within the invocation of the pool maintainer thread. For ex - If the
+     * main thread runs every 10s and the long-running transaction clean-up needs to be performed
+     * every 2 minutes, then we need to keep a track of when was the last time that this task
+     * executed and makes sure we only execute it every 2 minutes and not every 10 seconds.
+     */
+    @VisibleForTesting volatile Instant lastExecutionTime;
```
olavloite, line 1716: nit: Does this need to be `volatile`? `volatile` is only needed if multiple threads could potentially update the field, but I think that only the maintenance thread will ever update this field.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1792,6 +1846,75 @@ private void replenishPool() {
...
+              }
+              if ((options.closeInactiveTransactions()
+                      || options.warnAndCloseInactiveTransactions())
+                  && session.state != SessionState.CLOSING) {
+                final boolean isRemoved = removeFromPool(session);
+                if (isRemoved) {
+                  session.isRemovedFromPool = true;
+                  numLeakedSessionsRemoved++;
+                  if (longRunningSessionRemovedListener != null) {
+                    longRunningSessionRemovedListener.apply(session);
```
olavloite, line 1908: Just to be sure: The `longRunningSessionRemovedListener` field is something that only the client library itself can set (and is only used for testing), right? The reason I'm checking is that we are inside a `synchronized (lock)`, and if this field can be set by a user, then it could also in theory be executing a blocking or otherwise heavy operation, which we do not want while holding the lock.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -223,6 +259,132 @@ private enum ActionOnSessionLeak {
     FAIL
   }
 
+  @VisibleForTesting
+  enum ActionOnInactiveTransaction {
+    WARN,
+
```
olavloite, line 265: nit: remove empty line

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -369,6 +543,71 @@ public Builder setBlockIfPoolExhausted() {
       return this;
     }
 
+    /**
+     * If there are inactive transactions, log warning messages with the origin of such transactions
+     * to aid debugging. A transaction is classified as inactive if it executes for more than a
+     * system defined configuration.
```
olavloite, line 549:
```suggestion
     * system defined duration.
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -369,6 +543,71 @@ public Builder setBlockIfPoolExhausted() {
...
+          InactiveTransactionRemovalOptions.newBuilder()
+              .setActionOnInactiveTransaction(ActionOnInactiveTransaction.WARN)
+              .build();
+      return this;
+    }
+
+    /**
+     * If there are inactive transactions, release the resources consumed by such transactions. A
+     * transaction is classified as inactive if it executes for more than a system defined
+     * configuration. The option would also produce necessary warning logs through which it can be
```
olavloite, line 567:
```suggestion
     * duration. The option would also produce necessary warning logs through which it can be
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -369,6 +543,71 @@ public Builder setBlockIfPoolExhausted() {
...
+     * debugged as to what resources were released due to this option.
+     *
+     * <p>Use the option {@link Builder#setWarnIfInactiveTransactions()} if you only want to log
+     * warnings about long-running sessions.
```
olavloite, line 571:
```suggestion
     * warnings about long-running transactions.
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -369,6 +543,71 @@ public Builder setBlockIfPoolExhausted() {
...
+          InactiveTransactionRemovalOptions.newBuilder()
+              .setActionOnInactiveTransaction(ActionOnInactiveTransaction.WARN_AND_CLOSE)
+              .build();
+      return this;
+    }
+
+    /**
+     * If there are inactive transactions, release the resources consumed by such transactions. A
+     * transaction is classified as inactive if it executes for more than a system defined
+     * configuration.
```
olavloite, line 586:
```suggestion
     * duration.
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -369,6 +543,71 @@ public Builder setBlockIfPoolExhausted() {
...
+     *
+     * <p>Use the option {@link Builder#setWarnIfInactiveTransactions()} if you only want to log
+     * warnings about long-running sessions.
+     *
+     * <p>Use the option {@link Builder#setWarnAndCloseIfInactiveTransactions()} if you want to log
+     * warnings along with closing the long-running sessions.
```
olavloite, line 592:
```suggestion
     * warnings along with closing the long-running transactions.
```

olavloite (approved):
LGTM (with some minor nits and questions)

Thanks for your perseverance on this one.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Yes, `longRunningSessionRemovedListener` is package private and only used within our tests. It cannot be set by external customers.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1366,6 +1367,22 @@ final class PooledSession implements Session {
     private volatile SpannerException lastException;
     private volatile boolean allowReplacing = true;
 
+    /**
+     * Property to mark if the session is eligible to be long-running. This can only be true if the
```
asthamohta, line 1371: Can we add here what long running is, as in the 60 minute limit?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2015,18 +2147,26 @@ int getNumberOfSessionsInUse() {
...
       if (isClosed()) {
         decrementPendingClosures(1);
-        return;
+        return false;
       }
       session.markClosing();
       allSessions.remove(session);
-      numIdleSessionsRemoved++;
-    }
-    if (idleSessionRemovedListener != null) {
```
asthamohta, line 2028: Are we no longer going to recycle idle sessions?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1792,6 +1846,75 @@ private void replenishPool() {
...
+            final Duration durationFromLastUse = Duration.between(session.lastUseTime, currentTime);
+            if (!session.eligibleForLongRunning
+                && durationFromLastUse.compareTo(
+                        inactiveTransactionRemovalOptions.getIdleTimeThreshold())
+                    > 0) {
+              if ((options.warnInactiveTransactions() || options.warnAndCloseInactiveTransactions())
+                  && !session.isLeakedExceptionLogged) {
+                logger.log(
+                    Level.WARNING,
+                    String.format("Removing long running session => %s", session.getName()),
```
asthamohta, line 1896: Can we have a stack for the transaction here to help trace which txn is causing the customer this issue?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
asthamohta: Found the answer below. removing the comment

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -223,6 +259,135 @@ private enum ActionOnSessionLeak {
...
+  static class InactiveTransactionRemovalOptions {
+
+    /** Option to set the behaviour when there are inactive transactions. */
+    private ActionOnInactiveTransaction actionOnInactiveTransaction;
+
+    /**
+     * Frequency for closing inactive transactions. Between two consecutive task executions, it's
+     * ensured that the duration is greater or equal to this duration.
+     */
+    private Duration executionFrequency;
```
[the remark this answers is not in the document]
rajatbhatta: Please update the setters and getters for the field appropriately, including method names.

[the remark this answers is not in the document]
arpan14: We changed this variable naming from `recurrenceDuration` to `executionFrequency` based on previous review feedback. I agree that this is not aligning with the textbook definition of frequency but the way I interpret it is - this variable answers "How frequently does the task run? Is it every 2 minutes? Is it every 2 hours".

I hope it's not bringing a lot of confusion.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -223,6 +259,135 @@ private enum ActionOnSessionLeak {
...
 
+  @VisibleForTesting
+  enum ActionOnInactiveTransaction {
+    WARN,
+    WARN_AND_CLOSE,
+    CLOSE
+  }
+
+  /** Configuration options for task to clean up inactive transactions. */
+  static class InactiveTransactionRemovalOptions {
```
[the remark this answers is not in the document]
arpan14: "Long-Running" is one of the reasons why transaction can be inactive. Tomorrow if there are other reasons why a transaction can be in-active it will still remain abstracted in inner layers and won't be seen as a pool option.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: @rajatbhatta Yes, you are right. `sessionFuture.leakedException` will have the stack trace of the transaction. @asthamohta We have all the needed information to debug.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1572,6 +1597,15 @@ void markUsed() {
...
     }
 
+    /**
+     * Method to mark a session that is to be occupied by a possibly long-running transaction. Any
+     * transaction that is expected to be long-running (for ex - Partitioned DML, Batch Read) must
+     * use this method.
+     */
+    void markEligibleForLongRunning() {
+      eligibleForLongRunning = true;
+    }
```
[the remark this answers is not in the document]
arpan14: Thanks for pointing out. I'll try to explain the small differences here.

* Yeah, I have added a method since there can be more than one transaction types (PDML, Batch Reads) that can be eligible for long running. And hence, it will be cleaner for all such transaction types to call a single method. Today, BatchReads don't use the session pool hence it looks as if PDML is the only type that calls `markEligibleForLongRunning` . But in general the method can be used by other transaction types.
* We don't need a method `eligibleForLongRunning = false` since its a common logic that's present in two places for all transactions a) during initialisation and b) during Session#close() method. Transactions don't need to set this flat to false.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: * I was avoiding adding the exact configuration value into docs since it may change in future and i'd like to avoid the docs getting stale. 
* That said, in the documentation of `SessionPool#PoolMaintainer` class we have explicitly given some examples where we have mentioned 60 minutes.

Let me know if the readability sounds ok.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -223,6 +259,135 @@ private enum ActionOnSessionLeak {
     FAIL
   }
 
+  @VisibleForTesting
+  enum ActionOnInactiveTransaction {
```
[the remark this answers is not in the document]
arpan14: @rajatbhatta The `InactiveTransactionRemovalOptions` class has the necessary documentation. Also we have added documentation for the field `private ActionOnInactiveTransaction actionOnInactiveTransaction;`

Documentation for each enum option would have made sense. But we have added documentation on the respective public methods (`setWarnIfInactiveTransactions`, `setWarnAndCloseIfInactiveTransactions`, `setCloseIfInactiveTransactions`) which are responsible for setting this enum option.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1366,6 +1367,22 @@ final class PooledSession implements Session {
...
+     */
+    private volatile boolean eligibleForLongRunning = false;
+
+    /**
+     * Property to mark if the session is no longer part of the session pool For ex - A session
+     * which is long-running gets cleaned up and removed from the pool.
+     */
+    private volatile boolean isRemovedFromPool = false;
+
+    private volatile boolean isLeakedExceptionLogged = false;
```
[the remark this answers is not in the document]
arpan14: Added doc.

asthamohta (approved):

rajatbhatta (approved):
LGTM

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -334,8 +334,7 @@ static InactiveTransactionRemovalOptions.Builder newBuilder() {
     }
 
     static class Builder {
-      private ActionOnInactiveTransaction actionOnInactiveTransaction =
-          ActionOnInactiveTransaction.WARN;
+      private ActionOnInactiveTransaction actionOnInactiveTransaction;
```
olavloite, line 337: Just to be sure: Are we sure that all code can safely handle that this value is now `null`?

olavloite (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
arpan14: Yes, all places safely handle the null value.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -369,6 +545,70 @@ public Builder setBlockIfPoolExhausted() {
...
+     * transaction is classified as inactive if it executes for more than a system defined duration.
+     * The option would also produce necessary warning logs through which it can be debugged as to
+     * what resources were released due to this option.
+     *
+     * <p>Use the option {@link Builder#setWarnIfInactiveTransactions()} if you only want to log
+     * warnings about long-running transactions.
+     *
+     * @return this builder for chaining
+     */
+    Builder setWarnAndCloseIfInactiveTransactions() {
```
psinghbay1, line 577: why are we making these methods package protected?

psinghbay1:
> 

Any reason we are keeping these methods package protected in builder class?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
arpan14: @psinghbay1 This feature is not yet launched. Hence it is package protected in this PR. We will be enabling this in the next couple of weeks.

PR #2441, opened 2023-05-18
build: enable multiScm for releases

chingor13:


chingor13:
synthtool templates will update this one

PR #2442, opened 2023-05-19
fix: avoid unbalanced session pool creation

olavloite:
A query storm at the startup of the client library before the session pool had initialized could cause the creation of an unbalanced session pool. This again would put a large batch of sessions using the same gRPC channel at the head of the pool, which could then continously be used by the application.


google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolStressTest.java
```diff
@@ -92,6 +93,7 @@ private void setupSpanner(DatabaseId db) {
     when(spannerOptions.getNumChannels()).thenReturn(4);
     when(spannerOptions.getDatabaseRole()).thenReturn("role");
     SessionClient sessionClient = mock(SessionClient.class);
+    when(sessionClient.getSpanner()).thenReturn(mockSpanner);
```
arpan14, line 96: Curious as to why we need this mocking with current changes and did not require it earlier.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2423,6 +2471,48 @@ private void releaseSession(PooledSession session, Position position) {
     }
   }
 
+  private boolean isUnbalanced(PooledSession session) {
```
arpan14, line 2474: Can we add `@VisibleForTesting` and add a few more unit tests for this method? It's doing a bunch of things which is currently untested with the existing tests.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2396,18 +2427,35 @@ private void maybeCreateSession() {
...
+          // more efficient.
+          session.releaseToPosition = Position.FIRST;
+        }
+        switch (session.releaseToPosition) {
           case RANDOM:
             if (!sessions.isEmpty()) {
+              // A session should only be added at a random position the first time it is added to
+              // the pool. All following releases into the pool should happen at the front of the
+              // pool.
+              session.releaseToPosition = Position.FIRST;
```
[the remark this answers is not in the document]
olavloite: I don't think that adding it as a branch at L2451 would work. The basic logic is:
1. The session should normally only be added to the pool at a random position once.
2. So we need an if/switch case that determines whether it is currently `RANDOM`, and if it is, then it should be reset to  `FIRST` so it is not randomized the next time.

I've changed the switch case to an if-else, as we currently only support `FIRST` and `RANDOM`, so a switch does not really make sense. I think that makes the logic a bit clearer as well.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2423,6 +2471,48 @@ private void releaseSession(PooledSession session, Position position) {
...
+      if (channel == otherSession.getChannel()) {
+        count++;
+        if (count > 2) {
+          return true;
+        }
+      }
+    }
+    // Ideally, the use of a channel in the checked out sessions is exactly
+    // numCheckedOut / numChannels
+    // We check whether we are more than a factor two away from that perfect distribution.
```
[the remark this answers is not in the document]
olavloite: Benchmarking showed that this was a reasonable trade-off. There are two competing scenarios here:
1. A low-QPS scenario benefits from as little randomization as possible, because in that case it helps to reuse the same session for most queries (better cache hit ratio), while there is no real probability of overloading a single channel. Being too aggressive showed higher latencies for this scenario.
2. A high-QPS scenario benefits from as good distribution as possible, to prevent overloading a single channel.

We don't keep track of the QPS in the session pool, so the best substitute that we have for it is the number of checked out sessions. The factor 2 was chosen after trying varying factors and also other strategies, as this gave good results both for low and high QPS. (Although I've not tried with every single number between let's say 1.0 and 5.0)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2423,6 +2471,48 @@ private void releaseSession(PooledSession session, Position position) {
     }
   }
 
+  private boolean isUnbalanced(PooledSession session) {
+    // Don't bother with any randomization if the number of checked out sessions is low, as it is
```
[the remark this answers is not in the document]
olavloite: Changed to 're-balance'

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolMaintainerTest.java
```diff
@@ -111,9 +113,11 @@ private SessionImpl setupMockSession(final SessionImpl session) {
   }
 
   private SessionPool createPool() throws Exception {
+    // Allow sessions to be added to the head of the pool in all cases in this test, as it is
+    // otherwise impossible to know which session exactly is getting pinged at what point in time.
```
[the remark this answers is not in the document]
olavloite: That is actually not a typo. The reason that we need to do this is that this test runs the maintenance task in a tight loop, and has a number of checks that verifies which sessions were __pinged__. Those checks do not work if the order of the sessions in the pool is random.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2423,6 +2471,48 @@ private void releaseSession(PooledSession session, Position position) {
...
+    // better to re-use sessions as much as possible in a low-QPS scenario.
+    if (sessions.isEmpty() || checkedOutSessions.size() <= 2) {
+      return false;
+    }
+    int numChannels = sessionClient.getSpanner().getOptions().getNumChannels();
```
olavloite, line 2480: This line is invoked during the running of the `SessionPoolStressTest`, and is the reason that we need some additional mocking.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolStressTest.java
olavloite: The `isUnbalanced` method that was added to the session pool calls this method. See https://github.com/googleapis/java-spanner/pull/2442/files#r1318112669

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolStressTest.java
```diff
@@ -226,22 +229,26 @@ public void stressTest() throws Exception {
...
+            clock,
+            Position.RANDOM);
     pool.idleSessionRemovedListener =
         pooled -> {
           String name = pooled.getName();
-          synchronized (lock) {
-            sessions.remove(name);
-            return null;
-          }
+          // We do not take the test lock here, as we already hold the session pool lock. Taking the
```
olavloite, line 240: This change is unrelated to the session pool re-balancing, but it fixes a potential deadlock in this test. The deadlock was caused by the fact that the session pool maintenance task has taken the `SessionPool#lock` when it comes here. Trying to also take the `lock` of this test case could cause a deadlock if one of the tests in this file had already taken this lock, and then tried to execute an action on the session pool that would also require the `SessionPool#lock`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
olavloite: Yeah, good point. I've changed it a bit so we have a static method that does the actual calculation. That method is easier to test than one that depends on all the state fields in the SessionPool class. And I've added different tests for that method.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2472,38 +2466,79 @@ private void releaseSession(PooledSession session, boolean isNewSession) {
...
           return true;
         }
       }
     }
     // Ideally, the use of a channel in the checked out sessions is exactly
     // numCheckedOut / numChannels
     // We check whether we are more than a factor two away from that perfect distribution.
-    // If we are, then we randomize.
+    // If we are, then we re-balance.
+    count = 0;
```
olavloite, line 2538: This was a bug in the previous version that was found with the additional test cases.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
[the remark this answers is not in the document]
arpan14: Thanks for the explanation. The reasoning sounds good.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2396,33 +2427,127 @@ private void maybeCreateSession() {
...
+        } else if (session.releaseToPosition == Position.RANDOM
+            && !isNewSession
+            && checkedOutSessions.size() <= 2) {
+          // Do not randomize if there are few other sessions checked out and this session has been
+          // used. This ensures that this session will be re-used for the next transaction, which is
+          // more efficient.
+          session.releaseToPosition = Position.FIRST;
+        }
+        if (session.releaseToPosition == Position.RANDOM && !sessions.isEmpty()) {
+          // A session should only be added at a random position the first time it is added to
```
arpan14, line 2453: Due to the condition `!sessions.isEmpty()`, apart from the first session, all other sessions will enter the `else block` and get added to the front of the queue. If this is intended, then isn't this comment inconsistent?

How are we measuring if a session was added for the first time or has been added before. Shouldn't that be a different state variable that get's toggled the first time the session was added to the pool?

`Or are you saying session.releaseToPosition == Position.RANDOM` returns true only when session was released first time. I guess all other releases will toggle `session.releaseToPosition` to `Position.FIRST`

Assuming that session.releaseToPosition = Position.RANDOM indicates first time insertion isn't great right. And it can easily break the algorithm. 

WDYT?

olavloite: This comment was indeed outdated. It was initially written before adding the `isUnbalanced(..)` part to the solution, as  I initially thought that it would be enough to only randomize the order when a session was initially added to the pool. That was not enough, because it could easily happen that a session would be used multiple times before it would actually be released into the pool (that is; because there was always a waiter for a session, which meant that the session was given to that waiter instead of being released to the pool). In addition, it could still be that an initial randomization would give an unbalanced pool.

I've updated the comment to reflect the current situation.

> Due to the condition `!sessions.isEmpty()`, apart from the first session, all other sessions will enter the else block and get added to the front of the queue. If this is intended, then isn't this comment inconsistent?

Hmmm.... I'm not sure I understand your question correctly. The `!sessions.isEmpty()` ensures that __only__ the first session to be released into the pool enters the else block. We do that because adding it at a random place in an empty list does not make sense.

> How are we measuring if a session was added for the first time or has been added before. Shouldn't that be a different state variable that get's toggled the first time the session was added to the pool?

The original implementation would only set it to `RANDOM` when the session was created, and then reset it to `FIRST` here. So in that implementation, it was a valid indicator for whether it was the first time it was added. With the addition of checking for an unbalanced pool, that has changed (but the comment unfortunately not), and a session can be added at a random position multiple times, but the value is always reset to `FIRST` after it has been added, as that is the default as long as the pool remains balanced.

arpan14: Thanks. I had got confused thinking each session when it enters the queue first has to be put in front of the queue. But I think you meant only the first session. Now it's clear.

arpan14 (approved):

PR #2447, opened 2023-05-19
cleanup: remove old region tag (spanner_postgresql_delete_dml_returning)

Shabirmean:
### Fixes N/A

- Related to PR: https://github.com/googleapis/java-spanner/pull/2446
- After [cl/533581792](https://critique.corp.google.com/cl/533581792) is merged we must merge this PR to remove the old region tag


Shabirmean:

PR #2451, opened 2023-05-20
cleanup: remove old region tag (spanner_delete_dml_returning)

Shabirmean:
### Fixes N/A

- Related to PR: https://github.com/googleapis/java-spanner/pull/2450
- After [cl/533585863](https://critique.corp.google.com/cl/533585863) is merged we must merge this PR to remove the old region tag


Shabirmean:

PR #2462, opened 2023-05-23
feat: delay transaction start option

olavloite:
Adds an opt-in to delay the actual start of a read/write transaction until the first write operation. This reduces lock contention and can reduce latency for read/write transactions that execute (many) read operations before any write operations, at the expense of a lower transaction isolation level. Typical workloads that benefit from this option are ORMs (e.g. Hibernate). The application must be able to handle the lower isolation level.

This feature will automatically propagate into clients that use the Connection API, such as the JDBC driver and PGAdapter.


google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ReadWriteTransaction.java
```diff
@@ -225,13 +237,22 @@ public boolean isReadOnly() {
   }
 
   @Override
-  void checkValidTransaction() {
+  void checkValidTransaction(ParsedStatement statement) {
```
thiagotnunes, line 240: nit: `checkValidationTransaction` mutates state within the transaction (it did before as well with the future), so perhaps it is a good chance to rename the method? (very nitty, so feel free to ignore)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ReadWriteTransaction.java
```diff
@@ -622,7 +651,8 @@ public void onSuccess(long[] result) {}
   @Override
   public ApiFuture<Void> writeAsync(Iterable<Mutation> mutations) {
     Preconditions.checkNotNull(mutations);
-    checkValidTransaction();
+    // We actually don't need a transaction yet, as mutations are buffered until commit.
+    checkValidTransaction(null);
```
thiagotnunes, line 655: nit: can we avoid passing `null` here? Do we need this call to set the `transactionStarted` or can we avoid this call?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ReadWriteTransaction.java
```diff
@@ -644,11 +674,21 @@ public Void call() {
...
   @Override
   public ApiFuture<Void> commitAsync() {
-    checkValidTransaction();
+    checkValidTransaction(COMMIT_STATEMENT);
     state = UnitOfWorkState.COMMITTING;
     commitResponseFuture = SettableApiFuture.create();
     ApiFuture<Void> res;
-    if (retryAbortsInternally) {
+    // Check if this transaction actually needs to commit anything.
+    if (txContextFuture == null) {
```
thiagotnunes, line 682: Just double checking my understanding: `checkValidTransaction(COMMIT_STATEMENT)` will initialize the `txContextFuture` here, so it won't be `null` for blind writes, correct?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/AbstractMockServerTest.java
```diff
@@ -114,6 +114,8 @@ public abstract class AbstractMockServerTest {
       Statement.of("INSERT INTO TEST (ID, NAME) VALUES (1, 'test aborted')");
   public static final Statement INSERT_RETURNING_STATEMENT =
       Statement.of("INSERT INTO TEST (ID, NAME) VALUES (1, 'test aborted') THEN RETURN *");
+  public static final Statement PG_INSERT_RETURNING_STATEMENT =
```
thiagotnunes, line 117: Unrelated to delayed transactions?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/DelayTransactionStartUntilFirstWriteMockServerTest.java
```diff
@@ -0,0 +1,575 @@
...
+      ExecuteSqlRequest request = mockSpanner.getRequestsOfType(ExecuteSqlRequest.class).get(0);
+      assertTrue(request.hasTransaction());
+      assertTrue(request.getTransaction().hasBegin());
+      assertTrue(request.getTransaction().getBegin().hasReadWrite());
+      assertEquals(1, mockSpanner.countRequestsOfType(CommitRequest.class));
+    }
+  }
+
+  @Test
+  public void testSingleQuery() {
```
thiagotnunes, line 135: nit: rename to something a bit more clear. It seems we test that we never commit nor rollback, because there are no writes (thus we are using a singleUse read only transaction?)

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/DelayTransactionStartUntilFirstWriteMockServerTest.java
```diff
@@ -0,0 +1,575 @@
...
+        executeRandomQuery(connection);
+        if (commit) {
+          connection.commit();
+        } else {
+          connection.rollback();
+        }
+
+        assertEquals(2, mockSpanner.countRequestsOfType(ExecuteSqlRequest.class));
+        assertFalse(mockSpanner.getRequestsOfType(ExecuteSqlRequest.class).get(0).hasTransaction());
+        assertFalse(mockSpanner.getRequestsOfType(ExecuteSqlRequest.class).get(1).hasTransaction());
```
thiagotnunes, line 174: Why does the second query has a transaction? I would expect both of them to be using a singleUse (because there were no writes)?

thiagotnunes (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ReadWriteTransaction.java
olavloite: That's indeed a good point. Renamed to `checkOrCreateValidTransaction`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ReadWriteTransaction.java
olavloite: We need to set the `transactionStarted`, and we need to verify that the transaction is in a valid state, so we can't skip it altogether. Also; the method takes a `ParsedStatement` as an input argument so it can check the type of statement, and we don't have a SQL statement that is equivalent with writing/buffering mutations.

What we can however do is slightly rearrange the methods so we have a separate method only for verifying the valid state of the transaction and marking it as started, but skipping the creation of the underlying transaction. I've changed it to that.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ReadWriteTransaction.java
olavloite: Correct, although it will only be initialized for blind writes if the transaction actually has any mutations. So a transaction that does not do anything except calling `Commit` will be a no-op (it already was).

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/AbstractMockServerTest.java
olavloite: I added this so we could include tests that execute DML with a RETURNING clause in both dialects. The reason that it is relevant is that these statements are executed using the `executeQuery` method, and we should ensure that these statements trigger the start of the underlying transaction, even though they are executed through the `executeQuery` method.

So this statement is referenced in the `DelayTransactionStartUntilFirstWriteMockServerTest`

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/DelayTransactionStartUntilFirstWriteMockServerTest.java
olavloite: I renamed the tests in this class to more clearly reflect that they test transactions with a single query, ... etc.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/DelayTransactionStartUntilFirstWriteMockServerTest.java
olavloite: I think you misread the assertion... Or I'm missing something. The line says `assertFalse(... hasTransaction())`.

So your assumption is right. Both are single-use read-only.

arpan14 (approved):

PR #2471, opened 2023-05-30
perf: only capture the call stack if the call is actually async

olavloite:
Only capture the call stack of a caller of one of the Connection API methods if the top-level call was an async call. For synchronous calls, we do not need to additionally capture the call stack, as any exception will include the current call stack of the synchronous method call.

### DIFF
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ExceptionMockServerTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ExceptionMockServerTest.java
@@ -1,0 +1,163 @@
+/*
+ * Copyright 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.spanner.connection;
+
+import static com.google.cloud.spanner.SpannerApiFutures.get;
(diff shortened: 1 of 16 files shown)


google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ExceptionMockServerTest.java
```diff
@@ -0,0 +1,163 @@
...
+public class ExceptionMockServerTest extends AbstractMockServerTest {
+
+  @After
+  public void clearRequests() {
+    mockSpanner.clearRequests();
+  }
+
+  @Test
+  public void testUpdateAsyncException() {
+    Statement update = Statement.of("update foo set bar=1 where baz=1");
```
rajatbhatta, line 47: Can we add these (update, select, exception) as as private static final members within this class? Seems like these can be re-used across the test methods.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ExceptionMockServerTest.java
```diff
@@ -0,0 +1,163 @@
...
+      SpannerException exception =
+          assertThrows(SpannerException.class, () -> connection.executeUpdate(update));
+      assertNotNull(exception.getSuppressed());
+      assertEquals(0, exception.getSuppressed().length);
+    }
+  }
+
+  @Test
+  public void testQueryAsyncException() {
+    Statement update = Statement.of("select * from foo");
```
rajatbhatta, line 87: nit: update -> select. These can be modified in other parts within this file too, and can be private static final variables within this class.

rajatbhatta (approved):
LGTM, apart from a couple of nits. Thanks for working on this!

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ExceptionMockServerTest.java
olavloite: Done

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ExceptionMockServerTest.java
olavloite: Done

PR #2495, opened 2023-06-13
feat: add support for Proto Columns to Connection API

harshachinta:
Adds support for Proto Columns feature in the Connection API. This is needed to support Proto Columns in the JDBC driver [PR-1252](https://github.com/googleapis/java-spanner-jdbc/pull/1252).


google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
```diff
@@ -623,6 +633,55 @@ public void setStatementTag(String tag) {
     this.statementTag = tag;
   }
 
+  @Override
+  public byte[] getProtoDescriptors() {
+    ConnectionPreconditions.checkState(!isClosed(), CLOSED_ERROR_MSG);
+    if (this.protoDescriptorsFilePath != null) {
```
olavloite, line 639: I think it's better to do this slightly differently:
1. The check should be `this.protoDescriptors == null && this.protoDescriptorsFilePath != null`
2. At the end of this method, do not clear the field `this.protoDescriptorsFilePath`
3. In the setter methods for Proto Descriptors, we should make sure to clear both fields when one of them is set.

The advantage of the above is that you retain the fact that the current proto descriptors were set using a file name. That gives you more opportunities to include the file name in any error message that we might need to return in the future (for example if it turns out that the proto descriptor was invalid).

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
```diff
@@ -623,6 +633,55 @@ public void setStatementTag(String tag) {
...
+    this.protoDescriptors = protoDescriptors;
+    this.protoDescriptorsFilePath = null;
+  }
+
+  void setProtoDescriptorsFilePath(@Nonnull String protoDescriptorsFilePath) {
+    Preconditions.checkNotNull(protoDescriptorsFilePath);
+    ConnectionPreconditions.checkState(!isClosed(), CLOSED_ERROR_MSG);
+    ConnectionPreconditions.checkState(
+        !isBatchActive(), "Proto descriptors cannot be set when a batch is active");
+    this.protoDescriptorsFilePath = protoDescriptorsFilePath;
```
olavloite, line 682: See above; I think that we should set `protoDescriptors=null` here

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
```diff
@@ -1444,7 +1505,11 @@ private void popUnitOfWorkFromTransactionStack() {
   }
 
   private ApiFuture<Void> executeDdlAsync(CallType callType, ParsedStatement ddl) {
-    return getCurrentUnitOfWorkOrStartNewUnitOfWork().executeDdlAsync(callType, ddl);
+    ApiFuture<Void> result =
+        getCurrentUnitOfWorkOrStartNewUnitOfWork().executeDdlAsync(callType, ddl);
+    // reset proto descriptors after executing a DDL statement
+    this.protoDescriptors = null;
```
olavloite, line 1511: Also reset protoDescriptorsFilePath

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
```diff
@@ -1535,6 +1600,10 @@ public ApiFuture<long[]> runBatchAsync() {
       }
       return ApiFutures.immediateFuture(new long[0]);
     } finally {
+      if (isDdlBatchActive()) {
+        // reset proto descriptors after executing a DDL batch
+        this.protoDescriptors = null;
```
olavloite, line 1605: Also reset protoDescriptorsFilePath

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java
```diff
@@ -493,6 +496,28 @@ public StatementResult statementShowTransactionIsolationLevel() {
...
+
+  @Override
+  public StatementResult statementSetProtoDescriptorsFilePath(String filePath) {
+    Preconditions.checkNotNull(filePath);
+    getConnection().setProtoDescriptorsFilePath(filePath);
+    return noResult(SET_PROTO_DESCRIPTORS_FILE_PATH);
+  }
+
+  @Override
+  public StatementResult statementShowProtoDescriptors() {
```
olavloite, line 514: I think that it would be good to also have a `SHOW PROTO_DESCRIPTORS_FILE_PATH` command to be consistent.

google-cloud-spanner/src/main/resources/com/google/cloud/spanner/connection/ClientSideStatements.json
```diff
@@ -474,6 +474,45 @@
...
 		}
+	},
+	{
+		"name": "SET PROTO_DESCRIPTORS = '<proto_descriptor>'",
+		"executorName": "ClientSideStatementSetExecutor",
+		"resultType": "NO_RESULT",
+		"statementType": "SET_PROTO_DESCRIPTORS",
+		"regex": "(?is)\\A\\s*set\\s+proto_descriptors\\s*(?:=)\\s*(.*)\\z",
+		"method": "statementSetProtoDescriptors",
+		"exampleStatements": ["set proto_descriptors='protodescriptorsbase64'"],
```
olavloite, line 485: It surprises me that this works with the generated tests. Will the Base64 decoder not complain about this?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ClientSideStatementValueConverters.java
```diff
@@ -494,4 +495,46 @@ public String convert(String value) {
...
+      try {
+        return Base64.getDecoder().decode(value);
+      } catch (IllegalArgumentException e) {
+        return null;
+      }
+    }
+  }
+
+  /** Converter for converting String that take in file path as input to String */
+  static class ProtoDescriptorsFileConverter implements ClientSideStatementValueConverter<String> {
```
gauravpurohit06, line 523: We supporting reading from the file here? If Yes where should file be located just on resources or local path ? 
We had similar discussion on https://github.com/googleapis/java-spanner/pull/2277#discussion_r1111874247.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ClientSideStatementValueConverters.java
```diff
@@ -494,4 +495,46 @@ public String convert(String value) {
...
+    public Class<String> getParameterClass() {
+      return String.class;
+    }
+
+    @Override
+    public String convert(String filePath) {
+      if (filePath == null || filePath.length() == 0) {
+        return null;
+      }
+      return filePath;
```
gauravpurohit06, line 537: What this converter class is doing? It seems like we are returning the `filePath` only after checking if it's null or not.

gauravpurohit06 (changes requested):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ClientSideStatementValueConverters.java
```diff
@@ -494,4 +495,46 @@ public String convert(String value) {
...
+
+    public ProtoDescriptorsConverter(String allowedValues) {}
+
+    @Override
+    public Class<byte[]> getParameterClass() {
+      return byte[].class;
+    }
+
+    @Override
+    public byte[] convert(String value) {
```
rajatbhatta, line 510: Let's also add a @NonNull annotation to the argument here. Similarly in other places wherever a null value is not expected.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/Connection.java
```diff
@@ -382,6 +383,25 @@ default String getStatementTag() {
...
+   *
+   * @param protoDescriptors The proto descriptors to use with the next DDL statement (single or
+   *     batch) that will be executed on this connection.
+   */
+  default void setProtoDescriptors(@Nonnull byte[] protoDescriptors) {
+    throw new UnsupportedOperationException();
+  }
+
+  /**
+   * @return The proto descriptor that will be used with the next DDL statement (single or batch)
```
rajatbhatta, line 398:
```suggestion
   * @return The proto descriptor that will be used only with the next DDL statement (single or batch)
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
```diff
@@ -623,6 +633,55 @@ public void setStatementTag(String tag) {
...
+        !isBatchActive(), "Proto descriptors cannot be set when a batch is active");
+    this.protoDescriptors = protoDescriptors;
+    this.protoDescriptorsFilePath = null;
+  }
+
+  void setProtoDescriptorsFilePath(@Nonnull String protoDescriptorsFilePath) {
+    Preconditions.checkNotNull(protoDescriptorsFilePath);
+    ConnectionPreconditions.checkState(!isClosed(), CLOSED_ERROR_MSG);
+    ConnectionPreconditions.checkState(
+        !isBatchActive(), "Proto descriptors cannot be set when a batch is active");
```
rajatbhatta, line 681:
```suggestion
        !isBatchActive(), "Proto descriptors file path cannot be set when a batch is active");
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ClientSideStatementValueConverters.java
```diff
@@ -494,4 +495,46 @@ public String convert(String value) {
...
+    public ProtoDescriptorsFileConverter(String allowedValues) {}
+
+    @Override
+    public Class<String> getParameterClass() {
+      return String.class;
+    }
+
+    @Override
+    public String convert(String filePath) {
+      if (filePath == null || filePath.length() == 0) {
```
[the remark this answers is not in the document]
harshachinta: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
```diff
@@ -623,6 +633,55 @@ public void setStatementTag(String tag) {
...
+        String fileName = protoDescriptorsFile.getName();
+        int dotIndex = fileName.lastIndexOf('.');
+        // check if the file is a .pb extension file.
+        if (dotIndex < 0
+            || dotIndex > fileName.length() - 1
+            || !fileName.substring(dotIndex + 1).equals("pb")) {
+          throw SpannerExceptionFactory.newSpannerException(
+              ErrorCode.INVALID_ARGUMENT, "Proto descriptor file should be a .pb extension file");
+        }
+        if (!protoDescriptorsFile.isFile()) {
```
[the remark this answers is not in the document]
harshachinta: Agreed. Modified the exception to SpannerException

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
```diff
@@ -623,6 +633,55 @@ public void setStatementTag(String tag) {
...
+          throw new IOException("File does not exist.");
+        }
+
+        InputStream pdStream = new FileInputStream(protoDescriptorsFile);
+        this.protoDescriptors = ByteArray.copyFrom(pdStream).toByteArray();
+      } catch (Exception e) {
```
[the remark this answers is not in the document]
harshachinta: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
harshachinta: Thanks for the suggestion, Knut. This makes more sense. I have modified the logic as above.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
harshachinta: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
harshachinta: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java
harshachinta: Sure. I have added a new client side statement `SHOW VARIABLE PROTO_DESCRIPTORS_FILE_PATH`. This statement would return the proto descriptors file path as output.

google-cloud-spanner/src/main/resources/com/google/cloud/spanner/connection/ClientSideStatements.json
harshachinta: In this particular example, `protodescriptorsbase64` has characters that are valid base64 characters. However, if we add some space or any invalid character then the decoder throws an exception.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
```diff
@@ -623,6 +633,55 @@ public void setStatementTag(String tag) {
...
+      try {
+        File protoDescriptorsFile = new File(this.protoDescriptorsFilePath);
+        String fileName = protoDescriptorsFile.getName();
+        int dotIndex = fileName.lastIndexOf('.');
+        // check if the file is a .pb extension file.
+        if (dotIndex < 0
+            || dotIndex > fileName.length() - 1
+            || !fileName.substring(dotIndex + 1).equals("pb")) {
+          throw SpannerExceptionFactory.newSpannerException(
+              ErrorCode.INVALID_ARGUMENT, "Proto descriptor file should be a .pb extension file");
```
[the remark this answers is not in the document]
harshachinta: Thanks for bringing this up, Knut. I was under the assumption that proto descriptor file must be a `.pb` file. To clarify that hypothesis, I did a small experiment now and it turned out my assumption was wrong. It can be any file and it worked with `.txt` as well.
This validation is no longer valid and hence removing it.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ClientSideStatementValueConverters.java
```diff
@@ -494,4 +495,46 @@ public String convert(String value) {
...
+
+    @Override
+    public byte[] convert(String value) {
+      if (value.length() == 0 || value.equalsIgnoreCase("null")) {
+        return null;
+      }
+      try {
+        return Base64.getDecoder().decode(value);
+      } catch (IllegalArgumentException e) {
+        return null;
```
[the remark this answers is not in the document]
harshachinta: From value converters we don't throw any exceptions and instead return `null`. The [getParameterValue](https://github.com/googleapis/java-spanner/blob/397519e0a8707078cb57979c68d1ba8728400405/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ClientSideStatementSetExecutor.java#L80) code throws an [exception](https://github.com/googleapis/java-spanner/blob/397519e0a8707078cb57979c68d1ba8728400405/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ClientSideStatementSetExecutor.java#L88) saying invalid argument for the client side statement when it sees a `null` value.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ClientSideStatementValueConverters.java
harshachinta: Yes we are supporting reading from a file path given through a client side statement. This is needed to support ORM's and other frameworks. This discussion is available in the design doc.
Just a small brief, in [PR#2277](https://github.com/googleapis/java-spanner/pull/2277#discussion_r1111874247), we were treating file as a resource which brings in complications as they rely on class path. However, in this method we instead used File system to read from relative or absolute path.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ClientSideStatementValueConverters.java
```diff
@@ -494,4 +495,46 @@ public String convert(String value) {
...
+    public ProtoDescriptorsConverter(String allowedValues) {}
+
+    @Override
+    public Class<byte[]> getParameterClass() {
+      return byte[].class;
+    }
+
+    @Override
+    public byte[] convert(String value) {
+      if (value.length() == 0 || value.equalsIgnoreCase("null")) {
```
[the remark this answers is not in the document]
harshachinta: As we are not reusing this anywhere in the method I don't see a need to create a variable for it.
Also it was done the same way for rest of the converters [Sample](https://github.com/googleapis/java-spanner/blob/397519e0a8707078cb57979c68d1ba8728400405/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ClientSideStatementValueConverters.java#L158)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/DdlClient.java
```diff
@@ -51,6 +54,13 @@ Builder setDatabaseAdminClient(DatabaseAdminClient client) {
       return this;
     }
 
+    Builder setProjectId(String projectId) {
+      Preconditions.checkArgument(
+          !Strings.isNullOrEmpty(projectId), "Empty projectId is not allowed");
```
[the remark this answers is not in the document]
harshachinta: We have used the same error for other variables like `instanceId` etc ([Link](https://github.com/googleapis/java-spanner/blob/397519e0a8707078cb57979c68d1ba8728400405/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/DdlClient.java#L56)) 
Following the same to maintain consistency.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
harshachinta: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
```diff
@@ -623,6 +632,55 @@ public void setStatementTag(String tag) {
...
+              String.format(
+                  "File %s is not a valid proto descriptors file", this.protoDescriptorsFilePath));
+        }
+        InputStream pdStream = new FileInputStream(protoDescriptorsFile);
+        this.protoDescriptors = ByteArray.copyFrom(pdStream).toByteArray();
+      } catch (SpannerException exception) {
+        throw exception;
+      } catch (Exception exception) {
+        throw SpannerExceptionFactory.newSpannerException(
+            ErrorCode.INVALID_ARGUMENT, exception.getMessage());
```
olavloite, line 654: nit: add `exception` as a `cause` to the exception we are throwing.

google-cloud-spanner/src/main/resources/com/google/cloud/spanner/connection/ClientSideStatements.json
olavloite: Ah, makes sense. Thanks for clarifying.

olavloite (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
```diff
@@ -623,6 +632,55 @@ public void setStatementTag(String tag) {
...
+        if (!protoDescriptorsFile.isFile()) {
+          throw SpannerExceptionFactory.newSpannerException(
+              ErrorCode.INVALID_ARGUMENT,
+              String.format(
+                  "File %s is not a valid proto descriptors file", this.protoDescriptorsFilePath));
+        }
+        InputStream pdStream = new FileInputStream(protoDescriptorsFile);
+        this.protoDescriptors = ByteArray.copyFrom(pdStream).toByteArray();
+      } catch (SpannerException exception) {
+        throw exception;
```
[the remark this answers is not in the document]
harshachinta: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
harshachinta: Done

PR #2505, opened 2023-06-20
feat: Adding support for databoost

gauravpurohit06:
Enable databoost feature for all customers and removing the beta api annotation.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
@@ -16,7 +16,6 @@
 
 package com.google.cloud.spanner;
 
-import com.google.api.core.BetaApi;
 import com.google.common.base.Preconditions;
 import com.google.spanner.v1.RequestOptions.Priority;
 import java.io.Serializable;
@@ -157,10 +156,8 @@ public final class Options implements Serializable {
 
   /**
    * If this is for a partitioned read & query and this field is set to `true`, the request will be
-   * executed via Spanner independent compute resources. The method is available in Beta mode (and
-   * is not generally available now).
+   * executed via Spanner independent compute resources.
    */
-  @BetaApi
   public static DataBoostQueryOption dataBoostEnabled(Boolean dataBoostEnabled) {
     return new DataBoostQueryOption(dataBoostEnabled);
   }


rajatbhatta:
Can we re-enable integration tests through this PR and merge it once those pass (most likely on 26 Jun)?

gauravpurohit06:
> Can we re-enable integration tests through this PR and merge it once those pass (most likely on 26 Jun)?

Yes we can and lets wait till 26  as feature is already available in beta mode.

rajatbhatta (approved):
LGTM (once the Kokoro integration tests pass)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
```diff
@@ -156,11 +155,9 @@ public static ListOption pageSize(int pageSize) {
...
 
   /**
-   * If this is for a partitioned read and query and this field is set to `true`, the request will
-   * be executed via Spanner independent compute resources. The method is available in Beta mode
-   * (and is not generally available now).
+   * If this is for PartitionedRead or PartitionedQuery and this field is set to `true`, the request will be
+   * executed via Spanner independent compute resources.
    */
-  @BetaApi
   public static DataBoostQueryOption dataBoostEnabled(Boolean dataBoostEnabled) {
```
rajatbhatta, line 161:
```suggestion
  public static ReadAndQueryOption dataBoostEnabled(Boolean dataBoostEnabled) {
```
As we are removing the BetaApi annotation now, can we change DataBoostQueryOption to ReadAndQueryOption (similar to how it's for other options)? I understand it's technically a breaking change but I feel as it was a BetaApi, changes may be acceptable (without a major version bump). WDYT, @olavloite ?

olavloite: No, I would not be in favor of that. As you say; this is a breaking change, and it could break existing software, however new it might be. I especially worry about integrations with Beam and similar that often dynamically load other dependencies, such as Cloud Spanner, and that these might break in unexpected ways if we change this now.

rajatbhatta: Makes sense!

PR #2506, opened 2023-06-20
samples: add support for BatchWriteAtLeastOnce

rajatbhatta:
Note: Github checks are expected to fail until https://github.com/googleapis/java-spanner/pull/2520 is released, and a new `libraries-bom` version is out.


samples/snippets/src/main/java/com/example/spanner/BatchWriteAtLeastOnceSample.java
```diff
@@ -0,0 +1,133 @@
...
+    // TODO(developer): Replace these variables before running the sample.
+    final String projectId = "my-project";
+    final String instanceId = "my-instance";
+    final String databaseId = "my-database";
+    batchWriteAtLeastOnce(projectId, instanceId, databaseId);
+  }
+
+  static void batchWriteAtLeastOnce(String projectId, String instanceId, String databaseId) {
+    try (Spanner spanner =
+        SpannerOptions.newBuilder().setProjectId(projectId).setHost("https://staging-wrenchworks.sandbox.googleapis.com").build().getService()) {
```
arpan14, line 105: I think you haven't hard-coded the host intentionally. It should be removed.

rajatbhatta: Sorry, missed removing it. Removed it now. Thanks for pointing it out.

arpan14 (approved):

samples/snippets/src/main/java/com/example/spanner/BatchWriteAtLeastOnceSample.java
```diff
@@ -0,0 +1,133 @@
...
+import com.google.cloud.spanner.SpannerOptions;
+import com.google.common.collect.ImmutableList;
+import com.google.rpc.Code;
+import com.google.spanner.v1.BatchWriteResponse;
+
+public class BatchWriteAtLeastOnceSample {
+
+  /***
+   * Assume DDL for the underlying database:
+   * <p>
```
olavloite, line 37: Code samples in Javadoc should use a `<pre>{@code ... }` construct. See https://github.com/googleapis/java-spanner/blob/59cec9b9cdad169bd8de8ab7b264b04150dda7fb/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java#L61 for an example

samples/snippets/src/main/java/com/example/spanner/BatchWriteAtLeastOnceSample.java
```diff
@@ -0,0 +1,133 @@
...
+      // commit timestamp and status.
+      for (BatchWriteResponse response : responses) {
+        if (response.getStatus().getCode() == Code.OK_VALUE) {
+          System.out.printf(
+              "Mutation group indexes %s have been applied with commit timestamp %s",
+              response.getIndexesList(), response.getCommitTimestamp());
+        } else {
+          System.out.printf(
+              "Mutation group indexes %s could not be applied with error code %s",
+              response.getIndexesList(), Code.forNumber(response.getStatus().getCode()));
```
olavloite, line 126: nit: Is it possible to also get the error message and print that? Getting for example an `INVALID_ARGUMENT` error code without any further details will make it hard to debug the error.

samples/snippets/src/main/java/com/example/spanner/BatchWriteAtLeastOnceSample.java
```diff
@@ -0,0 +1,135 @@
...
+import com.google.rpc.Code;
+import com.google.spanner.v1.BatchWriteResponse;
+
+public class BatchWriteAtLeastOnceSample {
+
+  /***
+   * Assume DDL for the underlying database:
+   * <pre>{@code
+   * CREATE TABLE Singers (
+   * SingerId   INT64 NOT NULL,
```
olavloite, line 39: nit: add a couple of spaces for indentation

samples/snippets/src/test/java/com/example/spanner/BatchWriteAtLeastOnceSampleIT.java
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  @Test
+  public void testBatchWriteAtLeastOnce() throws Exception {
+    final String out =
+        SampleRunner.runSample(() -> BatchWriteAtLeastOnceSample.batchWriteAtLeastOnce(
+            projectId, instanceId, databaseId));
+    assertTrue(
+        out.contains("have been applied with commit timestamp")
+            || out.contains("could not be applied with error code"));
```
olavloite, line 60: We don't expect an error here, do we? So in that case I think we should let the test case fail with the error instead of letting the test pass here.

If it is possible that the call fails with a transient error (e.g. UNAVAILABLE or ABORTED), then I think that the sample should show how customers should handle that.

arpan14: Yeah I agree we should exclude `out.contains("could not be applied with error code")` and let the test case fail. Modified this.

olavloite (approved):

PR #2512, opened 2023-06-26
deps: update dependency com.google.cloud:google-cloud-shared-dependencies to v3.12.0.with temp exclusions.

zhumin8:
Temp exclusions are needed to avoid dependency version discrepancy due to 3.23.2. 
For more context: https://github.com/googleapis/sdk-platform-java/pull/1791#issuecomment-1601587713
@suztomo  Can you take a look if this temp fix looks okay?

Closes #2509

### DIFF
--- a/google-cloud-spanner-executor/pom.xml
+++ b/google-cloud-spanner-executor/pom.xml
@@ -49,6 +49,14 @@
     <dependency>
       <groupId>com.google.protobuf</groupId>
       <artifactId>protobuf-java-util</artifactId>
+      <!--   Temp exclusion. Remove next protobuf upgrade (to 3.23.3 or higher)
+         context: https://github.com/googleapis/sdk-platform-java/pull/1791#issuecomment-1601587713-->
+      <exclusions>
+        <exclusion>
+          <groupId>com.google.j2objc</groupId>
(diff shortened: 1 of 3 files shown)


google-cloud-spanner-executor/pom.xml
```diff
@@ -82,6 +90,14 @@
       <groupId>com.google.api.grpc</groupId>
       <artifactId>proto-google-cloud-spanner-executor-v1</artifactId>
       <version>1.4.0</version>
+      <!--   Temp exclusion. Remove next protobuf upgrade (to 3.23.3 or higher)
```
arpan14, line 93: 1. How would we ensure that this temp exclusion gets removed? Do we have a good internal tracking mechanism?
2. How do we know that this may not introduce any other side effect? Any side effect that can be caused by excluding `j2objc-annotations` dependency? Or is this transitively getting included?

arpan14: * Sure thanks for the clarification and having this transitively included ensures there are no others side effects. 
* Yes, internal ticket works to track removing this workaround.

arpan14 (approved):

PR #2520, opened 2023-07-06
feat: add support for BatchWriteAtLeastOnce

rajatbhatta:


google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java
```diff
@@ -191,6 +193,77 @@ CommitResponse writeWithOptions(Iterable<Mutation> mutations, TransactionOption.
   CommitResponse writeAtLeastOnceWithOptions(
       Iterable<Mutation> mutations, TransactionOption... options) throws SpannerException;
 
+  /**
+   * Applies batch of mutation groups in a collection of efficient transactions. The mutation groups
+   * are applied non-atomically in an unspecified order and thus, they must be independent of each
+   * other. Partial failure is possible, i.e., some mutation groups may have been applied
+   * successfully, while some may have failed. The results of individual batches are streamed into
```
olavloite, line 200: Will there be exactly one BatchWriteResponse for each MutationGroup?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java
```diff
@@ -191,6 +193,77 @@ CommitResponse writeWithOptions(Iterable<Mutation> mutations, TransactionOption.
...
+   * the response as and when the batches are applied.
+   *
+   * <p>Since this method does not feature replay protection, it may attempt to apply {@code
+   * mutation groups} more than once; if the mutation groups are not idempotent, this may lead to a
+   * failure being reported when the mutation group was applied once. For example, an insert may
```
olavloite, line 205: The current text seems a bit strange. Should it have been this?

```suggestion
   * failure being reported when the mutation group was applied more than once. For example, an insert may
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java
```diff
@@ -191,6 +193,77 @@ CommitResponse writeWithOptions(Iterable<Mutation> mutations, TransactionOption.
...
+   * was called. For this reason, most users of the library will prefer to use {@link
+   * #write(Iterable)} instead. However, {@code batchWriteAtLeastOnce()} method may be appropriate
+   * for non-atomically committing multiple mutation groups in a single RPC with low latency.
+   *
+   * <p>Example of BatchWriteAtLeastOnce
+   *
+   * <pre>{@code
+   * ServerStream<BatchWriteResponse> responses =
+   *     dbClient.batchWriteAtLeastOnceWithOptions(
+   *         ImmutableList.of(MUTATION_GROUP1, MUTATION_GROUP2));
```
olavloite, line 216: Could we add the definition of `MUTATION_GROUP1` as well here? I don't know how/what a mutation group would look like, and how I would define one.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java
```diff
@@ -191,6 +193,77 @@ CommitResponse writeWithOptions(Iterable<Mutation> mutations, TransactionOption.
...
+   * for (BatchWriteResponse response : responses) {
+   *   // Do something when a response is received.
+   * }
+   * }</pre>
+   *
+   * @return ServerStream\<BatchWriteResponse>
```
olavloite, line 222: nit: remove entirely, it does not seem like it adds any information? Or otherwise at least remove the `\`

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java
```diff
@@ -191,6 +193,77 @@ CommitResponse writeWithOptions(Iterable<Mutation> mutations, TransactionOption.
...
+   */
+  ServerStream<BatchWriteResponse> batchWriteAtLeastOnce(Iterable<MutationGroup> mutationGroups)
```
olavloite, line 224: When should I create multiple MutationGroups instead of just one large mutation group? What's the tradeoff between a large number of small MutationGroups vs a few large MutationGroups? Is there any limits on the number of mutations that I may have in a MutationGroup?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java
```diff
@@ -191,6 +193,77 @@ CommitResponse writeWithOptions(Iterable<Mutation> mutations, TransactionOption.
...
+   *
+   * <ul>
+   *   <li>{@link Options#priority(com.google.cloud.spanner.Options.RpcPriority)}: The {@link
+   *       RpcPriority} to use for the batch write request.
+   *   <li>{@link Options#tag(String)}: The transaction tag to use for the batch write request.
+   * </ul>
+   *
+   * @return ServerStream\<BatchWriteResponse>
+   */
+  ServerStream<BatchWriteResponse> batchWriteAtLeastOnceWithOptions(
```
olavloite, line 264: I think it would be better to just have one `batchWriteAtLeastOnce` method. That method can take options. The reason that we have two different versions for some other methods was that some were added before we had options, and adding an options argument to an existing method would be a breaking change.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClientImpl.java
```diff
@@ -106,6 +108,28 @@ public CommitResponse writeAtLeastOnceWithOptions(
...
+
+  @Override
+  public ServerStream<BatchWriteResponse> batchWriteAtLeastOnceWithOptions(
+      final Iterable<MutationGroup> mutationGroups, final TransactionOption... options)
+      throws SpannerException {
+    Span span = tracer.spanBuilder(READ_WRITE_TRANSACTION).startSpan();
+    try (Scope s = tracer.withSpan(span)) {
+      return runWithSessionRetry(
+          session -> session.batchWriteAtLeastOnceWithOptions(mutationGroups, options));
+    } catch (RuntimeException e) {
```
olavloite, line 125: Why do we only catch `RuntimeException` here, instead of `Exception`?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/MutationGroup.java
```diff
@@ -0,0 +1,58 @@
...
+ * Represents a group of Cloud Spanner mutations to be committed together.
+ */
+public class MutationGroup {
+  private final List<Mutation> mutations;
+
+  private MutationGroup(List<Mutation> mutations) {
+    this.mutations = mutations;
+  }
+
+  public static MutationGroup of(Mutation... mutations) {
```
olavloite, line 35: nit: add Javadoc

olavloite, line 35: Maybe also add a static creator method that accepts an `Iterable<Mutation>`?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/MutationGroup.java
```diff
@@ -0,0 +1,58 @@
...
+    Preconditions.checkArgument(mutations.length > 0, "Should pass in at least one mutation.");
+    return new MutationGroup(Arrays.asList(mutations));
```
olavloite, line 37: Maybe `ImmutableList.of(mutations)`?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/MutationGroup.java
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  public List<Mutation> getMutations() {
```
olavloite, line 40: Maybe make this package-private? Or otherwise at least use an `ImmutableList` to ensure that `MutationGroup` is immutable?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/MutationGroup.java
```diff
@@ -0,0 +1,58 @@
...
+    return mutations;
+  }
+
+  static BatchWriteRequest.MutationGroup toProto(final MutationGroup mutationGroup) {
+    List<com.google.spanner.v1.Mutation> mutationsProto = new ArrayList<>();
+    Mutation.toProto(mutationGroup.getMutations(), mutationsProto);
+    return BatchWriteRequest.MutationGroup.newBuilder().addAllMutations(mutationsProto).build();
```
olavloite, line 47: Maybe it's a bit too much of pre-optimization, but this seems to make an unnecessary copy of the list of mutations:
1. First it creates an ArrayList and then `Mutation.toProto(...)` adds each mutation to that list.
2. Then after that we copy that ArrayList to the `BatchWriteRequest`.
 
I think it would be more efficient (and maybe also simpler) to just iterate through the list of mutations, convert one and then directly add them to the BatchWriteRequest.Builder.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionImpl.java
```diff
@@ -195,6 +198,43 @@ public CommitResponse writeAtLeastOnceWithOptions(
...
+    try (Scope s = tracer.withSpan(span)) {
+      return spanner.getRpc().batchWriteAtLeastOnce(requestBuilder.build(), this.options);
+    } catch (Throwable e) {
+      TraceUtil.setWithFailure(span, e);
+      throw SpannerExceptionFactory.newSpannerException(e);
+    } finally {
+      span.end(TraceUtil.END_SPAN_OPTIONS);
+    }
+  }
+
```
olavloite, line 237: This method seems to be very similar to `writeAtLeastOnceWithOptions`. Could we combine the two so they share more of the code?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITWriteTest.java
```diff
@@ -161,7 +183,7 @@ public static void setUpDatabase()
     if (!EmulatorSpannerHelper.isUsingEmulator()) {
       Database postgreSQLDatabase =
           env.getTestHelper()
-              .createTestDatabase(Dialect.POSTGRESQL, Collections.singletonList(POSTGRESQL_SCHEMA));
+              .createTestDatabase(Dialect.POSTGRESQL, Arrays.asList(POSTGRESQL_SCHEMA));
```
olavloite, line 186: Any specific reason for this?

rajatbhatta: The schema was earlier a String. It's now a String[], after introduction of DDL for new table T1.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/MutationGroup.java
rajatbhatta: Done.

rajatbhatta: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java
rajatbhatta: I've now added a detailed sample of how to create the mutation groups, and then pass it to the batchWriteAtLeastOnce method.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java
rajatbhatta: Removed entirely.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/MutationGroup.java
rajatbhatta: I now have `mutations` variable as an `ImmutableList<Mutation>` in the `MutationGroup` class, and return it directly inside `getMutations()`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionImpl.java
rajatbhatta: Moved the `RequestOptions` building logic to a separate private method.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/MutationGroup.java
rajatbhatta: Changed it to `ImmutableList.copyOf(mutations)`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClientImpl.java
rajatbhatta: I'm not 100% sure of the rationale behind it. It's something that is used uniformly across other methods as well (`writeWithOptions`, `writeAtLeastOnceWithOptions`, `runAsync`, etc).

rajatbhatta: Are you aware of the reason?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/MutationGroup.java
rajatbhatta: `Mutation.toProto()` goes through the list of mutations and has some additional logic to convert it into a list of mutation protos (which may reduce the size of the list also). That logic is not trivial. I think, simply iterating through the list of mutations, converting into mutation protos one by one, and adding to the BatchWriteRequest builder is not gonna directly work.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java
olavloite: Can we change this to something like: 'The mutation groups can be applied more than once. This can lead to failures if the mutation groups are non-idempotent. For example, an insert that is replayed can return an `AlreadyExists` error.'

I assume that the intended meaning of the word 'once' in the current text '...failure being reported when the mutation group was applied once.' is something like 'it had already been applied once by an earlier attempt'. But that is not how I read it as a new reader. I read it as 'It will return `AlreadyExists` if it is applied excatly one time', which did not make sense to me.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClientImpl.java
olavloite: No. So for consistency I agree that we should keep it like this for now.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/MutationGroup.java
olavloite: Sounds reasonable.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITWriteTest.java
olavloite: Ah, of course. Thanks!

olavloite (approved):
LGTM, with a couple of nits on the documentation. This is a non-trivial feature, so making sure that we make the documentation as clear as possible will help our users make the most out of this.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java
rajatbhatta: Yes, done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java
rajatbhatta: Makes sense, made changes to doc.

PR #2540, opened 2023-07-24
feat: support partitioned queries + data boost in Connection API

olavloite:
Adds support for Partitioned Queries and Data Boost in the Connection API. This enables the use of these features in the JDBC driver and PGAdapter.

This PR builds on #2556 which is a small refactoring that allows client-side statements to use statement parameters.


google-cloud-spanner/src/main/java/com/google/cloud/spanner/Statement.java
```diff
@@ -86,6 +86,12 @@ private Builder(Statement statement) {
           statement.queryOptions == null ? null : statement.queryOptions.toBuilder().build();
     }
 
+    /** Replaces the current SQL of this builder with the given string. */
+    public Builder withSql(String sql) {
```
arpan14, line 90: Nit - Should we name this `replace` or `replaceWithSql` ?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
```diff
@@ -1234,6 +1372,19 @@ public ApiFuture<long[]> executeBatchUpdateAsync(Iterable<Statement> updates) {
     return internalExecuteBatchUpdateAsync(CallType.ASYNC, parsedStatements);
   }
 
+  private QueryOption[] mergeDataBoost(QueryOption... options) {
```
arpan14, line 1375: * Nit: Should we break databoost as a separate PR? So that its easy to use as a reference later? This PR can probably be about just adding support for partitioned reads?
* I am ok to review this together, but breaking it just adds a future reference PR on how databoost support was added. This suggestion is only if it does not result in a lot of re-work for you.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
```diff
@@ -1234,6 +1372,19 @@ public ApiFuture<long[]> executeBatchUpdateAsync(Iterable<Statement> updates) {
...
+    if (this.dataBoostEnabled) {
```
arpan14, line 1376: Query - Just checking if its safe to not generate a query option when `dataBoostEnabled` is false? Wouldn't we want to differentiate the case when customer explicitly marked the option as `false` vs case where customer did not pass the property in connection string?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Statement.java
olavloite: I chose the `withSql` name to be consistent with the existing `withQueryOptions` method on this class, which also replaces any existing value. I'm happy to change it to something else if you feel strongly about this.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/PartitionId.java
```diff
@@ -0,0 +1,105 @@
...
+    PartitionId id = new PartitionId(transactionId, partition);
+    ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
+    try (ObjectOutputStream objectOutputStream =
+        new ObjectOutputStream(new GZIPOutputStream(byteArrayOutputStream))) {
+      objectOutputStream.writeObject(id);
+    } catch (Exception exception) {
+      throw SpannerExceptionFactory.newSpannerException(exception);
+    }
+    return Base64.getUrlEncoder().encodeToString(byteArrayOutputStream.toByteArray());
+  }
```
arpan14, line 73: Can we have a general utility to decode and encode (using Gzip)? That could be re-used at more than one place? We can use some generics to model more than one Input/Output?

google-cloud-spanner/src/main/resources/com/google/cloud/spanner/connection/ClientSideStatements.json
```diff
@@ -158,6 +158,33 @@
...
 		"exampleStatements": ["show variable delay_transaction_start_until_first_write"]
 	},
+	{
+		"name": "PARTITION <sql>",
+		"executorName": "ClientSideStatementPartitionExecutor",
+		"resultType": "RESULT_SET",
+		"statementType": "PARTITION",
+		"regex": "(?is)\\A\\s*partition(\\s+|\\()(.*)\\z",
+		"method": "statementPartition",
+		"exampleStatements": []
```
arpan14, line 168: Referring back to a previous comment of mine where I asked for a few examples, can we add a few examples here which can be used to understand the `ClientSideStatementPartitionExecutor` code better?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/MergedResultSet.java
```diff
@@ -0,0 +1,399 @@
...
+    }
+  }
+
+  static class PartitionExecutorResult {
+    private final Struct data;
+    private final Throwable exception;
+    private final Type type;
+    private final ResultSetMetadata metadata;
+
+    static PartitionExecutorResult data(Struct data) {
```
arpan14, line 111: Wouldn't a builder pattern be better suited for such object construction? Otherwise for every member that we introduce it will be difficult to multiplex such method to partially build the object.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/MergedResultSet.java
```diff
@@ -0,0 +1,399 @@
...
+import java.util.concurrent.LinkedBlockingDeque;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * {@link MergedResultSet} is a {@link ResultSet} implementation that combines the results from
+ * multiple queries. Each query uses its own {@link RowProducer} that feeds rows into the {@link
+ * MergedResultSet}. The order of the records in the {@link MergedResultSet} is not guaranteed.
+ */
+class MergedResultSet extends ForwardingStructReader implements PartitionedQueryResultSet {
```
arpan14, line 44: * Have we explored re-using some elements present in `AsyncResultSetImpl` ? Or in another way are there ways in which we can make `MergedResultSet` and `AsyncResultSetImpl` have some common code? 
* Also, there is not much difference in the way we would like to implement batch read in client library and connection API, apart from the fact that connection API supports a bunch of other configurations in connection string that client library does not support. 

Am I missing the other major differences between these two implementations of `ForwardingStructReader` ?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Statement.java
arpan14: In this case the naming convention is more like existing `append()` method. So going by that we can name this `replace()`.

arpan14 (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/PartitionId.java
arpan14: Not pre-creating a library. But basically breaking down this class into two classes - `Partition` and say `GzipEncodingUtility`. `GzipEncodingUtility` will use generics to take in dynamic input/output types. `Partition` class could internally use `GzipEncodingUtility` to do what it's doing currently.

PR #2544, opened 2023-07-25
fix: apply stream wait timeout

### LINKED ISSUE #2494: Hanging indefinitely
Hi, this issue is related to the support case 45040288.

A short while ago we had multiple of our services that use Spanner, both with the latest Spanner version at the time (6.42.1) but some using slightly older ones as well, hang indefinitely on [this particular row](https://github.com/googleapis/java-spanner/blob/977b5851b9cc1a538dedf57d006483dc58ec90b9/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java#L990). The only solution when that happened was to restart the service, which would fix the issue - even though it often happened again after a while.

This happened quite often during a two week period towards at least one particular Spanner instance, and we haven't seen it since a couple of weeks now. We have also not been able to determine the exact conditions under which this issue presents itself - We do not believe it to be a recent code change, as we saw it on some services that haven't been updated for months. The code that froze was just a regular `resultSet.next()` call, preceeded by a `singleUse().read(...)`, with appropriate closing etc.

Since we don't have a reliable way to reproduce this issue, I believe the most straightforward way to solve it would be to implement a timeout mechanic like mentioned in the TODO-comment in the code already.

olavloite:
Use the streamWaitTimeout that has been set on the call context when polling from the gRPC stream. This prevents the stream from blocking forever if for some reason the stream is no longer delivering data, and also no error is propagated to the client.

The default stream wait timeout that is set for all call contexts is 30 mins. This value can be overridden by configuring a custom call context for a specific query.

Fixes #2494

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
@@ -982,6 +982,8 @@ abstract class AbstractResultSet<R> extends AbstractStructReader implements Resu
           streamWaitTimeoutValue = streamWaitTimeout.getNano();
           streamWaitTimeoutUnit = TimeUnit.NANOSECONDS;
         }
+        // Note that if the stream-wait-timeout is zero, we won't set a timeout at all.
+        // That is consistent with ApiCallContext#withStreamWaitTimeout(Duration.ZERO).
       }
     }
 
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
@@ -25,6 +25,7 @@ import static com.google.common.base.Preconditions.checkState;
 import com.google.api.client.util.BackOff;
 import com.google.api.client.util.ExponentialBackOff;
 import com.google.api.gax.retrying.RetrySettings;
+import com.google.api.gax.rpc.ApiCallContext;
 import com.google.cloud.ByteArray;
 import com.google.cloud.Date;
 import com.google.cloud.Timestamp;
@@ -74,6 +75,7 @@ import java.util.logging.Logger;
 import java.util.stream.Collectors;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
+import org.threeten.bp.Duration;
 
 /** Implementation of {@link ResultSet}. */
 abstract class AbstractResultSet<R> extends AbstractStructReader implements ResultSet {
@@ -944,6 +946,8 @@ abstract class AbstractResultSet<R> extends AbstractStructReader implements Resu
 
     private SpannerRpc.StreamingCall call;
     private volatile boolean withBeginTransaction;
+    private TimeUnit streamWaitTimeoutUnit;
+    private long streamWaitTimeoutValue;
     private SpannerException error;
 
     @VisibleForTesting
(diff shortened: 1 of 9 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
```diff
@@ -965,6 +969,20 @@ protected final SpannerRpc.ResultStreamConsumer consumer() {
     public void setCall(SpannerRpc.StreamingCall call, boolean withBeginTransaction) {
       this.call = call;
       this.withBeginTransaction = withBeginTransaction;
+      ApiCallContext callContext = call.getCallContext();
+      Duration streamWaitTimeout = callContext == null ? null : callContext.getStreamWaitTimeout();
+      if (streamWaitTimeout != null) {
+        // Determine the timeout unit to use. This reduces the precision to seconds if the timeout
+        // value is more than 1 second, which is lower than the precision that would normally be
+        // used by the stream watchdog (which uses a precision of 10 seconds by default).
```
rajatbhatta, line 977: > This reduces the precision to seconds if the timeout value is more than 1 second.

Not sure if I understand the above correctly.

IIUC, if the timeout value is 2 seconds 1000 nanoseconds (which is greater than 1 second), it'll set `streamWaitTimeoutValue` to `2` and `streamWaitTimeoutUnit` to `TimeUnit.SECONDS`. It'll just trim the nanoseconds, if there's any.

Can we update this comment if the above understanding is correct?

Let me know if I misunderstood something.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
```diff
@@ -983,11 +1001,15 @@ public boolean isWithBeginTransaction() {
...
-        next = stream.take();
+        if (streamWaitTimeoutUnit != null) {
+          next = stream.poll(streamWaitTimeoutValue, streamWaitTimeoutUnit);
+          if (next == null) {
+            throw SpannerExceptionFactory.newSpannerException(
+                ErrorCode.DEADLINE_EXCEEDED, "stream wait timeout");
+          }
+        } else {
+          next = stream.take();
+        }
```
rajatbhatta, line 1012: The default stream wait timeout that is set for all call contexts is 30 mins. This value can be overridden by configuring a custom call context for a specific query. So, just wanted to understand when will we encounter this else condition (`streamWaitTimeoutUnit == null`) better.

I can think of a case where user explicitly sets a `0` stream wait timeout value in which case the `streamWaitTimeoutUnit` will be `null` here, and IMO, we should directly throw a `DEADLINE_EXCEEDED` error without making any `stream.take()` call.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
rajatbhatta: Okay I think I get it now. The comment makes sense now. Resolving.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
olavloite: A user could do one of the following:
1. Configure a custom call context with a null stream timeout by calling `context.withStreamWaitTimeout(null)`. That would however in our specific setup not cause a null timeout to be used, because we merge the default context with the call-specific context. Any null values on the call-specific context will not overwrite any of the defaults.
2. As you mentioned: Setting a zero timeout will cause the `streamWaitTimeoutUnit` to be `null`. That is intentional, as the `streamWaitTimeout` should be considered disabled if set to zero. See https://github.com/googleapis/sdk-platform-java/blob/f9d920d076db36dc095f7ace9c571301122ce959/gax-java/gax/src/main/java/com/google/api/gax/rpc/ApiCallContext.java#L95

olavloite: I've added a test to verify this behavior, and some comments to clarify.

rajatbhatta: Thanks for clarifying!

rajatbhatta (approved):

PR #2559, opened 2023-07-31
samples: Bit Reversed Sequence support

rajatbhatta:


samples/snippets/src/test/java/com/example/spanner/SequenceSampleIT.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.Collections;
+import java.util.concurrent.TimeUnit;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/** Integration tests for DML Returning samples for GoogleStandardSql dialect. */
+@RunWith(JUnit4.class)
+public class SequenceSampleIT extends SampleTestBase {
```
arpan14, line 33: If it's not too much effort, can we have a single integration test and parameterise it for PG/Google SQL dialects. If no, can you add what are the differences in the way both dialects are tested?

rajatbhatta: Thanks for the suggestion. I've now parameterized the tests. Please take a look.

arpan14 (approved):

PR #2564, opened 2023-08-02
test: add benchmark for long-running sessions clean up task.

arpan14:
Adding benchmarks for long running sessions cleanup feature introduced as part of - https://github.com/googleapis/java-spanner/pull/2419 .

### DIFF
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/LongRunningSessionsBenchmark.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/LongRunningSessionsBenchmark.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 Google LLC
+ * Copyright 2023 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/LongRunningSessionsBenchmark.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/LongRunningSessionsBenchmark.java
@@ -33,6 +33,7 @@ import java.util.List;
 import java.util.Random;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 import org.openjdk.jmh.annotations.AuxCounters;
 import org.openjdk.jmh.annotations.Benchmark;
 import org.openjdk.jmh.annotations.BenchmarkMode;
@@ -77,6 +78,7 @@ public class LongRunningSessionsBenchmark {
     private StandardBenchmarkMockServer mockServer;
     private Spanner spanner;
     private DatabaseClientImpl client;
+    private AtomicInteger longRunningSessions;
 
     @Param({"100"})
     int minSessions;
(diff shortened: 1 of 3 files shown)


google-cloud-spanner/src/test/java/com/google/cloud/spanner/LongRunningSessionsBenchmark.java
```diff
@@ -0,0 +1,317 @@
+/*
+ * Copyright 2020 Google LLC
```
olavloite, line 2: nit:
```suggestion
 * Copyright 2023 Google LLC
```

google-cloud-spanner/src/test/java/com/google/cloud/spanner/LongRunningSessionsBenchmark.java
```diff
@@ -0,0 +1,317 @@
...
+                      .setInactiveTransactionRemovalOptions(inactiveTransactionRemovalOptions)
+                      .build())
+              .build();
+
+      spanner = options.getService();
+      client =
+          (DatabaseClientImpl)
+              spanner.getDatabaseClient(DatabaseId.of(TEST_PROJECT, TEST_INSTANCE, TEST_DATABASE));
+      // Wait until the session pool has initialized.
+      while (client.pool.getNumberOfSessionsInPool()
```
olavloite, line 135: nit: you can use `SessionPoolOptions#setWaitForMinSessions(Duration waitForMinSessions)` for this.

(I guess this is a copy-paste from one of the other benchmarks. The above-mentioned method was added later.)

google-cloud-spanner/src/test/java/com/google/cloud/spanner/LongRunningSessionsBenchmark.java
```diff
@@ -0,0 +1,317 @@
...
+    for (int i = 0; i < totalReads; i++) {
+      futures.add(
+          service.submit(
+              () -> {
+                Thread.sleep(RND.nextInt(RND_WAIT_TIME_BETWEEN_REQUESTS));
+                try (ResultSet rs =
+                    client.singleUse().executeQuery(StandardBenchmarkMockServer.SELECT1)) {
+                  while (rs.next()) {
+                    // introduce random sleep times to have long-running sessions
+                    randomWait();
```
olavloite, line 272: Pure theoretically this random function could end up never adding a long wait time if `random.nextBool` returns false every time. Would it maybe make sense to let this method update an atomic integer every time it sets  a long wait time, and then at the end of the test verify that the number of sessions that was removed is equal to that atomic integer (instead of just checking that it's greater than zero, which could in theory fail)?

olavloite (approved):

google-cloud-spanner/src/test/java/com/google/cloud/spanner/LongRunningSessionsBenchmark.java
arpan14: That's a very good idea. Let me do this. It will make the benchmark even more precise.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/LongRunningSessionsBenchmark.java
arpan14: Yes, I have done this.

Issue #2205, opened 2022-12-22
Java 8 Windows CI status failing


rajatbhatta:
Java 8 Windows CI Status is "Failing".

Error:

```
[ERROR] /T:/src/github/java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/nativeimage/SpannerFeature.java:[20,38] cannot access org.graalvm.nativeimage.hosted.Feature
  bad class file: C:\Users\kbuilder\.m2\repository\org\graalvm\sdk\graal-sdk\22.3.0\graal-sdk-22.3.0.jar(org/graalvm/nativeimage/hosted/Feature.class)
    class file has wrong version 55.0, should be 52.0
    Please remove or make sure it appears in the correct subdirectory of the classpath.
```

Seems to be related to native images.

mpeddada1:
@rajatbhatta Hm this looks familiar. It was previously addressed through https://github.com/googleapis/java-spanner/pull/1870/files so I'm curious as to why we're seeing this. Do you happen to have the PR that is running into this?

rajatbhatta:
@mpeddada1: Thanks for looking into this. We've this latest nightly build that failed: https://source.cloud.google.com/results/invocations/47498ffc-9e1e-4c2f-916e-3453ff20b7ae/targets/cloud-devrel%2Fclient-libraries%2Fjava%2Fjava-spanner%2Fnightly%2Fjava8-win.

mpeddada1:
Thank you! It looks like the `JAVA11_HOME` and `JAVA8_HOME` environment variables are not being set for this job. Troubleshooting continued at #2207.

mpeddada1:
Closed via #2207 and #2210

PR #1870, opened 2022-05-05
ci: using JDK 11 to compile and JDK 8 to run junit (8)

suztomo:
Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [ ] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/java-spanner/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [ ] Ensure the tests and linter pass
- [ ] Code coverage does not decrease (if any source code was changed)
- [ ] Appropriate docs were updated (if necessary)

Fixes #<issue_number_goes_here> ☕️

If you write sample code, please follow the [samples format](
https://github.com/GoogleCloudPlatform/java-docs-samples/blob/main/SAMPLE_FORMAT.md).

PR #2573, opened 2023-08-07
fix: fix kokoro windows java8 ci

harshachinta:
Windows build fails due to incorrect JAVA11_HOME env. This PR corrects it to latest version to fix the failing window builds.
image

image

### DIFF
--- a/.kokoro/build.bat
+++ b/.kokoro/build.bat
@@ -17,6 +17,6 @@
 
 set JAVA8_HOME=%JAVA_HOME:"=%
 choco install -y openjdk11
-set JAVA11_HOME=C:\Program Files\Eclipse Adoptium\jdk-11.0.18.10-hotspot\
+set JAVA11_HOME=C:\Program Files\Eclipse Adoptium\jdk-11.0.19.7-hotspot\
 
 "C:\Program Files\Git\bin\bash.exe" %~dp0build.sh
--- a/.kokoro/build.bat
+++ b/.kokoro/build.bat
@@ -15,8 +15,4 @@
 :: downstream client libraries before they are released.
 :: See documentation in type-shell-output.bat
 
-set JAVA8_HOME=%JAVA_HOME:"=%
-choco install -y openjdk11
-set JAVA11_HOME=C:\Program Files\Eclipse Adoptium\jdk-11.0.20.8-hotspot\
-
 "C:\Program Files\Git\bin\bash.exe" %~dp0build.sh
--- a/.kokoro/build.bat
+++ b/.kokoro/build.bat
@@ -17,6 +17,6 @@
 
 set JAVA8_HOME=%JAVA_HOME:"=%
 choco install -y openjdk11
-set JAVA11_HOME=C:\Program Files\Eclipse Adoptium\jdk-11.0.19.7-hotspot\
+set JAVA11_HOME=C:\Program Files\Eclipse Adoptium\jdk-11.0.20.8-hotspot\
 
 "C:\Program Files\Git\bin\bash.exe" %~dp0build.sh
--- a/README.md
+++ b/README.md
@@ -50,20 +50,20 @@ If you are using Maven without the BOM, add this to your dependencies:
 If you are using Gradle 5.x or later, add this to your dependencies:
 
 ```Groovy
-implementation platform('com.google.cloud:libraries-bom:26.21.0')
+implementation platform('com.google.cloud:libraries-bom:26.22.0')
 
 implementation 'com.google.cloud:google-cloud-spanner'
 ```
 If you are using Gradle without BOM, add this to your dependencies:
 
 ```Groovy
-implementation 'com.google.cloud:google-cloud-spanner:6.44.0'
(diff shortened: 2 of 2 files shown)


.kokoro/build.bat
```diff
@@ -15,8 +15,4 @@
 :: downstream client libraries before they are released.
 :: See documentation in type-shell-output.bat
 
-set JAVA8_HOME=%JAVA_HOME:"=%
-choco install -y openjdk11
-set JAVA11_HOME=C:\Program Files\Eclipse Adoptium\jdk-11.0.18.10-hotspot\
-
```
harshachinta, line 21: @rajatbhatta 
We generally update these jdk versions to latest manually, whenever a windows build fail. But do we need these explicitly getting installed and set JAVA11_HOME?

The build.sh file already sets these `JAVA8_HOME` and `JAVA11_HOME`
https://github.com/googleapis/java-spanner/blob/08af6714055b5770164db66ec3c54c3a35b58724/.kokoro/build.sh#L53-L54
https://github.com/googleapis/java-spanner/blob/08af6714055b5770164db66ec3c54c3a35b58724/.kokoro/build.sh#L28-L29

I compared the windows build by removing and having this and it seems it is able to capture the latest JDK by default.

`With out above lines`
https://screenshot.googleplex.com/8k9CiGKsgEknz8a

`With above lines after updating to latest version`
https://screenshot.googleplex.com/xBRpe9UXnmXjAPw

Can you please give your view here?

rajatbhatta: CC: @mpeddada1 for their inputs on this..

olavloite: I have no idea why that is explicitly being installed here. But the outcome of the build seems to indicate that we don't need it.
Also; Our JDBC driver build setup also does not include it: https://github.com/googleapis/java-spanner-jdbc/blob/main/.kokoro/build.bat

mpeddada1: I can provide a bit of context here. The graal-sdk dependency, which is used to provide native image configurations requires a minimum version of JDK 11 otherwise it results in a compilation error:
```
  bad class file: C:\Users\kbuilder\.m2\repository\org\graalvm\sdk\graal-sdk\22.3.0\graal-sdk-22.3.0.jar(org/graalvm/nativeimage/hosted/Feature.class)
[ERROR]     class file has wrong version 55.0, should be 52.0
```
Since we still need to verify against JDK 8 for some jobs, this [change](https://github.com/googleapis/java-spanner/pull/1870/files#diff-6ec64aad727811781a78bc1dd7e4fd68ac708ccff197213425ae5933df4f0a49) was made to compile tests against JDK 11 but run them against JDK 8.  This logic requires both JAVA11_HOME and JAVA8_HOME env variables to be [provided](https://github.com/googleapis/java-spanner/pull/1870/files#r880579798) so when they initially weren't included in the Windows java 8 job, we started seeing the compilation error mentioned above (https://github.com/googleapis/java-spanner/issues/2205). IIRC, the `setJava` method only assigns the `PATH` to the `JAVA8_HOME` value but doesn't contain any logic to compute it. 

@harshachinta in the screenshot, I see that the windows tests are running in JDK 11. Are we no longer validating with JDK 8? If we're only testing with JDK 11 then we should no longer need these lines.

olavloite: @harshachinta Any thoughts on the last question?

harshachinta: @mpeddada1 
I see that we are running tests on JDK 8 and JDK 11. https://github.com/googleapis/java-spanner/blob/fbf1df9f3fb12faaead8634b88fd4843cbdedf5b/.kokoro/build.sh#L26

Running on JDK 11 - https://screenshot.googleplex.com/87wspyzBrPNqRJE
Running on JDK 8 - https://screenshot.googleplex.com/uFos4MuHKEBYXNg

**Note:**
However, I observed that the JDK 8 and JDK 11 is already set but we are overwriting them by setting them again in the build.bat file.
https://github.com/googleapis/java-spanner/blob/fbf1df9f3fb12faaead8634b88fd4843cbdedf5b/.kokoro/build.bat#L20

From the below screenshot the environment variables `JAVA8_HOME` and `JAVA11_HOME` are already set.
https://screenshot.googleplex.com/7AUaPVjtWkS3ups
We are doing redundant work of downloading them again and in most cases leads to version mismatch due to the old version being hardcoded in the [build.bat](https://github.com/googleapis/java-spanner/blob/fbf1df9f3fb12faaead8634b88fd4843cbdedf5b/.kokoro/build.bat#L20) file

I think we no longer need these and can remove it. 
@olavloite @mpeddada1 Can you please let me know your views on this?

mpeddada1: Thanks for the detailed explanation @harshachinta! re `I see that we are running tests on JDK 8 and JDK 11.` and `the environment variables JAVA8_HOME and JAVA11_HOME are already set`: given this observation, it looks like we no longer need these environment variables.

mpeddada1 (approved):

PR #2579, opened 2023-08-10
fix: use streaming read/query settings for stream retry

olavloite:
Use the streaming read/query settings from the SpannerStubSettings to determine when and how to retry a failure halfway a streaming call.


google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
```diff
@@ -1099,24 +1104,58 @@ abstract static class ResumableStreamIterator extends AbstractIterator<PartialRe
...
-    }
-
-    private static ExponentialBackOff newBackOff() {
+      this.streamingRetrySettings = Preconditions.checkNotNull(streamingRetrySettings);
+      this.retryableCodes = Preconditions.checkNotNull(retryableCodes);
+      this.backOff = newBackOff();
+    }
+
+    private ExponentialBackOff newBackOff() {
+      if (Objects.equals(streamingRetrySettings, DEFAULT_STREAMING_RETRY_SETTINGS)) {
```
olavloite, line 1122: This ensures that we fall back to the current behavior if the user has not set any custom streaming read settings. This is safer, as the current default is to use a max 1000ms retry delay, and applying the actual streaming settings would change that.

rajatbhatta: What if the user explicitly sets the settings as the default streaming retry settings? How will we distinguish it from not being set?

olavloite: We would need to keep track of a separate field to see if they had called one of the setters. However, I don't think we need to do that. The default streaming retry settings are settings that disable retries, and use a zero duration for retry delays. This is not allowed on the `ExponentialBackoff` class, as it requires the delays to be at least 1ms, which means that we would not be able to honour those settings anyway. So I think it is therefore reasonable to fall back to the current behavior of using a min 10ms and max 1000ms retry-delay for those cases. If the user wants a retry-delay that is as small as possible, then they can always create a `RetrySettings` with a 1ns retry delay instead of one with a zero retry delay.

rajatbhatta (approved):
LGTM

PR #2580, opened 2023-08-11
fix: always allow metadata queries

olavloite:
Internal metadata queries should always be allowed, regardless of the type of transaction that is currently running or any special mode that has been set on the connection. Internal metadata queries are system queries that are generated by a connection to return metadata to an application, for example methods like getTables() in JDBC.


google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
```diff
@@ -133,17 +133,6 @@ enum BatchMode {
     DML
   }
 
-  /**
```
olavloite, line 136: This was dead code. Only the class in the public `Connection` interface was used.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/DdlBatch.java
```diff
@@ -121,29 +118,6 @@ public ApiFuture<ResultSet> executeQueryAsync(
       final ParsedStatement statement,
       AnalyzeMode analyzeMode,
       QueryOption... options) {
-    if (options != null) {
```
olavloite, line 124: This special handling is no longer needed, as it is generically handled in the connection.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/DdlBatchTest.java
```diff
@@ -157,25 +157,6 @@ public void testExecuteCreateDatabase() {
                     .parse(Statement.of("CREATE DATABASE foo"))));
   }
 
-  @Test
-  public void testExecuteMetadataQuery() {
```
olavloite, line 161: This test specifically checked whether a DDL batch could execute an internal metadata query. That is no longer needed, as a DDL batch no longer handles that special case itself. Instead, this is handled directly in the connection for all types of transactions.

olavloite, line 161: This test specifically checked whether a DDL batch could execute an internal metadata query. That is no longer needed, as a DDL batch no longer handles that special case itself. Instead, this is handled directly in the connection for all types of transactions.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
```diff
@@ -1514,22 +1519,31 @@ private ApiFuture<long[]> internalExecuteBatchUpdateAsync(
...
+      this.currentUnitOfWork = createNewUnitOfWork(false);
     }
     return this.currentUnitOfWork;
   }
 
   @VisibleForTesting
-  UnitOfWork createNewUnitOfWork() {
-    if (isAutocommit() && !isInTransaction() && !isInBatch()) {
+  UnitOfWork createNewUnitOfWork(boolean isInternalMetadataQuery) {
+    if (isInternalMetadataQuery || isAutocommit() && !isInTransaction() && !isInBatch()) {
```
rajatbhatta, line 1544:
```suggestion
    if (isInternalMetadataQuery || (isAutocommit() && !isInTransaction() && !isInBatch())) {
```
Makes it clearer.

rajatbhatta (approved):
LGTM, apart from just one nit.

PR #2591, opened 2023-08-15
chore: deduplicate code for batch updates

olavloite:
De-duplicate code that is equal for both the sync and async version of the executeBatchUpdate methods.


bharadwaj-aditya (approved):
LGTM

PR #2592, opened 2023-08-15
chore: add execute method with allowed return types

olavloite:
Adds an execute method to the Connection API that allows the caller to specify the allowed result types. This can be used by driver implementations, such as JDBC, to use a single execute method in the Connection API, while still making sure that the method only executes statements that it should.
The current executeUpdate method in the Connection API does not overlap completely with semantics of executeUpdate in JDBC, as JDBC allows any statement type that does not return a ResultSet to be executed with that method. The Connection API only allows statements that return an update count.
Instead of modifying the executeUpdate method in the Connection API to match the semantics of JDBC (which would be a breaking change), this method can be used generically for all execute*** methods in the JDBC driver, which again can be used to clean up some code there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,13 +57,13 @@ implementation 'com.google.cloud:google-cloud-spanner'
 If you are using Gradle without BOM, add this to your dependencies:
 
 ```Groovy
-implementation 'com.google.cloud:google-cloud-spanner:6.45.1'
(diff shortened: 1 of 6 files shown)


bharadwaj-aditya (approved):
LGTM

PR #2606, opened 2023-08-31
feat: disable dynamic code loading properties by default

olavloite:
Disable Connection URL properties that dynamically invoke code by default. These properties can now only be used if the corresponding System property has been enabled. This gives users control over whether they want to enable these properties in their applications or not.

You should only enable the use of these properties as long as untrusted end users cannot dynamically set these. That is; If your application is a public service that allows end users to set a Connection URL, then you should not enable the use of these properties, as it would allow a user to try to specify a credentials or channel provider class that is not a valid provider. The constructor of the selected class would in that case still be invoked.

### DIFF
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionOptionsTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionOptionsTest.java
@@ -36,18 +36,19 @@ import com.google.cloud.spanner.SpannerOptions;
 import com.google.common.io.BaseEncoding;
 import com.google.common.io.Files;
 import java.io.File;
-import java.io.IOException;
-import java.io.UnsupportedEncodingException;
+import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.Objects;
 import org.junit.Test;
+import org.junit.function.ThrowingRunnable;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
 
 @RunWith(JUnit4.class)
 public class ConnectionOptionsTest {
   private static final String FILE_TEST_PATH =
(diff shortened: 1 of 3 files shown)


google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/CredentialsProviderTest.java
manu2, line 1: Should we add a test for WithoutEnablingProperty() that checks the exception message for credentialsProvider as well?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionOptionsTest.java
```diff
@@ -727,116 +728,208 @@ public void testInvalidCredentials() {
...
     String uri =
         String.format(
             "cloudspanner:/projects/test-project/instances/test-instance/databases/test-database?credentialsProvider=%s",
             TestCredentialsProvider.class.getName());
+    SpannerException exception =
+        assertThrows(
+            SpannerException.class, () -> ConnectionOptions.newBuilder().setUri(uri).build());
+    assertEquals(ErrorCode.FAILED_PRECONDITION, exception.getErrorCode());
+    assertEquals(
+        "FAILED_PRECONDITION: credentialsProvider can only be used if the system property ENABLE_CREDENTIALS_PROVIDER has been set to true. "
```
olavloite, line 871: The error that you get when setting a CredentialsProvider without enabling the property is verified here.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/CredentialsProviderTest.java
olavloite: We test that here: https://github.com/googleapis/java-spanner/pull/2606/files#r1318465848

manu2 (approved):

PR #2607, opened 2023-09-01
feat: OpenTelemetry Metrics and Traces

surbhigarg92:
This PR contains opentelemetry Session metrics and grpc metrics implementation


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2034,8 +2047,12 @@ private enum Position {
    * be created.
    */
   static SessionPool createPool(
-      SpannerOptions spannerOptions, SessionClient sessionClient, List<LabelValue> labelValues) {
+      SpannerOptions spannerOptions,
+      SessionClient sessionClient,
+      List<LabelValue> labelValues,
+      Attributes attributes) {
     final SessionPoolOptions sessionPoolOptions = spannerOptions.getSessionPoolOptions();
+    spannerOptions.getOpenTelemetry();
```
olavloite, line 2055: Why do we call this without doing anything with the return value?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2068,17 +2087,22 @@ static SessionPool createPool(
         sessionClient,
         clock,
         Metrics.getMetricRegistry(),
-        SPANNER_DEFAULT_LABEL_VALUES);
+        SPANNER_DEFAULT_LABEL_VALUES,
+        null,
+        null);
   }
 
+  // main
```
olavloite, line 2095: Remove? Or otherwise expand on the comment.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2761,4 +2790,77 @@ private void initMetricsCollection(MetricRegistry metricRegistry, List<LabelValu
         // TODO: Remove metric.
         ignored -> 0L);
   }
+
+  private void initOpenTelemetryMetricsCollection(
+      OpenTelemetry openTelemetry, Attributes attributes) {
+
+    if (openTelemetry != null) {
```
olavloite, line 2797: I think it would be better to invert this if condition and do an early return. So like this:

```java
if (openTelemetry == null) {
  return;
}
```

That reduces the indentation of this method.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2761,4 +2790,77 @@ private void initMetricsCollection(MetricRegistry metricRegistry, List<LabelValu
...
+      Meter meter = openTelemetry.getMeter(MetricRegistryConstants.Scope);
+
+      meter
+          .gaugeBuilder(MAX_ALLOWED_SESSIONS_NAME)
+          .setDescription(MAX_ALLOWED_SESSIONS_DESCRIPTION)
+          .setUnit(COUNT)
+          .buildWithCallback(
+              measurement -> {
+                measurement.record(options.getMaxSessions(), attributes);
```
olavloite, line 2806: When is this callback being called? Because `options.getMaxSessions()` will never change during the lifetime of the session pool, so if there is something like `buildWithConstant(...)` or similar, then that would probably be a better fit here.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2761,4 +2790,77 @@ private void initMetricsCollection(MetricRegistry metricRegistry, List<LabelValu
...
+          attributesBuilder.put(SESSIONS_TYPE, NUM_SESSIONS_IN_USE_NAME).build();
+      Attributes attributesAvailableSessions =
+          attributesBuilder.put(SESSIONS_TYPE, NUM_SESSIONS_AVAILABLE_NAME).build();
+      meter
+          .upDownCounterBuilder(NUM_SESSIONS_IN_POOL_NAME)
+          .setDescription(NUM_SESSIONS_IN_POOL_DESCRIPTION)
+          .setUnit(COUNT)
+          .buildWithCallback(
+              measurement -> {
+                measurement.record(this.numSessionsInUse, attributesInUseSessions);
```
olavloite, line 2834: Access to properties like `this.numSessionsInUse` and `this.sessions` is normally guarded with a `synchronized (lock)` block to prevent concurrent reads/writes to take place. Do we need that for this as well? (When are these callbacks being called?)
And if we were to add `synchronized (lock)` in these callbacks: Do the threads that call these callbacks take any other locks as well? If yes, can we guarantee that there won't be a deadlock?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
```diff
@@ -218,9 +220,20 @@ public DatabaseClient getDatabaseClient(DatabaseId db) {
                 LabelValue.create(db.getDatabase()),
                 LabelValue.create(db.getInstanceId().getName()),
                 LabelValue.create(GaxProperties.getLibraryVersion(getOptions().getClass())));
+
+        AttributesBuilder attributesBuilder = Attributes.builder();
+        attributesBuilder.put("client_id", clientId);
+        attributesBuilder.put("database", db.getDatabase());
+        attributesBuilder.put("instance_id", db.getInstanceId().getName());
```
olavloite, line 227: This seems a bit inconsistent at first glance:
1. We add `database` with the ID of the database.
2. We add `instance_id` with the name of the instance.

Normally, we use this terminology:
* `projects/my-project/instance/my-instance`: This is an instance name. The instance ID is `my-instance`

Is this intentional because other Google Cloud products do it this way? Or would it maybe make more sense to either add `project_id`, `instance_id` and `database_id` all as separate attributes?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -701,6 +704,7 @@ public static class Builder
     private String compressorName;
     private String emulatorHost = System.getenv("SPANNER_EMULATOR_HOST");
     private boolean leaderAwareRoutingEnabled = true;
+    private OpenTelemetry openTelemetry = null;
```
olavloite, line 707: nit:
```suggestion
    private OpenTelemetry openTelemetry;
```

(null is the default value)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1205,6 +1216,7 @@ public SpannerOptions build() {
             this.grpcGcpExtensionEnabled ? GRPC_GCP_ENABLED_DEFAULT_CHANNELS : DEFAULT_CHANNELS;
       }
 
+      SpannerRpcMetrics.initializeRPCMetrics(this.openTelemetry);
```
olavloite, line 1219: I don't think we want this here (although I don't really know what it does). But it seems strange to initialize some metrics as a side-effect of building an instance of `SpannerOptions`. At this time, we have still not opened a connection to `Spanner`. I would expect this rather to happen when you call `SpannerOptions#getService()` (it does not need to be in that method, but maybe in the constructor of any of the objects that are being created by that call)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerRpcMetrics.java
```diff
@@ -0,0 +1,55 @@
...
+import io.opentelemetry.api.metrics.LongCounter;
+import io.opentelemetry.api.metrics.LongHistogram;
+import io.opentelemetry.api.metrics.Meter;
+
+public class SpannerRpcMetrics {
+  private static LongHistogram gfeLatencies = null;
+  private static LongCounter gfeHeaderMissingCount = null;
+
+  static void initializeRPCMetrics(OpenTelemetry openTelemetry) {
+
```
olavloite, line 29: nit: remove empty line

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerRpcMetrics.java
```diff
@@ -0,0 +1,55 @@
...
+    if (openTelemetry != null) {
```
olavloite, line 30: nit: invert if statement and do an early return

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerRpcMetrics.java
```diff
@@ -0,0 +1,55 @@
...
+          meter
+              .counterBuilder(MetricRegistryConstants.SPANNER_GFE_HEADER_MISSING_COUNT_NAME)
+              .setDescription(MetricRegistryConstants.SPANNER_GFE_HEADER_MISSING_COUNT_DESCRIPTION)
+              .setUnit(MetricRegistryConstants.COUNT)
+              .build();
+    }
+  }
+
+  public static void gfeLatencyRecorder(long value, Attributes attributes) {
+    if (gfeLatencies != null) gfeLatencies.record(value, attributes);
```
olavloite, line 49: nit: always use curly braces for if statements, even if the if-block only contains a single line.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerRpcMetrics.java
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  public static void gfeHeaderMissingCountRecorder(long value, Attributes attributes) {
+    if (gfeHeaderMissingCount != null) gfeHeaderMissingCount.add(value, attributes);
```
olavloite, line 53: nit: use curly braces

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerRpcMetrics.java
```diff
@@ -0,0 +1,55 @@
...
+      gfeHeaderMissingCount =
+          meter
+              .counterBuilder(MetricRegistryConstants.SPANNER_GFE_HEADER_MISSING_COUNT_NAME)
+              .setDescription(MetricRegistryConstants.SPANNER_GFE_HEADER_MISSING_COUNT_DESCRIPTION)
+              .setUnit(MetricRegistryConstants.COUNT)
+              .build();
+    }
+  }
+
+  public static void gfeLatencyRecorder(long value, Attributes attributes) {
```
olavloite, line 48: Does this really need to be public? If so, it should have some Javadoc that explains what it does. Same for the other method in this class.

Could we maybe also reconsider the name of the method. It seems that it is recording something (so it is actually executing an action). Could we let the method name reflect that as well. So something like 'recordGfeLatency`?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerRpcMetrics.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+package com.google.cloud.spanner;
+
+import io.opentelemetry.api.OpenTelemetry;
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.api.metrics.LongCounter;
+import io.opentelemetry.api.metrics.LongHistogram;
+import io.opentelemetry.api.metrics.Meter;
+
+public class SpannerRpcMetrics {
```
olavloite, line 24: Does this class need to be public? (See also questions on the methods below).

If it is public for technical reasons, for example because we need to access it from a different package within the client library, but users should normally not access it, then we should add an `@InternalApi` annotation to the class.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/HeaderInterceptor.java
```diff
@@ -139,4 +147,32 @@ private TagContext getTagContext(Metadata headers, String method) {
...
+        }
+        if (matcher.group("database") != null) {
+          databaseId = matcher.group("database");
+        }
+      } else {
+        LOGGER.log(LEVEL, "Error parsing google cloud resource header: " + googleResourcePrefix);
+      }
+    }
+    AttributesBuilder attributesBuilder = Attributes.builder();
+    attributesBuilder.put("database", databaseId);
```
olavloite, line 171: For consistency: Name this `database_id` as well?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/HeaderInterceptor.java
```diff
@@ -139,4 +147,32 @@ private TagContext getTagContext(Metadata headers, String method) {
     }
     return getTagContext(method, projectId, instanceId, databaseId);
   }
+
+  private Attributes getMetricAttributes(Metadata headers, String method) {
```
olavloite, line 151: (I see that this is consistent with how `getTagContext` works, so I understand where it is coming from)

But this feels like a relatively expensive operation that we execute for __every__ request that we execute. Both this method and `getTagContext` match a regular expression against one of the headers to get the project, instance, and database, and then build a data structure. That structure is then only used if a number of conditions are met (metric collection has been enabled + server timing key is returned).

Maybe for a separate PR, but I would suggest that we look into:
1. parsing the project, instance, and database only once from the header
2. Only parse that information and build the data structure if we know that they will actually be used (so basically move it into the if statements above)

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolTest.java
```diff
@@ -1551,6 +1588,193 @@ public void testSessionMetrics() throws Exception {
...
+                return null;
+              });
+      // Wait until the background thread is actually waiting for a session.
+      latch.await();
+      // Wait until the request has timed out.
+      int waitCount = 0;
+      while (pool.getNumWaiterTimeouts() == 0L && waitCount < 1000) {
+        Thread.sleep(5L);
+        waitCount++;
+      }
```
olavloite, line 1694: (Just checking) Would it be possible to replace this busy-wait with some kind of test setup that advances the `FakeClock` and causes a timeout that way?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolTest.java
```diff
@@ -1551,6 +1588,193 @@ public void testSessionMetrics() throws Exception {
...
+    }
+  }
+
+  private static void verifyMetricData(
+      Collection<MetricData> metricDataCollection, String metricName, int size, long value) {
+    Collection<MetricData> metricDataFiltered =
+        metricDataCollection.stream()
+            .filter(x -> x.getName().equals(metricName))
+            .collect(Collectors.toList());
+    ;
```
olavloite, line 1733: nit: remove (here and elsewhere in this class)

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolTest.java
```diff
@@ -1551,6 +1588,193 @@ public void testSessionMetrics() throws Exception {
...
+    assertThat(metricDataFiltered.size()).isEqualTo(size);
```
olavloite, line 1734: nit: we prefer the use of `assertEquals(..)` over `assertThat(..).isEqualTo(..)`. This test class is relatively old, so I know that there are a lot of `assertThat(...)` calls here, but we are trying to remove those step-by-step.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/spi/v1/SpannerRpcMetricsTest.java
```diff
@@ -0,0 +1,304 @@
...
+    mockSpannerNoHeader.reset();
+  }
+
+  @Test
+  public void testGfeLatencyExecuteSql() throws InterruptedException {
+    databaseClient
+        .readWriteTransaction()
+        .run(transaction -> transaction.executeUpdate(UPDATE_FOO_STATEMENT));
+
+    ;
```
olavloite, line 228: nit: remove

google-cloud-spanner/src/test/java/com/google/cloud/spanner/spi/v1/SpannerRpcMetricsTest.java
```diff
@@ -0,0 +1,304 @@
...
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public class SpannerRpcMetricsTest {
+
+  private static final String STATIC_OAUTH_TOKEN = "STATIC_TEST_OAUTH_TOKEN";
```
olavloite, line 67: This seems to be copy-pasted from a different test class. Do we need all of this? And if so, would it in that case make more sense to combine the two test classes, either by creating an abstract base class that they can both extend from, or by literally combining them into one class?

Or can we just remove these?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/spi/v1/SpannerRpcMetricsTest.java
```diff
@@ -0,0 +1,304 @@
...
+            input -> {
+              input.usePlaintext();
+              return input;
+            })
+        .setHost("http://" + endpoint)
+        // Set static credentials that will return the static OAuth test token.
+        .setCredentials(STATIC_CREDENTIALS)
+        // Also set a CallCredentialsProvider. These credentials should take precedence above
+        // the static credentials.
+        .setCallCredentialsProvider(() -> MoreCallCredentials.from(VARIABLE_CREDENTIALS))
```
olavloite, line 273: I don't think you need any of these. Instead, you can just call `setCredentials(NoCredentials.getInstance())`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
surbhigarg92: These attributes names and values are kept same as what we had with OpenCensus. As these attributes might be used by customers to build dashboards and alerts, they would have to do these changes. 

For now, they will only change lets say the source, read metrics data from opentelemetry instead of opencensus.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
surbhigarg92: The frequency of callback will depend on what exporter customers will be using with what export interval. 

For instance cloudmonitoringexporter will be pushing data every 10 sec as default.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerRpcMetrics.java
surbhigarg92: This needs to be public, as it needs to be used in HeaderInterceptor class which is in different package. Added the @InetrnalApi annotation to it.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerRpcMetrics.java
surbhigarg92: This needs to be public, as it needs to be used in HeaderInterceptor class which is in different package. Added the @InetrnalApi annotation to it.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/spi/v1/SpannerRpcMetricsTest.java
surbhigarg92: Removed .

google-cloud-spanner/src/test/java/com/google/cloud/spanner/spi/v1/SpannerRpcMetricsTest.java
surbhigarg92: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
surbhigarg92: I double checked this with OpenTelemetry team. What we are doing here is the correct way of handling such data, even if its constant. 
Even if its a constant value, a callback is required every few seconds to get the current value, as this data is not stored forever.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
surbhigarg92: This will initialise gRPC metrics `gfe_latency` and `gfe_header_count`. These are gRPC metrics which will be applicable for every gRPC call and data for this is added in `headerIntercepetor` class. 

I was looking a way to initialise these metrics before any gRPC call can happen.  Customers will use SpannerOptions to initialise the `OpenTelemetry` object, once the object is passed we can initialise these metrics. 

What would be the best place to do that ? Should it be in `SpannerOptions#getService()` ? Are we sure this method will be called with all the possible ways customer might use to initialise spanner object ?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/HeaderInterceptor.java
surbhigarg92: 1) IMK, One spanner client can have multiple database object, so we need ti parse Header to get database every time ? 
We can save project and instance though and use it again. 
2) You mean we should call this `getMetricAttributes` in the `processHeader` method, right ?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
olavloite: ack

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
olavloite: ack

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/HeaderInterceptor.java
olavloite: Ack

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/HeaderInterceptor.java
```diff
@@ -73,12 +76,12 @@ public <ReqT, RespT> ClientCall<ReqT, RespT> interceptCall(
       @Override
       public void start(Listener<RespT> responseListener, Metadata headers) {
         TagContext tagContext = getTagContext(headers, method.getFullMethodName());
+        Attributes attributes = getMetricAttributes(headers, method.getFullMethodName());
```
olavloite, line 79: Can we combine the header parsing of these two calls into one? For example create a separate method only to extract the full `DatabaseId` from the header, and then use that `DatabaseId` instance?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
olavloite: I'm not necessarily worried about the frequency, more about potential race conditions, and maybe also not so much about these specific cases. But we also call `sessions.size()` for example without taking a lock. `sessions.size()` is undefined if it is called while other threads are modifying the collection.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
olavloite: `SpannerOptions#getService()` is always called before a user can call any gRPC methods. But I think that in this case  it would be better to put this call in the `SpannerImpl` constructor.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerRpcMetrics.java
olavloite: ack

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/HeaderInterceptor.java
surbhigarg92: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/HeaderInterceptor.java
surbhigarg92: done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
surbhigarg92: done

README.md
```diff
@@ -158,6 +158,9 @@ the Cloud Spanner Java client.
 
 ## OpenCensus Metrics
 
+> Note: OpenCensus project is deprecated. See [Sunsetting OpenCensus](https://opentelemetry.io/blog/2023/sunsetting-opencensus/).
+We recommend migrating to OpenTelemetry, the successor project.
```
olavloite, line 162: nit: add a link to `OpenTelemetry`

google-cloud-spanner/pom.xml
```diff
@@ -238,6 +239,40 @@
...
+    <dependency>
+      <groupId>io.opentelemetry</groupId>
+      <artifactId>opentelemetry-context</artifactId>
+      <version>${opentelemetry.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.opentelemetry</groupId>
+      <artifactId>opentelemetry-sdk</artifactId>
+      <version>${opentelemetry.version}</version>
+      <scope>test</scope>
```
olavloite, line 256: Are you sure we want to use `test` scope for this? `test` means that we add these dependencies because we need them in our tests. If that is the case, then we should move this down to the other test dependencies further below.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
```diff
@@ -348,6 +356,7 @@ void initTransaction() {
         if (transactionId != null) {
           return;
         }
+        OpenTelemetryTraceUtil.addEvent(openTelemetrySpan, "Creating Transaction");
         span.addAnnotation("Creating Transaction");
```
olavloite, line 360: Would it make sense to create a util method somewhere that takes both an OpenTelemetry span and an OC span as an argument, and then add an event/annotation to both? That way we know that we always use the same event name, and we can keep this on one line.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
```diff
@@ -374,9 +383,15 @@ void initTransaction() {
...
           transactionId = transaction.getId();
+          OpenTelemetryTraceUtil.addEvent(
+              openTelemetrySpan,
+              "Transaction Creation Done",
+              OpenTelemetryTraceUtil.getTransactionAnnotations(transaction));
           span.addAnnotation(
               "Transaction Creation Done", TraceUtil.getTransactionAnnotations(transaction));
         } catch (SpannerException e) {
+          OpenTelemetryTraceUtil.addEventWithException(
+              openTelemetrySpan, "Transaction Creation Failed", e);
```
olavloite, line 394: Same here also for these: Would it be possible to create some util methods for this, so we can just call one method, and then let the util method call both OT and OC?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
```diff
@@ -1213,6 +1223,7 @@ public void close(@Nullable String message) {
       if (stream != null) {
         stream.close(message);
         span.end(TraceUtil.END_SPAN_OPTIONS);
+        OpenTelemetryTraceUtil.endSpan(openTelemetrySpan);
```
olavloite, line 1226: Same also for end as for the other span method: Could we combine these in a util method?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/MetricRegistryConstants.java
```diff
@@ -84,4 +95,13 @@ class MetricRegistryConstants {
   static final String NUM_RELEASED_SESSIONS_DESCRIPTION =
       "The number of sessions released by the user and pool maintainer.";
   static final String NUM_SESSIONS_IN_POOL_DESCRIPTION = "The number of sessions in the pool.";
+
+  static final String Scope = "cloud.google.com/java/spanner";
+  static final String SPANNER_GFE_LATENCY_NAME = "gfe_latency";
+  static final String SPANNER_GFE_LATENCY_DESCRIPTION =
+      "Latency between Google's network receiving an RPC and reading back the first byte of the response";
```
olavloite, line 102: Should we consider using the word 'request' instead of 'RPC'? So like this:
```suggestion
      "Latency between Google's network receiving a request and reading back the first byte of the response";
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/MetricRegistryConstants.java
```diff
@@ -84,4 +95,13 @@ class MetricRegistryConstants {
...
+  static final String SPANNER_GFE_HEADER_MISSING_COUNT_NAME = "gfe_header_missing_count";
+  static final String SPANNER_GFE_HEADER_MISSING_COUNT_DESCRIPTION =
+      "Number of RPC responses received without the server-timing header, most likely means that the RPC never reached Google's network";
```
olavloite, line 105: Here also: 'request' instead of 'RPC'?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/PartitionedDmlTransaction.java
```diff
@@ -141,6 +141,10 @@ public void invalidate() {
   @Override
   public void setSpan(Span span) {}
 
+  // No-op method needed to implement SessionTransaction interface.
```
olavloite, line 144: nit: Either put this comment inside the method body, or leave it here but change it to Javadoc

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1346,8 +1367,12 @@ PooledSession get(final boolean eligibleForLongRunning) {
           // ignore the exception as it will be handled by the call to super.get() below.
         }
         if (res != null) {
-          res.markBusy(span);
+          res.markBusy(span, openTelemetrySpan);
           span.addAnnotation(sessionAnnotation(res));
+          OpenTelemetryTraceUtil.addEvent(
+              openTelemetrySpan,
+              "Using Session",
+              Attributes.builder().put("sessionId", res.getName()).build());
```
olavloite, line 1375: Are these the same? (Does `sessionAnnotation(res)` return the same as we are adding to OT?)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2887,4 +2953,77 @@ private void initMetricsCollection(MetricRegistry metricRegistry, List<LabelValu
...
+    }
+
+    Meter meter = openTelemetry.getMeter(MetricRegistryConstants.Scope);
+    meter
+        .gaugeBuilder(MAX_ALLOWED_SESSIONS_NAME)
+        .setDescription(MAX_ALLOWED_SESSIONS_DESCRIPTION)
+        .setUnit(COUNT)
+        .buildWithCallback(
+            measurement -> {
+              measurement.record(options.getMaxSessions(), attributes);
```
olavloite, line 2970: nit: Can we add a small comment here noting that 'although this is a constant value, OT requires us to define this as a callback instead of setting a constant only once' (or something similar)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -134,6 +135,7 @@ public class SpannerOptions extends ServiceOptions<Spanner, SpannerOptions> {
   private final CloseableExecutorProvider asyncExecutorProvider;
   private final String compressorName;
   private final boolean leaderAwareRoutingEnabled;
+  private static OpenTelemetry openTelemetry;
```
olavloite, line 138: `static`? That seems strange. Should this not be a normal property?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1185,6 +1187,12 @@ public Builder disableLeaderAwareRouting() {
       return this;
     }
 
+    /** Sets OpenTelemetry object to be used for Spans, Metrics, Views. */
+    public Builder setOpenTelemetry(OpenTelemetry openTelemetry) {
```
olavloite, line 1191: See also my comment above: This sets a static property. That means that calling this not only affects this `SpannerOptions`, but the entire JVM. If that is intentional, then we should change this also into a static method and document it as such.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
```diff
@@ -1052,6 +1114,13 @@ private <T> T runInternal(final TransactionCallable<T> txCallable) {
                     .putAll(TraceUtil.getExceptionAnnotations(e))
                     .put("Attempt", AttributeValue.longAttributeValue(attempt.longValue()))
                     .build());
+            OpenTelemetryTraceUtil.addEvent(
+                openTelemetrySpan,
+                "Transaction Attempt Failed in Commit",
+                Attributes.builder()
+                    .put("Status", ((SpannerException) e).getErrorCode().toString())
```
olavloite, line 1121: nit: the cast to `SpannerException` should not be necessary here. It is already declared as a `SpannerException`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerRpcViews.java
```diff
@@ -91,8 +91,12 @@ public class SpannerRpcViews {
    * measures the latency between Google's network receives an RPC and reads back the first byte of
    * the response. gfe_header_missing_count is a counter of the number of RPC responses without a
    * server-timing header.
+   *
+   * @deprecated The OpenCensus project is deprecated. Use OpeTelemetry to get gfe_latency and
+   *     gfe_header_missing_count metrics as default.
```
olavloite, line 96: nit (here and elsewhere):
```suggestion
   *     gfe_header_missing_count metrics.
```

google-cloud-spanner/src/test/java/com/google/cloud/spanner/OpenTelemetrySpanTest.java
```diff
@@ -0,0 +1,262 @@
+/*
+ * Copyright 2020 Google LLC
```
olavloite, line 2: nit:

```suggestion
 * Copyright 2023 Google LLC
```

google-cloud-spanner/src/test/java/com/google/cloud/spanner/OpenTelemetrySpanTest.java
```diff
@@ -0,0 +1,262 @@
...
+          spans.stream()
+              .allMatch(
+                  name ->
+                      name.equals("CloudSpannerOperation.BatchCreateSessionsRequest")
+                          || name.equals("CloudSpannerOperation.ExecuteStreamingQuery")
+                          || name.equals("CloudSpannerOperation.BatchCreateSessions")
+                          || name.equals("CloudSpanner.ReadOnlyTransaction")
+                          || name.equals("SessionPool.WaitForSession"));
+      assertTrue(hasSpanData);
+      assertThat(spans.size()).isEqualTo(5);
```
olavloite, line 184: nit (here and elsewhere): prefer `assertEquals` over `assertThat(...).isEqualTo(...)`

google-cloud-spanner/src/test/java/com/google/cloud/spanner/OpenTelemetrySpanTest.java
```diff
@@ -0,0 +1,262 @@
...
+    mockSpanner.putStatementResult(StatementResult.update(UPDATE_STATEMENT, UPDATE_COUNT));
+    mockSpanner.putStatementResult(
+        StatementResult.exception(
+            INVALID_UPDATE_STATEMENT,
+            Status.INVALID_ARGUMENT.withDescription("invalid statement").asRuntimeException()));
+    String uniqueName = InProcessServerBuilder.generateName();
+    server =
+        InProcessServerBuilder.forName(uniqueName)
+            // We need to use a real executor for timeouts to occur.
+            .scheduledExecutorService(new ScheduledThreadPoolExecutor(1))
```
olavloite, line 110: nit: you can probably remove this. These tests do not test any timeouts.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2887,4 +2921,77 @@ private void initMetricsCollection(MetricRegistry metricRegistry, List<LabelValu
...
+    }
+    Attributes attributesInUseSessions =
+        attributesBuilder.put(SESSIONS_TYPE, NUM_SESSIONS_IN_USE_NAME).build();
+    Attributes attributesAvailableSessions =
+        attributesBuilder.put(SESSIONS_TYPE, NUM_SESSIONS_AVAILABLE_NAME).build();
+    meter
+        .upDownCounterBuilder(NUM_SESSIONS_IN_POOL_NAME)
+        .setDescription(NUM_SESSIONS_IN_POOL_DESCRIPTION)
+        .setUnit(COUNT)
+        .buildWithCallback(
```
blakeli0, line 2964: I don't think a counter has to be built with call back, is it due to async counter has better performance?

google-cloud-spanner/pom.xml
surbhigarg92: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
surbhigarg92: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
surbhigarg92: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/MetricRegistryConstants.java
surbhigarg92: This is the existing description message for the metrics which might already be used by customers. Should we change this ?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
surbhigarg92: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
surbhigarg92: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
surbhigarg92: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/HeaderInterceptor.java
```diff
@@ -137,6 +146,26 @@ private TagContext getTagContext(Metadata headers, String method) {
...
+        .putLocal(INSTANCE_ID, TagValue.create(spannerProperties.instanceId))
+        .putLocal(DATABASE_ID, TagValue.create(spannerProperties.databaseId))
+        .putLocal(METHOD, TagValue.create(method))
+        .build();
+  }
+
+  private Attributes getMetricAttributes(String method, SpannerProperties spannerProperties) {
+    AttributesBuilder attributesBuilder = Attributes.builder();
+    attributesBuilder.put("database", spannerProperties.databaseId);
+    attributesBuilder.put("instance_id", spannerProperties.instanceId);
```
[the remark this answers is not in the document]
surbhigarg92: Agree. But here we are migrating existing metrics, these attributes were already present. So if we remove these attributes now, it will be a breaking change for customers. We will keep this in mind with new metrics

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
surbhigarg92: Not performance, but the values will keep changing. So async callbacks makes more sense here.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/HeaderInterceptor.java
```diff
@@ -98,27 +114,20 @@ private void processHeader(Metadata metadata, TagContext tagContext) {
           measureMap.put(SPANNER_GFE_LATENCY, latency);
           measureMap.put(SPANNER_GFE_HEADER_MISSING_COUNT, 0L);
           measureMap.record(tagContext);
+
+          SpannerRpcMetrics.recordGfeLatency(latency, attributes);
```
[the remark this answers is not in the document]
surbhigarg92: No. For OpenTelemetry metrics and Spans to works , customers would need to create OpenTelemetry object and register as global else it will be no-op. 
For Shim customers need not create opentelemetry object.

google-cloud-spanner/pom.xml
```diff
@@ -16,6 +16,7 @@
   <properties>
     <site.installationModule>google-cloud-spanner</site.installationModule>
     <opencensus.version>0.31.1</opencensus.version>
+    <opentelemetry.version>1.30.1</opentelemetry.version>
```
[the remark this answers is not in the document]
surbhigarg92: Sure

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
```diff
@@ -375,9 +380,15 @@ void initTransaction() {
           }
           transactionId = transaction.getId();
           span.addAnnotation(
-              "Transaction Creation Done", TraceUtil.getTransactionAnnotations(transaction));
+              "Transaction Creation Done",
+              ImmutableMap.of(
+                  "id",
```
olavloite, line 385: nit: is it intentional that this is `id` and the other is `Timestamp` (i.e. all lower case vs. starting with upper case)?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
```diff
@@ -419,10 +430,14 @@ void initTransaction() {
   }
 
   @Override
-  public void setSpan(Span span) {
+  public void setSpan(ISpan span) {
     this.span = span;
   }
 
+  /** No-op method needed to implement SessionTransaction interface. */
```
olavloite, line 437: Would it not make more sense to change the interface definition to accept an `ISpan` instead?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
```diff
@@ -1230,11 +1221,9 @@ protected PartialResultSet computeNext() {
...
           span.addAnnotation(
               "Starting/Resuming stream",
-              ImmutableMap.of(
-                  "ResumeToken",
-                  AttributeValue.stringAttributeValue(
-                      resumeToken == null ? "null" : resumeToken.toStringUtf8())));
-          try (Scope s = tracer.withSpan(span)) {
+              "ResumeToken",
+              resumeToken == null ? "null" : resumeToken.toStringUtf8());
+          try (IScope sss = tracer.withSpan(span)) {
```
olavloite, line 1226: nit: 
```suggestion
          try (IScope scope = tracer.withSpan(span)) {
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClientImpl.java
```diff
@@ -96,152 +92,163 @@ public Timestamp writeAtLeastOnce(final Iterable<Mutation> mutations) throws Spa
...
   public ReadContext singleUse() {
-    Span span = tracer.spanBuilder(READ_ONLY_TRANSACTION).startSpan();
-    try (Scope s = tracer.withSpan(span)) {
+    ISpan span = tracer.spanBuilder(READ_ONLY_TRANSACTION);
+    try (IScope s = tracer.withSpan(span)) {
       return getSession().singleUse();
     } catch (RuntimeException e) {
-      TraceUtil.endSpanWithFailure(span, e);
+      span.setStatus(e);
+      span.end();
```
olavloite, line 129: We are only calling `span.end()` in case of an exception, but not when it finishes successfully. Is that intentional, or should this have been in a `finally` block?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClientImpl.java
```diff
@@ -96,152 +92,163 @@ public Timestamp writeAtLeastOnce(final Iterable<Mutation> mutations) throws Spa
...
   public ReadContext singleUse(TimestampBound bound) {
-    Span span = tracer.spanBuilder(READ_ONLY_TRANSACTION).startSpan();
-    try (Scope s = tracer.withSpan(span)) {
+    ISpan span = tracer.spanBuilder(READ_ONLY_TRANSACTION);
+    try (IScope s = tracer.withSpan(span)) {
       return getSession().singleUse(bound);
     } catch (RuntimeException e) {
-      TraceUtil.endSpanWithFailure(span, e);
+      span.setStatus(e);
+      span.end();
```
olavloite, line 141: Same here and elsewhere: Should this be in a finally block?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DualScope.java
```diff
@@ -0,0 +1,40 @@
...
+ * limitations under the License.
+ */
+
+package com.google.cloud.spanner;
+
+import io.opencensus.common.Scope;
+
+class DualScope implements IScope {
+
+  private Scope openCensusScope;
```
olavloite, line 23: nit: these fields can both be final

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DualSpan.java
```diff
@@ -0,0 +1,140 @@
...
+import io.opencensus.trace.Status;
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.api.common.AttributesBuilder;
+import io.opentelemetry.api.trace.StatusCode;
+import java.util.HashMap;
+import java.util.Map;
+
+class DualSpan implements ISpan {
+
+  public static EndSpanOptions END_SPAN_OPTIONS =
```
olavloite, line 33: `public` can be removed, and it should probably be `final`:
```suggestion
  static final EndSpanOptions END_SPAN_OPTIONS =
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DualSpan.java
```diff
@@ -0,0 +1,140 @@
...
+      EndSpanOptions.builder().setSampleToLocalSpanStore(true).build();
+  private Span openCensusSpan;
+  private io.opentelemetry.api.trace.Span openTelemetrySpan;
```
olavloite, line 36: final

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DualSpan.java
```diff
@@ -0,0 +1,140 @@
...
+
+  public DualSpan(Span openCensusSpan, io.opentelemetry.api.trace.Span openTelemetrySpan) {
+    this.openCensusSpan = openCensusSpan;
+    this.openTelemetrySpan = openTelemetrySpan;
```
olavloite, line 40:
```suggestion
    this.openCensusSpan = Preconditions.checkNotNull(openCensusSpan);
    this.openTelemetrySpan = Preconditions.checkNotNull(openTelemetrySpan);
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DualSpan.java
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  public Span getOpenCensusSpan() {
```
olavloite, line 43:
```suggestion
  Span getOpenCensusSpan() {
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DualSpan.java
```diff
@@ -0,0 +1,140 @@
...
+    return openCensusSpan;
+  }
+
+  public io.opentelemetry.api.trace.Span getOpenTelemetrySpan() {
```
olavloite, line 47:
```suggestion
  io.opentelemetry.api.trace.Span getOpenTelemetrySpan() {
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/ISpan.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+package com.google.cloud.spanner;
+
+import io.opencensus.trace.Status;
+import java.util.Map;
+
+interface ISpan {
+
+  void addAnnotation(String message, Map<String, Object> attributes);
```
olavloite, line 24: Would you mind adding some javadoc to this method? Most of these methods are self-explanatory, but this one isn't.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/MetricRegistryConstants.java
olavloite: No, good point.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionImpl.java
```diff
@@ -89,6 +87,8 @@ interface SessionTransaction {
     void invalidate();
     /** Registers the current span on the transaction. */
     void setSpan(Span span);
```
olavloite, line 89: Can we remove this one?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TraceWrapper.java
```diff
@@ -0,0 +1,88 @@
...
+
+import io.opencensus.trace.Span;
+import io.opencensus.trace.Tracer;
+import io.opentelemetry.context.Context;
+
+class TraceWrapper {
+
+  private Tracer openCensusTracer;
+
+  public TraceWrapper(Tracer openCensusTracer) {
```
olavloite, line 27: You can make this constructor and all methods in this class package-private (that is; remove the `public` keyword), as the class itself is package-private.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TraceWrapper.java
```diff
@@ -0,0 +1,88 @@
...
+            .startSpan();
+
+    io.opentelemetry.api.trace.Span otSpan;
+
+    if (dualParentSpan.getOpenTelemetrySpan() != null)
+      otSpan =
+          SpannerOptions.getOpenTelemetry()
+              .getTracer(MetricRegistryConstants.Scope)
+              .spanBuilder(spanName)
+              .setParent(Context.current().with(dualParentSpan.getOpenTelemetrySpan()))
```
olavloite, line 59: Will `Context.current().with(dualParentSpan.getOpenTelemetrySpan())` change the current context? If so, is that intentional, and when do we reset it to its previous value?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/HeaderInterceptor.java
```diff
@@ -98,27 +114,20 @@ private void processHeader(Metadata metadata, TagContext tagContext) {
...
-        .currentBuilder()
-        .putLocal(PROJECT_ID, TagValue.create(projectId))
-        .putLocal(INSTANCE_ID, TagValue.create(instanceId))
-        .putLocal(DATABASE_ID, TagValue.create(databaseId))
-        .putLocal(METHOD, TagValue.create(method))
-        .build();
-  }
-
-  private TagContext getTagContext(Metadata headers, String method) {
+  private SpannerProperties setProjectPropertes(Metadata headers) {
```
olavloite, line 130: nit:
```suggestion
  private static SpannerProperties createProjectPropertes(Metadata headers) {
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerRpcViews.java
```diff
@@ -91,8 +92,13 @@ public class SpannerRpcViews {
    * measures the latency between Google's network receives an RPC and reads back the first byte of
    * the response. gfe_header_missing_count is a counter of the number of RPC responses without a
    * server-timing header.
+   *
+   * @deprecated The OpenCensus project is deprecated. Use OpeTelemetry to get gfe_latency and
+   *     gfe_header_missing_count metrics.
    */
   @VisibleForTesting
+  @ObsoleteApi(
+      "The OpenCensus project is deprecated. Use OpeTelemetry to get gfe_latency and gfe_header_missing_count metrics.")
```
olavloite, line 101: nit (here and elsewhere):

```suggestion
      "The OpenCensus project is deprecated. Use OpenTelemetry to get gfe_latency and gfe_header_missing_count metrics.")
```

README.md
```diff
@@ -158,6 +158,9 @@ the Cloud Spanner Java client.
 
 ## OpenCensus Metrics
```
olavloite, line 159: Will we be adding instructions for how to use `OpenTelemetry` in a separate PR?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
surbhigarg92: Its a public method, changing this would result in breaking changes. So instead of changing I created a new overload with `ISpan` , please recommend if anything can be done here ?

olavloite: Yeah, that is a Java-strangeness. All methods in an interface are considered public. But if the interface itself is not public, then those methods are still not accessible from outside the package, even though they are marked as public. And in this case the interface `SessionTransaction` is package-private, which again means that we can safely change the API.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TraceWrapper.java
surbhigarg92: Whenever we do `startSpan` the newly created span gets set to the context. Here we are setting span in current context as the parent span of the newly cerated span.

README.md
surbhigarg92: Yes, I was planing to create a separate PR for documentation. We would be writing a blog as well, the link will also need to be added here. We might also need to add a sample for the same. 
So was planning to cover it in a separate PR.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
olavloite: Oh, that's an interesting find. `PartitionedDmlTransaction` should never have been `public` in the first place... AFAICT, this is also the only public implementation of this interface. And changing it to package-private works for the rest of the client library, so that is probably something that we should consider doing in a separate PR.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/ISpan.java
surbhigarg92: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionImpl.java
surbhigarg92: done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
surbhigarg92: Done

README.md
olavloite: ack

olavloite (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerMetrics.java
```diff
@@ -0,0 +1,42 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.spanner.spi.v1;
+
+public class SpannerMetrics {
+
+  private static boolean sessionMetricsEnabled = false;
+  private static boolean rpcMetricsEnabled = false;
```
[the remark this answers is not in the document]
surbhigarg92: Yes. That is correct

PR #2618, opened 2023-09-11
test: add integration tests for large batch dml updates.

arpan14:
There are additional optimisations added to support batch dml with a large batch size. These tests are added as part of an outage action item to test batch update against all environment.

### DIFF
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITBatchDmlTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITBatchDmlTest.java
@@ -45,16 +45,13 @@ import org.junit.experimental.categories.Category;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
 
-/**
- * Integration tests for DML.
- */
+/** Integration tests for DML. */
 @Category(ParallelIntegrationTest.class)
 @RunWith(JUnit4.class)
 public final class ITBatchDmlTest {
 
   private static Database db;
-  @ClassRule
-  public static IntegrationTestEnv env = new IntegrationTestEnv();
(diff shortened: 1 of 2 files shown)


google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITBatchDmlTest.java
```diff
@@ -189,4 +193,47 @@ public void errorBatchDmlAlreadyExist() {
...
+    }
+
+    for (int i = 0; i < 30; i++) {
+      stmts.add(Statement.of("DELETE FROM T WHERE T.K = 'boo" + i + "';"));
+    }
+
+    final TransactionCallable<long[]> callable = transaction -> transaction.batchUpdate(stmts);
+    TransactionRunner runner = client.readWriteTransaction();
+    long[] rowCounts = runner.run(callable);
+    assertThat(rowCounts.length).isEqualTo(10);
```
olavloite, line 211: It surprises me if this actually succeeds. The length should be 70. The fact that the last 30 statements delete 30 of the 40 rows that were inserted should still be counted as a statement that is executed, and the update count of each statement should be 1.

nit: prefer `assertEquals` over `assertThat`

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITBatchDmlTest.java
```diff
@@ -189,4 +193,47 @@ public void errorBatchDmlAlreadyExist() {
...
+  }
```
olavloite, line 212: Could we also add an assertion for the update count, and not only the length of the update count array?

olavloite (approved):

PR #2624, opened 2023-09-14
feat: Add support for Managed Autoscaler

claire921:
- Add the ability to create / update an instance with autoscaling config
  - Note that now user can only specify one of the node_count, processing_units or autoscaling_config as the compute capacity for an instance.
- Add the ability to see autoscaling config with an instance

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceAdminClientImpl.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceAdminClientImpl.java
@@ -193,9 +193,18 @@ class InstanceAdminClientImpl implements InstanceAdminClient {
   @Override
   public OperationFuture<Instance, CreateInstanceMetadata> createInstance(InstanceInfo instance)
       throws SpannerException {
+    boolean valid_capacity_with_autoscaling =
+        instance.getAutoscalingConfig() != null
+            && instance.getNodeCount() == 0
+            && instance.getProcessingUnits() == 0;
+    boolean valid_capacity_without_autoscaling =
+        (instance.getAutoscalingConfig() == null)
+            && (instance.getNodeCount() == 0 || instance.getProcessingUnits() == 0);
+
     Preconditions.checkArgument(
-        instance.getNodeCount() == 0 || instance.getProcessingUnits() == 0,
-        "Only one of nodeCount and processingUnits can be set when creating a new instance");
+        valid_capacity_with_autoscaling || valid_capacity_without_autoscaling,
+        "Only one of nodeCount, processingUnits or autoscalingConfig can be set when creating a new instance");
+
     String projectName = PROJECT_NAME_TEMPLATE.instantiate("project", projectId);
     OperationFuture<com.google.spanner.admin.instance.v1.Instance, CreateInstanceMetadata>
         rawOperationFuture =
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceAdminClientImpl.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceAdminClientImpl.java
@@ -193,18 +193,6 @@ class InstanceAdminClientImpl implements InstanceAdminClient {
   @Override
   public OperationFuture<Instance, CreateInstanceMetadata> createInstance(InstanceInfo instance)
       throws SpannerException {
-    boolean valid_capacity_with_autoscaling =
-        instance.getAutoscalingConfig() != null
-            && instance.getNodeCount() == 0
-            && instance.getProcessingUnits() == 0;
-    boolean valid_capacity_without_autoscaling =
-        (instance.getAutoscalingConfig() == null)
-            && (instance.getNodeCount() == 0 || instance.getProcessingUnits() == 0);
-
-    Preconditions.checkArgument(
-        valid_capacity_with_autoscaling || valid_capacity_without_autoscaling,
-        "Only one of nodeCount, processingUnits or autoscalingConfig can be set when creating a new instance");
-
     String projectName = PROJECT_NAME_TEMPLATE.instantiate("project", projectId);
     OperationFuture<com.google.spanner.admin.instance.v1.Instance, CreateInstanceMetadata>
         rawOperationFuture =
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceInfo.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceInfo.java
@@ -112,9 +112,7 @@ public class InstanceInfo {
      * instance. Exactly one of processing units, node count, or autoscaling config must be set when
      * creating a new instance.
      */
-    public Builder setAutoscalingConfig(AutoscalingConfig autoscalingConfig) {
-      throw new UnsupportedOperationException("Unimplemented");
-    }
+    public abstract Builder setAutoscalingConfig(AutoscalingConfig autoscalingConfig);
 
     public abstract Builder setState(State state);
 
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceInfo.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceInfo.java
@@ -112,7 +112,9 @@ public class InstanceInfo {
      * instance. Exactly one of processing units, node count, or autoscaling config must be set when
      * creating a new instance.
      */
-    public abstract Builder setAutoscalingConfig(AutoscalingConfig autoscalingConfig);
+    public Builder setAutoscalingConfig(AutoscalingConfig autoscalingConfig) {
+      throw new UnsupportedOperationException("Unimplemented");
+    }
 
     public abstract Builder setState(State state);
 
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceInfo.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceInfo.java
@@ -23,6 +23,7 @@ import com.google.cloud.Timestamp;
 import com.google.common.base.MoreObjects;
 import com.google.common.collect.ImmutableMap;
 import com.google.protobuf.FieldMask;
+import com.google.spanner.admin.instance.v1.AutoscalingConfig;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
@@ -35,13 +36,16 @@ public class InstanceInfo {
     DISPLAY_NAME("display_name"),
     NODE_COUNT("node_count"),
     PROCESSING_UNITS("processing_units"),
+    AUTOSCALING_CONFIG("autoscaling_config"),
     LABELS("labels");
 
     static InstanceField[] defaultFieldsToUpdate(InstanceInfo info) {
-      if (info.getNodeCount() > 0) {
-        return new InstanceField[] {DISPLAY_NAME, NODE_COUNT, LABELS};
+      if (info.getAutoscalingConfig() != null) {
+        return new InstanceField[] {DISPLAY_NAME, AUTOSCALING_CONFIG, LABELS};
+      } else if (info.getNodeCount() > 0) {
+        return new InstanceField[] {DISPLAY_NAME, AUTOSCALING_CONFIG, NODE_COUNT, LABELS};
       } else {
-        return new InstanceField[] {DISPLAY_NAME, PROCESSING_UNITS, LABELS};
+        return new InstanceField[] {DISPLAY_NAME, AUTOSCALING_CONFIG, PROCESSING_UNITS, LABELS};
       }
     }
 
@@ -87,18 +91,28 @@ public class InstanceInfo {
     }
 
     /**
-     * Sets the number of nodes for the instance. Exactly one of processing units or node count must
-     * be set when creating a new instance.
+     * Sets the number of nodes for the instance. Exactly one of processing units, node count or
+     * autoscaling config must be set when creating a new instance.
      */
     public abstract Builder setNodeCount(int nodeCount);
 
     /**
-     * Sets the number of processing units for the instance. Exactly one of processing units or node
-     * count must be set when creating a new instance. Processing units must be between 1 and 999
-     * (inclusive) when creating a new instance with node count = 0. Processing units from 1000 and
-     * up must always be a multiple of 1000 (that is equal to an integer number of nodes).
+     * Sets the number of processing units for the instance. Exactly one of processing units, node
+     * count, or autoscaling config must be set when creating a new instance. Processing units must
+     * be between 1 and 999 (inclusive) when creating a new instance with node count = 0. Processing
+     * units from 1000 and up must always be a multiple of 1000 (that is equal to an integer number
+     * of nodes).
      */
     public Builder setProcessingUnits(int processingUnits) {
+      throw new UnsupportedOperationException("Unimplemented");
+    }
+
+    /**
+     * Sets the autoscaling config for the instance, which will enable the autoscaling for this
+     * instance. Exactly one of processing units, node count, or autoscaling config must be set when
+     * creating a new instance.
+     */
+    public Builder setAutoscalingConfig(AutoscalingConfig autoscalingConfig) {
       throw new UnsupportedOperationException("Unimplemented");
     }
 
@@ -117,6 +131,7 @@ public class InstanceInfo {
     private String displayName;
     private int nodeCount;
     private int processingUnits;
+    private AutoscalingConfig autoscalingConfig;
     private State state;
     private Map<String, String> labels;
     private Timestamp updateTime;
@@ -133,6 +148,7 @@ public class InstanceInfo {
       this.displayName = instance.displayName;
       this.nodeCount = instance.nodeCount;
       this.processingUnits = instance.processingUnits;
+      this.autoscalingConfig = instance.autoscalingConfig;
       this.state = instance.state;
       this.labels = new HashMap<>(instance.labels);
       this.updateTime = instance.updateTime;
@@ -176,6 +192,12 @@ public class InstanceInfo {
     }
 
     @Override
+    public BuilderImpl setAutoscalingConfig(AutoscalingConfig autoscalingConfig) {
+      this.autoscalingConfig = autoscalingConfig;
+      return this;
+    }
+
+    @Override
     public BuilderImpl setState(State state) {
       this.state = state;
       return this;
@@ -204,6 +226,7 @@ public class InstanceInfo {
   private final String displayName;
   private final int nodeCount;
   private final int processingUnits;
+  private final AutoscalingConfig autoscalingConfig;
   private final State state;
   private final ImmutableMap<String, String> labels;
   private final Timestamp updateTime;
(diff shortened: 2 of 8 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceInfo.java
```diff
@@ -87,21 +91,31 @@ Builder setCreateTime(Timestamp createTime) {
...
     public Builder setProcessingUnits(int processingUnits) {
       throw new UnsupportedOperationException("Unimplemented");
     }
 
+    /**
+     * Sets the autoscaling config for the instance, which will enable the autoscaling for this
+     * instance. Exactly one of processing units, node count, or autoscaling config must be set when
+     * creating a new instance.
+     */
+    public Builder setAutoscalingConfig(AutoscalingConfig autoscalingConfig) {
```
arpan14, line 115: This method can be made abstract similar to `setNodeCount`

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceInfo.java
```diff
@@ -35,13 +36,16 @@ public enum InstanceField implements FieldSelector {
...
+    AUTOSCALING_CONFIG("autoscaling_config"),
     LABELS("labels");
 
     static InstanceField[] defaultFieldsToUpdate(InstanceInfo info) {
-      if (info.getNodeCount() > 0) {
-        return new InstanceField[] {DISPLAY_NAME, NODE_COUNT, LABELS};
+      if (info.getAutoscalingConfig() != null) {
+        return new InstanceField[] {DISPLAY_NAME, AUTOSCALING_CONFIG, LABELS};
+      } else if (info.getNodeCount() > 0) {
+        return new InstanceField[] {DISPLAY_NAME, AUTOSCALING_CONFIG, NODE_COUNT, LABELS};
```
arpan14, line 46: Why does this array contain  both `NODE_COUNT` and `AUTOSCALING_CONFIG` ? It should contain just `NODE_COUNT` ? Could you check if any unit tests broke here or this was a silent error?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceInfo.java
```diff
@@ -35,13 +36,16 @@ public enum InstanceField implements FieldSelector {
...
       } else {
-        return new InstanceField[] {DISPLAY_NAME, PROCESSING_UNITS, LABELS};
+        return new InstanceField[] {DISPLAY_NAME, AUTOSCALING_CONFIG, PROCESSING_UNITS, LABELS};
```
arpan14, line 48: Is it expected to contain both `AUTOSCALING_CONFIG` and  `PROCESSING_UNITS` ?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceAdminClientImpl.java
```diff
@@ -193,9 +193,18 @@ public Operation fromProto(Operation proto) {
   @Override
   public OperationFuture<Instance, CreateInstanceMetadata> createInstance(InstanceInfo instance)
       throws SpannerException {
+    boolean valid_capacity_with_autoscaling =
+        instance.getAutoscalingConfig() != null
```
arpan14, line 197: Is this validation just specific to `createInstance` RPC? What about other RPCs ? I am wondering if these validations are better suited when we invoke `setAutoscalingConfig` method than within RPC.

WDYT?

arpan14: Also, how are such validations supported in auto-generated languages (for ex - Golang) ?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceInfo.java
claire921: Same above.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceInfo.java
```diff
@@ -87,21 +91,31 @@ Builder setCreateTime(Timestamp createTime) {
     }
 
     /**
-     * Sets the number of nodes for the instance. Exactly one of processing units or node count must
-     * be set when creating a new instance.
+     * Sets the number of nodes for the instance. Exactly one of processing units, node count or
+     * autoscaling config must be set when creating a new instance.
```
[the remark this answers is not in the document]
claire921: > Can we update the description of the PR to contain such gotchas around the new configuration being introduced?

Added, PTAL

> Are there any validations that are present in client library today (before this change) where we ensure that only one of processing units or node count is set?

Yes, we verify that in the InstanceAdminClientImpl.java: https://github.com/googleapis/java-spanner/blob/main/google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceAdminClientImpl.java#L197-L198

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceInfo.java
claire921: Ah I see, updated. Should the `setProcessingUnits` also be abstract then?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceAdminClientImpl.java
arpan14: Sure, let's get rid of the validation all together and rely on the backend validation. Some simpler validation on primitive types can be considered to be moved to the protos. This in general is preferred since we are gradually moving in a direction on how we can simplify our handwritten clients (similar to our auto-generated clients). So, this would be a step in that direction.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceInfo.java
arpan14: There are a few edge cases and open question here:

1. when `getAutoscalingConfig != null`, we are updating just `AUTOSCALING_CONFIG`. Shouldn't we have checks to include `NODE_COUNT` and `PROCESSING_UNITS` (to be set to 0) ?
2. If `getNodeCount > 0`, we are updating just `NODE_COUNT`. What if an instance already had `PROCESSING_UNITS` ? I assume the backend will throw a validation error. Similar question for the `else` condition. 
3. If we are not marking `PROCESSING_UNITS` or `NODE_COUNT` as 0 in the second check, then why are we explicitly marking `AUTOSCALING_CONFIG` as null?

claire921: For #1, customers can only autoscaling_config once autoscaler is enabled. node_count and processing_units are output_only params. An instance can still have node_count and processing_units, it is just customers cannot modify them. This is consistent with CBT autoscaling API: https://screenshot.googleplex.com/6mi2r2QUKMfTk7x

For #2, this seems to be the logic before the autoscaler related changes. My understanding is that this method is to provide a default field mask when user did not specify one. I'm not sure why we did this in the first place -- as you pointed out, if an instance is specified with both nodes and processing units, this can be wrong. I'm following an existing pattern here.

For #3, if the autoscaling_config is not null, that means this is an autoscaler instance, and users should not specify processing_units or node_count as part of the field mask. If it is null, then the user intention will be to update a non-autoscaling instance.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/InstanceAdminClientImplTest.java
```diff
@@ -294,21 +314,23 @@ public void testCreateInstanceWithProcessingUnits() throws Exception {
...
+    when(rpc.createInstance("projects/" + PROJECT_ID, INSTANCE_ID, getAutoscalingInstanceProto()))
+        .thenReturn(rawOperationFuture);
+    OperationFuture<Instance, CreateInstanceMetadata> operation =
+        client.createInstance(
+            InstanceInfo.newBuilder(InstanceId.of(PROJECT_ID, INSTANCE_ID))
+                .setInstanceConfigId(InstanceConfigId.of(PROJECT_ID, CONFIG_ID))
+                .setAutoscalingConfig(getAutoscalingInstanceProto().getAutoscalingConfig())
+                .build());
+    assertTrue(operation.isDone());
+    assertEquals(INSTANCE_NAME, operation.get().getId().getName());
```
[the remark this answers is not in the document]
claire921: Added, PTAL

google-cloud-spanner/src/test/java/com/google/cloud/spanner/InstanceTest.java
```diff
@@ -87,13 +113,36 @@ public void buildInstance() {
...
+                    .setMaxNodes(100))
+            .build();
 
     Instance instance =
         new Instance.Builder(instanceClient, dbClient, id)
             .setInstanceConfigId(configId)
             .setDisplayName("test instance")
             .setNodeCount(1)
             .setProcessingUnits(2000)
+            .setAutoscalingConfig(autoscalingConfig1)
```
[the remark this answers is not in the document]
claire921: Good point, added a new test only for autoscaling config. PTAL.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITInstanceAdminTest.java
```diff
@@ -121,6 +116,51 @@ public void updateInstance() throws Exception {
...
+            .build();
+    OperationFuture<Instance, UpdateInstanceMetadata> op =
+        instanceClient.updateInstance(toUpdate, InstanceInfo.InstanceField.AUTOSCALING_CONFIG);
+    Instance newInstance = op.get();
+    assertThat(newInstance.getAutoscalingConfig()).isEqualTo(autoscalingConfig);
+
+    Instance newInstanceFromGet =
+        instanceClient.getInstance(env.getTestHelper().getInstanceId().getInstance());
+    assertThat(newInstanceFromGet).isEqualTo(newInstance);
+
```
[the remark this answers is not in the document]
claire921: Added a comment, PTAL.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITInstanceAdminTest.java
```diff
@@ -21,14 +21,9 @@
...
 
 import com.google.api.gax.longrunning.OperationFuture;
-import com.google.cloud.spanner.Instance;
-import com.google.cloud.spanner.InstanceAdminClient;
-import com.google.cloud.spanner.InstanceConfig;
-import com.google.cloud.spanner.InstanceInfo;
-import com.google.cloud.spanner.IntegrationTestEnv;
-import com.google.cloud.spanner.Options;
-import com.google.cloud.spanner.ParallelIntegrationTest;
+import com.google.cloud.spanner.*;
```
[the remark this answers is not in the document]
claire921: Reverted.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/InstanceAdminClientImplTest.java
```diff
@@ -294,21 +314,26 @@ public void testCreateInstanceWithProcessingUnits() throws Exception {
   }
 
   @Test
-  public void testCreateInstanceWithBothNodeCountAndProcessingUnits() throws Exception {
```
arpan14, line 297: Are we renaming or removing this test `testCreateInstanceWithBothNodeCountAndProcessingUnits` ? If yes, why?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/InstanceAdminClientImplTest.java
```diff
@@ -294,21 +314,26 @@ public void testCreateInstanceWithProcessingUnits() throws Exception {
...
+                CreateInstanceMetadata.getDefaultInstance());
+    when(rpc.createInstance("projects/" + PROJECT_ID, INSTANCE_ID, getAutoscalingInstanceProto()))
+        .thenReturn(rawOperationFuture);
+    OperationFuture<Instance, CreateInstanceMetadata> operation =
+        client.createInstance(
+            InstanceInfo.newBuilder(InstanceId.of(PROJECT_ID, INSTANCE_ID))
+                .setInstanceConfigId(InstanceConfigId.of(PROJECT_ID, CONFIG_ID))
+                .setAutoscalingConfig(getAutoscalingInstanceProto().getAutoscalingConfig())
+                .build());
+    assertTrue(operation.isDone());
```
arpan14, line 332: Nit: store operation.get() in a local object say `expectedOutput` and then use it in every `assertEquals` statement instead of doing operation.get() repeatedly.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/InstanceAdminClientImplTest.java
```diff
@@ -364,22 +398,119 @@ public void testUpdateInstanceProcessingUnits() throws Exception {
...
+            .build();
+    OperationFuture<Instance, UpdateInstanceMetadata> operationWithFieldMask =
+        client.updateInstance(
+            instanceInfo,
+            InstanceInfo.InstanceField.AUTOSCALING_CONFIG,
+            InstanceInfo.InstanceField.PROCESSING_UNITS);
+    assertTrue(operationWithFieldMask.isDone());
+    assertEquals(INSTANCE_NAME, operationWithFieldMask.get().getId().getName());
+
+    when(rpc.updateInstance(
```
arpan14, line 421: * Refactor this to a new test. A single test should have a single outcome. Also, let's not modify the behaviour of any existing tests.
* In general I see a lot of modifications to existing tests where we are adding a new `when().then()` mocking and then doing an additional `update`. Can you describe why we are doing it?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/InstanceAdminClientImplTest.java
claire921: This test was disabled since we removed the check in google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceAdminClientImpl.java: ``` Preconditions.checkArgument(
        instance.getNodeCount() == 0 || instance.getProcessingUnits() == 0,
        "Only one of nodeCount and processingUnits can be set when creating a new instance"); ```

Per earlier conversation, we decided to remove the check to reduce the hand-written layer.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/InstanceAdminClientImplTest.java
claire921: ```Refactor this to a new test. A single test should have a single outcome. Also, let's not modify the behaviour of any existing tests. ```
Yeah sorry this was for my local test -- reverted.

``` In general I see a lot of modifications to existing tests where we are adding a new when().then() mocking and then doing an additional update. Can you describe why we are doing it? ```
This pattern seems to be there before: https://screenshot.googleplex.com/VzKP68q9X5CZSjq ? I was following the same pattern.

File: google-cloud-spanner/src/test/java/com/google/cloud/spanner/InstanceAdminClientImplTest.java (hunk 10 above)
Comment (arpan14) [reply google-cloud-spanner/src/test/java/com/google/cloud/spanner/InstanceAdminClientImplTest.java:297]: Ah got it.

File: google-cloud-spanner/src/test/java/com/google/cloud/spanner/InstanceAdminClientImplTest.java [hunk 13]
```diff
@@ -408,14 +531,53 @@ public void testUpdateInstanceWithNodeCountAndProcessingUnits() throws Exception
...
         .thenReturn(rawOperationFuture);
     InstanceInfo instanceInfo =
         InstanceInfo.newBuilder(InstanceId.of(INSTANCE_NAME))
             .setInstanceConfigId(InstanceConfigId.of(CONFIG_NAME))
             .setNodeCount(3)
             .setProcessingUnits(3000)
+            .setAutoscalingConfig(getAutoscalingConfigProto())
             .build();
     OperationFuture<Instance, UpdateInstanceMetadata> operationWithFieldMask =
         client.updateInstance(instanceInfo);
```
arpan14, line 583: * Can we add more assertions on what we expect the node count, processing unit and autoscaling config to contain, similar to how we are asserting `INSTANCE_NAME` ?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/InstanceAdminClientImplTest.java
```diff
@@ -408,14 +531,53 @@ public void testUpdateInstanceWithNodeCountAndProcessingUnits() throws Exception
...
+    InstanceInfo instanceInfo =
+        InstanceInfo.newBuilder(InstanceId.of(INSTANCE_NAME))
+            .setInstanceConfigId(InstanceConfigId.of(CONFIG_NAME))
+            .setNodeCount(3)
+            .setProcessingUnits(3000)
+            .build();
+    OperationFuture<Instance, UpdateInstanceMetadata> operationWithFieldMask =
+        client.updateInstance(instanceInfo);
+    assertTrue(operationWithFieldMask.isDone());
+    assertEquals(INSTANCE_NAME, operationWithFieldMask.get().getId().getName());
```
arpan14, line 547: Could you add more assertions on fields being updated?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/InstanceAdminClientImplTest.java
```diff
@@ -379,7 +413,96 @@ public void testUpdateInstanceProcessingUnits() throws Exception {
...
-                .addAllPaths(Arrays.asList("display_name", "processing_units", "labels"))
+                .addAllPaths(
+                    Arrays.asList(
+                        "display_name", "autoscaling_config", "processing_units", "labels"))
+                .build()))
+        .thenReturn(rawOperationFuture);
+    OperationFuture<Instance, UpdateInstanceMetadata> operation =
+        client.updateInstance(instanceInfo);
+    assertTrue(operation.isDone());
+    assertEquals(INSTANCE_NAME, operation.get().getId().getName());
```
arpan14, line 424: Can we add few more assertions on value of nodecount, processing units and autoscaling config. Applicable at other tests as well.

arpan14 (approved):

google-cloud-spanner/src/test/java/com/google/cloud/spanner/InstanceAdminClientImplTest.java
claire921: The assertions are probably not needed, because we mock the RPC to return an instance only with processing_units: https://screenshot.googleplex.com/7fF5jTJZ8hbQ2nn. So if we add assertions, the assertions will only be to test the mocking of the RPC, instead of the actual code.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/InstanceAdminClientImplTest.java
claire921: Same above

google-cloud-spanner/src/test/java/com/google/cloud/spanner/InstanceAdminClientImplTest.java
claire921: Same above.

PR #2630, opened 2023-09-17
docs: improve timeout and retry sample

olavloite:
The timeout and retry sample used 60 seconds for both initial, max, and total timeout. It also used a 1.0 multiplier for the timeout value. This made it impossible to explain how to set an increasing RPC timeout value. It also rendered adding the DEADLINE_EXCEEDED error code as a retryable code superfluous.

### DIFF
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
@@ -3419,6 +3419,48 @@ public class DatabaseClientImplTest {
     }
   }
 
+  @Test
+  public void testSampleRetrySettings() {
+    String sql =
+        "INSERT INTO Singers (SingerId, FirstName, LastName)\n"
+            + "VALUES (20, 'George', 'Washington')";
+    mockSpanner.putStatementResult(StatementResult.update(Statement.of(sql), 1L));
+
+    SpannerOptions.Builder builder =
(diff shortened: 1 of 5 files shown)


google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
```diff
@@ -3419,6 +3419,48 @@ public void testRetryOnResourceExhausted() {
     }
   }
 
+  @Test
+  public void testSampleRetrySettings() {
```
arpan14, line 3423: Can we carve it into a separate test class similar to what we do for most other samples code? It improves the discoverability of the test.

arpan14 (approved):

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
arpan14: Nit: Consider adding the new test in a separate class. `google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java` is already overloaded for testing too many things.

olavloite: Moved the sample test to a separate class (and created an abstract base class for creating mock server tests, so it's easier to set up more of those test classes).

PR #2632, opened 2023-09-18
feat: support setting core pool size for async API in system property

### LINKED ISSUE #2631: Set number of threads for async executor provider as a system property
The `ExecutorProvider` that is used for the Async API in the client library can be set using the `SpannerOptions$Builder#setAsyncExecutorProvider` [method](https://github.com/googleapis/java-spanner/blob/5ebf067a249acad249e36ef7cb51fa0009ac72f7/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java#L1111). This makes it possible for direct users of the Async API to determine the type of executor and the number of threads to use. This configuration is however not accessible to anyone using an abstraction on top of the client library, such as the [Spanner R2DBC driver](https://github.com/GoogleCloudPlatform/cloud-spanner-r2dbc), unless that abstraction has also added support for this configuration property to their own API.

When no `ExecutorProvider` has been set, the Java client library will use a default 8-thread `Executor`. The '8 threads' is currently hard-coded. We should change this hard-coded value into one that can be set with a system property, so users of both the client library and any driver that builds on top of it can set a different default.

olavloite:
Add support for setting the core pool size for the thread pool that is used for the Async API using a system property. This allows users to change this property without having to change code, and to set the property if they are using a framework that builds on top of the client library, but that does not have a configuration option for setting a custom executor provider.

Fixes #2631


google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerOptionsTest.java
```diff
@@ -913,6 +914,49 @@ public void testCustomAsyncExecutorProvider() {
     assertSame(service, options.getAsyncExecutorProvider().getExecutor());
   }
 
+  @Test
+  public void testAsyncExecutorProviderCoreThreadCount() throws Exception {
+    assertEquals(8, SpannerOptions.getDefaultAsyncExecutorProviderCoreThreadCount());
```
arpan14, line 919: nit: Should we add a check to ensure that the system property was originally unset?

arpan14: I see that there is a handling of this scenario within the `finally` block of `runWithSystemProperty` method, but thought if explicitly setting null would be better to ensure that the default behaviour is due to null property and not because somehow the property was already set to 8.

arpan14 (approved):

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerOptionsTest.java
olavloite: I added a test for both `null` and an empty string.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -524,7 +524,30 @@ public static FixedCloseableExecutorProvider create(ScheduledExecutorService exe
...
   @VisibleForTesting
   static CloseableExecutorProvider createDefaultAsyncExecutorProvider() {
-    return createAsyncExecutorProvider(8, 60L, TimeUnit.SECONDS);
+    return createAsyncExecutorProvider(
+        getDefaultAsyncExecutorProviderCoreThreadCount(), 60L, TimeUnit.SECONDS);
+  }
+
+  @VisibleForTesting
+  static int getDefaultAsyncExecutorProviderCoreThreadCount() {
+    String propertyName = "SPANNER_ASYNC_NUM_CORE_THREADS";
```
psinghbay1, line 533: magic string!

should we maintain list of constants/enums property names?

PR #2603, opened 2023-08-25
feat: allow non-default service accounts

ko3a4ok:

PR #2635, opened 2023-09-19
chore: allow non-default service account for DP

olavloite:
By default, direct path will only be attempted if the application uses the default service account of the Google Compute Engine that it is running on. This pull request sets the option to allow the use of a non-default service account in combination with direct path. This is necessary both for our tests, that use a separate service account than the default service account of the GCE instance that the test is running on, and for end users who might want to use a separate service account for their application.

The pull request additionally ensures that direct path is only attempted in the following cases:
1. The credential that is being used is not an instance of `NoCredentials`. The latter is used for plain text connections to the emulator and to the in-memory mock Spanner server that is used for testing.
2. The `disableDirectPath()` option has not been set. This option is used for some specific tests that run against the in-memory Spanner mock server, and that use (test) custom `CredentialProvider` implementations. The option is also an escape hatch for applications that for whatever reason would want to disable direct path. The option is marked as a `@BetaApi`, so we have the option of removing it in the future. The method cannot be package-private, as the tests that rely on it live in a different package.

Replaces #2603


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1220,6 +1225,12 @@ public Builder disableLeaderAwareRouting() {
       return this;
     }
 
+    @BetaApi
+    public Builder disableDirectPath() {
```
arpan14, line 1229: * Instead of a method, wouldn't it be better to use an internal system property? AFAIU, we only require this for our test infrastructure right now. 
* Also, my understanding is customers would never want to disable directpath, given they don't need to understand about internal proxy connections between GCP and Borg. If yes, that also pushes us to not expose a public method and instead use a system property?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -339,9 +340,16 @@ public GapicSpannerRpc(final SpannerOptions options) {
...
               // Attempts direct access to spanner service over gRPC to improve throughput,
               // whether the attempt is allowed is totally controlled by service owner.
-              .setAttemptDirectPath(true);
+              // We'll only attempt DirectPath if we are using real credentials.
+              // NoCredentials is used for plain text connections, for example when connecting to
+              // the emulator.
+              .setAttemptDirectPath(
+                  options.isAttemptDirectPath()
+                      && !Objects.equals(
+                          options.getScopedCredentials(), NoCredentials.getInstance()));
```
arpan14, line 352: `!Objects.equals(
                          options.getScopedCredentials(), NoCredentials.getInstance())`

This sounds more like a identification strategy for emulator. Can't we use the env variable used to identify emulator instead?

olavloite: We need this for more than just the emulator. This is also used for the in-memory Spanner mock server tests that also use `NoCredentials`. Also; It is possible to connect to the emulator without setting the environment variable. You can also just directly set the host name.

arpan14 (approved):

ko3a4ok:
thanks a lot for the fix!

PR #2641, opened 2023-09-26
feat: add session pool option for modelling a timeout around session acquisition.

arpan14:
Requests are made to wait till a session is available. We are introducing a session pool option `acquireSessionTimeout` and a public method `setAcquireSessionTimeout` in the `SessionPoolOptions` class that will allow to specify a timeout value for requests. Post this timeout application will receive a `RESOURCE_EXHAUSTED` error. If this option is not passed, the client library will wait for 60s by default.

* Sample Usage
```
SessionPoolOptions sessionPoolOptions =
        SessionPoolOptions.newBuilder().setAcquireSessionTimeout(Duration.ofMillis(60)).build();
SpannerOptions.Builder builder =
        SpannerOptions.newBuilder()
            .setProjectId("my-project")
            .setSessionPoolOption(sessionPoolOptions);
```

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
@@ -1665,8 +1665,6 @@ class SessionPool {
         try (Scope waitScope = tracer.withSpan(span)) {
           PooledSession s =
               pollUninterruptiblyWithTimeout(currentTimeout, options.getAcquireSessionTimeout());
-          // TODO clean up this code since acquireSessionTimeout will never be null and hence
-          // we won't have a use-case when s == null
           if (s == null) {
             // Set the status to DEADLINE_EXCEEDED and retry.
             numWaiterTimeouts.incrementAndGet();
@@ -1695,8 +1693,6 @@ class SessionPool {
       try {
         while (true) {
           try {
-            // TODO refactor this code since eventually acquireSessionTimeout will always have a
-            // default value and won't be null
             return acquireSessionTimeout == null
                 ? waiter.get(timeoutMillis, TimeUnit.MILLISECONDS)
                 : waiter.get(acquireSessionTimeout.toMillis(), TimeUnit.MILLISECONDS);
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
@@ -1708,7 +1708,9 @@ class SessionPool {
                   ErrorCode.RESOURCE_EXHAUSTED,
                   "Timed out after waiting "
                       + acquireSessionTimeout.toMillis()
-                      + "ms for acquiring session.");
+                      + "ms for acquiring session. To mitigate error increase the timeout duration at"
+                      + " session pool option `acquireSessionTimeout` or increase the "
+                      + "number of sessions in the session pool.");
             }
             return null;
           } catch (ExecutionException e) {
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
@@ -63,7 +63,6 @@ import com.google.common.base.Preconditions;
 import com.google.common.base.Supplier;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.ImmutableSet;
 import com.google.common.util.concurrent.ForwardingListenableFuture;
 import com.google.common.util.concurrent.ForwardingListenableFuture.SimpleForwardingListenableFuture;
 import com.google.common.util.concurrent.ListenableFuture;
@@ -120,17 +119,6 @@ class SessionPool {
   private static final Logger logger = Logger.getLogger(SessionPool.class.getName());
   private static final Tracer tracer = Tracing.getTracer();
   static final String WAIT_FOR_SESSION = "SessionPool.WaitForSession";
-  static final ImmutableSet<ErrorCode> SHOULD_STOP_PREPARE_SESSIONS_ERROR_CODES =
-      ImmutableSet.of(
-          ErrorCode.UNKNOWN,
-          ErrorCode.INVALID_ARGUMENT,
-          ErrorCode.PERMISSION_DENIED,
-          ErrorCode.UNAUTHENTICATED,
-          ErrorCode.RESOURCE_EXHAUSTED,
-          ErrorCode.FAILED_PRECONDITION,
-          ErrorCode.OUT_OF_RANGE,
-          ErrorCode.UNIMPLEMENTED,
-          ErrorCode.INTERNAL);
 
   /**
    * If the {@link SessionPoolOptions#getWaitForMinSessions()} duration is greater than zero, waits
@@ -1675,7 +1663,10 @@ class SessionPool {
       while (true) {
         Span span = tracer.spanBuilder(WAIT_FOR_SESSION).startSpan();
         try (Scope waitScope = tracer.withSpan(span)) {
-          PooledSession s = pollUninterruptiblyWithTimeout(currentTimeout);
+          PooledSession s =
+              pollUninterruptiblyWithTimeout(currentTimeout, options.getAcquireSessionTimeout());
+          // TODO clean up this code since acquireSessionTimeout will never be null and hence
+          // we won't have a use-case when s == null
           if (s == null) {
             // Set the status to DEADLINE_EXCEEDED and retry.
             numWaiterTimeouts.incrementAndGet();
@@ -1685,6 +1676,11 @@ class SessionPool {
             return s;
           }
         } catch (Exception e) {
+          if (e instanceof SpannerException
+              && ErrorCode.RESOURCE_EXHAUSTED.equals(((SpannerException) e).getErrorCode())) {
+            numWaiterTimeouts.incrementAndGet();
+            tracer.getCurrentSpan().setStatus(Status.RESOURCE_EXHAUSTED);
+          }
           TraceUtil.setWithFailure(span, e);
           throw e;
         } finally {
@@ -1693,15 +1689,27 @@ class SessionPool {
       }
     }
 
-    private PooledSession pollUninterruptiblyWithTimeout(long timeoutMillis) {
+    private PooledSession pollUninterruptiblyWithTimeout(
+        long timeoutMillis, Duration acquireSessionTimeout) {
       boolean interrupted = false;
       try {
         while (true) {
           try {
-            return waiter.get(timeoutMillis, TimeUnit.MILLISECONDS);
+            // TODO refactor this code since eventually acquireSessionTimeout will always have a
+            // default value and won't be null
+            return acquireSessionTimeout == null
+                ? waiter.get(timeoutMillis, TimeUnit.MILLISECONDS)
+                : waiter.get(acquireSessionTimeout.toMillis(), TimeUnit.MILLISECONDS);
           } catch (InterruptedException e) {
             interrupted = true;
           } catch (TimeoutException e) {
+            if (acquireSessionTimeout != null) {
+              throw SpannerExceptionFactory.newSpannerException(
+                  ErrorCode.RESOURCE_EXHAUSTED,
+                  "Timed out after waiting "
+                      + acquireSessionTimeout.toMillis()
+                      + "ms for acquiring session.");
+            }
             return null;
           } catch (ExecutionException e) {
             throw SpannerExceptionFactory.newSpannerException(e.getCause());
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
@@ -1704,9 +1704,8 @@ class SessionPool {
                   ErrorCode.RESOURCE_EXHAUSTED,
                   "Timed out after waiting "
                       + acquireSessionTimeout.toMillis()
-                      + "ms for acquiring session. To mitigate error increase the timeout duration at"
-                      + " session pool option `acquireSessionTimeout` or increase the "
-                      + "number of sessions in the session pool.");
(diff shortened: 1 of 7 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1693,15 +1689,27 @@ public PooledSession get() {
...
       boolean interrupted = false;
       try {
         while (true) {
           try {
-            return waiter.get(timeoutMillis, TimeUnit.MILLISECONDS);
+            // TODO refactor this code since eventually acquireSessionTimeout will always have a
+            // default value and won't be null
+            return acquireSessionTimeout == null
+                ? waiter.get(timeoutMillis, TimeUnit.MILLISECONDS)
+                : waiter.get(acquireSessionTimeout.toMillis(), TimeUnit.MILLISECONDS);
```
olavloite, line 1702: We should add a check to the `SpannerOptions$Builder#setAcquireSessionTimeout` method that ensures that no one ever sets a timeout that is bigger than can be converted to milliseconds. Otherwise, this method could return an `ArithmeticException`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1693,15 +1689,27 @@ public PooledSession get() {
...
 
-    private PooledSession pollUninterruptiblyWithTimeout(long timeoutMillis) {
+    private PooledSession pollUninterruptiblyWithTimeout(
+        long timeoutMillis, Duration acquireSessionTimeout) {
       boolean interrupted = false;
       try {
         while (true) {
           try {
-            return waiter.get(timeoutMillis, TimeUnit.MILLISECONDS);
+            // TODO refactor this code since eventually acquireSessionTimeout will always have a
```
olavloite, line 1698: Why eventually? Are we not setting a default non-null value already in this PR?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1675,7 +1663,10 @@ public PooledSession get() {
       while (true) {
         Span span = tracer.spanBuilder(WAIT_FOR_SESSION).startSpan();
         try (Scope waitScope = tracer.withSpan(span)) {
-          PooledSession s = pollUninterruptiblyWithTimeout(currentTimeout);
+          PooledSession s =
+              pollUninterruptiblyWithTimeout(currentTimeout, options.getAcquireSessionTimeout());
+          // TODO clean up this code since acquireSessionTimeout will never be null and hence
```
olavloite, line 1668: Is there any reason that we cannot do the cleanup in this PR already? I don't think that will make this PR unreasonably big.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -120,17 +119,6 @@ class SessionPool {
   private static final Logger logger = Logger.getLogger(SessionPool.class.getName());
   private static final Tracer tracer = Tracing.getTracer();
   static final String WAIT_FOR_SESSION = "SessionPool.WaitForSession";
-  static final ImmutableSet<ErrorCode> SHOULD_STOP_PREPARE_SESSIONS_ERROR_CODES =
```
olavloite, line 123: I assume that this is an unrelated cleanup of code that was already no longer in use, no?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1693,15 +1689,27 @@ public PooledSession get() {
...
+            return acquireSessionTimeout == null
+                ? waiter.get(timeoutMillis, TimeUnit.MILLISECONDS)
+                : waiter.get(acquireSessionTimeout.toMillis(), TimeUnit.MILLISECONDS);
           } catch (InterruptedException e) {
             interrupted = true;
           } catch (TimeoutException e) {
+            if (acquireSessionTimeout != null) {
+              throw SpannerExceptionFactory.newSpannerException(
+                  ErrorCode.RESOURCE_EXHAUSTED,
+                  "Timed out after waiting "
```
olavloite, line 1709: nit: We should add a hint to this error message to instruct the user how they can increase the timeout and/or increase the number of sessions in the pool, if they feel that this should have worked.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -538,6 +549,9 @@ public Builder setFailIfPoolExhausted() {
     /**
      * If all sessions are in use and there is no more room for creating new sessions, block for a
      * session to become available. Default behavior is same.
+     *
+     * <p>By default the requests are blocked for 60s. If we wish to block for a different period
```
olavloite, line 553:
```suggestion
     * <p>By default the requests are blocked for 60s and will fail with a `SpannerException` with error code `ResourceExhausted` if this timeout is exceeded. If you wish to block for a different period
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -695,6 +709,15 @@ public Builder setWaitForMinSessions(Duration waitForMinSessions) {
       return this;
     }
 
+    /**
+     * If greater than zero, we wait for said duration when no sessions are available in the {@link
+     * SessionPool}. When no configuration is passed, we default to 60s to acquire a session.
```
olavloite, line 714:
```suggestion
     * SessionPool}. When no configuration is passed, the default is a 60s timeout.
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -695,6 +709,15 @@ public Builder setWaitForMinSessions(Duration waitForMinSessions) {
...
+     */
+    public Builder setAcquireSessionTimeout(Duration acquireSessionTimeout) {
+      this.acquireSessionTimeout = acquireSessionTimeout;
```
olavloite, line 717:
```suggestion
      this.acquireSessionTimeout = Preconditions.checkNotNull(acquireSessionTimeout);
```

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
```diff
@@ -2207,6 +2211,107 @@ public void testBatchCreateSessionsPermissionDenied() {
...
+                  // the application.
+                  SpannerException e = assertThrows(SpannerException.class, rs::next);
+                  assertEquals(ErrorCode.DEADLINE_EXCEEDED, e.getErrorCode());
+                  System.out.printf("finished test %d\n", counter.incrementAndGet());
+
+                  return null;
+                }));
+      }
+      service.shutdown();
+      assertEquals(5000, Futures.allAsList(futures).get().size());
```
olavloite, line 2258: This seems like a very heavy test. It's good that we have tested it in this way, but do we want to run this during every build?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
```diff
@@ -2207,6 +2211,107 @@ public void testBatchCreateSessionsPermissionDenied() {
...
+                  assertEquals(ErrorCode.RESOURCE_EXHAUSTED, e.getErrorCode());
+                  System.out.printf("finished test %d\n", counter.incrementAndGet());
+
+                  return null;
+                }));
+      }
+      service.shutdown();
+      assertEquals(5000, Futures.allAsList(futures).get().size());
+    }
+  }
```
olavloite, line 2313: Same here as above: do we want to run this during every build?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
```diff
@@ -2207,6 +2211,107 @@ public void testBatchCreateSessionsPermissionDenied() {
...
+
```
olavloite, line 2314: One thing that we are not testing here, but that is the most likely scenario to happen, is the case where:
1. `BatchCreateSessions` works as expected (that is; fast).
2. The session pool is exhausted because all the sessions have been checked out (and maybe leaked).
3. A request for another session comes in and then fails because no sessions become available.

Could we add a test for that as well. The simplest way is probably to create a session pool with just 1 or 2 sessions, and then check those sessions out by starting for example a read-only transaction and not closing it.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolOptionsTest.java
```diff
@@ -186,4 +186,19 @@ public void setNegativeIdleTimeThreshold() {
     SessionPoolOptions.newBuilder()
         .setInactiveTransactionRemovalOptions(inactiveTransactionRemovalOptions);
   }
+
+  @Test
+  public void setAcquireSessionTimeout() {
+    SessionPoolOptions sessionPoolOptions =
```
olavloite, line 192: We should add some checks for the corner case timeout values and also verify that we can get it as milliseconds (or that it fails in the correct way):
1. `null` timeout
2. `Duration.ofMillis(0)`
3. `Duration.ofSeconds(Long.MAX_VALUE)`

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Yes this is unrelated dead-code.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
arpan14: * Refactored this test to a new class `BatchCreateSessionsSlowTest` and annotated the class with `@Category(SlowTest.class)`. This will ensure this does not get triggered for each build.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolOptionsTest.java
arpan14: * Sure there was an existing code path that was testing a similar behaviour. I have used that as a reference to add a new test `blockAndTimeoutOnPoolExhaustion_withAcquireSessionTimeout`. 
* For the existing test `blockAndTimeoutOnPoolExhaustion`, I had to explicitly set `acquireSessionTimeout` value as null since otherwise it was going into the new code path and thus failing. This kind of demonstrates how this change is a breaking change and how the customers can mitigate by setting it to `null`.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
arpan14: * Sure there was an existing code path that was testing a similar behaviour. I have used that as a reference to add a new test `blockAndTimeoutOnPoolExhaustion_withAcquireSessionTimeout`. 
* For the existing test `blockAndTimeoutOnPoolExhaustion`, I had to explicitly set `acquireSessionTimeout` value as null since otherwise it was going into the new code path and thus failing. This kind of demonstrates how this change is a breaking change and how the customers can mitigate by setting it to `null`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Added the hint

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1693,15 +1687,27 @@ public PooledSession get() {
...
+                      + acquireSessionTimeout.toMillis()
+                      + "ms for acquiring session. To mitigate error increase the timeout duration at"
+                      + " session pool option `acquireSessionTimeout` or increase the "
```
olavloite, line 1708: super nit: could we literally call out the method they should call? So something like `Call SessionPoolOptions#setAcquireSessionTimeout(Duration) to set a higher timeout.`

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -695,6 +716,26 @@ public Builder setWaitForMinSessions(Duration waitForMinSessions) {
       return this;
     }
 
+    /**
+     * If greater than zero, we wait for said duration when no sessions are available in the {@link
+     * SessionPool}. When no configuration is passed, the default is a 60s timeout. To avoid setting
+     * a default value, set the value as null.
```
olavloite, line 722: nit:

```suggestion
     * SessionPool}. The default is a 60s timeout. Set the value to null to disable the timeout.
```

google-cloud-spanner/src/test/java/com/google/cloud/spanner/BatchCreateSessionsSlowTest.java
```diff
@@ -0,0 +1,210 @@
...
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import org.junit.runners.MethodSorters;
+import org.threeten.bp.Duration;
+
+@Category(SlowTest.class)
+@RunWith(JUnit4.class)
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
```
olavloite, line 54: nit: copy-paste? I don't think you need this ordering for these tests.

olavloite (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Added a validation.

PR #2655, opened 2023-10-05
feat: enable session leaks prevention by cleaning up long-running tra…

arpan14:
Adding a couple of option in `SessionPoolOptions` for auto-detecting long running transactions, logging the stack-trace and removing such transactions from consuming more resources. There can be scenarios where transactions are unexpectedly taking more time and hence can consume the `session` for longer durations. With this change, customers get two new options which can be set while providing `SessionPoolOptions`
1. `setWarnIfInactiveTransactions()` - Calling this method will generate warning logs and help in identifying faulty code which is leading to incorrect consumption of sessions.
2. `setWarnAndCloseIfInactiveTransactions()` - Calling this method will close the faulty transactions that are consuming the resources. We will also be generating the warning logs which can be later referred to know the transactions that were closed through this option.

Reference Feature PR - https://github.com/googleapis/java-spanner/pull/2419

### DIFF
--- a/session-and-channel-pool-configuration.md
+++ b/session-and-channel-pool-configuration.md
@@ -285,7 +285,7 @@ This will cause the following to happen internally in the client library:
 ### Session Leak
 A Spanner object of the Client Library has a limit on the number of maximum sessions. For example the 
 default value of `MaxSessions` in the Java Client Library is 400. You can configure these values at the time of 
-Client side Database object creation by passing in the `SessionPoolOptions`. When all the sessions are checked 
+creating a `Spanner` instance by setting custom `SessionPoolOptions`. When all the sessions are checked 
 out of the session pool, every new transaction has to wait until a session is returned to the pool. 
 If a session is never returned to the pool (hence causing a session leak), the transactions will have to wait 
 indefinitely and your application will be blocked.
--- a/session-and-channel-pool-configuration.md
+++ b/session-and-channel-pool-configuration.md
@@ -291,10 +291,10 @@ If a session is never returned to the pool (hence causing a session leak), the t
 indefinitely and your application will be blocked.
 
 #### Common Root Causes
-The most common reason for a session leak is that a transaction was started by the application but 
-never committed or rolled back. What you should do is simply call commit on the transaction at the 
-end of your transaction code. Spanner has two types of transactions, read only and read-write 
-transactions. When we perform a read in a read-write transaction we still need to commit it.
+The most common reason for session leaks in the Java client library are:
+1. Not closing a `ResultSet` that is returned by `executeQuery`. Always put `ResultSet` objects in a try-with-resources block, or take other measures to ensure that the `ResultSet` is always closed.
+2. Not closing a `ReadOnlyTransaction` when you no longer need it. Always put `ReadOnlyTransaction` objects in a try-with-resources block, or take other measures to ensure that the `ReadOnlyTransaction` is always closed.
+3. Not closing a `TransactionManager` when you no longer need it. Always put `TransactionManager` objects in a try-with-resources block, or take other measures to ensure that the `TransactionManager` is always closed.
 
 As shown in the example below, the `try-with-resources` block releases the session after it is complete. 
 If you don't use `try-with-resources` block, unless you explicitly call the `close()` method on all resources 
--- a/session-and-channel-pool-configuration.md
+++ b/session-and-channel-pool-configuration.md
@@ -281,3 +281,85 @@ This will cause the following to happen internally in the client library:
 1. The `TransactionRunner` will automatically commit the transaction if the supplied user code
    finished without any errors. The `Commit` RPC that is invoked uses a thread from the default gRPC
    thread pool.
+
+### Session Leak
+A Spanner object of the Client Library has a limit on the number of maximum sessions. For example the 
+default value of `MaxSessions` in the Java Client Library is 400. You can configure these values at the time of 
+Client side Database object creation by passing in the `SessionPoolOptions`. When all the sessions are checked 
+out of the session pool, every new transaction has to wait until a session is returned to the pool. 
+If a session is never returned to the pool (hence causing a session leak), the transactions will have to wait 
+indefinitely and your application will be blocked.
+
+#### Common Root Causes
+The most common reason for a session leak is that a transaction was started by the application but 
+never committed or rolled back. What you should do is simply call commit on the transaction at the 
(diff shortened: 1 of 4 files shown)


session-and-channel-pool-configuration.md
```diff
@@ -281,3 +281,85 @@ This will cause the following to happen internally in the client library:
...
+  while (resultSet.next()) {
+    // use the results.
+  }
+}
+```
+
+#### Debugging and Resolving Session Leaks
+
+##### Logging
+Enabled by default, the logging option shares warn logs when you have exhausted >95% of your session pool.
```
asthamohta, line 318: I think for java we won't have this log right so we can remove the line?

asthamohta (approved):

session-and-channel-pool-configuration.md
```diff
@@ -281,3 +281,85 @@ This will cause the following to happen internally in the client library:
 1. The `TransactionRunner` will automatically commit the transaction if the supplied user code
    finished without any errors. The `Commit` RPC that is invoked uses a thread from the default gRPC
    thread pool.
+
+### Session Leak
+A Spanner object of the Client Library has a limit on the number of maximum sessions. For example the
```
olavloite, line 286: nit:
```suggestion
A `DatabaseClient` object of the Client Library has a limit on the number of maximum sessions. For example the 
```

session-and-channel-pool-configuration.md
```diff
@@ -281,3 +281,85 @@ This will cause the following to happen internally in the client library:
 1. The `TransactionRunner` will automatically commit the transaction if the supplied user code
    finished without any errors. The `Commit` RPC that is invoked uses a thread from the default gRPC
    thread pool.
+
+### Session Leak
+A Spanner object of the Client Library has a limit on the number of maximum sessions. For example the 
+default value of `MaxSessions` in the Java Client Library is 400. You can configure these values at the time of 
+Client side Database object creation by passing in the `SessionPoolOptions`. When all the sessions are checked
```
olavloite, line 288:
```suggestion
creating a `Spanner` instance by setting custom `SessionPoolOptions`. When all the sessions are checked 
```

session-and-channel-pool-configuration.md
```diff
@@ -281,3 +281,85 @@ This will cause the following to happen internally in the client library:
...
+Client side Database object creation by passing in the `SessionPoolOptions`. When all the sessions are checked 
+out of the session pool, every new transaction has to wait until a session is returned to the pool. 
+If a session is never returned to the pool (hence causing a session leak), the transactions will have to wait 
+indefinitely and your application will be blocked.
+
+#### Common Root Causes
+The most common reason for a session leak is that a transaction was started by the application but 
+never committed or rolled back. What you should do is simply call commit on the transaction at the 
+end of your transaction code. Spanner has two types of transactions, read only and read-write 
+transactions. When we perform a read in a read-write transaction we still need to commit it.
```
olavloite, line 297:
```suggestion
The most common reason for session leaks in the Java client library are:
1. Not closing a `ResultSet` that is returned by `executeQuery`. Always put `ResultSet` objects in a try-with-resources block, or take other measures to ensure that the `ResultSet` is always closed.
2. Not closing a `ReadOnlyTransaction` when you no longer need it. Always put `ReadOnlyTransaction` objects in a try-with-resources block, or take other measures to ensure that the `ReadOnlyTransaction` is always closed.
3. Not closing a `TransactionManager` when you no longer need it. Always put `TransactionManager` objects in a try-with-resources block, or take other measures to ensure that the `TransactionManager` is always closed.
```

session-and-channel-pool-configuration.md
```diff
@@ -281,3 +281,85 @@ This will cause the following to happen internally in the client library:
...
+
+As shown in the example below, the `try-with-resources` block releases the session after it is complete. 
+If you don't use `try-with-resources` block, unless you explicitly call the `close()` method on all resources 
+such as `ResultSet` the session is not released back to the pool. If the transaction does not run within 
+`try-with-resources` block and if we don't close the resources explicitly, we will have a session leak.
```
olavloite, line 302:
```suggestion
such as `ResultSet`, the session is not released back to the pool.
```

session-and-channel-pool-configuration.md
```diff
@@ -281,3 +281,85 @@ This will cause the following to happen internally in the client library:
...
+##### Logging
+Enabled by default, the logging option shares warn logs when you have exhausted >95% of your session pool. 
+This could mean two things, either you need to increase the max sessions in your session pool (as the number 
+of queries run using the client side database object is greater than your session pool can serve) or you may 
+have a session leak.
+
+To help debug which transactions may be causing this session leak, the logs will also contain stack traces of 
+transactions which have been running longer than expected. The logs are pushed to a destination based on 
+how the log exporter is configured for the host application.
+
```
olavloite, line 326: Are we logging in the Java client? Based on the comment above, I would expect 'No'.

If we are logging: I'm pretty sure that we are not pushing those logs to any 'destination based on how the log exporter is configured'.

olavloite (approved):

session-and-channel-pool-configuration.md
arpan14: Some of the other client libraries such as Node/Python, emit a log every-time the application uses > 95% of the sessions. In Java, we use the condition as a pre-condition for the algo but don't emit any log for this condition. We only emit a one of the below logs when, 
a) WARN LOG - when there is an actual long-running transaction. This includes bunch of other pre-conditions conditions (lastUsedTime, type of Transaction, etc.)
OR b) WARN LOG - We have automatically cleaned a long running transaction

Now what I mean as pushing log to the exporter is - the logs will be stored in a destination as per what is configured for the application. For ex - Some apps may configure cloud logging while others could configure logs to be stored on separate files on the app server. 

The note about the exporter is not too relevant for Java where there is a default logging behaviour and I think it was more relevant for other languages like Node where there was a logging capability was absent. In Node, the default logging was console logs or you could define some other exporter.

LMK if this comment can be better framed or excluded in context of Java client.

olavloite: We're using `java.util.logging` for this, right? Then I would just explicitly call out that. So something like 'The warning is logged using java.util.logging.'. I think that most people know what they need to do to configure that, and can otherwise search for it.

Flyboy-60 (commented):
****

Flyboy-60 (commented):
[](https://github.com/googleapis/java-spanner/pull/2655)

PR #2656, opened 2023-10-05
chore: add a system property to set session pool ordering

olavloite:
Adds support for using a System property to set the default session pool ordering. The default ordering is LIFO. This can be changed to FIFO and RANDOM for users who need/want that.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
@@ -17,8 +17,10 @@
 package com.google.cloud.spanner;
 
 import com.google.cloud.spanner.SessionPool.Clock;
+import com.google.cloud.spanner.SessionPool.Position;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
+import java.util.Locale;
 import java.util.Objects;
 import org.threeten.bp.Duration;
 
@@ -62,6 +64,7 @@ public class SessionPoolOptions {
   private final boolean autoDetectDialect;
   private final Duration waitForMinSessions;
   private final Duration acquireSessionTimeout;
+  private final Position releaseToPosition;
 
   /** Property for allowing mocking of session maintenance clock. */
   private final Clock poolMaintainerClock;
@@ -86,6 +89,7 @@ public class SessionPoolOptions {
     this.autoDetectDialect = builder.autoDetectDialect;
     this.waitForMinSessions = builder.waitForMinSessions;
     this.acquireSessionTimeout = builder.acquireSessionTimeout;
+    this.releaseToPosition = builder.releaseToPosition;
     this.inactiveTransactionRemovalOptions = builder.inactiveTransactionRemovalOptions;
     this.poolMaintainerClock = builder.poolMaintainerClock;
   }
@@ -114,6 +118,7 @@ public class SessionPoolOptions {
         && Objects.equals(this.autoDetectDialect, other.autoDetectDialect)
         && Objects.equals(this.waitForMinSessions, other.waitForMinSessions)
         && Objects.equals(this.acquireSessionTimeout, other.acquireSessionTimeout)
+        && Objects.equals(this.releaseToPosition, other.releaseToPosition)
         && Objects.equals(
             this.inactiveTransactionRemovalOptions, other.inactiveTransactionRemovalOptions)
         && Objects.equals(this.poolMaintainerClock, other.poolMaintainerClock);
@@ -138,6 +143,7 @@ public class SessionPoolOptions {
         this.autoDetectDialect,
         this.waitForMinSessions,
         this.acquireSessionTimeout,
+        this.releaseToPosition,
         this.inactiveTransactionRemovalOptions,
         this.poolMaintainerClock);
   }
@@ -252,6 +258,10 @@ public class SessionPoolOptions {
   @VisibleForTesting
   Duration getAcquireSessionTimeout() {
     return acquireSessionTimeout;
+  }
+
+  Position getReleaseToPosition() {
+    return releaseToPosition;
   }
 
   public static Builder newBuilder() {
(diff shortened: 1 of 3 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -440,9 +450,23 @@ public static class Builder {
     private boolean autoDetectDialect = false;
     private Duration waitForMinSessions = Duration.ZERO;
     private Duration acquireSessionTimeout = Duration.ofSeconds(60);
+    private Position releaseToPosition = getReleaseToPositionFromSystemProperty();
 
     private Clock poolMaintainerClock;
 
+    private static Position getReleaseToPositionFromSystemProperty() {
+      // NOTE: This System property is a beta feature. Support for it can be removed in the future.
+      String key = "SESSION_POOL_RELEASE_TO_POSITION";
```
c2nes, line 459: Maybe namespace this as `com.google.cloud.spanner.session_pool_release_to_position`?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolTest.java
```diff
@@ -236,6 +241,139 @@ public void poolLifo() {
...
+          session2.close();
+          session1.close();
+
+          // Verify that we get the sessions in FIFO order, so in this order:
+          // 1. session2
+          // 2. session1
+          Session session3 = pool.getSession().get();
+          Session session4 = pool.getSession().get();
+          assertEquals(session2, session3);
+          assertEquals(session4, session1);
```
arpan14, line 278: Nit: swap the expected/actual value

```suggestion
          assertEquals(session1, session4);
```

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolTest.java
```diff
@@ -236,6 +241,139 @@ public void poolLifo() {
...
+                  allSessions.get(0).stream()
+                      .map(PooledSessionFuture::get)
+                      .collect(Collectors.toList());
+              List<Session> secondTime =
+                  allSessions.get(1).stream()
+                      .map(PooledSessionFuture::get)
+                      .collect(Collectors.toList());
+              switch (position) {
+                case FIRST:
+                  // FIFO:
```
arpan14, line 328: Isn't this LIFO? Since the last element `N` will be added to the first, and we always poll from the front, we would always have the last one polled first.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolTest.java
```diff
@@ -236,6 +241,139 @@ public void poolLifo() {
...
+                  // First check out all sessions, so we have 1, 2, 3, 4, ..., N
+                  // Then release them all back into the pool in the same order (1, 2, 3, 4, ...,
+                  // N).
+                  // That will give us the list N, ..., 4, 3, 2, 1 because each session is added at
+                  // the
+                  // front of the pool.
+                  assertEquals(firstTime, Lists.reverse(secondTime));
+                  break;
+                case LAST:
+                  // LIFO:
```
arpan14, line 338: Similarly isn't this FIFO? Given everything gets added to the end of the queue, whatever was added first, will be removed first. No?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolTest.java
olavloite: Eh, yes. Hence my comment about the naming being a bit confusing, I managed to confuse myself as well here.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolTest.java
olavloite: Yep

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -440,9 +450,23 @@ public static class Builder {
...
+      if (System.getProperties().containsKey(key)) {
+        try {
+          return Position.valueOf(System.getProperty(key).toUpperCase(Locale.ENGLISH));
```
arpan14, line 462: Would it be useful to add some logging for the case when this environment variable is set? Just so that if we run into any un-intended cases where a few other customers discover this option and set this, we at-least know if this option got set?

Logging is the minimum we can do, apart from adding to the trace?

olavloite: We wouldn't be able to see the logging, except if we add a gRPC header, and I don't think it's worth the latter. I don't worry about anyone finding and enabling this option on their own (and if they do, they probably also turn it off if they feel that it does not work well for them).

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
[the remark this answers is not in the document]
olavloite: As discussed offline, we'll go for the namespaced name `com.google.cloud.spanner.session_pool_release_to_position`.

olavloite: Yes, that is more idiomatic for System properties. All caps is for environment variables.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolTest.java
olavloite: Done

arpan14 (approved):

c2nes:
(retroactive LGTM from me as well)

PR #2662, opened 2023-10-09
fix: Noop in case there is no change in  autocommit value for setAutocommit() method

ankiaga:
Noop in case there is no change in  autocommit value for setAutocommit() method

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
@@ -401,10 +401,10 @@ class ConnectionImpl implements Connection {
 
   @Override
   public void setAutocommit(boolean autocommit) {
+    ConnectionPreconditions.checkState(!isClosed(), CLOSED_ERROR_MSG);
     if (isAutocommit() == autocommit) {
       return;
     }
-    ConnectionPreconditions.checkState(!isClosed(), CLOSED_ERROR_MSG);
     ConnectionPreconditions.checkState(!isBatchActive(), "Cannot set autocommit while in a batch");
     ConnectionPreconditions.checkState(
         !isTransactionStarted(), "Cannot set autocommit while a transaction is active");
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
@@ -401,6 +401,9 @@ class ConnectionImpl implements Connection {
 
   @Override
   public void setAutocommit(boolean autocommit) {
+    if (isAutocommit() == autocommit) {
+      return;
+    }
     ConnectionPreconditions.checkState(!isClosed(), CLOSED_ERROR_MSG);
     ConnectionPreconditions.checkState(!isBatchActive(), "Cannot set autocommit while in a batch");
     ConnectionPreconditions.checkState(
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
@@ -472,9 +472,13 @@ public class ConnectionImplTest {
       assertThat(subject.isAutocommit(), is(true));
       subject.execute(Statement.of("begin transaction"));
 
-      SpannerException exception = assertThrows(SpannerException.class, () -> subject.setAutocommit(false));
+      SpannerException exception =
+          assertThrows(SpannerException.class, () -> subject.setAutocommit(false));
       assertEquals(ErrorCode.FAILED_PRECONDITION, exception.getErrorCode());
-      assertTrue(exception.getMessage().contains("Cannot set autocommit while in a temporary transaction"));
+      assertTrue(
+          exception
+              .getMessage()
+              .contains("Cannot set autocommit while in a temporary transaction"));
     }
   }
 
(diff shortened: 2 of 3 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
```diff
@@ -401,6 +401,9 @@ public boolean isClosed() {
 
   @Override
   public void setAutocommit(boolean autocommit) {
+    if (isAutocommit() == autocommit) {
```
olavloite, line 404: Move this `if` block to after the line with `ConnectionPreconditions.checkState(!isClosed(), CLOSED_ERROR_MSG);`. Calling this method, even if the value is the same as now, should throw an exception. The current implementation will actually do that as well, as the `isAutocommit()` call will check it, but that is not clear from the current implementation, so for readability it would be better to move this a bit down.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
```diff
@@ -412,6 +412,22 @@ public void testExecuteSetAutocommitOff() {
...
 
+  @Test
+  public void testExecuteAutocommitNoop() {
+    try (ConnectionImpl subject =
+        createConnection(
+            ConnectionOptions.newBuilder()
+                .setCredentials(NoCredentials.getInstance())
+                .setUri(URI)
+                .build())) {
+      assertThat(subject.isAutocommit(), is(true));
```
olavloite, line 423: We should also add tests for the following scenarios:
1. Calling `setAutoCommit(false)` when `autocommit=false` and the transaction has not yet been started (that is; there has been no statement executed on the connection yet).
1. Calling `setAutoCommit(false)` when `autocommit=false` and the transaction has been started (that is; there has been at least one statement executed on the connection).
1. Calling `setAutoCommit(true)` when `autocommit=true`, but the user has first executed `connection.execute(Statement.of("begin transaction"))`

The last test case should also be extended to include calling `setAutoCommit(false)` and be tested both before and after a different statement has been executed in the transaction. The latter means testing all reasonable variations of:

```
set autocommit=true|false;
begin transaction;
[select * from my_table];  -- Include/exclude this statement in the test to force the actual start of the transaction
set autocommit=true|false;
```

olavloite:
A couple of general comments on the pull request itself:
1. Try to make the first line of the commit message 80 characters or less. Alternatively; manually edit the title of the pull request so it's not broken into a part that is in the title and a part that is in the description.
2. The semantic prefix should be written using all lower-case letters (so `fix:` instead of `Fix:`)
3. Add a description to the pull request. GitHub creates this automatically if you format your commit message as follows:

```
fix: my pull request header

My Pull Request description that may span
multiple lines. If your pull request also fixes a
specific issue, then you should end your commit
message with a suffix that says 'Fixes #github_issue_number'.

Fixes #2662
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
ankiaga: Fixed

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
ankiaga: Added tests for different combinations

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
```diff
@@ -412,6 +412,140 @@ public void testExecuteSetAutocommitOff() {
...
+      subject.execute(Statement.of("begin transaction"));
+
+      subject.setAutocommit(false);
+      fail("Cannot set autocommit while in a temporary transaction");
```
olavloite, line 476: nit: use `assertThrows(..)` for this.

I know that there are examples in this test class that use this setup with a try-fail block. Those are however left-overs from when this library had to support Java 7, and we did not have support for lambda expressions.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
```diff
@@ -412,6 +412,140 @@ public void testExecuteSetAutocommitOff() {
...
+
+  @Test
+  public void testSetAutocommitToTrue_inAutoCommitAndInTransaction_noop() {
+    try (ConnectionImpl subject =
+        createConnection(
+            ConnectionOptions.newBuilder()
+                .setCredentials(NoCredentials.getInstance())
+                .setUri(URI)
+                .build())) {
+      assertThat(subject.isAutocommit(), is(true));
```
olavloite, line 439: No change needed in this PR, but for next PR: We prefer the use of `assertTrue(..)` and other standard JUnit assertions over `Truth` assertions. The examples using `assertThat(..)` are also leftovers from pre-Java 8 times.

olavloite (approved):
LGTM with a couple of minor nits.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
ankiaga: Changed to assertTrue everywhere

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionImplTest.java
ankiaga: Done

PR #2663, opened 2023-10-09
test: swap expected and actual values in DML returning tests.

arpan14:
While debugging flaky issue - https://github.com/googleapis/java-spanner/issues/2629 , it was noticed that `ITDmlReturningTest` is not using the right convention for `assertEquals` method. This PR swaps the value so that debugging readability is improved.

For ex - The following error message in the above issue is confusing due to this swap between expected/actual values.
```
java.lang.AssertionError: expected:<1> but was:<2>
```


harshachinta (approved):
LGTM

PR #2672, opened 2023-10-12
chore: add option for retrying on different channel 

olavloite:
Adds an option to retry the BeginTransaction RPC on a different channel
than the original channel.


google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/RetryOnNewChannelConfiguration.java
```diff
@@ -0,0 +1,113 @@
...
+          System.getProperty(
+              "com.google.cloud.spanner.retry_begin_transaction_on_new_channel", "false"));
+  private final int retryDelayBeginTransactionOnNewChannel =
+      getIntProperty("com.google.cloud.spanner.retry_delay_begin_transaction_on_new_channel", 5);
+  // This is the maximum number of times that the custom retry logic will attempt to execute the
+  // RPC.
+  private final int maxAttemptsBeginTransactionOnNewChannel =
+      getIntProperty("com.google.cloud.spanner.max_attempts_begin_transaction_on_new_channel", 4);
+
+  // This is the number of times that Gax will retry the RPC on a given channel when retrying on a
```
arpan14, line 37: Given its an experimental feature do we really want to expose settings to set retry configuration? Can't we start just by exposing a default setting (of low duration). Wouldn't `com.google.cloud.spanner.retry_begin_transaction_on_new_channel` be enough?

We anyway don't foresee system properties as the long-term way of passing such attributes.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerOptionsTest.java
```diff
@@ -951,6 +951,18 @@ public void testAsyncExecutorProviderCoreThreadCount() throws Exception {
                 propertyName, "", SpannerOptions::getDefaultAsyncExecutorProviderCoreThreadCount));
   }
 
+  static void runWithSystemProperty(String propertyName, String propertyValue, Runnable runnable)
```
arpan14, line 954: Where is this being used?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITWriteTest.java
```diff
@@ -91,6 +91,15 @@
 @Category(ParallelIntegrationTest.class)
 @RunWith(Parameterized.class)
 public class ITWriteTest {
+  private static final String RETRY_BEGIN_TRANSACTION_ON_NEW_CHANNEL_PROPERTY =
+      "com.google.cloud.spanner.retry_begin_transaction_on_new_channel";
+  private static final String originalValueForBeginTransactionOnNewChannel =
+      System.getProperty(RETRY_BEGIN_TRANSACTION_ON_NEW_CHANNEL_PROPERTY);
+
+  static {
+    System.setProperty(RETRY_BEGIN_TRANSACTION_ON_NEW_CHANNEL_PROPERTY, "true");
```
arpan14, line 100: Instead of setting this for all integration tests, shouldn't we keep the existing tests unchanged, and add maybe one new integration test with this property?

arpan14:
@olavloite closing this PR as we did not choose to implement this.

PR #2557, opened 2023-07-28
feat: Add maxBatchingDelayMs Option.

nginsberg-google:
Adds support for max batching delay option in Commit requests.

This feature allows users to express how much latency they are willing for spanner to add to attempt to batch their writes. Generally speaking, the more latency they are willing to tolerate, the more efficiently Spanner can serve the writes.

PR #2716, opened 2023-10-31
refactor: Options class to carve a new class that can be auto-generated.

arpan14:
**DO NOT MERGE, PROTOTYPE/EXPERIMENT ONLY**

This is a proto-type of how an auto-generated class can be used in a backward-compatible way in clients (without breaking anything for the customer. For this prototype, we are picking a recent feature `maxBatchingDelayMs` which is a new option in the `CommitRequest` proto message (Reference feature PR - https://github.com/googleapis/java-spanner/pull/2557/files).

Details of changes in PR -

- `google-cloud-spanner/src/main/java/com/google/cloud/spanner/AutoGeneratedOptions.java` : A new class which holds each auto-generated option. Assume that this class is auto-generated by GAPIC. Things that we need to handle within auto-gen,

- [ ] How will GAPIC parse the proto file and generate this class.
- [ ] How will GAPIC know which interface to implement i.e how do options get mapped to internal interfaces that already exist. This is important since these interfaces (for ex : `TransactionOption`) are publicly exposed. And hence, we cannot afford to break these contracts for customers.
- [ ] How will GAPIC avoid over-writing the entire file. i.e how do we ensure that the file gets updated with just new options. I'm assuming this should be easy to handle with existing parsing logic.

- `google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java` : Existing class that holds all options. This is a public class used directly within public APIs. Hence, this will continue to remain our only public interface through which customers pass Options to our APIs. To make this work, we are adding `public final class Options extends AutoGeneratedOptions`
- `google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionImpl.java` : This is where the `Options` are read and plumbed into the proto message. We will continue to have such changes in the long-term. In fact, ideally this would be the only hand-written change needed. Based on various type of requests, there maybe more than 1 place where such requests are being built. So we may need separate client side refactoring to ensure such proto objects are constructed in only one place. Such refactoring should be less ambiguous as well as easy to complete.
- `google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java` - A unit test which proves backward compatibility. In this unit test, we don't make use new class `AutoGeneratedOptions` . For all purposes `Options` / `TransactionOption` continues to remain our only public interfaces (as before)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/AutoGeneratedOptions.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+package com.google.cloud.spanner;
+
+/** Class to represent auto-generated RPC options. The options in this class
+ * are auto-generated using GAPIC auto-gen
+ */
+class AutoGeneratedOptions {
+
+  /** Specifies the maximum batching delay for the transaction. */
```
arpan14, line 24: How are comments generated?

PR #2724, opened 2023-11-06
chore: adding more logs with session name to debug DEADLINE_EXCEEDED errors.

arpan14:
* Logging the session name when we receive exceptions while executing a transaction.
* [DO NOT MERGE] This snipper is to be shared as a custom jar.


olavloite (approved):

arpan14:
Closing the PR since it is being shared privately as a .jar

PR #2736, opened 2023-11-20
feat: add PG OID support

tlhquynh:
Added support for PG OID type which is an int64 value with a PG_OID type annotation code.


tlhquynh:
Both aforementioned TODOs regarding class incompatible due to mismatched serialVersionUID and rebasing after #2703 is merged are done.

.java-version
```diff
@@ -0,0 +1 @@
+17.0.9
```
olavloite, line 1: Remove this file (and potentially add the file name to .gitignore to prevent it from being included in commits in the future)

.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
```
olavloite, line 1: Same as above: Remove file from GitHub and add the filename to .gitignore

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractStructReader.java
```diff
@@ -61,6 +61,8 @@ protected String getPgJsonbInternal(int columnIndex) {
     throw new UnsupportedOperationException("Not implemented");
   }
 
+  protected abstract long getPgOidInternal(int columnIndex);
```
olavloite, line 64: Add a default implementation for this and the other methods that are being added to this class. The default implementation should just throw UnsupportedOperationException. This prevents this change from being a breaking change, as anyone who has a class that extends AbstractStructReader does not get a compile error.

That should also remove the requirement to add the method names to clirr-ignored-differences.xml

google-cloud-spanner/clirr-ignored-differences.xml
```diff
@@ -192,6 +192,63 @@
     <className>com/google/cloud/spanner/StructReader</className>
     <method>java.util.List getPgJsonbList(java.lang.String)</method>
   </difference>
+  <!-- PG OID -->
+  <difference>
```
olavloite, line 196: It should not be necessary to add the internal methods to this file, if those methods are given a default implementation that just throws UnsupportedOperationException

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractStructReader.java
```diff
@@ -259,6 +265,19 @@ public String getPgJsonb(String columnName) {
...
+    checkNonNullOfType(columnIndex, Type.pgOid(), columnIndex);
+    return getPgOidInternal(columnIndex);
+  }
+
+  @Override
+  public long getPgOid(String columnName) {
+    int columnIndex = getColumnIndex(columnName);
+    checkNonNullOfType(columnIndex, Type.pgOid(), columnName);
+    return getPgOidInternal(columnIndex);
+  }
```
olavloite, line 279: Would it be worth considering not adding any of this at all, and instead instruct users that they have to get PG.OID values in the same way as INT64? So just call getLong(..) and the other corresponding methods. This is otherwise all just a copy-paste of those methods, and there's no additional validation of the values that we are getting.
I know that it is not what we have done for other types (e.g. we could in theory have done the same for JSON by telling users to just call getString(...)). But this might be the right time to start to prevent us from doing double work for future types, while it does not add any additional type safety or features for users.

@skuruppu Any thoughts on this?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/AllTypesMockServerTest.java
```diff
@@ -572,22 +613,40 @@ public void testInsertAllTypes() {
       ExecuteSqlRequest request = mockSpanner.getRequestsOfType(ExecuteSqlRequest.class).get(0);
       Map<String, Type> paramTypes = request.getParamTypesMap();
       Map<String, Value> params = request.getParams().getFieldsMap();
-      assertEquals(20, paramTypes.size());
-      assertEquals(20, params.size());
+      System.out.println("Dialect = " + dialect);
```
olavloite, line 616: nit: remove

olavloite (commented):
Looks generally good to me. The biggest question that I have is whether there is any value in adding all the getPgOid(..) methods, or whether it would be better to just say 'use getLong(..) for both normal INT64 and PG.OID values'.

.java-version
tlhquynh: Done.

.vscode/settings.json
tlhquynh: Done.

google-cloud-spanner/clirr-ignored-differences.xml
tlhquynh: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractStructReader.java
tlhquynh: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractStructReader.java
tlhquynh: Done.
I had an offline chat with @skuruppu and treating OID as Long is actually what we are doing for the other clients. I have removed all external PG.OID getters.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/AllTypesMockServerTest.java
tlhquynh: Removed.

google-cloud-spanner/clirr-ignored-differences.xml
olavloite: These entries in `clirr-ignored-differences.xml` for `StructReader` can be removed, as we don't have those methods anymore.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractStructReader.java
```diff
@@ -61,6 +61,10 @@ protected String getPgJsonbInternal(int columnIndex) {
     throw new UnsupportedOperationException("Not implemented");
   }
 
+  protected long getPgOidInternal(int columnIndex) {
```
olavloite, line 64: I think that you can remove this method (and the equivalent array methods below) as well, and instead just use the `getLongInternal` method.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractStructReader.java
```diff
@@ -165,15 +177,20 @@ public boolean getBoolean(String columnName) {
 
   @Override
   public long getLong(int columnIndex) {
-    checkNonNullOfCodes(columnIndex, Arrays.asList(Code.ENUM, Code.INT64), columnIndex);
-    return getLongInternal(columnIndex);
+    checkNonNullOfCodes(
+        columnIndex, Arrays.asList(Code.ENUM, Code.PG_OID, Code.INT64), columnIndex);
+    return getColumnType(columnIndex).getCode() == Code.PG_OID
+        ? getPgOidInternal(columnIndex)
```
olavloite, line 183: I think that this can be simplified to just calling `getLongInternal`. The same also applies to the array methods below.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/GrpcStruct.java
```diff
@@ -460,6 +468,12 @@ protected String getPgJsonbInternal(int columnIndex) {
     return (String) rowData.get(columnIndex);
   }
 
+  @Override
+  protected long getPgOidInternal(int columnIndex) {
```
olavloite, line 472: remove and rely on `getLongInternal` instead

google-cloud-spanner/src/main/java/com/google/cloud/spanner/GrpcStruct.java
```diff
@@ -771,6 +789,18 @@ protected List<String> getPgJsonbListInternal(int columnIndex) {
     return Collections.unmodifiableList((List<String>) rowData.get(columnIndex));
   }
 
+  @Override
+  protected Int64Array getPgOidListInternal(int columnIndex) {
```
olavloite, line 793: Remove and use `getLongListInternal` instead

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Value.java
```diff
@@ -823,6 +871,13 @@ public String getPgJsonb() {
     throw new UnsupportedOperationException("Not implemented");
   }
 
+  /**
+   * Returns the value of a {@code PG_OID}-typed instance.
+   *
+   * @throws IllegalStateException if {@code isNull()} or the value is not of the expected type
+   */
+  public abstract long getPgOid();
```
olavloite, line 879: As we expect users to call `ResultSet#getInt64(..)` for OID, we should do the same here. So remove this method and the array equivalent, and make sure that the `getInt64(..)` here also works for OID.

google-cloud-spanner/clirr-ignored-differences.xml
tlhquynh: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Value.java
tlhquynh: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/GrpcStruct.java
tlhquynh: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractStructReader.java
tlhquynh: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractStructReader.java
tlhquynh: Done.

olavloite (approved):

PR #2739, opened 2023-11-23
test: add a default benchmark for measuring baselines.

arpan14:
* Adding a benchmark (along with usual considerations) for measuring the latencies of some Java Client APIs. For any new feature, we usually first define the baseline by running benchmarks (without the feature).
* Adding few utility methods to measure the p99/p95/p50 latencies.
* Added benchmarking best-practices to avoid hotspot (by randomising keys), avoid cold start, etc.


google-cloud-spanner/src/test/java/com/google/cloud/spanner/BenchmarkingDataLoaderUtility.java
```diff
@@ -0,0 +1,173 @@
...
+ * data, report latency metrics, etc.
+ */
+@Category(SlowTest.class)
+@RunWith(JUnit4.class)
+public class BenchmarkingDataLoaderUtility {
+  @ClassRule public static IntegrationTestEnv env = new IntegrationTestEnv();
+  private static final String PROJECT_ID = "";
+  private static final String INSTANCE_ID = "";
+  private static final String DATABASE_ID = "";
+  private static final String SERVER_URL = "https://staging-wrenchworks.sandbox.googleapis.com";
```
olavloite, line 52: Should this be here?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/BenchmarkingDataLoaderUtility.java
```diff
@@ -0,0 +1,173 @@
...
+ * Hosts a bunch of utility methods/scripts that can be used while performing benchmarks to load
+ * data, report latency metrics, etc.
+ */
+@Category(SlowTest.class)
+@RunWith(JUnit4.class)
+public class BenchmarkingDataLoaderUtility {
+  @ClassRule public static IntegrationTestEnv env = new IntegrationTestEnv();
+  private static final String PROJECT_ID = "";
+  private static final String INSTANCE_ID = "";
+  private static final String DATABASE_ID = "";
```
olavloite, line 51: Are these maybe meant to be environment variable keys? If so, we should rename them so that is clearer (and give them an actual value, and make them actually look up an env var). Or is it something else?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/BenchmarkingDataLoaderUtility.java
```diff
@@ -0,0 +1,173 @@
...
+  /**
+   * A utility which bulk inserts 10^5 records into the database in batches. The method assumes that
+   * the instance/database/table is already created. It does not perform any admin operations.
+   *
+   * <p>Table schema used here: CREATE TABLE FOO ( id INT64 NOT NULL, BAZ INT64, BAR INT64, )
+   * PRIMARY KEY(id);
+   */
+  @Test
+  public void bulkInsertTestData() {
+    String insertQuery = "INSERT INTO FOO(id, BAZ, BAR) VALUES(@key, @val1, @val2)";
```
olavloite, line 84: Would it not make more sense to use either mutations or even the new batch write API for this?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/BenchmarkingDataLoaderUtility.java
```diff
@@ -0,0 +1,173 @@
...
+   * @return
+   * @throws Exception
+   */
+  public static List<Duration> collectResults(
+      final ListeningScheduledExecutorService service,
+      final List<ListenableFuture<List<Duration>>> results,
+      final int numOperations)
+      throws Exception {
+    service.shutdown();
+    if (!service.awaitTermination(60L, TimeUnit.MINUTES)) {
```
olavloite, line 123: Considering that it is a generic and public method: Would it make sense to add the timeout as a method argument, so the caller can determine what the max wait should be?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/BenchmarkingDataLoaderUtility.java
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  public static double percentile(int percentile, List<Duration> orderedResults) {
+    int index = percentile * orderedResults.size() / 100;
+    Duration value = orderedResults.get(index);
+    Double convertedValue = convertDurationToFractionInSeconds(value);
+    return convertedValue;
+  }
+
+  public static double avg(List<Duration> results) {
```
olavloite, line 159: Add a Javadoc explaining that the returned value is the average in seconds (the latter is the part that is not self-explanatory)

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DefaultBenchmark.java
```diff
@@ -0,0 +1,246 @@
...
+@Warmup(iterations = 1)
+public class DefaultBenchmark {
+  private static final String SELECT_QUERY = "SELECT ID FROM FOO WHERE ID = @id";
+  private static final String UPDATE_QUERY = "UPDATE FOO SET BAR=1 WHERE ID = @id";
+  private static final String ID_COLUMN_NAME = "id";
+  private static final int PARALLEL_THREADS = 1;
+  private static final int TOTAL_READS_PER_THREAD = 12000;
+  private static final int TOTAL_WRITES_PER_THREAD = 4000;
+  private static final int WARMUP_TRANSACTIONS = 1;
+  private static final int RANDOM_SEARCH_SPACE = 99999;
```
olavloite, line 76: Should all of this be a fixed value?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DefaultBenchmark.java
```diff
@@ -0,0 +1,246 @@
...
+
+  @State(Scope.Thread)
+  @AuxCounters(org.openjdk.jmh.annotations.AuxCounters.Type.EVENTS)
+  public static class BenchmarkState {
+
+    private static final String INSTANCE_ID = "";
+    private static final String DATABASE_ID = "";
+    private static final String SERVER_URL = "";
```
olavloite, line 84: This looks weird. Or is it something that the developer needs to add? If so, we should add a `// TODO(developer): Add your values here` comment

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DefaultBenchmark.java
```diff
@@ -0,0 +1,246 @@
...
+    private Spanner spanner;
+    private DatabaseClientImpl client;
+
+    @Setup(Level.Invocation)
```
olavloite, line 88: Are we sure that this is the right level for these benchmarks? The Javadoc for this level is quite telling (bring your own sword!): https://javadoc.io/static/org.openjdk.jmh/jmh-core/1.1.1/org/openjdk/jmh/annotations/Level.html#Invocation

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DefaultBenchmark.java
arpan14: They can be a fixed value so that every individual does not have to think about choosing the right numbers. I'll add a doc to each of this variable to explain what this means. And then, based on the requirement individuals may tweak the value.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/BenchmarkingDataLoaderUtility.java
arpan14: Done

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DefaultBenchmark.java
arpan14: Done

google-cloud-spanner/src/test/java/com/google/cloud/spanner/BenchmarkingDataLoaderUtility.java
arpan14: Yeah this needs to be set for us to set which test database is to be used for benchmarking. As per the setting the database will be loaded with test data.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/BenchmarkingDataLoaderUtility.java
arpan14: Not sure if I made it clear, but my aim here is that this script should make the lives of internal developers easier. At-least they don't have to setup a script each time. I have spoken to a few folks internally and they think such a utility makes it easy for them. Especially ones who don't have a lot of experience with Java.

So if this is an internal (for devs) utility, then we can just define the constants this way? Do you see any other advantage to setting it as env variables?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/BenchmarkingDataLoaderUtility.java
arpan14: Yeah, I can consider doing that.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/BenchmarkingDataLoaderUtility.java
arpan14: Sure, done

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DefaultBenchmark.java
arpan14: My requirement here was to re-initialise the session pool for every invocation. But I think we can change it to `@Iteration` given the cost of setting up the session would anyway get excluded in the top 1% results.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/BenchmarkingUtilityTest.java
```diff
@@ -0,0 +1,190 @@
...
+    client.write(ImmutableList.of(Mutation.delete("FOO", KeySet.all())));
+  }
+
+  @AfterClass
+  public static void afterClass() {
+    spanner.close();
+  }
+
+  /**
+   * A utility which bulk inserts 10^5 records into the database in batches. The method assumes that
```
olavloite, line 88: I think this should be 10^6 records. The method executes 100 batches with each 10,000 rows.

```suggestion
   * A utility which bulk inserts 10^6 records into the database in batches. The method assumes that
```

google-cloud-spanner/src/test/java/com/google/cloud/spanner/BenchmarkingUtilityTest.java
```diff
@@ -0,0 +1,190 @@
...
+ *
+ * <p>Table schema used here: CREATE TABLE FOO ( id INT64 NOT NULL, BAZ INT64, BAR INT64, )
+ * PRIMARY KEY(id);
+ */
+@Category(SlowTest.class)
+@RunWith(JUnit4.class)
+public class BenchmarkingUtilityTest {
+
+  @ClassRule
+  public static IntegrationTestEnv env = new IntegrationTestEnv();
```
olavloite, line 54: Is this used anywhere?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DefaultBenchmark.java
```diff
@@ -0,0 +1,282 @@
...
+  /**
+   * Measures the time needed to execute a burst of read and write requests.
+   *
+   * <p>Some read requests will be long-running and will cause session leaks. Such sessions will be
+   * removed by the session maintenance background task if SessionPool Option
+   * ActionOnInactiveTransaction is set as WARN_AND_CLOSE.
+   *
+   * <p>Some write requests will be long-running. The test asserts that no sessions are removed by
+   * the session maintenance background task with SessionPool Option ActionOnInactiveTransaction set
+   * as WARN_AND_CLOSE. This is because PDML writes are expected to be long-running.
```
olavloite, line 193: copy-paste error?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DefaultBenchmark.java
```diff
@@ -0,0 +1,282 @@
...
+      }
+    }
+    return watch.elapsed();
+  }
+
+  private List<java.time.Duration> runBenchmarkForUpdates(
+      final BenchmarkState server, int numberOfOperations) {
+    List<Duration> results = new ArrayList<>(numberOfOperations);
+    // Execute one update to make sure everything has been warmed up.
+    executeUpdate(server);
```
olavloite, line 245: Shouldn't this also call `executeWarmup`, so the number of warmup queries is respected for updates as well?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/BenchmarkingUtilityTest.java
```diff
@@ -0,0 +1,190 @@
...
+  }
+
+  /**
+   * A utility which bulk inserts 10^5 records into the database in batches. The method assumes that
+   * the instance/database/table is already created. It does not perform any admin operations.
+   *
+   * <p>Table schema used here: CREATE TABLE FOO ( id INT64 NOT NULL, BAZ INT64, BAR INT64, )
+   * PRIMARY KEY(id);
+   */
+  @Test
```
olavloite, line 94: Do we want this as an actual test that is executed during (nightly?) builds? If so, it should verify that the test actually succeeds by adding at least one assertion.

Or would it make more sense to move this method to the actual benchmark class, so it can more easily be executed together with a benchmark?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/BenchmarkingUtilityTest.java
arpan14: removed

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DefaultBenchmark.java
arpan14: Yes, removed.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DefaultBenchmark.java
arpan14: Yes, added this.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/BenchmarkingUtilityTest.java
arpan14: I have annotated `BenchmarkingUtilityTest` with `@Category(SlowTest.class)`. This will make sure its excluded from our pre-submits and nightly tests. Note that the data loading utility is a one time activity and need to be executed repeatedly along with the benchmarks.

Hence, I think it will be better to leave this as a standalone class.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DefaultBenchmark.java
```diff
@@ -55,17 +54,17 @@
  * <p>Below are a few considerations here: 1. We use all default options for this test because that
```
olavloite, line 54: super-nit (feel free to leave as is): Use `<ol>` and `<li>` tags to create an actual ordered (numbered) list.

olavloite (approved):
LGTM

PR #2759, opened 2023-12-15
test: refactor executor code with autogenerated protos

harshachinta:
The executor framework in Java depends on the proto from artifact uploaded through [cndb-client-testing-protos](https://github.com/googleapis/cndb-client-testing-protos/tree/main) repo.

Recently we have added OwlBot setup in #2754 to auto generate the protos similar to other spanner folders. #2756 is the result of this owlbot configuration where the protos for executor now lie with in this repo and gets updated automatically through owl-bot.

This PR replaces the references of earlier protos with the newly generated protos. The validation of this change has been performed on local systest which runs successfully.
https://screenshot.googleplex.com/B4dn7jKBi8Z6muM


google-cloud-spanner-executor/clirr-ignored-differences.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- see http://www.mojohaus.org/clirr-maven-plugin/examples/ignored-differences.html -->
+<differences>
+  <difference>
+    <differenceType>8001</differenceType>
+    <className>com/google/cloud/executor/spanner/CloudClientExecutor</className>
```
arpan14, line 6: why are we ignoring these differences?

google-cloud-spanner-executor/pom.xml
```diff
@@ -93,7 +93,10 @@
     <dependency>
       <groupId>com.google.api.grpc</groupId>
       <artifactId>proto-google-cloud-spanner-executor-v1</artifactId>
```
arpan14, line 95: Shouldn't we also remove `proto-google-cloud-spanner-executor-v1` dependency? I am assuming this the proto repository which we no longer need.

google-cloud-spanner-executor/src/main/java/com/google/cloud/spanner/executor/CloudClientExecutor.java
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package com.google.cloud.executor.spanner;
+package com.google.cloud.spanner.executor;
```
arpan14, line 17: Why are we modifying the package?

google-cloud-spanner-executor/clirr-ignored-differences.xml
harshachinta: This is because of change in the package name from earlier.
`com/google/cloud/executor/spanner` to `com/google/cloud/spanner/executor/`

This was done to maintain consistency with other class name in this repo
For ex: `com.google.cloud.spanner.*`

google-cloud-spanner-executor/pom.xml
harshachinta: Earlier we had this `artifactId` getting uploaded from [cndb-protos](https://github.com/googleapis/cndb-client-testing-protos/blob/db032a52db722a87522733cff3e314eed02ab600/pom.xml#L6) repo.
Now we will do it through the `proto-google-cloud-spanner-executor-v1 ` which is in [java-spanner](https://github.com/googleapis/java-spanner/tree/main/proto-google-cloud-spanner-executor-v1) itself.

google-cloud-spanner-executor/src/main/java/com/google/cloud/spanner/executor/CloudClientExecutor.java
harshachinta: Just to be consistent with the class names prefix we are maintaining in java-spanner repo.
Ex:
1. com.google.cloud.spanner.admin.database.v1
2. com.google.cloud.spanner.admin.instance.v1
3. com.google.cloud.spanner

google-cloud-spanner-executor/clirr-ignored-differences.xml
arpan14: Usually we don't exclude breaking changes. We only exclude things which `clirr` identified by mistake but we know its not breaking anything. In this case, `clirr` actually identified a breaking change.

google-cloud-spanner-executor/src/main/java/com/google/cloud/spanner/executor/CloudClientExecutor.java
arpan14: I understand the need to be consistent but not at the cost of introducing a breaking change. There are other policies a) major version upgrade, b) release with JDK upgrade which needs to be followed if we were to release a breaking change in Java. I think in this case, lets continue to have the same package structure. The lack of consistency is not a major problem (especially when I think of a breaking change).

google-cloud-spanner-executor/clirr-ignored-differences.xml
harshachinta: As discussed, we will not be modifying the package names to avoid breaking changes and ignore consistency.

google-cloud-spanner-executor/src/main/java/com/google/cloud/spanner/executor/CloudClientExecutor.java
harshachinta: Removed the changes as discussed.

arpan14 (approved):

PR #2766, opened 2023-12-27
feat: add support for Directed Read options

arpan14:
Customers can set the option at two level a) per RPC by passing `Options.directedRead`  or b) global setting in `SpannerOptions`. The option is not applicable for RW transactions. If customer sets the setting for a RW operation, they will receive an exception. 

**Sample Usage**
```
DatabaseClient client =
        spanner.getDatabaseClient(DatabaseId.of(TEST_PROJECT, TEST_INSTANCE, TEST_DATABASE));
    try (ResultSet resultSet =
        client.singleUse().executeQuery(SELECT1, Options.directedRead(DIRECTED_READ_OPTIONS1))) {
      while (resultSet.next()) {}
    }
```
```
Spanner spannerWithDirectedReadOptions =
        spanner
            .getOptions()
            .toBuilder()
            .setDirectedReadOption(DIRECTED_READ_OPTIONS2)
            .build()
            .getService();
    DatabaseClient client =
        spannerWithDirectedReadOptions.getDatabaseClient(
            DatabaseId.of(TEST_PROJECT, TEST_INSTANCE, TEST_DATABASE));
    TransactionRunner runner = client.readWriteTransaction();
    runner.run(
        transaction -> {
          try (ResultSet resultSet = transaction.executeQuery(SELECT1)) {
            while (resultSet.next()) {}
          }
          return null;
        });

```


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionImpl.java
```diff
@@ -255,6 +255,7 @@ public ReadContext singleUse(TimestampBound bound) {
             .setRpc(spanner.getRpc())
             .setDefaultQueryOptions(spanner.getDefaultQueryOptions(databaseId))
             .setDefaultPrefetchChunks(spanner.getDefaultPrefetchChunks())
+            .setDefaultDirectedReadOption(spanner.getOptions().getDirectedReadOptions())
```
olavloite, line 258: Should this also be set for 'BatchReadOnlyTransactionImpl'?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1153,6 +1159,19 @@ public Builder setAsyncExecutorProvider(CloseableExecutorProvider provider) {
...
     }
 
+    /**
+     * Sets the {@link DirectedReadOption} that specify which replicas or regions should be used for
+     * non-transactional reads or queries.
+     *
+     * <p>DirectedReadOptions set at the request level will take precedence over the options set
+     * using this method.
+     */
+    public Builder setDirectedReadOption(DirectedReadOptions directedReadOptions) {
```
olavloite, line 1169: Should we add a concrete example to this Javadoc for how to create a `DirectedReadOptions` instance? I do not feel that the `DirectedReadOptions` class and javadoc is self-explanatory.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
```diff
@@ -1542,6 +1601,79 @@ public void testExecuteReadWithTag() {
...
+        client.singleUse().read(READ_TABLE_NAME, KeySet.singleKey(Key.of(1L)), READ_COLUMN_NAMES)) {
+      while (resultSet.next()) {}
+    }
+
+    List<ReadRequest> requests = mockSpanner.getRequestsOfType(ReadRequest.class);
+    assertEquals(requests.size(), 1);
+    ReadRequest request = requests.get(0);
+    assertTrue(request.hasDirectedReadOptions());
+    assertEquals(DIRECTED_READ_OPTIONS2, request.getDirectedReadOptions());
+  }
```
olavloite, line 1648: Could we also add a test here to verify that setting a `DirectedReadOption` for a query overrides the default that is set on the `SpannerOptions`?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
```diff
@@ -2854,6 +2987,48 @@ public void testBackendPartitionQueryOptions() {
...
+              Options.directedRead(DIRECTED_READ_OPTIONS1));
+      try (ResultSet rs = transaction.execute(partitions.get(0))) {
+        // Just iterate over the results to execute the query.
+        while (rs.next()) {}
+      } finally {
+        transaction.cleanup();
+      }
+      // Check if the last query executed is a DeleteSessionRequest and the second last query
+      // executed is a ExecuteSqlRequest and was executed using a custom optimizer version and
+      // statistics package.
```
olavloite, line 3025: update the comment so it says that we verify that `DirectedReadOptions` are used.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
```diff
@@ -2854,6 +2987,48 @@ public void testBackendPartitionQueryOptions() {
       assertThat(executeSqlRequest.getQueryOptions().getOptimizerVersion()).isEqualTo("1");
       assertThat(executeSqlRequest.getQueryOptions().getOptimizerStatisticsPackage())
           .isEqualTo("custom-package");
+      assertThat(executeSqlRequest.getDirectedReadOptions()).isEqualTo(DIRECTED_READ_OPTIONS1);
+    }
+  }
+
+  @Test
+  public void testBackendPartitionReadOptions() {
+    // Use a Spanner instance with MinSession=0 and WriteFraction=0.0 to prevent background requests
```
olavloite, line 2996: The `WriteFraction=0.0` is no longer valid (we should remove it from the other test(s) as well)

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
```diff
@@ -2854,6 +2987,48 @@ public void testBackendPartitionQueryOptions() {
...
+      // Check if the last query executed is a DeleteSessionRequest and the second last query
+      // executed is a ExecuteSqlRequest and was executed using a custom optimizer version and
+      // statistics package.
+      List<AbstractMessage> requests = mockSpanner.getRequests();
+      assert requests.size() >= 2 : "required to have at least 2 requests";
+      assertThat(requests.get(requests.size() - 1)).isInstanceOf(DeleteSessionRequest.class);
+      assertThat(requests.get(requests.size() - 2)).isInstanceOf(ReadRequest.class);
+      ReadRequest readRequest = (ReadRequest) requests.get(requests.size() - 2);
+      assertThat(readRequest.getDirectedReadOptions()).isEqualTo(DIRECTED_READ_OPTIONS1);
     }
```
olavloite, line 3032: Assuming that default `DirectedReadOptions` should also apply to `BatchReadOnly` transactions; we should add a test for that as well.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITReadTest.java
```diff
@@ -77,6 +83,17 @@ public class ITReadTest {
   private static final Type TABLE_TYPE =
       Type.struct(
           StructField.of("key", Type.string()), StructField.of("stringvalue", Type.string()));
+  private static DirectedReadOptions DIRECTED_READ_OPTIONS =
+      DirectedReadOptions.newBuilder()
+          .setIncludeReplicas(
+              IncludeReplicas.newBuilder()
+                  .addReplicaSelections(
+                      ReplicaSelection.newBuilder()
+                          .setLocation("us-west1")
```
olavloite, line 92: Will this fail if someone tries to run this integration test on an instance config that does not have a read-only replica in `us-west1`? If so; could we handle that more gracefully? (Could we for example read the possible values from the actual instance config, and set one of those here?)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionImpl.java
harshachinta: +1
We need to set this option for `BatchReadOnlyTransaction` and also make change in `getExecuteSqlRequestBuilder` and `readInternalWithOptions` of AbstractReadContext class.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
```diff
@@ -623,6 +632,11 @@ ExecuteSqlRequest.Builder getExecuteSqlRequestBuilder(
     if (options.hasDataBoostEnabled()) {
       builder.setDataBoostEnabled(options.dataBoostEnabled());
     }
+    if (options.hasDirectedReadOptions()) {
+      builder.setDirectedReadOptions(options.directedReadOptions());
+    } else if (defaultDirectedReadOptions != null) {
```
harshachinta, line 637: I think we can replace this `else if` condition with `else`? I think by default builder has this field set to `null`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
rajatbhatta, line 1169: nit: `setDirectedReadOption` -> `setDirectedReadOptions`?

Please feel free to retain as is, if you feel the current naming is appropriate. No strong opinions, but let's just try to keep it consistent with `getDirectedReadOptions`?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -729,6 +733,7 @@ public static class Builder
     private String emulatorHost = System.getenv("SPANNER_EMULATOR_HOST");
     private boolean leaderAwareRoutingEnabled = true;
     private boolean attemptDirectPath = true;
+    private DirectedReadOptions directedReadOption;
```
rajatbhatta, line 736: `directedReadOption` -> `directedReadOptions`?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
```diff
@@ -117,6 +119,11 @@ B setClock(Clock clock) {
       return self();
     }
 
+    B setDefaultDirectedReadOption(DirectedReadOptions directedReadOptions) {
```
rajatbhatta, line 122: nit: `setDefaultDirectedReadOption` -> `setDefaultDirectedReadOptions`.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITDirectedReadsTest.java
```diff
@@ -0,0 +1,183 @@
...
+            .createTestDatabase("CREATE TABLE TEST (ID INT64, NAME STRING(100)) PRIMARY KEY (ID)");
+  }
+
+  @AfterClass
+  public static void tearDown() {
+    db.drop();
+  }
+
+  @Test
+  public void testReadWriteTransactionRunner_queryWithDirectedReadOptions_throwsError() {
```
harshachinta, line 83: nit: can we add `ViaRequest` to the method name?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITReadTest.java
harshachinta: If the provided replica is not available in that instance config, then the request will silently ignore the setting.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
arpan14: Sounds good, done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
arpan14: I wouldn't prefer to assume what the builder sets the default to. By this way we are being explicit that we honor default values if and only if the request values are not sent and the defaults are non-null.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionImpl.java
arpan14: Added this, as well as additional tests for client level options.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
arpan14: I have added one example of how the object is built. Our samples will also contain more usage patterns for each language. LMK if something else needs to be included.

olavloite: Thanks, this looks fine to me.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITReadTest.java
olavloite: It actually makes me worry a bit more, but more generally about the feature than about the test :-)

The fact that an invalid value is silently ignored means that the user does not get __any__ feedback about whether the feature is being used or not. If a user makes a typo for the location, then there is no feedback loop that will indicate that to the user, which again can make for very difficult debugging.
Could we file a request with the feature owner that there is some kind of feedback from the server about which replica was actually used to serve the read? Or do we already have that information in some of the metadata that we receive?

olavloite (approved):

harshachinta (approved):

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITReadTest.java
arpan14: @mayurkale22 Can you take a look here and answer this query? I think this is another reason why we should return the feedback in our backend response (similar to what we were discussing last evening)

mayurkale22: >> If the provided replica is not available in that instance config, then the request will silently ignore the setting.

In the earlier design, the plan was to ensure that the location and/or replica type specified in the directed reads settings does indeed belong to the instance configuration used, otherwise, return an `INVALID_ARGUMENT` error to indicate users that they are using the incorrect setting/location. However, this means that if the currently specified location(s) from the source instance configuration does not exist in the destination instance configuration, this would result in requests errors during the instance move/migration process. That's a poor user experience.

To support friction-free instance migration, we will eliminate the above constraint. So basically, if the location(s) provided in directed reads settings are not found in the instance configuration, the Spanner will fallback to default routing mechanism by sending the requests to the nearest available replica and hence this will ensure that the directed read requests do not fail during/after the instance migration. To tell users about incorrect location in the directed reads, we plan to expose a new metric. The metric can be used to understand misconfigured settings.

>> Could we file a request with the feature owner that there is some kind of feedback from the server about which replica was actually used to serve the read? 

We plan to expose this info in the metric for now. In the future we can expose in the API response itself.

PR #2767, opened 2023-12-29
perf: use virtual threads for Connection API executor and gRPC executors

olavloite:
Uses virtual threads instead of platform daemon threads for executor threads that are mainly I/O bound:
1. Connection executor threads
2. gRPC default executor threads
3. gRPC transport threads

Virtual threads are only supported on Java 21 and higher, and virtual threads are therefore only created when supported. For all other Java versions, standard platform threads are created instead.


olavloite:
Replaced by various other pull requests.

PR #2768, opened 2024-01-03
docs: samples and tests for instance APIs.

arpan14:
Adds samples and tests for auto-generated  Instance Admin APIs.


samples/install-without-bom/pom.xml
```diff
@@ -15,6 +15,7 @@
     <groupId>com.google.cloud.samples</groupId>
     <artifactId>shared-configuration</artifactId>
     <version>1.2.0</version>
+    <relativePath></relativePath>
```
arpan14, line 18: This change was done to mitigate below IDE error

```
'parent.relativePath' of POM com.google.cloud:spanner-snapshot:1.2.0 (/Users/arpanmishra/java-spanner/samples/snapshot/pom.xml) points at com.google.cloud:google-cloud-spanner-samples instead of com.google.cloud.samples:shared-configuration, please verify your project structure
```

samples/install-without-bom/pom.xml
```diff
@@ -144,7 +145,7 @@
           <systemPropertyVariables>
             <spanner.test.instance>java-client-integration-test</spanner.test.instance>
             <spanner.test.instance.mr>java-client-mr-integration-test</spanner.test.instance.mr>
-            <spanner.test.instance.config>nam15</spanner.test.instance.config>
+            <spanner.test.instance.config>nam11</spanner.test.instance.config>
```
arpan14, line 148: This change was needed because `nam15` does not have any optional replicas. And because of this existing integration tests were broken for `CreateInstanceConfigSample` class. Particularly below line
```
List<ReplicaInfo> readOnlyReplicas =
          ImmutableList.of(baseConfig.getOptionalReplicas().get(0));
```

I have modified this config and that fixes existing broken test.

samples/snippets/src/main/java/com/example/spanner/admin/generated/CreateInstanceConfigSample.java
```diff
@@ -0,0 +1,77 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.example.spanner.admin.generated;
+
+// [START spanner_create_instance_config]
```
olavloite, line 19: I don't know exactly what the plan is regarding when what will be merged, but we should make sure that we only have one set of region tags at any time in the main branch, as otherwise the documentation pages that automatically import the sample snippets will fail.

samples/snippets/src/main/java/com/example/spanner/admin/generated/CreateInstanceConfigSample.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+class CreateInstanceConfigSample {
+  static void createInstanceConfig() throws IOException {
+    // TODO(developer): Replace these variables before running the sample.
+    String projectId = "projects/my-project";
+    String baseInstanceConfig = "nam11";
+    String instanceConfigId = "custom-instance-config4";
+    String instanceConfigName = "projects/my-project/instanceConfigs/custom-instance-config4";
```
olavloite, line 38: nit: can we construct this string from the other variables in this method, so something like `String.format("projects/%s/instanceConfigs/%s",  projectId, instanceConfigId)`

samples/snippets/src/main/java/com/example/spanner/admin/generated/CreateInstanceConfigSample.java
```diff
@@ -0,0 +1,77 @@
...
+    createInstanceConfig(projectId, baseInstanceConfig, instanceConfigId, instanceConfigName);
+  }
+
+  static void createInstanceConfig(
+      String projectId, String baseInstanceConfig, String instanceConfigId, String instanceConfigName) throws IOException {
+    InstanceAdminClient instanceAdminClient = InstanceAdminClient.create();
```
olavloite, line 44: Not relevant in this PR, but I just thought of it when I saw this: We need additional samples for showing how to create the auto-generated clients when you want to use some non-standard options. The things that I can think of are:
1. Connecting to the emulator. Setting the `SPANNER_EMULATOR_HOST=localhost:9010` won't do it anymore. (Although we should consider creating some kind of standard helper method that customers can use that would do that, and add that method to the client library)
2. Connecting with custom credentials. That could then also be part of the above-mentioned helper method.

samples/snippets/src/main/java/com/example/spanner/admin/generated/CreateInstanceConfigSample.java
```diff
@@ -0,0 +1,77 @@
...
+     * The replicas for the custom instance configuration must include all the replicas of the base
+     * configuration, in addition to at least one from the list of optional replicas of the base
+     * configuration.
+     */
+    final List<ReplicaInfo> replicas =
+        Stream.concat(baseConfig.getReplicasList().stream(),
+            baseConfig.getOptionalReplicasList().stream().limit(1)).collect(Collectors.toList());
+    final InstanceConfig instanceConfig =
+        InstanceConfig.newBuilder().setName(instanceConfigName).setBaseConfig(baseInstanceConfig)
+            .setDisplayName(instanceConfigId).addAllReplicas(replicas).build();
```
olavloite, line 57: nit: 
```suggestion
            .setDisplayName("Custom Instance Configuration").addAllReplicas(replicas).build();
```

(or anything that shows that the display name can be anything)

samples/snippets/src/main/java/com/example/spanner/admin/generated/CreateInstanceConfigSample.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+class CreateInstanceConfigSample {
+  static void createInstanceConfig() throws IOException {
+    // TODO(developer): Replace these variables before running the sample.
+    String projectId = "projects/my-project";
```
olavloite, line 35: We should try to be 100% consistent with this, as it often causes confusion:
1. ProjectId (and other ids) are always without the prefix (e.g. in this case `"my-project"`)
2. ProjectName (and other names) are always fully qualified (e.g. in this case `"projects/my-project"`)

So either the variable name should be changed to `projectName`, or the value should be changed to `my-project`.

samples/snippets/src/main/java/com/example/spanner/admin/generated/CreateInstanceWithAutoscalingConfigExample.java
```diff
@@ -0,0 +1,84 @@
...
+            .setAutoscalingLimits(
+                AutoscalingConfig.AutoscalingLimits.newBuilder().setMinNodes(1).setMaxNodes(2))
+            .setAutoscalingTargets(
+                AutoscalingConfig.AutoscalingTargets.newBuilder()
+                    .setHighPriorityCpuUtilizationPercent(65)
+                    .setStorageUtilizationPercent(95))
+            .build();
+    Instance instance =
+        Instance.newBuilder()
+            .setAutoscalingConfig(autoscalingConfig)
```
olavloite, line 58: nit: Should we explicitly call out in a comment that we are not setting a node count or processing units here, as we are seting an AutoscalingConfig?

samples/snippets/src/main/java/com/example/spanner/admin/generated/CreateInstanceWithProcessingUnitsExample.java
```diff
@@ -0,0 +1,74 @@
+/*
+ * Copyright 2021 Google LLC
```
olavloite, line 2: nit: Copyright year (maybe we should set all of them to 2024?)

samples/snippets/src/main/java/com/example/spanner/admin/generated/DeleteInstanceConfigSample.java
```diff
@@ -0,0 +1,48 @@
...
+// [START spanner_delete_instance_config]
+import com.google.cloud.spanner.SpannerException;
+import com.google.cloud.spanner.admin.instance.v1.InstanceAdminClient;
+import com.google.spanner.admin.instance.v1.DeleteInstanceConfigRequest;
+import java.io.IOException;
+
+class DeleteInstanceConfigSample {
+  static void deleteInstanceConfig() throws IOException {
+    // TODO(developer): Replace these variables before running the sample.
+    String instanceConfigId = "projects/my-project/instanceConfigs/custom-user-config";
```
olavloite, line 28: nit: this is not an id, but a name

samples/snippets/src/main/java/com/example/spanner/admin/generated/GetInstanceConfigSample.java
```diff
@@ -0,0 +1,45 @@
...
+
+import com.google.cloud.spanner.admin.instance.v1.InstanceAdminClient;
+import com.google.spanner.admin.instance.v1.InstanceConfig;
+import java.io.IOException;
+
+public class GetInstanceConfigSample {
+
+  static void getInstanceConfig() throws IOException {
+    // TODO(developer): Replace these variables before running the sample.
+    final String instanceConfigName = "nam6";
```
olavloite, line 29: nit: not a name, but an id

samples/snippets/src/main/java/com/example/spanner/admin/generated/GetInstanceConfigSample.java
```diff
@@ -0,0 +1,45 @@
...
+    getInstanceConfig(instanceConfigName);
+  }
+
+  static void getInstanceConfig(String instanceConfigName) throws IOException {
+    final InstanceAdminClient instanceAdminClient = InstanceAdminClient.create();
+
+    final InstanceConfig instanceConfig = instanceAdminClient.getInstanceConfig(instanceConfigName);
```
olavloite, line 36: I don't think this method would work with the example value that we have in our sample (`nam6`), as that is not a fully qualified name.

samples/snippets/src/main/java/com/example/spanner/admin/generated/ListInstanceConfigOperationsSample.java
```diff
@@ -0,0 +1,62 @@
...
+
+  static void listInstanceConfigOperations(String projectName) throws IOException {
+    final InstanceAdminClient instanceAdminClient = InstanceAdminClient.create();
+    try {
+      System.out.printf(
+          "Getting list of instance config operations for project %s...\n",
+          projectName);
+      final ListInstanceConfigOperationsRequest request =
+          ListInstanceConfigOperationsRequest.newBuilder().setParent(projectName).setFilter("(metadata.@type=type.googleapis.com/"
+              + "google.spanner.admin.instance.v1.CreateInstanceConfigMetadata)").build();
```
olavloite, line 42: Does the class `CreateInstanceConfigMetadata` have a static method or constant that would also return this magic string? If so, could we use that in this sample?

samples/snippets/src/main/java/com/example/spanner/admin/generated/UpdateInstanceConfigSample.java
```diff
@@ -0,0 +1,66 @@
...
+    final com.google.cloud.spanner.admin.instance.v1.InstanceAdminClient instanceAdminClient =
+        com.google.cloud.spanner.admin.instance.v1.InstanceAdminClient.create();
+    final InstanceConfig instanceConfig =
+        InstanceConfig.newBuilder()
+            .setName(instanceConfigName)
+            .setDisplayName("updated custom instance config")
+            .putLabels("updated", "true").build();
+    final UpdateInstanceConfigRequest updateInstanceConfigRequest =
+        UpdateInstanceConfigRequest.newBuilder().setInstanceConfig(instanceConfig).setUpdateMask(
+            FieldMask.newBuilder().addAllPaths(ImmutableList.of("display_name", "labels")).build()).build();
```
olavloite, line 46: This is worth an explanatory comment: Why are we adding a `FieldMask`? Is it required?

samples/snippets/src/main/java/com/example/spanner/admin/generated/CreateInstanceConfigSample.java
arpan14: * So given we have created a new path, when we merge this PR our existing documentation pages will not fail. They will continue to fetch the old samples (which continue to exist in the current path)
* After we have merged all PRs and received all other approvals, there will be a last PR which will swap the paths of the sample. This PR will archive the old samples and these new samples will show up on public docs.
* I've verified this. Last week we merged the first Java PR and now on `main` we have two samples with the same tag - https://github.com/search?q=repo%3Agoogleapis%2Fjava-spanner+spanner_create_instance%5D&type=code and public doc is showing the old sample as expected - https://cloud.google.com/spanner/docs/create-manage-instances

samples/snippets/src/main/java/com/example/spanner/admin/generated/CreateInstanceConfigSample.java
arpan14: Sure, I've taken a note. Will add these new samples in a new PR.

samples/snippets/src/main/java/com/example/spanner/admin/generated/CreateInstanceConfigSample.java
arpan14: Yes, I have now modified all new samples to use the same convention (projectId, configId, instanceId) and using utilities like `ProjectName.of()`, `InstanceConfigName.of()`, `InstanceName.of()` . I see that this will be helpful as customers don't need to understand the difference between id and name.

samples/snippets/src/main/java/com/example/spanner/admin/generated/GetInstanceConfigSample.java
arpan14: Your'e right. The integration test was using the fully qualified name. I have now made this consistent.

samples/snippets/src/main/java/com/example/spanner/admin/generated/ListInstanceConfigOperationsSample.java
arpan14: No I did not find any such static methods or constants. Also in our existing samples, customers are already accustomed to passing the full string.

samples/snippets/src/main/java/com/example/spanner/admin/generated/UpdateInstanceConfigSample.java
arpan14: While this is available in the proto definition, I think it makes sense to add here as well since `FieldMask` is a new class which older customers don't know of.

samples/snippets/src/main/java/com/example/spanner/admin/generated/CreateInstanceWithAutoscalingConfigExample.java
arpan14: This is there is the proto documentation, but I don't see any harm in communicating again over here. Added it here as well.

samples/snippets/src/main/java/com/example/spanner/admin/generated/CreateInstanceConfigSample.java
```diff
@@ -0,0 +1,86 @@
...
+            .setBaseConfig(baseInstanceConfigName.toString())
+            .setDisplayName("Instance Configuration").addAllReplicas(replicas).build();
+    final CreateInstanceConfigRequest createInstanceConfigRequest =
+        CreateInstanceConfigRequest.newBuilder().setParent(ProjectName.of(projectId).toString())
+            .setInstanceConfigId(instanceConfigId).setInstanceConfig(instanceConfig).build();
+    try {
+      System.out.printf("Waiting for create operation for %s to complete...\n", instanceConfigName);
+      InstanceConfig instanceConfigResult =
+          instanceAdminClient.createInstanceConfigAsync(
+              createInstanceConfigRequest).get(5, TimeUnit.MINUTES);
```
lqiu96, line 74: Any significance of explicitly calling out 5 min for this LRO? This LRO (by default) has a total timeout of 5 minutes as specified here: https://github.com/googleapis/java-spanner/blob/58f94b200276d879f83e4432716b49baf3206226/google-cloud-spanner/src/main/java/com/google/cloud/spanner/admin/instance/v1/stub/InstanceAdminStubSettings.java#L916-L925

I think it would be beneficial to add a comment describing the underlying LRO behavior, but I don't think you need to explicitly call with `.get(5min)`.

samples/snippets/src/main/java/com/example/spanner/admin/generated/CreateInstanceWithAutoscalingConfigExample.java
```diff
@@ -0,0 +1,87 @@
...
+
+  static void createInstance() throws IOException {
+    // TODO(developer): Replace these variables before running the sample.
+    String projectId = "my-project";
+    String instanceId = "my-instance";
+    createInstance(projectId, instanceId);
+  }
+
+  static void createInstance(String projectId, String instanceId) throws IOException {
+    InstanceAdminClient instanceAdminClient = InstanceAdminClient.create();
```
lqiu96, line 40: I know you mentioned to me before that some samples have a way to close these clients once they're done running. I don't see that part here (and I'm unaware how it all get tied together). Just wanted to make sure that the clients are closed at the end.

samples/snippets/src/main/java/com/example/spanner/admin/generated/UpdateInstanceConfigSample.java
```diff
@@ -0,0 +1,80 @@
...
+        UpdateInstanceConfigRequest.newBuilder()
+            .setInstanceConfig(instanceConfig)
+            .setUpdateMask(
+                FieldMask.newBuilder().addAllPaths(ImmutableList.of("display_name", "labels"))
+                    .build()).build();
+    try {
+      System.out.printf("Waiting for update operation on %s to complete...\n", instanceConfigName);
+      InstanceConfig instanceConfigResult =
+          instanceAdminClient.updateInstanceConfigAsync(
+              updateInstanceConfigRequest).get(5, TimeUnit.MINUTES);
```
lqiu96, line 65: Same question with this.

samples/snippets/src/main/java/com/example/spanner/admin/generated/CreateInstanceWithAutoscalingConfigExample.java
arpan14: `I know you mentioned to me before that some samples have a way to close these clients once they're done running.`

I previously mentioned this for the integration tests code. I see over here we are not closing the resource. I'll add a `try-with-resources` block in all the samples.

samples/snippets/src/main/java/com/example/spanner/admin/generated/CreateInstanceConfigSample.java
arpan14: The code you have linked is for `RetrySetting`. A `get()` invocation on a `Future` object is a different thing. By default, the `get()` will block infinitely. And so it is preferred to specify a timeout.

samples/snippets/src/main/java/com/example/spanner/admin/generated/CreateInstanceConfigSample.java
olavloite: Ack

samples/snippets/src/main/java/com/example/spanner/admin/generated/CreateInstanceConfigSample.java
olavloite: Thanks 👍

olavloite (approved):

lqiu96 (approved):

PR #2775, opened 2024-01-09
docs: samples and tests for database Admin APIs.

arpan14:
Adds samples and tests for auto-generated Database Admin APIs.


samples/snippets/src/main/java/com/example/spanner/admin/generated/AddAndDropDatabaseRole.java
```diff
@@ -0,0 +1,75 @@
...
+    String databaseId = "my-database";
+    String parentRole = "parent_role";
+    String childRole = "child_role";
+    addAndDropDatabaseRole(projectId, instanceId, databaseId, parentRole, childRole);
+  }
+
+  static void addAndDropDatabaseRole(
+      String projectId, String instanceId, String databaseId, String parentRole, String childRole)
+      throws IOException {
+    final DatabaseAdminClient databaseAdminClient = DatabaseAdminClient.create();
```
olavloite, line 44: These samples do not get the admin clients from a `Spanner` instance. That means that anyone who follows this sample would need to configure the emulator manually. I think that we would want to update these samples to use clients that will automatically use the emulator.

Also, this client is not used in a try-with-resources block, and it is also not closed in any other way. Good practice in a case like this would be to close the client when you are done.

samples/snippets/src/main/java/com/example/spanner/admin/generated/AddJsonbColumnSample.java
```diff
@@ -0,0 +1,49 @@
...
+
+  static void addJsonbColumn(String projectId, String instanceId, String databaseId)
+      throws InterruptedException, ExecutionException, IOException {
+    final DatabaseAdminClient databaseAdminClient = DatabaseAdminClient.create();
+
+    // Wait for the operation to finish.
+    // This will throw an ExecutionException if the operation fails.
+    databaseAdminClient.updateDatabaseDdlAsync(
+        DatabaseName.of(projectId, instanceId, databaseId),
+        ImmutableList.of("ALTER TABLE Venues ADD COLUMN VenueDetails JSONB")).get();
```
olavloite, line 45: I think that we should add a comment to clarify that `jsonb` is only supported on PostgreSQL-dialect databases.

samples/snippets/src/main/java/com/example/spanner/admin/generated/CreateDatabaseWithVersionRetentionPeriodSample.java
```diff
@@ -0,0 +1,71 @@
...
+      String instanceId, String databaseId, String versionRetentionPeriod) throws IOException {
+    DatabaseAdminClient databaseAdminClient = DatabaseAdminClient.create();
+
+    try {
+      CreateDatabaseRequest request =
+          CreateDatabaseRequest.newBuilder()
+              .setParent(InstanceName.of(projectId, instanceId).toString())
+              .setCreateStatement("CREATE DATABASE `" + databaseId + "`")
+              .addAllExtraStatements(Lists.newArrayList("ALTER DATABASE " + "`" + databaseId + "`"
+                  + " SET OPTIONS ( version_retention_period = '" + versionRetentionPeriod + "' )"))
```
olavloite, line 54: I think that these statements would be more readable if we used `String.format(...)` instead of string concatenation.

samples/snippets/src/main/java/com/example/spanner/admin/generated/EnableFineGrainedAccess.java
```diff
@@ -0,0 +1,106 @@
...
+            .addAll(policy.getBindingsList())
+            .add(binding1)
+            .add(binding2)
+            .build();
+    Policy policyWithConditions =
+        Policy.newBuilder()
+            .setVersion(policyVersion)
+            .setEtag(policy.getEtag())
+            .addAllBindings(bindings)
+            .build();
```
olavloite, line 96: I assume that most of this is copy-pasted from a previous sample, and as such any comments on the contents are a bit out of scope in this case: But I think that this sample needs a lot more comments and explanation for all of this means. What is the difference between versions 0, 1, and 3. What do the bindings that are being set mean? etc.

samples/snippets/src/main/java/com/example/spanner/admin/generated/AddAndDropDatabaseRole.java
arpan14: I am awaiting the BOM release before I can raise one PR which will update all the samples to use the new methods that we created.

samples/snippets/src/main/java/com/example/spanner/admin/generated/EnableFineGrainedAccess.java
arpan14: Yeah, all of this is copy pasted from existing sample. I have created b/325368620 to fast follow up on this one.

samples/snippets/src/main/java/com/example/spanner/admin/generated/AddJsonbColumnSample.java
arpan14: Done

samples/snippets/src/main/java/com/example/spanner/admin/generated/CreateDatabaseWithVersionRetentionPeriodSample.java
arpan14: all of this is copy pasted from existing sample. I have created b/325368620 to fast follow up on this one.

samples/snippets/src/main/java/com/example/spanner/admin/generated/AddAndDropDatabaseRole.java
olavloite: Makes sense, thanks!

olavloite (approved):

PR #2155, opened 2022-11-09
feat: Support for Proto Messages & Enums

gauravpurohit06:
To add support for Proto Columns which includes messages and enums.

PR #2779, opened 2024-01-10
feat: add support for Proto Columns

harshachinta:
[DO NOT MERGE]

This PR has the changes from
1. https://github.com/googleapis/java-spanner/pull/2155
2. https://github.com/googleapis/java-spanner/pull/2212
3. https://github.com/googleapis/java-spanner/pull/2277


google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
```diff
@@ -699,6 +717,30 @@ public boolean isNull(int columnIndex) {
...
+          "Proto message may not be null. Use MyProtoClass.getDefaultInstance() as a parameter value.");
+      try {
+        return (T)
+            message
+                .toBuilder()
+                .mergeFrom(((LazyByteArray) rowData.get(columnIndex)).getByteArray().toByteArray())
+                .build();
+      } catch (InvalidProtocolBufferException e) {
+        throw SpannerExceptionFactory.asSpannerException(e);
+      }
```
olavloite, line 733: I think that this would be more efficient, as we would then not copy the byte array around as many times. Instead, it just creates a stream of bytes from the `base64` string that can be decoded directly.

```suggestion
      try {
        return (T)
            message
                .toBuilder()
                .mergeFrom(
                    Base64.getDecoder()
                        .wrap(
                            CharSource.wrap(((LazyByteArray) rowData.get(columnIndex)).base64String)
                                .asByteSource(StandardCharsets.UTF_8)
                                .openStream()))
                .build();
      } catch (IOException ioException) {
        throw SpannerExceptionFactory.asSpannerException(ioException);
      }
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
```diff
@@ -891,6 +943,56 @@ protected List<String> getJsonListInternal(int columnIndex) {
...
+        List<T> protoMessagesList = new ArrayList<>(bytesArray.size());
+        for (LazyByteArray protoMessageBytes : bytesArray) {
+          if (protoMessageBytes == null) {
+            protoMessagesList.add(null);
+          } else {
+            protoMessagesList.add(
+                (T)
+                    message
+                        .toBuilder()
+                        .mergeFrom(protoMessageBytes.getByteArray().toByteArray())
```
olavloite, line 966: Here also: I think we should apply the same optimization as mentioned above

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Type.java
```diff
@@ -127,6 +127,24 @@ public static Type pgJsonb() {
     return TYPE_PG_JSONB;
   }
 
+  /**
+   * To get the descriptor for the {@code PROTO} type.
+   *
+   * @param protoTypeFqn Proto full name
```
olavloite, line 133: Could we add an example for the type of string that we expect here?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Value.java
```diff
@@ -450,6 +536,85 @@ public static Value pgJsonbArray(@Nullable Iterable<String> v) {
...
+      @Nullable Iterable<AbstractMessage> v, Descriptor descriptor) {
+    if (v == null) {
+      return new ProtoMessageArrayImpl(true, null, descriptor.getFullName());
+    }
+
+    List<ByteArray> serializedArray = new ArrayList<>();
+    v.forEach(
+        (message) -> {
+          if (message != null) {
+            serializedArray.add(ByteArray.copyFrom(message.toByteArray()));
```
olavloite, line 557: nit: It would have been great if we could have done this conversion without copying the byte array twice (`message.toByteArray()` creates a `new byte[]`, and the `copyFrom` method then makes a copy of that again).

I had a quick look, and could not find anything, so feel free to ignore.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Value.java
```diff
@@ -1463,6 +1677,19 @@ public ByteArray getBytes() {
...
     }
 
+    @Override
+    public <T extends AbstractMessage> T getProtoMessage(T m) {
+      Preconditions.checkNotNull(
+          m,
+          "Proto message may not be null. Use MyProtoClass.getDefaultInstance() as a parameter value.");
+      checkNotNull();
+      try {
+        return (T) m.toBuilder().mergeFrom(value.getByteArray().toByteArray()).build();
```
olavloite, line 1687: Here also: Any possibility that we can do this conversion without copying to a byte array multiple times?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Value.java
```diff
@@ -1480,6 +1707,75 @@ void valueToString(StringBuilder b) {
...
+    }
+
+    @Override
+    public <T extends AbstractMessage> T getProtoMessage(T m) {
+      Preconditions.checkNotNull(
+          m,
+          "Proto message may not be null. Use MyProtoClass.getDefaultInstance() as a parameter value.");
+      checkNotNull();
+      try {
+        return (T) m.toBuilder().mergeFrom(value.toByteArray()).build();
```
olavloite, line 1729: Here also: Can we do a more direct conversion?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Value.java
```diff
@@ -1962,6 +2276,31 @@ public List<ByteArray> getBytesArray() {
...
+      try {
+        List<T> protoMessagesList = new ArrayList<>(value.size());
+        for (LazyByteArray protoMessageBytes : value) {
+          if (protoMessageBytes == null) {
+            protoMessagesList.add(null);
+          } else {
+            protoMessagesList.add(
+                (T)
+                    m.toBuilder()
+                        .mergeFrom(protoMessageBytes.getByteArray().toByteArray())
```
olavloite, line 2294: Here also: Would it be possible to apply the same kind of optimization as above?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/RandomResultSetGenerator.java
```diff
@@ -40,73 +45,104 @@
...
+
+    // TODO(harsha): add test corresponding to this in testGetAllTypesAsString()
+    // appendProtoTypes(types, dialect);
+    Type[] typeArray = new Type[types.size()];
+    typeArray = types.toArray(typeArray);
+    return typeArray;
+  }
+
+  /** To append Proto & Enum types * */
+  private static void appendProtoTypes(List<Type> types, Dialect dialect) {
```
olavloite, line 125: This does not seem to be used. Could we revert the changes in this class if they are not needed/used?

jitpack.yml
```diff
@@ -0,0 +1,2 @@
+jdk:
+  - openjdk11
```
olavloite, line 2: Is this something that we want to include in this PR? If so, why?

olavloite (approved):
Looks generally good to me, but with some requests for optimizations. Feel free to defer those to a later PR if this is urgent to get merged.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Value.java
harshachinta: @olavloite I have applied the optimization suggested above to this case too. Please let me know if this is correct.
```
value.getByteArray().toByteArray()
```
is changed to 
```
Base64.getDecoder()
                .wrap(
                    CharSource.wrap(value.getBase64String())
                        .asByteSource(StandardCharsets.UTF_8)
                        .openStream())
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
harshachinta: Thanks for suggesting this optimization, @olavloite. I agree with you here that we are copying the byte array multiple times which could affect performance when proto messages are big.
I have made the changes in the same PR.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
harshachinta: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Type.java
harshachinta: Added

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Value.java
harshachinta: I have added the optimization suggested above to this.

jitpack.yml
harshachinta: We don't need this file. I have removed it.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/RandomResultSetGenerator.java
harshachinta: We need this for one test. I have made the changes in this PR https://github.com/googleapis/java-spanner/pull/2845

PR #2789, opened 2024-01-19
chore: add support for virtual threads to Connection API

olavloite:
Adds support for using virtual threads in the Connection API. Virtual
threads can be enabled for two things:
1. As the StatementExecutor thread for each connection.
2. As the gRPC transport thread pool.

Both options can (for now) only be set in the Connection API.

Setting any of these options only has any effect if the application is
running on Java 21 or higher.


google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ReadWriteTransaction.java
```diff
@@ -80,6 +81,7 @@ class ReadWriteTransaction extends AbstractMultiUseTransaction {
   private static final String MAX_INTERNAL_RETRIES_EXCEEDED =
       "Internal transaction retry maximum exceeded";
   private static final int MAX_INTERNAL_RETRIES = 50;
+  private final ReentrantLock abortedLock = new ReentrantLock();
```
olavloite, line 84: This is changed to use a `ReentrantLock` instead of a synchronized block, because Java virtual threads cannot be unmounted when they are inside a synchronized block.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/ThreadFactoryUtil.java
```diff
@@ -0,0 +1,99 @@
...
+   * platform thread factory that creates daemon threads. Virtual threads are supported from JDK21.
+   *
+   * @param baseNameFormat the base name format for the threads, '-%d' will be appended to the
+   *     actual thread name format
+   * @param tryVirtualThreads whether to try to use virtual threads if available or not
+   * @return a {@link ThreadFactory} that produces virtual threads (Java 21 or higher) or daemon
+   *     threads
+   */
+  @InternalApi
+  public static ThreadFactory createVirtualOrDaemonThreadFactory(
```
ankiaga, line 43: Aren't virtual thread always daemon? Should we rename to createVirtualOrPlatformThreadFactory and also replace daemon by platform throughout PR if true

google-cloud-spanner/src/main/java/com/google/cloud/spanner/ThreadFactoryUtil.java
```diff
@@ -0,0 +1,99 @@
...
+    return new ThreadFactoryBuilder().setDaemon(true).setNameFormat(baseNameFormat + "-%d").build();
+  }
+
+  /**
+   * Tries to create a {@link ThreadFactory} that creates virtual threads. Returns null if virtual
+   * threads are not supported on this JVM.
+   */
+  @InternalApi
+  @Nullable
+  public static ThreadFactory tryCreateVirtualThreadFactory(String baseNameFormat) {
```
ankiaga, line 60: Does this needs to be public?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/ThreadFactoryUtil.java
olavloite: Virtual threads are indeed always daemon threads. The reason that I chose this name is to indicate that the thread factory will create either virtual or daemon (platform) threads. Put another way: Calling the method `createVirtualOrPlatformThreadFactory` does not indicate that the platform threads will be daemon threads. I renamed it to `createVirtualOrPlatformDaemonThreadFactory`

google-cloud-spanner/src/main/java/com/google/cloud/spanner/ThreadFactoryUtil.java
olavloite: Yes, because we want to use it in a number of places:
1. In this package to create a virtual thread executor for gRPC (if configured). For that, it could have been package-private.
2. In the `.connection` package, so we can use it to create the thread that is used by connections to execute statements.
3. In some of the libraries that depend on this client library. Specifically in PGAdapter and the JDBC driver.

ankiaga (approved):

PR #2790, opened 2024-01-19
chore: use a cache for the statement parser

olavloite:
Add a cache for the statement parser that is used in the Connection API. This will reduce the number of times that a SQL string needs to be parsed by the JDBC driver and PGAdapter.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/AbstractStatementParser.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/AbstractStatementParser.java
@@ -25,6 +25,10 @@ import com.google.cloud.spanner.Statement;
 import com.google.cloud.spanner.connection.StatementResult.ClientSideStatementType;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheStats;
+import com.google.common.cache.Weigher;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.spanner.v1.ExecuteSqlRequest.QueryOptions;
@@ -58,6 +62,13 @@ public abstract class AbstractStatementParser {
               SpannerStatementParser.class,
               Dialect.POSTGRESQL,
               PostgreSQLStatementParser.class);
+
+  @VisibleForTesting
+  static void resetParsers() {
+    synchronized (lock) {
+      INSTANCES.clear();
+    }
+  }
 
   /**
    * Get an instance of {@link AbstractStatementParser} for the specified dialect.
@@ -171,7 +182,7 @@ public abstract class AbstractStatementParser {
     private static ParsedStatement query(
         Statement statement, String sqlWithoutComments, QueryOptions defaultQueryOptions) {
       return new ParsedStatement(
-          StatementType.QUERY, statement, sqlWithoutComments, defaultQueryOptions, false);
+          StatementType.QUERY, null, statement, sqlWithoutComments, defaultQueryOptions, false);
     }
 
     private static ParsedStatement update(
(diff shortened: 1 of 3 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/AbstractStatementParser.java
```diff
@@ -361,8 +392,58 @@ ClientSideStatement getClientSideStatement() {
   static final Set<String> dmlStatements = ImmutableSet.of("INSERT", "UPDATE", "DELETE");
   private final Set<ClientSideStatementImpl> statements;
 
+  /** The default maximum size of the statement cache in Mb. */
+  public static final int DEFAULT_MAX_STATEMENT_CACHE_SIZE_MB = 20;
+
+  private static int getMaxStatementCacheSize() {
+    String stringValue = System.getProperty("spanner.statement_cache_size");
```
ankiaga, line 399: Should we also append append `mb` in the property name?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/AbstractStatementParser.java
```diff
@@ -361,8 +392,58 @@ ClientSideStatement getClientSideStatement() {
   static final Set<String> dmlStatements = ImmutableSet.of("INSERT", "UPDATE", "DELETE");
   private final Set<ClientSideStatementImpl> statements;
 
+  /** The default maximum size of the statement cache in Mb. */
+  public static final int DEFAULT_MAX_STATEMENT_CACHE_SIZE_MB = 20;
```
ankiaga, line 396: Why public?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/AbstractStatementParser.java
```diff
@@ -361,8 +392,58 @@ ClientSideStatement getClientSideStatement() {
...
+  private final Cache<String, ParsedStatement> statementCache;
+
   AbstractStatementParser(Set<ClientSideStatementImpl> statements) {
     this.statements = Collections.unmodifiableSet(statements);
+    int maxCacheSize = getMaxStatementCacheSize();
+    if (maxCacheSize > 0) {
+      CacheBuilder<String, ParsedStatement> cacheBuilder =
+          CacheBuilder.newBuilder()
+              // Set the max size to (approx) 20MB (by default).
+              .maximumWeight(getMaxStatementCacheSize() * 1024L * 1024L)
```
ankiaga, line 429: nit: use variable maxCacheSize

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/AbstractStatementParser.java
olavloite: This feature is used in the Connection API, which itself is not really a public API. Instead, the JDBC driver and PGAdapter are, and these sometimes want to use this default as well, for example so they can include it in automated help text and/or the list of properties that they support.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/AbstractStatementParser.java
```diff
@@ -361,8 +392,58 @@ ClientSideStatement getClientSideStatement() {
...
+    int maxCacheSize = getMaxStatementCacheSize();
+    if (maxCacheSize > 0) {
+      CacheBuilder<String, ParsedStatement> cacheBuilder =
+          CacheBuilder.newBuilder()
+              // Set the max size to (approx) 5MB (by default).
+              .maximumWeight(maxCacheSize * 1024L * 1024L)
+              // We do length*2 because Java uses 2 bytes for each char.
+              .weigher(
+                  (Weigher<String, ParsedStatement>)
+                      (key, value) -> 2 * key.length() + 2 * value.sqlWithoutComments.length())
```
ankiaga, line 433: Why are we only considering `value.sqlWithoutComments` as there are other attributes also stored as value? Is it because those have constant small size?

ankiaga (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/AbstractStatementParser.java
olavloite: Yes, exactly. The other properties are all small, and all have fixed size. The `statement` (so the original SQL statement) is also not stored as part of the value in the cache, as that is the key that we are using.

PR #2791, opened 2024-01-19
test: add tests for concatenated literals

olavloite:
Adds tests to verify that concatenated literals work with the statement parser.

### DIFF
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/StatementParserTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/StatementParserTest.java
@@ -1579,6 +1579,25 @@ public class StatementParserTest {
   }
 
   @Test
+  public void testConcatenatedLiterals() {
+    assumeTrue(dialect == Dialect.GOOGLE_STANDARD_SQL);
+
+    assertTrue(parser.isUpdateStatement("UPDATE foo SET name='foo' 'bar' WHERE ID=1"));
+    assertTrue(parser.isQuery("SELECT 'boo'   'bar' FROM my_table WHERE value='lit1''lit2'"));
+    assertTrue(
+        parser.checkReturningClause(
+            "insert into my_table (id, value) values (1, 'value' '1') then return id"));
+    assertTrue(parser.isUpdateStatement("UPDATE foo SET name=b'foo' b'bar' WHERE ID=1"));
+
+    assertEquals(
+        1,
+        parser.convertPositionalParametersToNamedParameters(
+                '?', "select 'foo?''bar?' where id=? and value='?'  '?'")
+            .numberOfParameters,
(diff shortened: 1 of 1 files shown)


google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/StatementParserTest.java
```diff
@@ -1578,6 +1578,25 @@ int skipMultiLineComment(String sql, int startIndex) {
     return PostgreSQLStatementParser.skipMultiLineComment(sql, startIndex, null);
   }
 
+  @Test
+  public void testConcatenatedLiterals() {
+    assumeTrue(dialect == Dialect.GOOGLE_STANDARD_SQL);
+
+    assertTrue(parser.isUpdateStatement("UPDATE foo SET name='foo' 'bar' WHERE ID=1"));
```
harshachinta, line 1585: @olavloite 
Can you help me understand what does concatenated literal mean and what is the use case for this?
From this example I guess `'foo' 'bar'` represents a concatenated literal.

Also where is this case handled in code in [AbstractStatementParser.java](https://github.com/googleapis/java-spanner/blob/main/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/AbstractStatementParser.java)

olavloite: Yes, `'foo' 'bar'` is a 'concatenated literal', and is a way to write literals that many other databases already support, and that Spanner will also (probably) soon support.

We were therefore asked to verify that it would not cause any problems in the Connection API. It is not really 'handled' in `AbstractStatementParser`, other than that:
1. When we parse SQL strings, we automatically skip through all string literals, and ignore any tokens there that might have contained something that would have been a query parameter outside of the literal. That is: `select * from my_table where value='some-string @id'` is correctly parsed as not having any query parameters, as `@id` is inside a string.
2. The parser does not care about what is between two literals. Meaning that `select 'foo' 'bar'` and 'select 'foo' || 'bar'` is basically treated the same by the parser. This was already the case, and these tests verify that, and also make sure that it stays that way in the future, as the tests would fail if someone made a change that would change that behavior.

harshachinta (approved):

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/StatementParserTest.java
olavloite: It will be the concatenation of `'foo'` and `'bar'`. The basic idea is that you can more easily split up long string literals into multiple parts, so it is easier to format a SQL statement,. So `select 'foo' 'bar'` `select 'foobar'` and

```
select
'foo'


'bar'
```

are all equivalent.

PR #2793, opened 2024-01-20
chore: fix session acquire timeout test

olavloite:
The test for session acquire timeout did not actually do what it was supposed to do, as it did not check that a timeout was actually registered. This again also caused it to busy-wait for 5 seconds to times (MinSessions=0 and MinSessions=1). This fixes both the actual test, and reduces the overall test time by about 10 seconds.


google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolTest.java
```diff
@@ -1614,15 +1614,16 @@ public void testSessionNotFoundPartitionedUpdate() {
...
   @Test
   public void testSessionMetrics() throws Exception {
     // Create a session pool with max 2 session and a low timeout for waiting for a session.
     options =
         SessionPoolOptions.newBuilder()
             .setMinSessions(1)
             .setMaxSessions(2)
-            .setMaxIdleSessions(0)
             .setInitialWaitForSessionTimeoutMillis(50L)
+            .setAcquireSessionTimeout(null)
```
harshachinta, line 1626: If we don't set this, then does it add a default timeout of 60 secs during acquiring a session?
Since this test validates the opencensus metrics I was thinking why do we need to add this?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolTest.java
```diff
@@ -1721,10 +1722,12 @@ public void testSessionMetrics() throws Exception {
     latch.await();
     // Wait until the request has timed out.
     int waitCount = 0;
-    while (pool.getNumWaiterTimeouts() == 0L && waitCount < 1000) {
-      Thread.sleep(5L);
+    while (pool.getNumWaiterTimeouts() == 0L && waitCount < 5000) {
```
harshachinta, line 1725: There is similar waits added in https://github.com/googleapis/java-spanner/blob/5ee5540ac580b75bae65dce9407bbf43c97855f6/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolTest.java#L1169
and
https://github.com/googleapis/java-spanner/blob/5ee5540ac580b75bae65dce9407bbf43c97855f6/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolTest.java#L1217.

Does this change apply there also?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolTest.java
olavloite: Yeah, this part is confusing, as we have made multiple changes to it through the years:
1. The `acquireSessionTimeout` is a 'real' timeout, as in: It will throw an error if it is exceeded.
2. In addition, we have the `InitialWaitForSessionTimeout` which is the original implementation that we had, and also what this test verifies. It does not throw an error if it is exceeded. Instead, it only adds an annotation to the trace.

So the reason that we are setting `acquireSessionTimeout` to null is to verify the correct behavior of the field `InitialWaitForSession`, as that field is ignored if there is an `acquireSessionTimeout`. See this line: https://github.com/googleapis/java-spanner/blob/ee477c2e7c509ce4b7c43da3b68c1433c59e46fb/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java#L1687

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolTest.java
olavloite: Those tests are not affected by the same problem as the test that this change is for. The 'testSessionMetrics` test had two problems:
1. It did not actually check for what it should test, as the timeout that it was supposed to verify was disabled.
2. The verify calls did not actually check that the thing that was supposed to happen, actually happened.
3. And as it was not triggering the condition it should, it waited the full 5 seconds, before finishing without checking anything.

The other tests do check what they should, and are also not waiting the full 5 seconds. But I agree that we should be consistent in the way we execute these waits, so I updated the wait loops to sleep for 1ms instead of 5ms.

harshachinta: Thanks for explaining this.

harshachinta (approved):

PR #2798, opened 2024-01-22
perf: optimize critical path for executeQuery

olavloite:
Removes some unnecessary calls from the critical path for executing a query.


olavloite:
Replaced by #3101

PR #2847, opened 2024-01-27
feat: support lazy decoding of query results

olavloite:
Adds an option for lazy decoding of query results. Currently, all values in a query result row are decoded from protobuf values to plain Java objects at the moment that the result set is advanced to the next row. This means that all values are decoded, regardless whether the application actually fetches these or not.

Lazy decoding also enables the possibility for (internal) consumers of a result set to access the protobuf value before it is converted to a plain Java object. This for example allows ChecksumResultSet to calculate the checksum based on the protobuf value, instead of a Java object, which can be more efficient.


google-cloud-spanner/src/main/java/com/google/cloud/spanner/GrpcStruct.java
```diff
@@ -193,6 +216,11 @@ public String toString() {
 
   boolean consumeRow(Iterator<com.google.protobuf.Value> iterator) {
     rowData.clear();
+    if (decodeMode == DecodeMode.LAZY_PER_ROW) {
+      rowDecoded = false;
+    } else if (decodeMode == DecodeMode.LAZY_PER_COL) {
+      colDecoded.clear();
+    }
```
ankiaga, line 223: nitpick: I guess we can also reset these values unconditionally

google-cloud-spanner/src/main/java/com/google/cloud/spanner/GrpcStruct.java
```diff
@@ -355,64 +396,96 @@ protected <T extends ProtocolMessageEnum> T getProtoEnumInternal(
...
+          decodeValue(
+              type.getStructFields().get(columnIndex).getType(),
+              (com.google.protobuf.Value) rowData.get(columnIndex)));
+      colDecoded.set(columnIndex);
+    }
+  }
+
   @Override
   protected Value getValueInternal(int columnIndex) {
+    ensureDecoded(columnIndex);
```
ankiaga, line 488: If DecodeMode is not direct then cant we just directly return the Value without decoding?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/GrpcStruct.java
olavloite: No (at least not in the current setup), as `colDecoded` is `null` in all cases, except when `decodeMode == DecodeMode.LAZY_PER_COL`. We only initialize that bitset if necessary (see line 194).

google-cloud-spanner/src/main/java/com/google/cloud/spanner/GrpcStruct.java
olavloite: I don't think so, but I'm also not sure whether there might be a misunderstanding here or not:
1. Here we are returning a `com.google.cloud.spanner.Value`. So a decoded Value instance as defined by the Spanner library, and not a protobuf value.
2. The underlying value, when decode mode is not DIRECT, could be a `com.google.protobuf.Value`. We need to decode that value into the corresponding Java object when we want to return a `com.google.cloud.spanner.Value` instance.

ankiaga (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/GrpcStruct.java
ankiaga: Understood, missed that we are returning a `com.google.cloud.spanner.Value`

PR #2848, opened 2024-01-27
perf: calculate checksum using protobuf values

olavloite:
Calculates the read/write transaction checksum for query results in the Connection API using the protobuf values instead of the decoded Java objects. This reduces both memory and CPU usage for the calculation.

The hash algorithm is also changed from `sha-256` to `md5`. The latter has a slightly higher collision probability, but still widely within the acceptable ranges for the use-case at hand, which is not security related, but only needed in order to determine equality between query results. Switching to `md5` also reduces CPU usage.


google-cloud-spanner/src/main/java/com/google/cloud/spanner/GrpcResultSet.java
```diff
@@ -49,6 +50,18 @@ class GrpcResultSet extends AbstractResultSet<List<Object>> {
...
+  @Override
+  public boolean canGetProtobufValue(int columnIndex) {
+    return !closed && currRow != null && currRow.canGetProtoValue(columnIndex);
+  }
+
+  @Override
+  public Value getProtobufValue(int columnIndex) {
+    checkState(!closed, "ResultSet is closed");
+    checkState(currRow != null, "next() call required");
+    return currRow.getProtoValueInternal(columnIndex);
```
ankiaga, line 62: Please correct me if I am wrong but IMO the idea of this function is to directly get the protobuf values without conversion to java object but I think we are first converting proto value to a java object and this function then converts the java object back to proto

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ReplaceableForwardingResultSet.java
```diff
@@ -77,6 +82,21 @@ public boolean next() throws SpannerException {
...
+    return !closed
+        && delegate instanceof ProtobufResultSet
+        && ((ProtobufResultSet) delegate).canGetProtobufValue(columnIndex);
+  }
+
+  @Override
+  public com.google.protobuf.Value getProtobufValue(int columnIndex) {
+    checkClosed();
+    Preconditions.checkState(
+        delegate instanceof ProtobufResultSet, "The result set does not support protobuf values");
```
ankiaga, line 96: Cant we call canGetProtobufValue() function to check if it returns true ?

olavloite: No, because `canGetProtobufValue()` is defined in the `ProtobufResultSet` interface, and this check is to verify that the given `ResultSet` is an instance of that (which it is not, which also means that we can't call that method).

olavloite: Oh, wait a second. I think I understand what you mean. Yes, in theory we could also call `this.canGetProtobufValue(..)` (I thought you meant `delegate.canGetProtobufValue(..)`). The reason that we don't do that is that is that `canGetProtobufValue` can also return false for other reasons, for example if the protobuf value is no longer available. That is not the check we want here. Here we want to check if the `ResultSet` at all supports protobuf values (not only: Can I get it __now__, which is what `canGetProtobufValue` returns)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/GrpcResultSet.java
olavloite: Your understanding of the intention here is correct. We do this in order to get the protobuf values directly without conversion.

The method however does get the protobuf value directly (or it fails if it is no longer available). `currRow` contains a list of the protobuf values in the following cases:
1. Decoding has been set to LAZY_PER_ROW and none of the columns have so far been decoded. The first time one column is decoded, all columns are decoded, and this method can no longer be called.
2. Decoding has been set to LAZY_PER_COL and this specific column has not yet been decoded.

If decoding is set to DIRECT, then this method will always fail. The `currRow.getProtovalueInternal` call will check that the protobuf value is still actually available.

ankiaga (approved):

PR #2855, opened 2024-01-31
feat: support Directed Read in Connection API

olavloite:
Adds support for setting Directed Read options in the Connection API. The value must be a valid JSON representation of a DirectedReadOptions protobuf instance.


google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/DirectedReadTest.java
```diff
@@ -0,0 +1,251 @@
...
+    try (Connection connection = createConnection()) {
+      connection.setAutocommit(false);
+      connection.setReadOnly(true);
+
+      connection.execute(
+          Statement.of(
+              String.format(
+                  "set %sdirected_read='%s'",
+                  getVariablePrefix(), DirectedReadOptionsUtil.toString(expected))));
+      // This uses a read/write transaction, which will ignore any DirectedReadOptions.
```
ankiaga, line 230: Why would this use a RW transaction as connection. is readOnly mode?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/DirectedReadTest.java
```diff
@@ -0,0 +1,251 @@
...
+      executeQuery(connection);
+      assertDirectedReadOptions(expected);
+    }
+  }
+
+  private void executeQuery(Connection connection) {
```
ankiaga, line 236: Would it be better to name it executeReadQuery?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/DirectedReadTest.java
```diff
@@ -0,0 +1,251 @@
...
+                                  .build())
+                          .build()))));
+      // This uses a read/write transaction, which will ignore any DirectedReadOptions.
+      executeQuery(connection);
+      assertDirectedReadOptions(DirectedReadOptions.getDefaultInstance());
+    }
+  }
+
+  @Test
+  public void testDirectedReadIsUsedInReadOnlyTransaction() {
```
ankiaga, line 208: Can we also test that DirectedReadOptions is persisted across different statements/transactions in a connection once set until reset?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/DirectedReadTest.java
olavloite: You're right, that's a copy-paste error. Fixed.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/DirectedReadTest.java
olavloite: I'm not sure why you feel that. It calls the method `executeQuery`. Or am I missing something here?

If any, I would maybe otherwise suggest `executeQueryAndConsumeResults` (the latter is why I put it in a separate method, so we didn't have to repeat the empty `while(resultSet.next())` block in every test method.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/DirectedReadTest.java
olavloite: Good point, added both for read-only transactions and auto-commit.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/Repeat.java
```diff
@@ -0,0 +1,33 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.spanner.connection;
+
+/** Simple util class for repeating (test) functions. */
+class Repeat {
```
ankiaga, line 20: This file should be in the test package

ankiaga (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/Repeat.java
olavloite: Indeed, thanks for noticing. Done.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/DirectedReadTest.java
olavloite: I've updated this method and added two more methods for executing both a read query and a DML query to ensure that we do the right thing in both cases.

(The original implementation did not do the right thing, as it would also include DirectedReadOptions for DML statements in auto-commit mode)

PR #2860, opened 2024-02-06
fix: use default query options with statement cache

olavloite:
Statements that were executed using the Connection API and that were not found in the statement cache (that is; it was the first time it was executed), would not use the default query options that had been set for the connection. This would mean that for example an optimizer version that had been set for the connection would not be used the first time a given query string would be executed using a connection. All subsequent executions of the statement would use specified optimizer version.

This change ensures that both the first and all following executions of the statement use the default query options that have been set.

### DIFF
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/QueryOptionsTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/QueryOptionsTest.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright 2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/QueryOptionsTest.java
```diff
@@ -0,0 +1,75 @@
...
+            assertEquals("10", request.getQueryOptions().getOptimizerVersion());
+
+            mockSpanner.clearRequests();
+          });
+    }
+  }
+
+  @Test
+  public void testUseOptimizerVersionFromStatement() {
+    try (Connection connection = createConnection()) {
```
ankiaga, line 50: Would it make sense to also pass optimizerVersion here and verify that its overridden?

olavloite: It's a good point. But this test is not intended to specifically verify that, and we also have specific tests that cover that here: https://github.com/googleapis/java-spanner/blob/aa9ad7f5a5e2405e8082a542916c3d1fa7d0fa25/google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionTest.java#L60

PR #2862, opened 2024-02-06
fix: cast for Proto type

harshachinta:
This PR fixes the following issues,
1. When export was done on `PROTO` type it was throwing a cast exception as follows,
```
Caused by: java.lang.ClassCastException: class com.google.cloud.spanner.AbstractResultSet$LazyByteArray cannot be cast to class com.google.cloud.ByteArray (com.google.cloud.spanner.AbstractResultSet$LazyByteArray and com.google.cloud.ByteArray are in unnamed module of loader 'app')

```
2. Add `ENUM` compatibility for `getLongArray` method.
3. `null` check for `PROTO` and `ENUM`

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
@@ -820,7 +820,8 @@ abstract class AbstractResultSet<R> extends AbstractStructReader implements Resu
         case INT64:
           return Value.int64(isNull ? null : getLongInternal(columnIndex));
         case ENUM:
-          return Value.protoEnum(getLongInternal(columnIndex), columnType.getProtoTypeFqn());
+          return Value.protoEnum(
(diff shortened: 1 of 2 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
```diff
@@ -485,7 +485,7 @@ private Object writeReplace() {
           case PROTO:
             builder
                 .set(fieldName)
-                .to(Value.protoMessage((ByteArray) value, fieldType.getProtoTypeFqn()));
+                .to(Value.protoMessage(((LazyByteArray) value).getByteArray(), fieldType.getProtoTypeFqn()));
```
olavloite, line 488: Do we also need a null check here?

olavloite (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
harshachinta: Thanks for pointing this. Added

PR #2864, opened 2024-02-06
feat(spanner): add support of using emulator using autogenerated admin clients

rahul2393:
Similar handwritten 
hook in Go https://github.com/googleapis/google-cloud-go/blob/main/spanner/admin/database/apiv1/init.go

- [ ] Verified that new proto chaneges won't overwrite the changes here.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/admin/database/v1/DatabaseAdminSettings.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/admin/database/v1/DatabaseAdminSettings.java
@@ -26,6 +26,7 @@ import com.google.api.core.ApiFunction;
 import com.google.api.core.BetaApi;
 import com.google.api.gax.core.GoogleCredentialsProvider;
 import com.google.api.gax.core.InstantiatingExecutorProvider;
+import com.google.api.gax.core.NoCredentialsProvider;
 import com.google.api.gax.grpc.InstantiatingGrpcChannelProvider;
 import com.google.api.gax.httpjson.InstantiatingHttpJsonChannelProvider;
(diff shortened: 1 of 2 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/admin/database/v1/DatabaseAdminSettings.java
```diff
@@ -314,6 +318,20 @@ public static ApiClientHeaderProvider.Builder defaultApiClientHeaderProviderBuil
 
   /** Returns a new gRPC builder for this class. */
   public static Builder newBuilder() {
+    String hostAndPort = System.getenv(SPANNER_EMULATOR_HOST_ENV_VAR);
```
harshachinta, line 321: @olavloite has suggested not to add logic on autogenerated files since there is a chance of this file getting overwritten during the next autogenerated PRs.
In NodeJS, we are writing a function similar to Python which can return a new autogenerated admin client with the emulator host set.

rahul2393: yes, there's similar requirement in other client libs. 
Bigtable https://github.com/googleapis/java-bigtable/blob/main/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminSettings.java#L56 has logic to use emulator in their classes.

Similar for Bigquery:  https://github.com/googleapis/java-bigtable/blob/main/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminSettings.java#L146-L158

arpan14:
Closing this in favor of - https://github.com/googleapis/java-spanner/pull/2878/

PR #2878, opened 2024-02-10
feat: support public methods to use autogenerated admin clients.

arpan14:
Supporting two new methods `createInstanceAdminClient()` and `createDatabaseAdminClient` in the Spanner public interface for customers to use the upgraded admin clients.

**Sample Usages**

```
DatabaseAdminClient dbAdminClient = spanner.createInstanceAdminClient();   
```
```
InstanceAdminClient instanceAdminClient = spanner.createDatabaseAdminClient();
```

Notes:

* The new public methods support all existing setting (set via environment variables). For ex - Emulator can be used in the same way as before by setting env variable `SPANNER_EMULATOR_HOST`.


rahul2393 (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Spanner.java
```diff
@@ -36,9 +42,34 @@ public interface Spanner extends Service<SpannerOptions>, AutoCloseable {
...
+  /*
+   * <!--SNIPPET get_dbadmin_client-->
+   * <pre>{@code
+   * SpannerOptions options = SpannerOptions.newBuilder().build();
+   * Spanner spanner = options.getService();
+   * DatabaseAdminClient dbAdminClient = spanner.databaseAdminClient();
+   * }</pre>
+   * <!--SNIPPET get_dbadmin_client-->
+   */
+  default com.google.cloud.spanner.admin.database.v1.DatabaseAdminClient databaseAdminClient() {
```
olavloite, line 63: This method returns an instance of `com.google.cloud.spanner.admin.database.v1.DatabaseAdminClient`. That interface implements `AutoCloseable`. That again means that most users will be inclined to put it in a `try-with-resources` block, or in some other way close it when they are done. If they do that, then the underlying `stub` that is used by the client will be closed, which again is a shared object, and means that all existing and also future instances that will be returned by this method will use a closed stub.

So we need to make a choice here whether:
1. We document that users should __never__ call `close()` on the instance that is returned.
2. OR: Create a new stub for each new instance that is returned by this method.
3. OR: Create/use a small wrapper object around the instance that we are returning here that overrides the default `close()` behavior. The modified behavior could either be to do nothing, or to throw an exception that indicates that this instance is managed by the `Spanner` instance that created it.

I think that option 3 is the safest option.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Spanner.java
```diff
@@ -36,9 +42,34 @@ public interface Spanner extends Service<SpannerOptions>, AutoCloseable {
    * }</pre>
    * <!--SNIPPET get_dbadmin_client-->
    */
+  @Deprecated
```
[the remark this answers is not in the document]
arpan14: * Sure, our internal documents mention "Soon to be deprecated" since internally readers may confuse this with the "Deprecation Policy" which means that the code will be actually deleted. To avoid that confusion we mentioned "soon to be deprecated".
* In the Java code, shouldn't we provide a way for our customers to know that these methods will indeed lose support soon. While the code is not deleted, we are providing customers a fair notice that the support for these methods could be removed in future. Unfortunately Java does not provide me a "Soon to be deprecated" annotation, else I agree it would have been a better choice.

I'm ok with removing this annotation, but wanted to know your thoughts on this.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
```diff
@@ -207,11 +209,23 @@ public DatabaseAdminClient getDatabaseAdminClient() {
     return dbAdminClient;
   }
 
+  @Override
+  public com.google.cloud.spanner.admin.database.v1.DatabaseAdminClient databaseAdminClient() {
+    DatabaseAdminStub databaseAdminStub = gapicRpc.getDatabaseAdminStub();
+    return com.google.cloud.spanner.admin.database.v1.DatabaseAdminClient.create(databaseAdminStub);
```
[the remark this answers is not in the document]
arpan14: * Sure, moved the initialization to the constructor. That will make sure we don't return new objects for each get call.
* On the naming, unfortunately I couldn't overload the method due to similar signature of existing method. I tried hacking this and creating a varargs alternative of existing method with a different signature,

```
  public com.google.cloud.spanner.admin.instance.v1.InstanceAdminClient getInstanceAdminClient(InstanceAdminSettings... settings) {
    return instanceClient;
  }
```

The advantage of this is, future customers can use the `InstanceAdminSettings` class to provide values directly. And it allows us to keep the same method signature as existing method i.e `getInstanceAdminClient`. But the con is, I'd prefer to keep things simple. I don't see a requirement right now on why we need to expose the `InstanceAdminSettings` object. If a future requirement comes we can always do it at a later point.

* Some alternate naming I could think [in order of preference] - 
  * `getSpannerDatabaseAdminClient()`, `getSpannerInstanceAdminClient()` 
  * `getAutogeneratedDatabaseAdminClient()`, `getAutogeneratedInstanceAdminClient()`
  * `getDBAdminClient()`, `getDBInstanceAdminClient()`

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Spanner.java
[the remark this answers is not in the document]
arpan14: Removed the `@Deprecated` annotation from this PR. Will take up the discussion offline and in a new PR.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Spanner.java
arpan14: Thanks for the catch! 

Thinking a bot more on the issue, I think its due to the dependency on the `stub` making it a shared resource when it should not have been one. Another approach which I explored was to make use of the shared `DatabaseAdminStubSettings` and `InstanceAdminStubSettings` objects and creating a new `stub` objects from these settings by doing something like `gapicRpc.getDatabaseAdminStubSettings().createStub()` . This will ensure we avoid the coupling and at the same time are able to re-use all the code of `GapicSpannerRpc` for creating the settings.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
```diff
@@ -135,6 +139,8 @@ static final class ClosedException extends RuntimeException {
     this.dbAdminClient = new DatabaseAdminClientImpl(options.getProjectId(), gapicRpc);
     this.instanceClient =
         new InstanceAdminClientImpl(options.getProjectId(), gapicRpc, dbAdminClient);
+    this.databaseAdminClient = createDatabaseAdminClient();
```
olavloite, line 142: But doesn't this still create a single `DatabaseAdminClient` that will be shared between everyone that calls `databaseAdminClient()`? Meaning that if there is one place in the code that calls `DatabaseAdminClient#close()`, then it applies to all instances (because they are the same) that have ever been gotten through `databaseAdminClient()`. Even future calls to `databaseAdminClient()` will return a closed instance.

We can do this, but then we should very clearly document the `databaseAdminClient()` method that it returns a shared instance that they should not close, unless they are going to close their application (and which they now need to do in addition to calling `Spanner#close()`)..

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Spanner.java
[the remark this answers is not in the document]
olavloite: I agree that there is a missing annotation in Java here, that would indicate something like 'you should not use this method for new code anymore, but you are also not in a rush to remove this in all your current code'. Marking this as `@Deprecated` in my mind goes too far, as it is a method that has been around since the very first version, and that is likely to be spread across a very large range of customer code. Marking something as `@Deprecated` is normally a hint that something will be removed soonish, and that users should actively port their code. I expect the latter to cause a lot of questions if we mark it as such without some gentle pre-warning first.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
arpan14: Based on offline conversations, revised the PR to now create an instance each time an instance is returned. This mitigated most of the edges cases that arise here.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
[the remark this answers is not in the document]
arpan14: We settled for the naming `createInstanceAdminClient` and `createDatabaseAdminClient` based on our design choice to return a new instance each time a client object is requested.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Spanner.java
arpan14: Based on offline conversations, we found an additional issue. For ex - refer the below snippet where we would still up see the shared stub behavior.

```
void doSomeAdminStuff() {
  try (DatabaseAdminClient client = spanner.databaseAdminClient()) {
    client.listDatabases();
    doSomethingElse();
    client.dropDatabase(...);
  }
}

void doSomethingElse() {
  try (DatabaseAdminClient client = spanner.databaseAdminClient()) {
    client.updateDatabase(...);
  }
}
```

I explored two other approaches based on @olavloite 's suggestion

1. Creating a `ManagedDatabaseAdminClient` which extends the `DatabaseAdminClient` class and overrides the `close()` method. This is not possible to implement since the method is marked as final and hence we cannot override it.
2. Using dynamic proxy classes (reference -  https://www.baeldung.com/java-dynamic-proxies#invocation-handler-via-lambda-expressions ). This is not possible since dynamic proxies only work for classes that implement some interface and in this case `DatabaseAdminClient` does not implement any interface that can supply all its methods.

arpan14: We are settling for an approach where we return a new instance each time a client object is requested. This makes sure we don't run into any edge cases (like discussed above). At the same time, we have captured in the method documentation that customers may choose to cache the instances based on their application side code.

olavloite (approved):
LGTM, assuming that the emulator tests succeed (the failure seemed completely unrelated).

Also one minor request: The pull request description needs to be updated to reflect the latest method names and how these work.

PR #2890, opened 2024-02-15
chore: flaky test

surbhigarg92:


google-cloud-spanner/src/test/java/com/google/cloud/spanner/OpenTelemetrySpanTest.java
```diff
@@ -403,10 +385,9 @@ public void multiUse() {
   }
 
   @Test
-  public void transactionRunner() {
+  public void transactionRunner() throws InterruptedException {
```
arpan14, line 388: Why are we adding `throws InterruptedException` ? I don't see anything changing in current tests.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/OpenTelemetrySpanTest.java
```diff
@@ -282,7 +266,11 @@ public void setUp() throws Exception {
             .setChannelProvider(channelProvider)
             .setOpenTelemetry(openTelemetry)
             .setCredentials(NoCredentials.getInstance())
-            .setSessionPoolOption(SessionPoolOptions.newBuilder().setMinSessions(0).build());
+            .setSessionPoolOption(
+                SessionPoolOptions.newBuilder()
+                    .setMinSessions(2)
```
arpan14, line 271: You can also have a `maxSessions` and make min=max to have a more deterministic behavior. Else if you get more than 2 requests concurrently, it will try to create more sessions. Are we making sure that current tests always utilize <= 2 sessions?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpanTest.java
```diff
@@ -286,22 +289,23 @@ public void singleUse() {
...
     assertThat(spans).containsEntry("CloudSpannerOperation.ExecuteStreamingQuery", true);
 
     List<String> expectedAnnotations =
         ImmutableList.of(
-            "Requesting 25 sessions",
-            "Request for 25 sessions returned 25 sessions",
-            "Creating 25 sessions",
+            "Requesting 2 sessions",
+            "Request for 2 sessions returned 2 sessions",
+            "Creating 2 sessions",
```
arpan14, line 299: Open Question for my understanding, given this is an unit test and not an integration test, I am assuming that the session pool functionality would be mocked.

So how does it matter if there are 2 sessions or 25 sessions? Do we need to mock a few session pool methods?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/OpenTelemetrySpanTest.java
surbhigarg92: Yes, current tests always utilize 1 session. I created 2 just for extra safe.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpanTest.java
surbhigarg92: The end call will be mocked, but these annotations are getting added at the time batch create session starts. So if the request is for 2 sessions, then the annotation will be added accordingly. 
`String.format("Requesting %d sessions", sessionCount)`

google-cloud-spanner/src/test/java/com/google/cloud/spanner/OpenTelemetrySpanTest.java
surbhigarg92: This was a miss. Updated

arpan14 (approved):

PR #2894, opened 2024-02-16
feat: support float32 type

arawind:
TODO: 
- [ ] Remove the isCloudDevel check once FLOAT32 is supported in production (will be done after merging)
- [x] ~Un-ignore the float32 parameter tests that are waiting for a backend fix.~


google-cloud-spanner/src/main/java/com/google/cloud/spanner/Mutation.java
```diff
@@ -386,8 +388,10 @@ private boolean areValuesEqual(List<Value> values, List<Value> otherValues) {
 
   private boolean isNaN(Value value) {
```
arpan14, line 389: It's getting difficult to read this method now with so many nested operators. Can we break the method into smaller methods (1 per type).

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/DirectExecuteResultSet.java
```diff
@@ -180,6 +180,12 @@ public long getLong(String columnName) {
     return delegate.getLong(columnName);
   }
 
+  @Override
```
arpan14, line 183: @olavloite Could you give this a look if these are all the changes that are needed in the Connection API or do we need to do more changes to support this new type?

arpan14: The rest of the changes LGTM.

arpan14 (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Mutation.java
arawind: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/StructReader.java
```diff
@@ -123,6 +123,18 @@ public interface StructReader {
    */
   long getLong(String columnName);
 
+  /**
+   * @param columnIndex index of the column
+   * @return the value of a non-{@code NULL} column with type {@link Type#float32()}.
+   */
+  float getFloat(int columnIndex);
```
olavloite, line 130: New methods in an interface should have a default implemention that throws an `UnsupportedOperationException`. See the example for `getJson` in this file.

The reason is that:
1. The current implementation will cause a compilation error for any external user that implements this interface, for example in tests. That is a breaking change.
2. Adding a default implementation defers any error until runtime and the actual usage of the method, which makes this change non-breaking.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/StructReader.java
```diff
@@ -361,6 +373,38 @@ default Value getValue(String columnName) {
    */
   List<Long> getLongList(String columnName);
 
+  /**
+   * @param columnIndex index of the column
+   * @return the value of a non-{@code NULL} column with type {@code Type.array(Type.float32())}.
+   * @throws NullPointerException if any element of the array value is {@code NULL}. If the array
+   *     may contain {@code NULL} values, use {@link #getFloatList(int)} instead.
+   */
+  float[] getFloatArray(int columnIndex);
```
olavloite, line 382: Same as above: These should have a default implementation that just throws an error.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Type.java
```diff
@@ -89,6 +91,14 @@ public static Type int64() {
     return TYPE_INT64;
   }
 
+  /**
+   * Returns the descriptor for the {@code FLOAT32} type: a floating point type with the same value
+   * domain as a Java {code float}.
```
olavloite, line 96: nit:

```suggestion
   * domain as a Java {@code float}.
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Value.java
```diff
@@ -1346,6 +1456,14 @@ public String getString() {
...
     }
 
+    @Override
+    public float getFloat32() {
+      checkNotNull();
+      Preconditions.checkState(
+          value.hasNumberValue(), "This value does not contain a number value");
+      return (float) value.getNumberValue();
+    }
+
```
olavloite, line 1466: This method is not needed and should be removed.

```suggestion
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Value.java
```diff
@@ -1492,6 +1610,46 @@ int valueHash() {
...
+    @Override
+    public float getFloat32() {
+      checkNotNull();
+      return value;
+    }
+
+    @Override
+    com.google.protobuf.Value valueToProto() {
+      return com.google.protobuf.Value.newBuilder().setNumberValue((double) value).build();
```
olavloite, line 1629: nit: remove unnecessary cast

```suggestion
      return com.google.protobuf.Value.newBuilder().setNumberValue(value).build();
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Value.java
```diff
@@ -1909,6 +2069,23 @@ public BigDecimal getNumeric() {
...
+        try {
+          valueAsFloat = Float.parseFloat(value);
+        } catch (NumberFormatException e) {
+          floatConversionError = e;
+          throw e;
+        }
+      }
+      return valueAsFloat;
+    }
+
```
olavloite, line 2088: I think we can (and should) remove this. There's no specific reason that we would need to get a `PgNumeric` as a float32. The reason that the `getFloat64()` method is here, is that `PgNumeric` is a bit different than the standard `Numeric` in GoogleSQL Spanner; `PgNumeric` also supports NaN, which cannot be represented as a BigDecimal.

```suggestion
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Value.java
```diff
@@ -1879,6 +2037,8 @@ void valueToString(StringBuilder b) {
   private static class PgNumericImpl extends AbstractObjectValue<String> {
     private BigDecimal valueAsBigDecimal;
     private NumberFormatException bigDecimalConversionError;
+    private Float valueAsFloat;
+    private NumberFormatException floatConversionError;
```
olavloite, line 2041: Remove (see also below)

```suggestion
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Value.java
```diff
@@ -2106,6 +2283,46 @@ int arrayHash() {
...
+    }
+
+    @Override
+    Float getValue(int i) {
+      return values[i];
+    }
+
+    @Override
+    com.google.protobuf.Value getValueAsProto(int i) {
+      return com.google.protobuf.Value.newBuilder().setNumberValue((double) values[i]).build();
```
olavloite, line 2317:
```suggestion
      return com.google.protobuf.Value.newBuilder().setNumberValue(values[i]).build();
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Value.java
```diff
@@ -2481,6 +2698,8 @@ private static class PgNumericArrayImpl extends AbstractArrayValue<String> {
 
     private List<BigDecimal> valuesAsBigDecimal;
     private NumberFormatException bigDecimalConversionError;
+    private List<Float> valuesAsFloat;
+    private NumberFormatException floatConversionError;
```
olavloite, line 2702: Remove (see also above)

```suggestion
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Value.java
```diff
@@ -2514,6 +2733,26 @@ public List<BigDecimal> getNumericArray() {
...
+                  .map(v -> v == null ? null : Float.valueOf(v))
+                  .collect(Collectors.toList());
+        } catch (NumberFormatException e) {
+          floatConversionError = e;
+          throw e;
+        }
+      }
+      return valuesAsFloat;
+    }
+
```
olavloite, line 2755: Remove (see above)

```suggestion
```

google-cloud-spanner/src/test/java/com/google/cloud/spanner/RandomResultSetGenerator.java
```diff
@@ -138,6 +143,9 @@ private void setRandomValue(Value.Builder builder, Type type) {
         case FLOAT64:
           builder.setNumberValue(random.nextDouble());
           break;
+        case FLOAT32:
+          builder.setNumberValue((double) random.nextFloat());
```
olavloite, line 147:
```suggestion
          builder.setNumberValue(random.nextFloat());
```

google-cloud-spanner/src/test/java/com/google/cloud/spanner/ValueTest.java
```diff
@@ -1512,6 +1604,20 @@ public void testValueToProto() {
...
             .build(),
         Value.int64Array(Arrays.asList(1L, null)).toProto());
+    assertEquals(
+        com.google.protobuf.Value.newBuilder()
+            .setListValue(
+                ListValue.newBuilder()
+                    .addAllValues(
+                        Arrays.asList(
+                            com.google.protobuf.Value.newBuilder()
+                                .setNumberValue((double) 3.14f)
```
olavloite, line 1614:
```suggestion
                                .setNumberValue(3.14f)
```

google-cloud-spanner/src/test/java/com/google/cloud/spanner/ValueTest.java
```diff
@@ -1426,6 +1511,13 @@ public void testValueToProto() {
         com.google.protobuf.Value.newBuilder().setNullValue(NullValue.NULL_VALUE).build(),
         Value.int64(null).toProto());
 
+    assertEquals(
+        com.google.protobuf.Value.newBuilder().setNumberValue((double) 3.14f).build(),
```
olavloite, line 1515:
```suggestion
        com.google.protobuf.Value.newBuilder().setNumberValue(3.14f).build(),
```

google-cloud-spanner/src/test/java/com/google/cloud/spanner/ValueTest.java
```diff
@@ -1667,6 +1773,29 @@ public void testValueToProto() {
...
+            .setListValue(
+                ListValue.newBuilder()
+                    .addValues(
+                        com.google.protobuf.Value.newBuilder()
+                            .setListValue(
+                                ListValue.newBuilder()
+                                    .addAllValues(
+                                        Arrays.asList(
+                                            com.google.protobuf.Value.newBuilder()
+                                                .setNumberValue((double) 3.14f)
```
olavloite, line 1787:
```suggestion
                                                .setNumberValue(3.14f)
```

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITFloat32Test.java
```diff
@@ -0,0 +1,448 @@
+/*
+ * Copyright 2022 Google LLC
```
olavloite, line 2: nit:
```suggestion
 * Copyright 2024 Google LLC
```

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITFloat32Test.java
```diff
@@ -0,0 +1,448 @@
...
+
+    // Assumes that the jobType contains the string "cloud-devel" to signal that
+    // the environment is cloud-devel.
+    return !isNullOrEmpty(jobType) && jobType.contains("cloud-devel");
+  }
+
+  @BeforeClass
+  public static void setUpDatabase()
+      throws ExecutionException, InterruptedException, TimeoutException {
+    if (!isUsingCloudDevel()) {
```
olavloite, line 107: Use `assumeTrue(isUsingCloudDevel())` to check and skip if it is not.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITFloat32Test.java
```diff
@@ -0,0 +1,448 @@
...
+    return readRow("T", lastKey, columns);
+  }
+
+  @Test
+  public void writeFloat32() {
+    assumeTrue("FLOAT32 is currently only supported in cloud-devel", isUsingCloudDevel());
+
+    write(baseInsert().set("Float32Value").to(2.0f).build());
+    Struct row = readLastRow("Float32Value");
+    assertThat(row.isNull(0)).isFalse();
```
olavloite, line 168: In new tests, please use `junit` assertions, so:

`assertEquals(...)`
`assertTrue(..)`
`assertFalse(..)`
`assertThrows(...)`

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITFloat32Test.java
```diff
@@ -0,0 +1,448 @@
...
+    assertThat(row.getFloat(0)).isWithin(0.0f).of(2.0f);
+  }
+
+  @Ignore("Needs a fix in the backend.")
+  @Test
+  public void writeFloat32NonNumbers() {
+    assumeTrue("FLOAT32 is currently only supported in cloud-devel", isUsingCloudDevel());
```
olavloite, line 175: This should not be needed, as the entire test class is conditional on that.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITFloat32Test.java
```diff
@@ -0,0 +1,448 @@
...
+            .build());
+    Struct row = readLastRow("Float32ArrayValue");
+    assertThat(row.isNull(0)).isFalse();
+    assertThat(row.getFloatList(0)).containsExactly(null, 1.0f, 2.0f).inOrder();
+    try {
+      row.getFloatArray(0);
+      fail("Expected exception");
+    } catch (NullPointerException ex) {
+      assertNotNull(ex.getMessage());
+    }
```
olavloite, line 238: Use `assertThrows(...)` instead of this construct. This type of construct was needed back when we had to support Java 7.

google-cloud-spanner/src/test/resources/com/google/cloud/spanner/connection/it/CreateMusicTables.sql
```diff
@@ -73,7 +73,7 @@ CREATE TABLE TableWithAllColumnTypes (
   ColDate		DATE		NOT NULL,
   ColTimestamp	TIMESTAMP	NOT NULL,
   ColCommitTS	TIMESTAMP	NOT NULL OPTIONS (allow_commit_timestamp=true),
-  
+
```
olavloite, line 76: Remove unrelated change

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractStructReader.java
```diff
@@ -43,6 +43,8 @@ public abstract class AbstractStructReader implements StructReader {
 
   protected abstract long getLongInternal(int columnIndex);
 
+  protected abstract float getFloatInternal(int columnIndex);
```
olavloite, line 46: These methods should also have a default implementation to prevent this from becoming a breaking change.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/StructReader.java
arawind: Thanks for the context. Makes sense, made the change.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractStructReader.java
arawind: Thanks, made the change

google-cloud-spanner/src/test/resources/com/google/cloud/spanner/connection/it/CreateMusicTables.sql
arawind: Done

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITFloat32Test.java
arawind: Ack. Apologies, this was a copy paste from an older test.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITFloat32Test.java
arawind: Ack. I intended to write another test that verifies that FLOAT32 is not supported in production, but then realized that it would be a bad idea as the client tests would break when the backend starts supporting the type.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITFloat32Test.java
arawind: Done! Removed the `com.google.common.truth.Truth` imports

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Value.java
arawind: Thanks, this seems to be a remnant of an earlier commit before GrpcResultSet

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Type.java
arawind: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/StructReader.java
arawind: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/StructReader.java
arawind: Can you PTAL at the CLIRR differences too to verify this? I had to add the Value.float32 to the ignored differences, but this seems to be the first addition in this class.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Mutation.java
```diff
@@ -385,9 +387,21 @@ private boolean areValuesEqual(List<Value> values, List<Value> otherValues) {
...
-        && value.getType().equals(Type.float64())
-        && Double.isNaN(value.getFloat64());
+    return !value.isNull() && (isFloat64NaN(value) || isFloat32NaN(value));
+  }
+
+  // Checks if the Float64 value is either a "Double" or a "Float" NaN.
+  // Refer the comment above `areValuesEqual` for more details.
+  private boolean isFloat64NaN(Value value) {
+    return value.getType().equals(Type.float64())
+        && (Double.isNaN(value.getFloat64()) || Float.isNaN((float) value.getFloat64()));
```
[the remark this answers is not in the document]
arawind: Thanks! I've made the change to match your request.

I'm a bit confused though -- I did try a similar approach in my [first change](https://github.com/googleapis/java-spanner/pull/2894/commits/1d64140e9e668c7e3ed4caf748f5b93f5f453e8d#diff-674fca14023d09c647d3f33423dcded27e8e7369cf65e3886d4caba086fecff2). Somehow the tests were only passing after I had added the other unnecessary checks.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITFloat32Test.java
arawind: Done

google-cloud-spanner/src/test/java/com/google/cloud/spanner/ValueTest.java
arawind: Done

google-cloud-spanner/src/test/java/com/google/cloud/spanner/ValueTest.java
arawind: Done

google-cloud-spanner/src/test/java/com/google/cloud/spanner/RandomResultSetGenerator.java
arawind: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Value.java
arawind: Thanks, done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Value.java
arawind: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Value.java
arawind: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/StructReader.java
olavloite: The clirr diffs look fine. This is however not the first addition to this class. See for example the addition of `jsonb`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Mutation.java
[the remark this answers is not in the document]
olavloite: In your original change, you had a method `isNaN`. Now it is `isFloat64NaN`. That also meant that you had to check for different things.

olavloite (approved):
LGTM (assuming that the test failure is unrelated)

arawind:
The cloud-devel failures are unrelated and all FLOAT32 integration tests have passed on cloud-devel.

From https://source.cloud.google.com/results/invocations/3dd1b019-0668-4964-a2bc-512009d741ef: 

```
[INFO] Tests run: 20, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 5.336 s -- in com.google.cloud.spanner.it.ITFloat32Test
```

PR #2896, opened 2024-02-19
chore: add support for OpenTelemetry metrics to Connection API

hengfengli:
Adds support for OpenTelemetry metrics in the Connection API.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
@@ -39,6 +39,7 @@ import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
 import com.google.common.collect.Sets;
 import com.google.spanner.v1.ExecuteSqlRequest.QueryOptions;
+import io.opentelemetry.api.OpenTelemetry;
 import java.io.IOException;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
@@ -484,6 +485,7 @@ public class ConnectionOptions {
     private List<StatementExecutionInterceptor> statementExecutionInterceptors =
         Collections.emptyList();
     private SpannerOptionsConfigurator configurator;
+    private OpenTelemetry openTelemetry;
 
     private Builder() {}
 
@@ -633,6 +635,11 @@ public class ConnectionOptions {
       return this;
     }
 
+    public Builder setOpenTelemetry(OpenTelemetry openTelemetry) {
+      this.openTelemetry = openTelemetry;
+      return this;
+    }
+
     /** @return the {@link ConnectionOptions} */
     public ConnectionOptions build() {
       Preconditions.checkState(this.uri != null, "Connection URI is required");
@@ -691,6 +698,7 @@ public class ConnectionOptions {
   private final boolean retryAbortsInternally;
   private final boolean useVirtualThreads;
   private final boolean useVirtualGrpcTransportThreads;
+  private final OpenTelemetry openTelemetry;
   private final List<StatementExecutionInterceptor> statementExecutionInterceptors;
   private final SpannerOptionsConfigurator configurator;
 
@@ -792,6 +800,7 @@ public class ConnectionOptions {
     this.retryAbortsInternally = parseRetryAbortsInternally(this.uri);
     this.useVirtualThreads = parseUseVirtualThreads(this.uri);
     this.useVirtualGrpcTransportThreads = parseUseVirtualGrpcTransportThreads(this.uri);
+    this.openTelemetry = builder.openTelemetry;
     this.statementExecutionInterceptors =
         Collections.unmodifiableList(builder.statementExecutionInterceptors);
     this.configurator = builder.configurator;
@@ -854,6 +863,14 @@ public class ConnectionOptions {
       }
     }
     return null;
+  }
+
+  /**
+   * @return an instance of OpenTelemetry. If OpenTelemetry object is not set then <code>null</code>
+   *     will be returned.
+   */
+  OpenTelemetry getOpenTelemetry() {
+    return this.openTelemetry;
   }
 
   SpannerOptionsConfigurator getConfigurator() {
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/SpannerPool.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/SpannerPool.java
@@ -29,6 +29,7 @@ import com.google.common.base.MoreObjects;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Ticker;
 import io.grpc.ManagedChannelBuilder;
+import io.opentelemetry.api.OpenTelemetry;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -156,6 +157,7 @@ public class SpannerPool {
     private final String databaseRole;
     private final boolean routeToLeader;
     private final boolean useVirtualGrpcTransportThreads;
+    private final OpenTelemetry openTelemetry;
 
     @VisibleForTesting
     static SpannerPoolKey of(ConnectionOptions options) {
@@ -183,6 +185,7 @@ public class SpannerPool {
       this.userAgent = options.getUserAgent();
       this.routeToLeader = options.isRouteToLeader();
       this.useVirtualGrpcTransportThreads = options.isUseVirtualGrpcTransportThreads();
+      this.openTelemetry = options.getOpenTelemetry();
     }
 
     @Override
@@ -201,7 +204,8 @@ public class SpannerPool {
           && Objects.equals(this.userAgent, other.userAgent)
           && Objects.equals(this.routeToLeader, other.routeToLeader)
           && Objects.equals(
-              this.useVirtualGrpcTransportThreads, other.useVirtualGrpcTransportThreads);
+              this.useVirtualGrpcTransportThreads, other.useVirtualGrpcTransportThreads)
+          && Objects.equals(this.openTelemetry, other.openTelemetry);
     }
 
     @Override
@@ -216,7 +220,8 @@ public class SpannerPool {
           this.databaseRole,
           this.userAgent,
           this.routeToLeader,
-          this.useVirtualGrpcTransportThreads);
+          this.useVirtualGrpcTransportThreads,
+          this.openTelemetry);
     }
   }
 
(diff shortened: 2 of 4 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
```diff
@@ -633,6 +644,12 @@ Builder setCredentials(Credentials credentials) {
       return this;
     }
 
+    @VisibleForTesting
+    Builder setOpenTelemetry(OpenTelemetry openTelemetry) {
```
olavloite, line 648: This should be a public setter, so that clients that build `ConnectionOptions` directly (and not only through a connection string) can pass in a specific `OpenTelemetry` instance. This is for example the case for PGAdapter.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/SpannerPool.java
```diff
@@ -349,6 +352,9 @@ Spanner createSpanner(SpannerPoolKey key, ConnectionOptions options) {
         .setDatabaseRole(options.getDatabaseRole())
         .setCredentials(options.getCredentials());
     builder.setSessionPoolOption(key.sessionPoolOptions);
+    if (options.isEnabledOpenTelemetryMetrics() && options.getOpenTelemetry() != null) {
```
olavloite, line 355: `options.getOpenTelemetry()` will in the current implementation never return null, so I think it would be better to remove this null check, and instead add a `NonNull` annotation to the `getOpenTelemetry()` method.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/SpannerPool.java
```diff
@@ -337,6 +337,9 @@ private void initialize() {
 
   @VisibleForTesting
   Spanner createSpanner(SpannerPoolKey key, ConnectionOptions options) {
+    if (options.isEnabledOpenTelemetryMetrics() && !ConnectionSpannerOptions.isEnabledOpenTelemetryMetrics()) {
+      ConnectionSpannerOptions.enableOpenTelemetryMetrics();
```
olavloite, line 341: This is a static method that enables the option globally. Theoretically, we could have the following situation:

1. Someone creates a `Connection` with `enableOpenTelemetryMetrics=false`. That means that a `Spanner` instance without an `OpenTelemetry` will be created.
2. Someone then creates another connection with `enableOpenTelemetryMetrics=true`. That in the current implementation will cause `ConnectionSpannerOptions.enableOpenTelemetryMetrics()` to be enabled, but as that property is not part of the `SpannerPoolKey`, no new `Spanner` instance will be created. Meaning that the `Connection` will not collect any metrics.

However, I also expect problems if you do add the property `enableOpenTelemetryMetrics` to the `SpannerPoolKey`, because then this method will again be called, even though there are already a `Spanner` instance that has been created without an `OpenTelemetry` instance.

So what we need to do is:
1. Add `enableOpenTelemetryMetrics` as an attribute to the `SpannerPoolKey`
2. Add tests that verify that all combinations of creating connections with and without metrics actually create the type of connection that they should, and that it works with the global `ConnectionSpannerOptions.enableOpenTelemetryMetrics()` setting.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
```diff
@@ -856,6 +877,18 @@ private static Integer parseIntegerProperty(String propertyName, String value) {
     return null;
   }
 
+  /**
+   * Returns an instance of OpenTelemetry. If OpenTelemetry object is not set
+   * then GlobalOpenTelemetry will be used as fallback.
+   */
+  OpenTelemetry getOpenTelemetry() {
```
olavloite, line 884: nit: annotate this method with `@NonNull` to indicate to callers that it will never return a null value.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/SpannerPool.java
hengfengli: I think the main problem is once we enable `SpannerOptions.enableOpenTelemetryMetrics` to true globally, we do not have a way to disable it. 

What if someone creates a `Connection` with `enableOpenTelemetryMetrics=true`. And then, someone creates a Connection with `enableOpenTelemetryMetrics=false`, who will get a `Spanner` instance with` OpenTelemetry` enabled. 

Multiple `Spanner` instances would share the same global `ConnectionSpannerOptions.enableOpenTelemetryMetrics()` setting.

hengfengli: Should we make `enableOpenTelemetryMetrics` in `ConnectionOptions` also be a global setting? If the caller of Connection API wants to enable OpenTelemetry metrics instead of putting this as a parameter in URI, they must call: 

```java
ConnectionOptions.enableOpenTelemetryMetrics();
...
ConnectionOptions.Builder connectionOptionsBuilder = ConnectionOptions.newBuilder().setUri(uri);

connectionOptionsBuilder.setOpenTelemetry(openTelemetry);

ConnectionOptions connectionOptions = connectionOptionsBuilder.build();
Connection spannerConnection = connectionOptions.getConnection();
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
hengfengli: Fixed.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/SpannerPool.java
hengfengli: Fixed.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
hengfengli: Updated.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/SpannerPool.java
olavloite: TLDR:

We should:
1. Add a `ConnectionOptions$Builder#setOpenTelemetry(OpenTelemetry openTelemetry)` method
2. Add this `ConnectionOptions.openTelemetry` property to the `SpannerPoolKey`, and set the `OpenTelemetry` instance on the `Spanner` instance that is built.
3. Not add a global `ConnectionOptions.enableOpenTelemetryMetrics()`. We don't need it. The users of the Connection API (JDBC + PGAdapter) should just call `SpannerOptions.enableOpenTelemetryMetrics()`.

Longer version:

First to clarify one thing: The `SpannerOptions.enableOpenTelemetryMetrics()` method only instructs the Java client to use OpenTelemetry __instead of__ OpenCensus for metrics. It does not itself create and register an OpenTelemetry instance etc.

We don't need to replicate that in the Connection API. Instead, a user of the Connection API should itself just call `SpannerOptions.enableOpenTelemetryMetrics()` directly before trying to use OpenTelemetry. This is probably something that they should do in any case, as none of them want to use OpenCensus. Note here that 'a user of the Connection API' in reality means 'JDBC' or 'PGAdapter'.

Regarding how to set the `OpenTelemetry` instance for a connection:
1. Adding a `ConnectionOptions$Builder#setOpenTelemetry(OpenTelemetry openTelemetry)` method makes sense. That property should then be propagated to the `SpannerOptions$Builder#setOpenTelemetry(...)` method.
4. That then also means that this new `OpenTelemetry` property needs to be added to the `SpannerPoolKey`, so that if two connections with different `OpenTelemetry` instances are created, they also get two different `Spanner` instances.

-----

The bigger challenge is how to create and configure the `OpenTelemetry` instance that should be set on a `ConnectionOptions` instance by the JDBC driver and PGAdapter. Note here that users of JDBC and PGAdapter do not always have programmatic access to the configuration (and none of them have access to `ConnectionOptions`). That is not something that needs to be solved in this PR, but we should make sure that we don't make a choice here that makes it impossible to use in JDBC and PGAdapter.

For PGAdapter, we have solved this by using the [AutoConfiguredOpenTelemetrySdk](https://github.com/GoogleCloudPlatform/pgadapter/blob/7af6e01aa1683381e254ab6860103898a636107d/src/main/java/com/google/cloud/spanner/pgadapter/Server.java#L107). This library reads configuration properties from System properties and/or environment variables that should be used for OpenTelemetry, and configures OpenTelemetry based on that. This OpenTelemetry instance can then be used for the `ConnectionOptions` instance that is created by PGAdapter.

For JDBC, I could imagine a couple of different options (all in parallel possible):
1. A public static method that sets a global `OpenTelemetry` object that should be used for all JDBC connections in the current JVM. So something like `JdbcDriver.setOpenTelemetry(OpenTelemetry openTelemetry)`. This method has to be called before creating any connections.
2. The standard way to get a JDBC connection is to call `DriverManager.getConnection(String url[, Properties properties])`. The `properties` map can also contain complex objects like `OpenTelemetry`. A user could then supply a specific `OpenTelemetry` that should be used for a connection in this `properties` argument. This would only be available to users who are using the JDBC driver directly, and not for users who are using for example Hibernate.
5. An `autoConfigOpenTelemetry=true` connection URL property. Setting this to true will activate similar logic as described for PGAdapter above: Using the `AutoConfigureOpenTelemetrySdk` an `OpenTelemetry` instance will be created and configured based on system properties and environment variables.

In all the above cases, the `OpenTelemetry` instance that has been configured will be passed on to the `ConnectionOptions` instance that is built.

hengfengli: Fixed. I wonder if JDBC should use the same approach as PGAdapter.

olavloite (approved):
LGTM, except for the comment on what to do when no `OpenTelemetry` has been set. I think that we should just return null, and let `SpannerOptions` determine the default.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
```diff
@@ -856,6 +867,19 @@ private static Integer parseIntegerProperty(String propertyName, String value) {
...
+   * Returns an instance of OpenTelemetry. If OpenTelemetry object is not set then
+   * GlobalOpenTelemetry will be used as fallback.
+   */
+  @Nonnull
+  OpenTelemetry getOpenTelemetry() {
+    if (this.openTelemetry != null) {
+      return this.openTelemetry;
+    } else {
+      return GlobalOpenTelemetry.get();
+    }
```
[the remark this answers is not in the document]
hengfengli: That's right. Fixed.

PR #2898, opened 2024-02-20
chore: add a kokoro presubmit workflow to run on cloud-devel.

arawind:


arpan14 (approved):

harshachinta:
@arpan14 
What is the need of adding a presubmit with cloud-devel? We never merge our PRs until the support is in Production.
We already have our presubmits flaky due to resource exhausted, and this will add up to that.

arpan14:
> @arpan14 What is the need of adding a presubmit with cloud-devel? We never merge our PRs until the support is in Production. We already have our presubmits flaky due to resource exhausted, and this will add up to that.

There are additional steps taken to ensure cloud-devel tests run against a different instance than production test. Additionally @arawind helped by separating the directpath tests to also run on a separate instance. All of this will reduce the current flakiness. 

Also we want to run the tests again cloud-devel as a non-required step to ensure when we merge changes with disabled integration tests (for ex - proto/FLOAT32), we at-least have ensured that such tests pass against cloud-devel and fail against production.

PR #2921, opened 2024-02-22
chore: refactor statement parsers to re-use more code

olavloite:
[WORK IN PROGRESS]

Combine some of the methods in the statement parsers into one method in the abstract base class, so more code can be re-used. This will also be used to support keeping the comments that are sent to the JDBC driver, so the comments can also be sent to Spanner.


olavloite:
Replaced by separate, smaller PRs.

PR #2927, opened 2024-02-23
chore: replace client creation methods for all new samples with auto-generated admin clients

arpan14:


samples/snippets/src/main/java/com/example/spanner/admin/generated/AddAndDropDatabaseRole.java
```diff
@@ -39,36 +40,41 @@ static void addAndDropDatabaseRole() throws IOException {
...
-    } catch (ExecutionException | TimeoutException e) {
-      System.out.printf(
-          "Error: AddAndDropDatabaseRole failed with error message %s\n", e.getMessage());
-      e.printStackTrace();
-    } catch (InterruptedException e) {
-      System.out.println(
-          "Error: Waiting for AddAndDropDatabaseRole operation to finish was interrupted");
+      String projectId, String instanceId, String databaseId,
+      String parentRole, String childRole) {
+    try (Spanner spanner =
```
olavloite, line 45: In these cases, you can combine these two `try-with-resources` blocks into one like this:

```java
try (Spanner spanner = SpannerOptions.newBuilder()
            .setProjectId(projectId)
            .build()
            .getService();
      DatabaseAdminClient databaseAdminClient = spanner.createDatabaseAdminClient()) {
   ....
}
```

That reduces the nested indentation in the sample.

samples/snippets/src/test/java/com/example/spanner/admin/generated/SampleTestBaseV2.java
```diff
@@ -113,6 +113,7 @@ public static void afterClass() throws InterruptedException {
       }
     }
     for (String backupId : idGenerator.getBackupIds()) {
+      System.out.println("Trying to delete " + backupId);
```
olavloite, line 116: nit: I assume that this and the other System.out.println will be removed before merging.

olavloite (approved):

samples/snippets/src/test/java/com/example/spanner/admin/generated/SampleTestBaseV2.java
arpan14: This is actually pretty useful. At the end of the test it prints if all the resources were cleaned up or not.

PR #2930, opened 2024-02-28
docs: add a sample to show how different admin clients (legacy and new) can be used together in same code block.

arpan14:


olavloite:

PR #2948, opened 2024-03-13
refactor: move skip methods to abstract parser

olavloite:
Move the PostgreSQL skip methods from the PostgreSQL parser to the abstract parser. This is step 1 in refactoring the GoogleSQL and PostgreSQL parser so they can share more code. The eventual goal is to allow the GoogleSQL parser to be able to handle SQL string without having to remove the comments from the string first.

This pull request only moves some methods from the class PostgreSQLStatementParser to AbstractStatementParser.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/AbstractStatementParser.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/AbstractStatementParser.java
@@ -40,6 +40,7 @@ import java.util.Set;
 import java.util.concurrent.Callable;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import javax.annotation.Nullable;
 
 /**
  * Internal class for the Spanner Connection API.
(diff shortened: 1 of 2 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/AbstractStatementParser.java
```diff
@@ -696,4 +697,169 @@ static int countOccurrencesOf(char c, String string) {
   public boolean checkReturningClause(String sql) {
     return checkReturningClauseInternal(sql);
   }
+
+  /**
+   * Returns true for characters that can be used as the first character in unquoted identifiers.
+   */
+  boolean isValidIdentifierFirstChar(char c) {
```
olavloite, line 704: All the methods that are added here, are only copy-pasted from `PostgreSQLStatementParser`. We do this as step 1 for a larger refactoring, so the next steps are easier to follow and review.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/PostgreSQLStatementParser.java
```diff
@@ -136,23 +135,6 @@ String removeCommentsAndTrimInternal(String sql) {
     return res.toString().trim();
   }
 
-  String parseDollarQuotedString(String sql, int index) {
```
olavloite, line 139: This and the other methods here that appear to be deleted, are just moved to `AbstractStatementParser`, which is the super class of this class. That means that functionally, nothing changes.

pratickchokhani (approved):

PR #2953, opened 2024-03-18
fix: return type of max commit delay option.

arpan14:
* Use `TransactionOption` as return type instead of `TransactionOption`


nginsberg-google (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
```diff
@@ -141,7 +141,7 @@ public static ReadQueryUpdateTransactionOption priority(RpcPriority priority) {
     return new PriorityOption(priority);
   }
 
-  public static ReadQueryUpdateTransactionOption maxCommitDelay(Duration maxCommitDelay) {
+  public static TransactionOption maxCommitDelay(Duration maxCommitDelay) {
```
olavloite, line 144: nit: Can we also add a short javadoc comment to this public method that indicates where/how to use it?

olavloite (approved):
LGTM, with a couple of nits:
1. We also need to add an entry to the `clirr-ignored-differences.xml` file.
2. The PR should be marked as `fix:`. This is something that should be included in the release notes to warn anyone who might already use this option.
3. There's something strange going on with your fork or branch, as it includes 32 other commits (they don't show up in the changes, so it does not really affect the PR, but it is something that should be cleaned up for future PRs)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
arpan14: I have asked Navarre from the backend team to add this since I didn't have the exact details on how our external docs are explaining this feature.

PR #2959, opened 2024-03-19
feat: add support for transaction-level exclusion from change streams

dengwe1:
Add support for excluding transactions from being recorded in the change streams by passing `Options.excludeTxnFromChangeStreams()` in the various write APIs:

- `writeWithOptions`
- `writeAtleastOnceWithOptions`
- `batchWriteAtLeastOnce`
- `executePartitionedUpdate`
- `readWriteTransaction`
- `transactionManager`
- `transactionManagerAsync`
- `runAsync`

Note: Samples and integration tests will be added in separate PRs.

### DIFF
--- a/google-cloud-spanner/clirr-ignored-differences.xml
+++ b/google-cloud-spanner/clirr-ignored-differences.xml
@@ -631,11 +631,4 @@
     <className>com/google/cloud/spanner/connection/Connection</className>
     <method>void setDirectedRead(com.google.spanner.v1.DirectedReadOptions)</method>
   </difference>
-
-  <difference>
-    <differenceType>7005</differenceType>
-    <className>com/google/cloud/spanner/DatabaseClient</className>
-    <method>long executePartitionedUpdate(com.google.cloud.spanner.Statement, com.google.cloud.spanner.Options$UpdateOption[])</method>
-    <to>long executePartitionedUpdate(com.google.cloud.spanner.Statement, com.google.cloud.spanner.Options$PartitionedUpdateOption[])</to>
-  </difference>
 </differences>
--- a/google-cloud-spanner/clirr-ignored-differences.xml
+++ b/google-cloud-spanner/clirr-ignored-differences.xml
@@ -631,4 +631,11 @@
     <className>com/google/cloud/spanner/connection/Connection</className>
     <method>void setDirectedRead(com.google.spanner.v1.DirectedReadOptions)</method>
   </difference>
+
+  <difference>
+    <differenceType>7005</differenceType>
+    <className>com/google/cloud/spanner/DatabaseClient</className>
+    <method>long executePartitionedUpdate(com.google.cloud.spanner.Statement, com.google.cloud.spanner.Options$UpdateOption[])</method>
+    <to>long executePartitionedUpdate(com.google.cloud.spanner.Statement, com.google.cloud.spanner.Options$PartitionedUpdateOption[])</to>
+  </difference>
 </differences>
--- a/google-cloud-spanner/clirr-ignored-differences.xml
+++ b/google-cloud-spanner/clirr-ignored-differences.xml
@@ -605,6 +605,14 @@
     <method>com.google.cloud.spanner.admin.instance.v1.stub.InstanceAdminStubSettings getInstanceAdminStubSettings()</method>
   </difference>
 
+  <!-- (Modify return type of maxCommitDelay. Feature not announced) -->
+  <difference>
+    <differenceType>7006</differenceType>
+    <className>com/google/cloud/spanner/Options</className>
+    <method>com.google.cloud.spanner.Options$ReadQueryUpdateTransactionOption maxCommitDelay(java.time.Duration)</method>
+    <to>com.google.cloud.spanner.Options$TransactionOption</to>
+  </difference>
+
   <difference>
     <differenceType>7005</differenceType>
     <className>com/google/cloud/spanner/PartitionedDmlTransaction</className>
--- a/.kokoro/presubmit/graalvm-native-17.cfg
+++ b/.kokoro/presubmit/graalvm-native-17.cfg
@@ -3,7 +3,7 @@
 # Configure the docker image for kokoro-trampoline.
 env_vars: {
   key: "TRAMPOLINE_IMAGE"
-  value: "gcr.io/cloud-devrel-public-resources/graalvm_sdk_platform_b:3.27.0"
+  value: "gcr.io/cloud-devrel-public-resources/graalvm_sdk_platform_b:3.28.1"
 }
 
 env_vars: {
--- a/.kokoro/presubmit/graalvm-native.cfg
+++ b/.kokoro/presubmit/graalvm-native.cfg
@@ -3,7 +3,7 @@
 # Configure the docker image for kokoro-trampoline.
 env_vars: {
   key: "TRAMPOLINE_IMAGE"
-  value: "gcr.io/cloud-devrel-public-resources/graalvm_sdk_platform_a:3.27.0"
+  value: "gcr.io/cloud-devrel-public-resources/graalvm_sdk_platform_a:3.28.1"
 }
 
 env_vars: {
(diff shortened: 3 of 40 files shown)


dengwe1:
@ShuranZhang @arpan14 Please review, thanks!

ShuranZhang (changes requested):
Overall looks good, I'll take a another deeper look in details tomorrow.

One major comment:
What will happen if users try to pass `Options.excludeTxnFromChangeStreams()` to request-level DML APIs which also take UpdateOptions as a parameter(ie:[executeUpdate](https://github.com/googleapis/java-spanner/blob/9962de399b102560dbef1c458510a549245f7701/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java#L740))? Seems like it will simply be a no-op since the new `UpdateTransactionOption` you added also extends `UpdateOptions`. I think it's better to prevent users from even trying to set `Options.excludeTxnFromChangeStreams()` for these request level DML APIs from the client library side, to avoid any potential confusions(ie: they were able to pass the option per-request and run the code, but records are still not excluded). 

For example, for above scenarios we might want to throw an error like “Option.excludeTxnFromChangeStreams() is not supported at request level”:
```
databaseClient
      .readWriteTransaction()
      .run(transaction -> {
        transaction.executeUpdate(
            Statement.of("UPDATE T1 SET Name='fool' WHERE id = 123"),Options.excludeTxnFromChangeStreams());
        return null;
      });

```
This should also apply for `executeUpdateAsync`, `batchUpdate` and `batchUpdateAsync`.

Thanks!

dengwe1:
> Overall looks good, I'll take a another deeper look in details tomorrow.
> 
> One major comment: What will happen if users try to pass `Options.excludeTxnFromChangeStreams()` to request-level DML APIs which also take UpdateOptions as a parameter(ie:[executeUpdate](https://github.com/googleapis/java-spanner/blob/9962de399b102560dbef1c458510a549245f7701/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java#L740))? Seems like it will simply be a no-op since the new `UpdateTransactionOption` you added also extends `UpdateOptions`. I think it's better to prevent users from even trying to set `Options.excludeTxnFromChangeStreams()` for these request level DML APIs from the client library side, to avoid any potential confusions(ie: they were able to pass the option per-request and run the code, but records are still not excluded).
> 
> For example, for above scenarios we might want to throw an error like “Option.excludeTxnFromChangeStreams() is not supported at request level”:
> 
> ```
> databaseClient
>       .readWriteTransaction()
>       .run(transaction -> {
>         transaction.executeUpdate(
>             Statement.of("UPDATE T1 SET Name='fool' WHERE id = 123"),Options.excludeTxnFromChangeStreams());
>         return null;
>       });
> ```
> 
> This should also apply for `executeUpdateAsync`, `batchUpdate` and `batchUpdateAsync`.
> 
> Thanks!

Great point! Yes, we only want to allow `Options.excludeTxnFromChangeStreams()` in the `executePartitionedUpdate` method, but not the other request-level `Update` methods.

I'm thinking of the following approach:
- Introduce a new interface `PartitionedUpdateOption` that is only applicable to the `executePartitionedUpdate` method.
- Modify the `executePartitionedUpdate` method to take `PartitionedUpdateOption` instead of `UpdateOption`.
- Let `UpdateOption` extend `PartitionedUpdateOption`, so that existing code that passes `UpdateOption` to `executePartitionedUpdate` would still work.
- Let `excludeTxnFromChangeStreams()` return a sub-type of both `PartitionedUpdateOption` and `TransactionOption`. Thus, `Options.excludeTxnFromChangeStreams()` can be passed to either `executePartitionedUpdate` or the other write APIs that accept `TransactionOption`. But it cannot be passed to the other `Update` methods that expect `UpdateOption`.

```
public interface PartitionedUpdateOption {}

public interface UpdateOption extends PartitionedUpdateOption {}

public interface PartitionedUpdateAndTransactionOption extends PartitionedUpdateOption, TransactionOption {}

public static PartitionedUpdateAndTransactionOption excludeTxnFromChangeStreams()
```

What do you think?

ShuranZhang (approved):

google-cloud-spanner/clirr-ignored-differences.xml
```diff
@@ -631,4 +631,11 @@
     <className>com/google/cloud/spanner/connection/Connection</className>
     <method>void setDirectedRead(com.google.spanner.v1.DirectedReadOptions)</method>
   </difference>
+
+  <difference>
```
olavloite, line 635: We cannot make this change, as it breaks binary compatibility of the client library.

olavloite (changes requested):

google-cloud-spanner/clirr-ignored-differences.xml
dengwe1: Thanks for pointing this out! Reverted this change.

dengwe1:
> > > Overall looks good, I'll take a another deeper look in details tomorrow.
> > > One major comment: What will happen if users try to pass `Options.excludeTxnFromChangeStreams()` to request-level DML APIs which also take UpdateOptions as a parameter(ie:[executeUpdate](https://github.com/googleapis/java-spanner/blob/9962de399b102560dbef1c458510a549245f7701/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java#L740))? Seems like it will simply be a no-op since the new `UpdateTransactionOption` you added also extends `UpdateOptions`. I think it's better to prevent users from even trying to set `Options.excludeTxnFromChangeStreams()` for these request level DML APIs from the client library side, to avoid any potential confusions(ie: they were able to pass the option per-request and run the code, but records are still not excluded).
> > > For example, for above scenarios we might want to throw an error like “Option.excludeTxnFromChangeStreams() is not supported at request level”:
> > > ```
> > > databaseClient
> > >       .readWriteTransaction()
> > >       .run(transaction -> {
> > >         transaction.executeUpdate(
> > >             Statement.of("UPDATE T1 SET Name='fool' WHERE id = 123"),Options.excludeTxnFromChangeStreams());
> > >         return null;
> > >       });
> > > ```
> > >   
> > > This should also apply for `executeUpdateAsync`, `batchUpdate` and `batchUpdateAsync`.
> > > Thanks!
> > 
> > 
> > Great point! Yes, we only want to allow `Options.excludeTxnFromChangeStreams()` in the `executePartitionedUpdate` method, but not the other request-level `Update` methods.
> > I'm thinking of the following approach:
> > 
> > * Introduce a new interface `PartitionedUpdateOption` that is only applicable to the `executePartitionedUpdate` method.
> > * Modify the `executePartitionedUpdate` method to take `PartitionedUpdateOption` instead of `UpdateOption`.
> > * Let `UpdateOption` extend `PartitionedUpdateOption`, so that existing code that passes `UpdateOption` to `executePartitionedUpdate` would still work.
> > * Let `excludeTxnFromChangeStreams()` return a sub-type of both `PartitionedUpdateOption` and `TransactionOption`. Thus, `Options.excludeTxnFromChangeStreams()` can be passed to either `executePartitionedUpdate` or the other write APIs that accept `TransactionOption`. But it cannot be passed to the other `Update` methods that expect `UpdateOption`.
> > 
> > ```
> > public interface PartitionedUpdateOption {}
> > 
> > public interface UpdateOption extends PartitionedUpdateOption {}
> > 
> > public interface PartitionedUpdateAndTransactionOption extends PartitionedUpdateOption, TransactionOption {}
> > 
> > public static PartitionedUpdateAndTransactionOption excludeTxnFromChangeStreams()
> > ```
> >   
> > What do you think?
> 
> Sounds good! Thanks for the explanation. Yeah it would be better in your approach, if the code itself does not compile rather than throwing an error afterwords. @arpan14 Could you please also take a look and confirm this is an acceptable approach? Thanks!

Thanks @olavloite for pointing out that this will break binary compatibility. I have reverted this change and added error handling in those `Update` methods.

olavloite (approved):

PR #2960, opened 2024-03-21
chore: add session pool options for multiplexed session.

arpan14:


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -271,6 +285,18 @@ long getRandomizePositionQPSThreshold() {
     return randomizePositionQPSThreshold;
   }
 
+  boolean getUseMultiplexedSession() {
```
olavloite, line 288: I assume that we are making these method package-private because we don't want to expose these methods publicly yet, but that they will be made public at a later time.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -669,6 +698,49 @@ Builder setPoolMaintainerClock(Clock poolMaintainerClock) {
...
+     * ensure that the multiplexed session is initialised. The default value for this is 10s.
+     *
+     * <p>If this is set to null, the client does not wait for session is initialised which means
+     * that the first few requests would see more latency since they will need to wait till the
+     * multiplexed session is initialised.
+     *
+     * @param waitForMultiplexedSession
+     * @return
+     */
+    Builder setWaitForMultiplexedSession(Duration waitForMultiplexedSession) {
```
olavloite, line 739: I think that this should somehow be combined with the `waitForMinSessions` option. It is illogical if the session pool blocks for multiplexed sessions, and not for other sessions, as it means that:
1. The first read operation will not have any additional latency.
2. The first write operations could have additional latency.

Or if we decide to keep them separate, then we should at least document this difference, so users can figure out why the first read/write transaction could get additional latency.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -669,6 +698,49 @@ Builder setPoolMaintainerClock(Clock poolMaintainerClock) {
...
+    }
+
+    /**
+     * This option is only useful when {@link SessionPoolOptions#useMultiplexedSession} is set to
+     * true. If greater than zero, we wait for the said duration at time of client initialization to
+     * ensure that the multiplexed session is initialised. The default value for this is 10s.
+     *
+     * <p>If this is set to null, the client does not wait for session is initialised which means
+     * that the first few requests would see more latency since they will need to wait till the
+     * multiplexed session is initialised.
```
olavloite, line 734:
```suggestion
    /**
     * This option is only used when {@link SessionPoolOptions#useMultiplexedSession} is set to
     * true. If greater than zero, calls to {@link Spanner#getDatabaseClient(DatabaseId)} will block
     * for up to the given duration while waiting for the multiplexed session to be created.
     * The default value for this is 10 seconds.
     *
     * <p>If this is set to null or zero, the client does not wait for the session to be created, which means
     * that the first read requests could see more latency, as they will need to wait until the
     * multiplexed session has been created.
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -669,6 +698,49 @@ Builder setPoolMaintainerClock(Clock poolMaintainerClock) {
...
+     *
+     * @param waitForMultiplexedSession
+     * @return
```
olavloite, line 737: nit: document or remove

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -669,6 +698,49 @@ Builder setPoolMaintainerClock(Clock poolMaintainerClock) {
...
+     * multiplexed session is sufficient to handle all concurrent traffic, so we don't require to
+     * provision any additional sessions.
+     *
+     * <p>When set to false, the client uses the regular session cached in the session pool for
+     * running 1 concurrent transaction per session. We require to provision sufficient sessions by
+     * making use of {@link SessionPoolOptions#minSessions} and {@link
+     * SessionPoolOptions#maxSessions} based on the traffic load. Failing to do so will result in
+     * higher latencies.
+     *
+     * @param useMultiplexedSession
```
olavloite, line 713: nit: document or remove

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -669,6 +698,49 @@ Builder setPoolMaintainerClock(Clock poolMaintainerClock) {
       return this;
     }
 
+    /**
+     * Sets whether the client should use multiplexed session or not. If set to true, the client
+     * optimises and runs multiple applicable requests concurrently on a single session. A single
+     * multiplexed session is sufficient to handle all concurrent traffic, so we don't require to
+     * provision any additional sessions.
```
olavloite, line 705:
```suggestion
     * multiplexed session is sufficient to handle all concurrent traffic.
```

olavloite (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
arpan14: Yes, you are right. Will make it public at a later point.

PR #2961, opened 2024-03-21
chore: add new members in SessionImpl for multiplexed session. Add a …

arpan14:


google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionImplTest.java
```diff
@@ -274,6 +276,49 @@ private static long utcTimeSeconds(int year, int month, int day, int hour, int m
...
+            session.singleUse(TimestampBound.strong());
+            try {
+              ResultSet resultSet =
+                  ctx1.read("Dummy", KeySet.all(), Collections.singletonList("C"));
+              assertNotNull(resultSet);
+            } catch (IllegalStateException ex) {
+              // in case of multiplexed session we should allow concurrent requests on same session
+              // and should not receive a IllegalStateException
+              Assert.fail();
+            }
```
olavloite, line 299: nit: do we need this catch block? I think that the test would also fail without it if we get an uncaught IllegalStateException.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionImplTest.java
```diff
@@ -515,4 +560,11 @@ public void multiUseReadOnlyTransactionReturnsMissingTransactionId() throws Pars
             () -> txn.readRow("Dummy", Key.of(), Collections.singletonList("C")));
     assertEquals(ErrorCode.INTERNAL, e.getErrorCode());
   }
+
+  private com.google.protobuf.Timestamp getCurrentGoogleTimestamp() {
```
olavloite, line 564: I think that `com.google.cloud.Timestamp.now().toProto() ` produces the same

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
```diff
@@ -754,7 +754,9 @@ void beforeReadOrQueryLocked() {
     // Note that transactions are invalidated under some circumstances on the backend, but we
     // implement the check more strictly here to encourage coding to contract rather than the
     // implementation.
-    checkState(isValid, "Context has been invalidated by a new operation on the session");
+    if (!session.getIsMultiplexed()) {
+      checkState(isValid, "Context has been invalidated by a new operation on the session");
+    }
```
olavloite, line 759: This works here now, but I'm not sure this is the actual thing that needs to be changed to allow multiple operations on the same session. The `isValid` flag is set to false when `invalidate()` is called. That is called from `SessionImpl#setActive(SessionTransaction)` method. That again sets the `activeTransaction` on a `SessionImpl`. That field is what I think needs to be refactored for multiplexed sessions, as `SessionImpl#activeTransaction` is not really a thing anymore, as one `SessionImpl` can now have multiple transactions. At least; the latter is what I get from the fact that we just create a normal `SessionImpl` instance when we create a multiplexed sessions.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionClient.java
```diff
@@ -224,6 +225,39 @@ SessionImpl createSession() {
...
+                  db.getName(),
+                  spanner.getOptions().getDatabaseRole(),
+                  spanner.getOptions().getSessionLabels(),
+                  null,
+                  true);
+      SessionImpl sessionImpl =
+          new SessionImpl(
+              spanner, session.getName(), session.getCreateTime(), session.getMultiplexed(), null);
+      consumer.onSessionReady(sessionImpl);
+    } catch (RuntimeException e) {
```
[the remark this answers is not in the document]
arpan14: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionClient.java
```diff
@@ -224,6 +225,39 @@ SessionImpl createSession() {
...
+  /**
+   * Create a multiplexed session and returns it to the given {@link SessionConsumer}. A multiplexed
+   * session is not affiliated with any GRPC channel. The given {@link SessionConsumer} is
+   * guaranteed to eventually get exactly 1 multiplexed session unless an error occurs. In case of
+   * an error on the gRPC calls, the consumer will receive one or more {@link
+   * SessionConsumer#onSessionCreateFailure(Throwable, int)} calls with the error.
+   *
+   * @param consumer The {@link SessionConsumer} to use for callbacks when sessions are available.
+   */
+  void createMultiplexedSession(SessionConsumer consumer) {
```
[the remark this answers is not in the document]
arpan14: Yes, I initially started by keeping the method behavior to be similar to existing `createSession` method which is also synchronous but it does not take any consumer. Thinking about how this method will be used a) in background thread where it will invoked in background but we would block for a session to get created because otherwise we cannot remove a stale session and b) In `BatchClient` where also a sync behavior is needed.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionClientTests.java
```diff
@@ -151,6 +152,82 @@ public void createAndCloseSession() {
...
+        .thenThrow(RuntimeException.class);
+    final AtomicInteger returnedSessionCount = new AtomicInteger();
+    final SessionConsumer consumer =
+        new SessionConsumer() {
+          @Override
+          public void onSessionReady(SessionImpl session) {}
+
+          @Override
+          public void onSessionCreateFailure(Throwable t, int createFailureForSessionCount) {
+            assertThat(t).isInstanceOf(RuntimeException.class);
```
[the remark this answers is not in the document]
arpan14: Done

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionImplTest.java
arpan14: I'm removing test and the current change in `AbstractReadContext` outside of this PR. I'll need to think a bit more on how `SessionTransaction` interface should be refactored to handle all use-cases.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionImplTest.java
arpan14: Changed.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionClient.java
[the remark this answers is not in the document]
olavloite: Let's leave it like this for now, but it might be that this should be changed at a later moment. Taking a `Consumer` as an input argument for a sync method is not really necessary. Instead, the method should just return the `SessionImpl`.
But considering that we are likely to use this method in the initialization in the Session pool, then I think that it will need to be changed to be async, so it can be called in the same way as `batchCreateSessions` can be called.

olavloite (approved):

PR #2973, opened 2024-03-26
chore: add multiplexed session implementations for CachedSession/SessionFuture  interfaces.

arpan14:
* Unit tests haven't been added in current PR. Will include in next PR which will contain the pool create/get methods for multiplexed session.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
@@ -574,6 +574,16 @@ class SessionPool {
     }
   }
 
+  class MultiplexedSessionReplacementHandler
+      implements SessionReplacementHandler<MultiplexedSessionFuture> {
+    @Override
+    public MultiplexedSessionFuture replaceSession(
+        SessionNotFoundException e, MultiplexedSessionFuture session) {
+      // TODO arpanmishra add this handling
+      return null;
+    }
+  }
+
   interface SessionNotFoundHandler {
     /**
      * Handles the given {@link SessionNotFoundException} by possibly converting it to a different
@@ -1420,6 +1430,246 @@ class SessionPool {
     }
   }
 
+  class MultiplexedSessionFuture extends SimpleForwardingListenableFuture<MultiplexedSession>
+      implements SessionFuture {
+
+    private volatile CountDownLatch initialized = new CountDownLatch(1);
+    private final ISpan span;
+
+    @VisibleForTesting
+    MultiplexedSessionFuture(ListenableFuture<MultiplexedSession> delegate, ISpan span) {
+      super(delegate);
+      this.span = span;
+    }
+
+    @Override
+    public Timestamp write(Iterable<Mutation> mutations) throws SpannerException {
+      return writeWithOptions(mutations).getCommitTimestamp();
+    }
+
+    @Override
+    public CommitResponse writeWithOptions(
+        Iterable<Mutation> mutations, TransactionOption... options) throws SpannerException {
+      try {
+        return get().writeWithOptions(mutations, options);
+      } finally {
+        close();
+      }
+    }
+
+    @Override
+    public Timestamp writeAtLeastOnce(Iterable<Mutation> mutations) throws SpannerException {
+      return writeAtLeastOnceWithOptions(mutations).getCommitTimestamp();
+    }
+
+    @Override
+    public CommitResponse writeAtLeastOnceWithOptions(
+        Iterable<Mutation> mutations, TransactionOption... options) throws SpannerException {
+      try {
+        return get().writeAtLeastOnceWithOptions(mutations, options);
+      } finally {
+        close();
+      }
+    }
+
+    @Override
+    public ServerStream<BatchWriteResponse> batchWriteAtLeastOnce(
+        Iterable<MutationGroup> mutationGroups, TransactionOption... options)
+        throws SpannerException {
+      try {
+        return get().batchWriteAtLeastOnce(mutationGroups, options);
+      } finally {
+        close();
+      }
+    }
+
+    @Override
+    public ReadContext singleUse() {
+      try {
+        return new AutoClosingReadContext<>(
+            session -> {
+              MultiplexedSession ms = session.get();
+              return ms.getDelegate().singleUse();
+            },
+            SessionPool.this,
+            multiplexedSessionReplacementHandler,
+            this,
+            true);
+      } catch (Exception e) {
+        close();
+        throw e;
+      }
+    }
+
+    @Override
+    public ReadContext singleUse(final TimestampBound bound) {
+      try {
+        return new AutoClosingReadContext<>(
+            session -> {
+              CachedSession ps = session.get();
+              return ps.getDelegate().singleUse(bound);
+            },
+            SessionPool.this,
+            multiplexedSessionReplacementHandler,
+            this,
+            true);
+      } catch (Exception e) {
+        close();
+        throw e;
+      }
+    }
+
+    @Override
+    public ReadOnlyTransaction singleUseReadOnlyTransaction() {
+      return internalReadOnlyTransaction(
+          session -> {
+            MultiplexedSession ms = session.get();
+            return ms.getDelegate().singleUseReadOnlyTransaction();
+          },
+          true);
+    }
+
+    @Override
+    public ReadOnlyTransaction singleUseReadOnlyTransaction(final TimestampBound bound) {
+      return internalReadOnlyTransaction(
+          session -> {
+            MultiplexedSession ms = session.get();
+            return ms.getDelegate().singleUseReadOnlyTransaction(bound);
+          },
+          true);
+    }
+
+    @Override
+    public ReadOnlyTransaction readOnlyTransaction() {
+      return internalReadOnlyTransaction(
+          session -> {
+            MultiplexedSession ms = session.get();
+            return ms.getDelegate().readOnlyTransaction();
+          },
+          false);
+    }
+
+    @Override
+    public ReadOnlyTransaction readOnlyTransaction(final TimestampBound bound) {
+      return internalReadOnlyTransaction(
+          session -> {
+            MultiplexedSession ms = session.get();
+            return ms.getDelegate().readOnlyTransaction(bound);
+          },
+          false);
+    }
+
+    private ReadOnlyTransaction internalReadOnlyTransaction(
+        Function<MultiplexedSessionFuture, ReadOnlyTransaction> transactionSupplier,
+        boolean isSingleUse) {
+      try {
+        return new AutoClosingReadTransaction<>(
+            transactionSupplier,
+            SessionPool.this,
+            multiplexedSessionReplacementHandler,
+            this,
+            isSingleUse);
+      } catch (Exception e) {
+        close();
+        throw e;
+      }
+    }
+
+    @Override
+    public TransactionRunner readWriteTransaction(TransactionOption... options) {
+      return new SessionPoolTransactionRunner<>(
+          this, multiplexedSessionReplacementHandler, options);
+    }
+
+    @Override
+    public TransactionManager transactionManager(TransactionOption... options) {
+      return new AutoClosingTransactionManager<>(
+          this, multiplexedSessionReplacementHandler, options);
+    }
+
+    @Override
+    public AsyncRunner runAsync(TransactionOption... options) {
+      return new SessionPoolAsyncRunner(this, multiplexedSessionReplacementHandler, options);
+    }
+
+    @Override
+    public AsyncTransactionManager transactionManagerAsync(TransactionOption... options) {
+      return new SessionPoolAsyncTransactionManager<>(
+          multiplexedSessionReplacementHandler, this, options);
+    }
+
+    @Override
+    public long executePartitionedUpdate(Statement stmt, UpdateOption... options) {
+      try {
+        return get().executePartitionedUpdate(stmt, options);
+      } finally {
+        close();
+      }
+    }
+
+    @Override
+    public String getName() {
+      return get().getName();
+    }
+
+    @Override
+    public void prepareReadWriteTransaction() {
+      get().prepareReadWriteTransaction();
+    }
+
+    @Override
+    public void close() {
+      try {
+        asyncClose().get();
+      } catch (InterruptedException e) {
+        throw SpannerExceptionFactory.propagateInterrupt(e);
+      } catch (ExecutionException e) {
+        throw SpannerExceptionFactory.asSpannerException(e.getCause());
+      }
+    }
+
+    @Override
+    public ApiFuture<Empty> asyncClose() {
+      MultiplexedSession delegate = getOrNull();
+      if (delegate != null) {
+        return delegate.asyncClose();
+      }
+      return ApiFutures.immediateFuture(Empty.getDefaultInstance());
+    }
+
+    private MultiplexedSession getOrNull() {
+      try {
+        return get();
+      } catch (Throwable t) {
+        return null;
+      }
+    }
+
+    @Override
+    public MultiplexedSession get() {
+      MultiplexedSession res = null;
+      try {
+        res = super.get();
+      } catch (Throwable e) {
+        // ignore the exception as it will be handled by the call to super.get() below.
+      }
+      if (res != null) {
+        res.markBusy(span);
+      }
+      initialized.countDown();
+
+      try {
+        // TODO arpanmishra@ do we really need this latch?
+        initialized.await();
+        return super.get();
+      } catch (ExecutionException e) {
+        throw SpannerExceptionFactory.newSpannerException(e.getCause());
+      } catch (InterruptedException e) {
+        throw SpannerExceptionFactory.propagateInterrupt(e);
+      }
+    }
+  }
+
   interface CachedSession extends Session {
 
     SessionImpl getDelegate();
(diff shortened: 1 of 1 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -574,6 +574,23 @@ public PooledSessionFuture replaceSession(
...
+        SessionNotFoundException e, MultiplexedSessionFuture session) {
+      /**
+       * For multiplexed sessions, we would never obtain a {@link SessionNotFoundException}. Hence,
+       * this method will ideally never be invoked.
+       */
+      logger.log(
+          Level.WARNING,
+          String.format(
+              "Replace session invoked for multiplexed " + "session => %s", session.getName()));
+      return session;
```
olavloite, line 590: I think that we should just propagate the `SessionNotFoundException` here. It is not something that should happen, but if it does happen for whatever reason, the current implementation is likely to just get into an eternal loop.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1420,6 +1437,247 @@ PooledSession get(final boolean eligibleForLongRunning) {
     }
   }
 
+  class MultiplexedSessionFuture extends SimpleForwardingListenableFuture<MultiplexedSession>
+      implements SessionFuture {
+
+    private volatile CountDownLatch initialized = new CountDownLatch(1);
```
olavloite, line 1443: This should be `final`, not `volatile`.
```suggestion
    private final CountDownLatch initialized = new CountDownLatch(1);
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1420,6 +1437,247 @@ PooledSession get(final boolean eligibleForLongRunning) {
...
+        close();
+      }
+    }
+
+    @Override
+    public ReadContext singleUse() {
+      try {
+        return new AutoClosingReadContext<>(
+            session -> {
+              MultiplexedSession ms = session.get();
```
olavloite, line 1498: According to the Java style guide, variable names should not be abbreviations. So `multiplexedSession` instead of `ms`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1420,6 +1437,247 @@ PooledSession get(final boolean eligibleForLongRunning) {
...
+              return ms.getDelegate().singleUse();
```
olavloite, line 1499: I don't remember what the definition of `MultiplexedSession` is, but is the call to `MultiplexedSession#getDelegate()` needed here? Or could it be just `multiplexedSession.singleUse()`?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1420,6 +1437,247 @@ PooledSession get(final boolean eligibleForLongRunning) {
...
+        throw e;
+      }
+    }
+
+    @Override
+    public ReadContext singleUse(final TimestampBound bound) {
+      try {
+        return new AutoClosingReadContext<>(
+            session -> {
+              CachedSession ps = session.get();
```
olavloite, line 1516:
```suggestion
              CachedSession cachedSession = session.get();
```

olavloite: But also: Why is it `CachedSession` here, and `MultiplexedSession` in the methods above?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1420,6 +1437,247 @@ PooledSession get(final boolean eligibleForLongRunning) {
...
+        return delegate.asyncClose();
+      }
+      return ApiFutures.immediateFuture(Empty.getDefaultInstance());
+    }
+
+    private MultiplexedSession getOrNull() {
+      try {
+        return get();
+      } catch (Throwable t) {
+        return null;
```
olavloite, line 1651: This needs a comment to explain why this is OK. (Is it really OK?)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1420,6 +1437,247 @@ PooledSession get(final boolean eligibleForLongRunning) {
...
+        span.addAnnotation("Awaiting multiplexed session to be initialized");
+        initialized.await();
+        span.addAnnotation("Multiplexed session initialized");
+        return super.get();
+      } catch (ExecutionException e) {
+        throw SpannerExceptionFactory.newSpannerException(e.getCause());
+      } catch (InterruptedException e) {
+        throw SpannerExceptionFactory.propagateInterrupt(e);
+      }
+    }
```
olavloite, line 1678: This looks strange. Why are we first doing a `initialized.countDown()` and then a `initialized.await()`? We know that initialized is a `CountDownLatch` with value 1, meaning that the count down here will always open the latch, which again means that the await won't do anything. Or am I missing something here?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1730,6 +1988,168 @@ public TransactionManager transactionManager(TransactionOption... options) {
     }
   }
 
+  class MultiplexedSession implements CachedSession {
+    SessionImpl delegate;
```
olavloite, line 1992: This should be final, as we do not allow replacing.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1730,6 +1988,168 @@ public TransactionManager transactionManager(TransactionOption... options) {
...
+
+    @Override
+    public void setAllowReplacing(boolean allowReplacing) {
+      // for multiplexed session there is only 1 session, there is nothing that can be replaced.
+      // hence this is no-op.
+    }
+
+    @Override
+    public void markBusy(ISpan span) {
+      this.delegate.setCurrentSpan(span);
```
olavloite, line 2013: Isn't `delegate` the single `SessionImpl` instance that we have for our only multiplexed session? Meaning that calling this will overwrite the value that was set by the previous `MultiplexedSession` instance, even when that other `MultiplexedSession` is still being used for something?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1730,6 +1988,168 @@ public TransactionManager transactionManager(TransactionOption... options) {
...
+    public void markUsed() {
+      // no-op for a multiplexed session since we don't track the last-used time
+      // in case of multiplexed session
+    }
+
+    @Override
+    public SpannerException setLastException(SpannerException exception) {
+      // multiplexed sessions run more than one transaction concurrently. we cannot store the
+      // exception state as that is not applicable to all transactions running on the session.
+      return exception;
```
olavloite, line 2026: Shouldn't this then throw something like an `UnsupportedOperationException`? Or is it OK to call this method for multiplexed sessions, even though we can't store the exception?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1730,6 +1988,168 @@ public TransactionManager transactionManager(TransactionOption... options) {
...
+
+    @Override
+    public long executePartitionedUpdate(Statement stmt, UpdateOption... options) {
+      throw SpannerExceptionFactory.newSpannerException(
+          ErrorCode.UNIMPLEMENTED, "Unimplemented with Multiplexed Session");
+    }
+
+    @Override
+    public String getName() {
+      return delegate.getName();
```
olavloite, line 2130: Would it make sense to create a unique name for `MultiplexedSession`s? For example generating a name that is equal to the time the instance was created + the actual underlying session name. Otherwise, this method will just return the same name for all sessions, meaning that it cannot in any way be used for identifying a single instance.

Or is this method only used in places where it makes more sense to show the underlying session name?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2192,7 +2618,8 @@ enum Position {
   private final CountDownLatch waitOnMinSessionsLatch;
   private final SessionReplacementHandler pooledSessionReplacementHandler =
       new PooledSessionReplacementHandler();
-
+  private final SessionReplacementHandler multiplexedSessionReplacementHandler =
```
olavloite, line 2621: nit: this could even be `private static final`

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2162,9 +2582,15 @@ enum Position {
   @GuardedBy("lock")
   private long numSessionsAcquired = 0;
 
+  @GuardedBy("lock")
+  private long numMultiplexedSessionsAcquired = 0;
```
olavloite, line 2586: This does not appear to be updated anywhere in this PR. Is that intentional? (I can imagine it is, as it does not include the code for acquiring the session, but checking just to be sure)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Yeah, I added it since I was also adding `numMultiplexedSessionsReleased`. the code that uses this constant will be in a separate PR.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: I couldn't find a use-case where it needs to show a unique name. Majority of usages in `SessionPool` class is to use this to show the actual session name in the traces/logs.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Yes, it will overwrite the value. But I also think the `setSpan` method is obsolete and won't have any impact on the traces. See this line in `ISpan span = tracer.spanBuilder(READ_ONLY_TRANSACTION);` for `singleUse` method.

So for each new singleUse() method call a new span is getting created. And I don't think calling this setter will impact how the traces would appear for concurrent transactions. I'll keep this comment open till I do a few more manual tests.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: I have removed this `CountDownLatch` for multiplexed session now. I don't think its needed anymore since we can expect the session to always be available because its created in sync. Additionally for cases where it's not available, it will be added to a waiters queue.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Such a `CoundDownLatch` is not needed for multiplexed session initialization since we can expect it to always be available/initalized unless there is something weird.

I have removed this variable and its usages.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Done, changed to `multiplexedSession`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Changed to `MultiplexedSession`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: * Assuming you are asking about the null return value. We would actually never return a null value since the exception won't be thrown. For multiplexed session, the Future wrapper is on top of an instance that is already available.
* Added a comment to elaborate this.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: We need `MultiplexedSession#getDelegate()` similar to what we do for `PooledSession` class. We require to invoke the `singleUse` method of `SessionImpl` class.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Sure, I will make this change to propagate the exception.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -574,6 +574,23 @@ public PooledSessionFuture replaceSession(
...
+    public MultiplexedSessionFuture replaceSession(
+        SessionNotFoundException e, MultiplexedSessionFuture session) {
+      /**
+       * For multiplexed sessions, we would never obtain a {@link SessionNotFoundException}. Hence,
+       * this method will ideally never be invoked.
+       */
+      logger.log(
+          Level.WARNING,
+          String.format(
+              "Replace session invoked for multiplexed " + "session => %s", session.getName()));
```
olavloite, line 589: nit: remove unnecessary string concatenation

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1420,6 +1437,242 @@ PooledSession get(final boolean eligibleForLongRunning) {
...
+      if (delegate != null) {
+        return delegate.asyncClose();
+      }
+      return ApiFutures.immediateFuture(Empty.getDefaultInstance());
+    }
+
+    private MultiplexedSession getOrNull() {
+      try {
+        return get();
+      } catch (Throwable t) {
```
olavloite, line 1648: nit: The convention is to name the variable `ignore` if you are going to ignore it.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
olavloite: Makes sense, thanks

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
olavloite: If it is not needed, then we should remove it. This would otherwise just cause confusion.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
olavloite: ack

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
olavloite: ack

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1420,16 +1438,254 @@ PooledSession get(final boolean eligibleForLongRunning) {
...
+        throw SpannerExceptionFactory.propagateInterrupt(e);
+      }
+    }
+  }
+
   interface CachedSession extends Session {
 
     SessionImpl getDelegate();
 
+    @ObsoleteApi("This method can be removed once we fully migrate to multiplexed sessions.")
```
olavloite, line 1681: nit: I think that you're absolutely right that this will become obsolete, and it is very good to document that already now, but I don't think we should do it with this annotation, as it is not actually obsolete yet. This might cause some IDEs or build systems to issue warnings. So could we change it to just a normal comment (or potentially a TODO comment).

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1420,16 +1438,254 @@ PooledSession get(final boolean eligibleForLongRunning) {
...
     void markBusy(ISpan span);
 
     void markUsed();
 
     SpannerException setLastException(SpannerException exception);
 
+    @ObsoleteApi("This method can be removed once we fully migrate to multiplexed sessions.")
```
olavloite, line 1688: nit: Same here as above

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1730,6 +1986,179 @@ public TransactionManager transactionManager(TransactionOption... options) {
...
+    }
+
+    @Override
+    public void close() {
+      synchronized (lock) {
+        numMultiplexedSessionsReleased++;
+        if (lastException != null && isDatabaseOrInstanceNotFound(lastException)) {
+          // Mark this session pool as no longer valid and then release the session into the pool as
+          // there is nothing we can do with it anyways.
+          synchronized (lock) {
```
olavloite, line 2145: nit: remove, we already hold the lock

olavloite (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Sure, makes sense. I have added a TODO.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Done.

PR #2975, opened 2024-03-26
chore: add a metric label is_multiplexed for multiplexed session.

arpan14:
Introduced a new metric label `is_multiplexed` using which customers can see the split of data between multiplexed sessions and regular sessions. This works when customer uses open telemetry.


surbhigarg92:
LGTM . But we should also modify test cases to reflect this new metric label. 

https://github.com/googleapis/java-spanner/blob/main/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolTest.java#L1986

olavloite:
Both the Windows build and the integration-tests-against-emulator build fail with this message:

```
SessionPoolTest.testOpenTelemetrySessionMetrics:1939->verifyMetricData:2004 expected:<0> but was:<2>
```

arpan14:
> Both the Windows build and the integration-tests-against-emulator build fail with this message:
> 
> ```
> SessionPoolTest.testOpenTelemetrySessionMetrics:1939->verifyMetricData:2004 expected:<0> but was:<2>
> ```
> 
> It seems however that the message is wrong, as expected and actual are mixed up in that test. It actually expects 2, and got 0.

Yes, the flakiness originated due the below assertion. This assumed that metric data always contains 1 datapoint. whereas now because of the label, there will be two data-points (1 for regular session and another for multiplexed session). The datapoint for multiplexed session will be 0, since we haven't enabled it yet. And for regular session, it will be same as what it is today.
`assertEquals(
        metricData.getLongSumData().getPoints().stream().findFirst().get().getValue(), value);`

olavloite (approved):

surbhigarg92 (approved):

PR #2978, opened 2024-03-26
feat: support statement tags in comments in Connection API

olavloite:
Adds support for supplying statement tags as comments in the Connection API. This enables the use of statement tags in frameworks that only support SQL options.


olavloite:
Replaced by #3005

PR #2982, opened 2024-03-28
chore: create create/get methods for multiplexed session.

arpan14:
Key changes

* `maintainMultiplexedSession` - method executed as part of background task.
* `getMultiplexedSessionWithFallback` - method to obtain a multiplexed session.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
@@ -2668,6 +2668,7 @@ class SessionPool {
 
   private final AtomicReference<SettableApiFuture<MultiplexedSessionFuture>>
       currentMultiplexedSessionReference = new AtomicReference<>(SettableApiFuture.create());
+  MultiplexedSessionFutureWrapper wrappedMultiplexedSessionFuture = null;
 
   @GuardedBy("lock")
   private final Set<PooledSession> allSessions = new HashSet<>();
@@ -3015,9 +3016,13 @@ class SessionPool {
   }
 
   SessionFutureWrapper getWrappedMultiplexedSessionFuture() {
-    return new MultiplexedSessionFutureWrapper(currentMultiplexedSessionReference.get());
+    return wrappedMultiplexedSessionFuture;
   }
 
+  /**
+   * This method is a blocking method. It will block until the underlying {@code
+   * SettableApiFuture<MultiplexedSessionFuture>} is resolved.
+   */
   MultiplexedSessionFuture getMultiplexedSession() {
     return (MultiplexedSessionFuture) getWrappedMultiplexedSessionFuture().get();
   }
@@ -3487,7 +3492,7 @@ class SessionPool {
         SettableApiFuture<MultiplexedSessionFuture> settableApiFuture = SettableApiFuture.create();
         settableApiFuture.set(new MultiplexedSessionFuture(settableFuture));
         currentMultiplexedSessionReference.set(settableApiFuture);
-
+        wrappedMultiplexedSessionFuture = new MultiplexedSessionFutureWrapper(settableApiFuture);
         if (oldSession != null) {
           logger.log(
               Level.INFO,
@@ -3510,6 +3515,8 @@ class SessionPool {
     public void onSessionCreateFailure(Throwable t, int createFailureForSessionCount) {
       synchronized (lock) {
         multiplexedSessionBeingCreated = false;
+        wrappedMultiplexedSessionFuture =
+            new MultiplexedSessionFutureWrapper(currentMultiplexedSessionReference.get());
       }
       logger.log(
           Level.WARNING,
@@ -3535,7 +3542,7 @@ class SessionPool {
         SettableApiFuture<MultiplexedSessionFuture> settableApiFuture =
             currentMultiplexedSessionReference.get();
         settableApiFuture.set(new MultiplexedSessionFuture(settableFuture));
-        currentMultiplexedSessionReference.set(settableApiFuture);
+        wrappedMultiplexedSessionFuture = new MultiplexedSessionFutureWrapper(settableApiFuture);
         multiplexedSessionBeingCreated = false;
       }
     }
@@ -3549,6 +3556,8 @@ class SessionPool {
     public void onSessionCreateFailure(Throwable t, int createFailureForSessionCount) {
       synchronized (lock) {
         multiplexedSessionBeingCreated = false;
+        wrappedMultiplexedSessionFuture =
+            new MultiplexedSessionFutureWrapper(currentMultiplexedSessionReference.get());
         currentMultiplexedSessionReference.get().setException(newSpannerException(t));
       }
     }
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
@@ -108,6 +108,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.GuardedBy;
 import org.threeten.bp.Duration;
@@ -169,6 +170,7 @@ class SessionPool {
   }
 
   private abstract static class CachedResultSetSupplier implements Supplier<ResultSet> {
+
     private ResultSet cached;
 
     abstract ResultSet load();
@@ -1673,17 +1675,12 @@ class SessionPool {
 
     @Override
     public MultiplexedSession get() {
-      MultiplexedSession res = null;
       try {
-        res = super.get();
-      } catch (Throwable e) {
-        // ignore the exception as it will be handled by the call to super.get() below.
-      }
-      if (res != null) {
-        res.markBusy(span);
-      }
-      try {
-        return super.get();
+        MultiplexedSession res = super.get();
+        if (res != null) {
+          res.markBusy(span);
+        }
+        return res;
       } catch (ExecutionException e) {
         throw SpannerExceptionFactory.newSpannerException(e.getCause());
       } catch (InterruptedException e) {
@@ -2163,6 +2160,8 @@ class SessionPool {
 
   private final class MultiplexedSessionWaiterFuture
       extends ForwardingListenableFuture<MultiplexedSession> {
+
+    private final Duration MAX_MULTIPLEXED_SESSION_WAIT_TIMEOUT = Duration.ofSeconds(60);
     private final SettableFuture<MultiplexedSession> waiter = SettableFuture.create();
 
     @Override
@@ -2183,19 +2182,15 @@ class SessionPool {
       while (true) {
         ISpan span = tracer.spanBuilder(WAIT_FOR_MULTIPLEXED_SESSION);
         try (IScope waitScope = tracer.withSpan(span)) {
-          MultiplexedSession s = pollUninterruptedlyWithTimeout(options.getAcquireSessionTimeout());
-          if (s == null) {
-            // Set the status to DEADLINE_EXCEEDED and retry.
-            numMultiplexedSessionWaiterTimeouts.incrementAndGet();
-            tracer.getCurrentSpan().setStatus(ErrorCode.DEADLINE_EXCEEDED);
-          } else {
-            return s;
-          }
+          final Duration timeout =
+              MoreObjects.firstNonNull(
+                  options.getAcquireSessionTimeout(), MAX_MULTIPLEXED_SESSION_WAIT_TIMEOUT);
+          return pollUninterruptedlyWithTimeout(timeout);
         } catch (Exception e) {
           if (e instanceof SpannerException
-              && ErrorCode.RESOURCE_EXHAUSTED.equals(((SpannerException) e).getErrorCode())) {
+              && ErrorCode.DEADLINE_EXCEEDED.equals(((SpannerException) e).getErrorCode())) {
             numMultiplexedSessionWaiterTimeouts.incrementAndGet();
-            tracer.getCurrentSpan().setStatus(ErrorCode.RESOURCE_EXHAUSTED);
+            tracer.getCurrentSpan().setStatus(ErrorCode.DEADLINE_EXCEEDED);
           }
           span.setStatus(e);
           throw e;
@@ -2222,14 +2217,11 @@ class SessionPool {
           } catch (InterruptedException e) {
             interrupted = true;
           } catch (TimeoutException e) {
-            if (acquireSessionTimeout != null) {
-              throw SpannerExceptionFactory.newSpannerException(
-                  ErrorCode.RESOURCE_EXHAUSTED,
-                  String.format(
-                      "Timed out after waiting %s ms to acquire multiplexed session.",
-                      acquireSessionTimeout.toMillis()));
-            }
-            return null;
+            throw SpannerExceptionFactory.newSpannerException(
+                ErrorCode.DEADLINE_EXCEEDED,
+                String.format(
+                    "Timed out after waiting %s ms to acquire multiplexed session.",
+                    acquireSessionTimeout.toMillis()));
           } catch (ExecutionException e) {
             throw SpannerExceptionFactory.newSpannerException(e.getCause());
           }
@@ -2594,38 +2586,45 @@ class SessionPool {
     void maintainMultiplexedSession(Instant currentTime) {
       try {
         if (options.getUseMultiplexedSession()) {
-          Iterator<MultiplexedSession> iterator = multiplexedSessions.iterator();
-          while (iterator.hasNext()) {
-            final MultiplexedSession session = iterator.next();
-            final Duration durationFromCreationTime =
-                Duration.between(session.getDelegate().getCreateTime(), currentTime);
-            if (durationFromCreationTime.compareTo(
-                        options.getMultiplexedSessionMaintenanceDuration())
-                    > 0
-                && numMultiplexedSessionsBeingCreated == 0) {
-              logger.log(
-                  Level.INFO,
-                  String.format(
-                      "Replacing Multiplexed Session => %s since it's created before maintenance "
-                          + "window => %s days",
-                      session.getName(),
-                      options.getMultiplexedSessionMaintenanceDuration().toDays()));
-              createMultiplexedSessions();
-            }
+          if (multiplexedSessions.size() == 0 && numMultiplexedSessionsBeingCreated == 0) {
+            // this means that multiplexed session initialization during startup failed.
+            createMultiplexedSession();
+          } else {
+            List<MultiplexedSession> staleSessions = getStaleMultiplexedSession(currentTime);
+            if (!staleSessions.isEmpty()) {
+              Set<String> staleSessionNames =
+                  staleSessions.stream()
+                      .map(MultiplexedSession::getName)
+                      .collect(Collectors.toSet());
+              if (multiplexedSessions.size() == staleSessions.size()
+                  && numMultiplexedSessionsBeingCreated == 0) {
+                /**
+                 * All current multiplexed sessions are stale. We would create a new multiplexed
+                 * session only when all other sessions have gone stale.
+                 */
+                createMultiplexedSession();
+              }
 
-            // if there is are > 1 multiplexed sessions, this means there are some stale sessions
-            // since a multiplexed session is only created if a previous one is stale.
-            // remove all stale sessions. Note that there may be active transaction which are
-            // running
-            // on the stale sessions, hence we do not issue a DeleteSession RPC.
-            synchronized (lock) {
-              while (multiplexedSessions.size() > 1) {
-                final MultiplexedSession lastSession = multiplexedSessions.pollLast();
-                logger.log(
-                    Level.INFO,
-                    String.format("Removed Multiplexed Session => %s", lastSession.getName()));
-                if (multiplexedSessionRemovedListener != null) {
-                  multiplexedSessionRemovedListener.apply(lastSession);
+              /**
+               * Remove stale multiplexed sessions ensuring there is at-least 1 session in the pool.
+               * If there is only 1 session in the pool and it is stale, we would continue to keep
+               * it till a new session is created. Note that there may be active transaction which
+               * are running on the stale sessions, hence we do not issue a DeleteSession RPC.
+               */
+              synchronized (lock) {
+                while (multiplexedSessions.size() > 1) {
+                  final MultiplexedSession lastSession = multiplexedSessions.peekLast();
+                  if (staleSessionNames.contains(lastSession.getName())) {
+                    multiplexedSessions.pollLast();
+                    logger.log(
+                        Level.INFO,
+                        String.format(
+                            "Removed Multiplexed Session => %s created at => %s",
+                            lastSession.getName(), lastSession.getDelegate().getCreateTime()));
+                    if (multiplexedSessionRemovedListener != null) {
+                      multiplexedSessionRemovedListener.apply(lastSession);
+                    }
+                  }
                 }
               }
             }
@@ -2634,6 +2633,21 @@ class SessionPool {
       } catch (final Throwable t) {
         logger.log(Level.WARNING, "Failed to maintain multiplexed session", t);
       }
+    }
+
+    List<MultiplexedSession> getStaleMultiplexedSession(Instant currentTime) {
+      List<MultiplexedSession> staleSessions = new ArrayList<>();
+      Iterator<MultiplexedSession> iterator = multiplexedSessions.iterator();
+      while (iterator.hasNext()) {
+        final MultiplexedSession session = iterator.next();
+        final Duration durationFromCreationTime =
+            Duration.between(session.getDelegate().getCreateTime(), currentTime);
+        if (durationFromCreationTime.compareTo(options.getMultiplexedSessionMaintenanceDuration())
+            > 0) {
+          staleSessions.add(session);
+        }
+      }
+      return staleSessions;
     }
   }
 
@@ -3023,7 +3037,7 @@ class SessionPool {
         createSessions(options.getMinSessions(), true);
       }
       if (options.getUseMultiplexedSession()) {
-        createMultiplexedSessions();
+        createMultiplexedSession();
       }
     }
   }
@@ -3491,6 +3505,13 @@ class SessionPool {
   }
 
   @VisibleForTesting
+  int getNumMultiplexedSessionWaiters() {
+    synchronized (lock) {
+      return multiplexedSessionWaiters.size();
+    }
+  }
+
+  @VisibleForTesting
   int totalSessions() {
     synchronized (lock) {
       return allSessions.size();
@@ -3542,15 +3563,14 @@ class SessionPool {
     }
   }
 
-  private void createMultiplexedSessions() {
+  private void createMultiplexedSession() {
     logger.log(Level.FINE, String.format("Creating multiplexed sessions"));
     synchronized (lock) {
       try {
         numMultiplexedSessionsBeingCreated++;
         sessionClient.createMultiplexedSession(multiplexedSessionConsumer);
-      } catch (Throwable t) {
-        numMultiplexedSessionsBeingCreated--;
-        handleMultiplexedSessionsFailure(newSpannerException(t));
+      } catch (Throwable ignore) {
+        // such an exception will never be thrown. the exception will be passed onto the consumer.
       }
     }
   }
@@ -3604,6 +3624,7 @@ class SessionPool {
 
     @Override
     public void onSessionCreateFailure(Throwable t, int createFailureForSessionCount) {
+      numMultiplexedSessionsBeingCreated--;
       handleMultiplexedSessionsFailure(newSpannerException(t));
     }
   }
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
@@ -2483,8 +2483,7 @@ class SessionPool {
       try {
         if (options.getUseMultiplexedSession()) {
           synchronized (lock) {
-            if (multiplexedSessionFuture.get().get().get() != null
-                && isMultiplexedSessionStale(currentTime)) {
+            if (getMultiplexedSession().get() != null && isMultiplexedSessionStale(currentTime)) {
               /*
                This will attempt to create a new multiplexed session. if successfully created then
                the existing session will be replaced. Note that there maybe active transactions
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
@@ -70,7 +70,6 @@ import com.google.common.base.Supplier;
 import com.google.common.collect.ImmutableList;
 import com.google.common.util.concurrent.ForwardingListenableFuture;
 import com.google.common.util.concurrent.ForwardingListenableFuture.SimpleForwardingListenableFuture;
-import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.MoreExecutors;
 import com.google.common.util.concurrent.SettableFuture;
@@ -124,7 +123,6 @@ class SessionPool {
   private final TraceWrapper tracer;
   static final String WAIT_FOR_SESSION = "SessionPool.WaitForSession";
   static final String WAIT_FOR_MULTIPLEXED_SESSION = "SessionPool.WaitForMultiplexedSession";
-
 
   /**
    * If the {@link SessionPoolOptions#getWaitForMinSessions()} duration is greater than zero, waits
@@ -1180,7 +1178,6 @@ class SessionPool {
     return new PooledSessionFuture(future, span);
   }
 
-
   private MultiplexedSessionFuture createMultiplexedSessionFuture(
       ListenableFuture<MultiplexedSession> future, ISpan span) {
     return new MultiplexedSessionFuture(future, span);
@@ -2002,7 +1999,6 @@ class SessionPool {
   }
 
   class MultiplexedSession implements CachedSession {
-
     final SessionImpl delegate;
     private volatile SpannerException lastException;
 
@@ -2164,6 +2160,7 @@ class SessionPool {
       return ApiFutures.immediateFuture(Empty.getDefaultInstance());
     }
   }
+
   private final class MultiplexedSessionWaiterFuture
       extends ForwardingListenableFuture<MultiplexedSession> {
     private final SettableFuture<MultiplexedSession> waiter = SettableFuture.create();
@@ -2189,7 +2186,7 @@ class SessionPool {
           MultiplexedSession s = pollUninterruptedlyWithTimeout(options.getAcquireSessionTimeout());
           if (s == null) {
             // Set the status to DEADLINE_EXCEEDED and retry.
-            numWaiterTimeouts.incrementAndGet();
+            numMultiplexedSessionWaiterTimeouts.incrementAndGet();
             tracer.getCurrentSpan().setStatus(ErrorCode.DEADLINE_EXCEEDED);
           } else {
             return s;
@@ -2197,7 +2194,7 @@ class SessionPool {
         } catch (Exception e) {
           if (e instanceof SpannerException
               && ErrorCode.RESOURCE_EXHAUSTED.equals(((SpannerException) e).getErrorCode())) {
-            numWaiterTimeouts.incrementAndGet();
+            numMultiplexedSessionWaiterTimeouts.incrementAndGet();
             tracer.getCurrentSpan().setStatus(ErrorCode.RESOURCE_EXHAUSTED);
           }
           span.setStatus(e);
@@ -2209,11 +2206,12 @@ class SessionPool {
     }
 
     /**
-     * Method which allows to obtain a multiplexed session after blocking for a configurable duration
-     * {@link SessionPoolOptions#getAcquireSessionTimeout()}. Note that this duration becomes
-     * obsolete in case we have set {@link SessionPoolOptions#getWaitForMultiplexedSession()}. Because
-     * {@link SessionPoolOptions#getWaitForMultiplexedSession()} will ensure that the multiplexed
-     * session is initialized and available during application start up.
+     * Method which allows to obtain a multiplexed session after blocking for a configurable
+     * duration {@link SessionPoolOptions#getAcquireSessionTimeout()}. Note that this duration
+     * becomes obsolete in case we have set {@link
+     * SessionPoolOptions#getWaitForMultiplexedSession()}. Because {@link
+     * SessionPoolOptions#getWaitForMultiplexedSession()} will ensure that the multiplexed session
+     * is initialized and available during application start up.
      */
     private MultiplexedSession pollUninterruptedlyWithTimeout(Duration acquireSessionTimeout) {
       boolean interrupted = false;
@@ -2227,7 +2225,8 @@ class SessionPool {
             if (acquireSessionTimeout != null) {
               throw SpannerExceptionFactory.newSpannerException(
                   ErrorCode.RESOURCE_EXHAUSTED,
-                  String.format("Timed out after waiting %s ms to acquire multiplexed session.",
+                  String.format(
+                      "Timed out after waiting %s ms to acquire multiplexed session.",
                       acquireSessionTimeout.toMillis()));
             }
             return null;
@@ -2422,6 +2421,7 @@ class SessionPool {
         this.prevNumSessionsAcquired = SessionPool.this.numSessionsAcquired;
       }
       Instant currTime = clock.instant();
+      maintainMultiplexedSession(currTime);
       removeIdleSessions(currTime);
       // Now go over all the remaining sessions and see if they need to be kept alive explicitly.
       keepAliveSessions(currTime);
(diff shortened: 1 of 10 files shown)


rahul2393 (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2501,6 +2603,73 @@ private void removeLongRunningSessions(
...
+              if (multiplexedSessions.size() == staleSessions.size()
+                  && numMultiplexedSessionsBeingCreated == 0) {
+                /**
+                 * All current multiplexed sessions are stale. We would create a new multiplexed
+                 * session only when all other sessions have gone stale.
+                 */
+                createMultiplexedSession();
+              }
+
+              /**
```
olavloite, line 2629: nit (here and elsewhere): Javadoc (`/** */`) should only be used to document fields, methods, classes etc. For general comments that are clarifying code, use either `/* ... */` or `//` style comments. (These Javadoc comments are picked up by code style checkers as dangling Javadoc comments and generating warnings)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2574,9 +2743,16 @@ enum Position {
   @GuardedBy("lock")
   private final Queue<WaiterFuture> waiters = new LinkedList<>();
 
+  @GuardedBy("lock")
+  private final Queue<MultiplexedSessionWaiterFuture> multiplexedSessionWaiters =
```
olavloite, line 2747: I think that we can simplify and remove this whole waiter if we change multiplexed session(s) from being held in a queue to be held in a `SettableFuture`. Waiting for the multiplexed session would then be reduced to simply calling `Future.get()`, which then naturally blocks until ready, and then returns either the result or the error that happened.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2574,9 +2743,16 @@ enum Position {
...
+      new LinkedList<>();
+
   @GuardedBy("lock")
   private int numSessionsBeingCreated = 0;
 
+  @GuardedBy("lock")
+  private int numMultiplexedSessionsBeingCreated = 0;
```
olavloite, line 2754: Change to a boolean to make it clear that we only have 1 multiplexed session. This value would otherwise only ever by 0 or 1.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2605,6 +2781,16 @@ enum Position {
   private long numLeakedSessionsRemoved = 0;
 
   private AtomicLong numWaiterTimeouts = new AtomicLong();
+  private AtomicLong numMultiplexedSessionWaiterTimeouts = new AtomicLong();
```
olavloite, line 2784: Remove this. The multiplexed session is either:
1. Created and available.
2. Still being created. The 'waiter' should just wait for the configured timeout.
3. Failed. We should return an error.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2749,6 +2939,8 @@ private SessionPool(
     this.initOpenTelemetryMetricsCollection(openTelemetry, attributes);
     this.waitOnMinSessionsLatch =
         options.getMinSessions() > 0 ? new CountDownLatch(1) : new CountDownLatch(0);
+    this.multiplexedSessions = new ArrayDeque<>();
+    this.multiplexedSessionsInitialized = new CountDownLatch(1);
```
olavloite, line 2943: nit: initialize these final fields where they are declared, as they do not depend on any of the input of this constructor

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -3276,13 +3525,27 @@ private int numWaiters() {
...
+
   @VisibleForTesting
   int totalSessions() {
     synchronized (lock) {
       return allSessions.size();
     }
   }
 
+  @VisibleForTesting
+  int totalMultiplexedSessions() {
```
olavloite, line 3543: Remove this, as it gives the impression that there might be more than 1 multiplexed session

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -3343,6 +3628,28 @@ private void createSessions(final int sessionCount, boolean distributeOverChanne
...
+        numMultiplexedSessionsBeingCreated--;
+        while (!multiplexedSessionWaiters.isEmpty()) {
+          multiplexedSessionWaiters.poll().put(multiplexedSession);
+        }
+      }
+    }
+
+    @Override
+    public void onSessionCreateFailure(Throwable t, int createFailureForSessionCount) {
+      numMultiplexedSessionsBeingCreated--;
```
olavloite, line 3648: This should also unblock any threads blocked in `waitOnMultiplexedSession` and propagate the error to those threads.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
```diff
@@ -279,6 +279,7 @@ public DatabaseClient getDatabaseClient(DatabaseId db) {
                 labelValues,
                 attributesBuilder.build());
         pool.maybeWaitOnMinSessions();
+        pool.waitOnMultiplexedSession();
```
olavloite, line 282: Note that we are in an `synchronized (this)` block here, and that (in the current implementation) blocking is the default behavior. That means that we are blocking here while also holding a lock. That also means that multiple concurrent calls to `Spanner.getDatabaseClient(..)` will be serialized and wait for each other's multiplexed session to be created.
(Preferably, we should also move the `pool.maybeWaitOnMinSessions()` call to outside the block.)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2154,6 +2179,82 @@ public ApiFuture<Empty> asyncClose() {
     }
   }
 
+  private final class MultiplexedSessionWaiterFuture
+      extends ForwardingListenableFuture<MultiplexedSession> {
+
+    private final Duration MAX_MULTIPLEXED_SESSION_WAIT_TIMEOUT = Duration.ofSeconds(60);
```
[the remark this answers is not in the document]
olavloite: I still think that we should remove most (all?) of this, because almost all of the code here is not used anyways, unless the user has changed some default settings. See my further comments below.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2154,6 +2157,80 @@ public ApiFuture<Empty> asyncClose() {
...
+      waiter.setException(e);
+    }
+
+    @Override
+    public MultiplexedSession get() {
+      ISpan span = tracer.spanBuilder(WAIT_FOR_MULTIPLEXED_SESSION);
+      try (IScope waitScope = tracer.withSpan(span)) {
+        final Duration timeout =
+            MoreObjects.firstNonNull(
+                options.getAcquireSessionTimeout(), MAX_MULTIPLEXED_SESSION_WAIT_TIMEOUT);
```
olavloite, line 2185: Note that:
1. We will only enter a waiting state if the initial `CreateSession` call is still in flight. In all other cases, a multiplexed session will be available.
2. MAX_MULTIPLEXED_SESSION_WAIT_TIMEOUT is 60s.
3. The default timeout for the `CreateSession` RPC is 30s.
4. So the `CreateSession` RPC will timeout before this will timeout, unless the user has modified the default timeout value for `CreateSession`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2154,6 +2157,80 @@ public ApiFuture<Empty> asyncClose() {
...
+
+    /**
+     * Method which allows to obtain a multiplexed session after blocking for a configurable
+     * duration {@link SessionPoolOptions#getAcquireSessionTimeout()}. Note that this duration
+     * becomes obsolete in case we have set {@link
+     * SessionPoolOptions#getWaitForMultiplexedSession()}. Because {@link
+     * SessionPoolOptions#getWaitForMultiplexedSession()} will ensure that the multiplexed session
+     * is initialized and available during application start up.
+     */
+    private MultiplexedSession pollUninterruptedlyWithTimeout(Duration acquireSessionTimeout) {
```
olavloite, line 2208: This poll method made sense for regular sessions, because those could be blocked during normal application operation, waiting for a session to be returned by another thread. Now, waiting will only potentially happen at startup while `CreateSession` is in flight, and will never be longer than the timeout value for the `CreateSession` RPC (which by default is lower than the fixed timeout value that is used here).

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2501,6 +2579,46 @@ private void removeLongRunningSessions(
...
+          synchronized (lock) {
+            if (multiplexedSession.get() != null && isMultiplexedSessionStale(currentTime)) {
+              logger.log(
+                  Level.INFO,
+                  String.format(
+                      "Removed Multiplexed Session => %s created at => %s",
+                      multiplexedSession.get().getName(),
+                      multiplexedSession.get().getDelegate().getCreateTime()));
+              if (multiplexedSessionRemovedListener != null) {
+                multiplexedSessionRemovedListener.apply(multiplexedSession.get());
```
olavloite, line 2595: This should not be called here, as the existing session is not yet being removed.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2501,6 +2579,46 @@ private void removeLongRunningSessions(
...
+
+    boolean isMultiplexedSessionStale(Instant currentTime) {
+      final MultiplexedSession session = multiplexedSession.get();
+      final Duration durationFromCreationTime =
+          Duration.between(session.getDelegate().getCreateTime(), currentTime);
+      if (durationFromCreationTime.compareTo(options.getMultiplexedSessionMaintenanceDuration())
+          > 0) {
+        return true;
+      }
+      return false;
```
olavloite, line 2620: nit: just return the value from the comparison, instead of using an if statement

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
olavloite: See my comments further above. I don't think this metric makes sense for multiplexed sessions. Threads won't be waiting for a multiplexed session in the same way as for regular sessions, meaning that this metric would only potentially be increased during startup (and then also only if the user has changed the default timeout configuration).

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2605,6 +2730,14 @@ enum Position {
   private long numLeakedSessionsRemoved = 0;
 
   private AtomicLong numWaiterTimeouts = new AtomicLong();
+  private AtomicLong numMultiplexedSessionWaiterTimeouts = new AtomicLong();
+
+  /**
+   * A recently created multiplexed session is always added at the front/head of the queue. A stale
+   * session is removed from the last/back of the queue.
+   */
```
olavloite, line 2738: nit: out-of-date javadoc

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2922,6 +3066,42 @@ boolean isValid() {
...
+        if (session != null) {
+          span.addAnnotation("Acquired multiplexed session", "sessionId", session.getName());
+          incrementNumSessionsInUse(true);
+        } else {
+          span.addAnnotation("No multiplexed session available. Adding to waiter queue.");
+          maybeCreateMultiplexedSession();
+          waiter = new MultiplexedSessionWaiterFuture();
+          multiplexedSessionWaiters.add(waiter);
+        }
+        return checkoutMultiplexedSession(span, session, waiter);
```
olavloite, line 3098: I think that all of this could be replaced by a simple `return currentMultiplexedSessionFuture`. That is:
1. The `SessionPool` would have a field of type `AtomicReference<SettableApiFuture<MultiplexedSessions>>` that holds the current multiplexed session. During startup/initial creation, that future is still pending. In all other cases, it will be a future that is done, and that contains the current multiplexed session.
2. This method can then always just return the single instance of that future, instead of having to create additional objects every time a multiplexed session is 'checked out'. AFAICT, the only state that we maintain in `MultiplexedSessionFuture` is the `span`, but that is only used to set the span on the underlying `SessionImpl`, which is always the same instance, meaning that we don't need a separate `MultiplexedSessionFuture` instance for each checkout. (And it also means that we should get rid of the `span` field on that class, because it now does something that is clearly wrong and confusing.)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -3321,6 +3532,30 @@ private boolean canCreateSession() {
     }
   }
 
+  private void maybeCreateMultiplexedSession() {
+    synchronized (lock) {
+      if (!numMultiplexedSessionsBeingCreated) {
+        logger.log(Level.FINE, String.format("Creating multiplexed sessions"));
+        try {
+          numMultiplexedSessionsBeingCreated = true;
```
olavloite, line 3540: nit: rename this field to `multiplexedSessionBeingCreated`

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -3343,6 +3578,27 @@ private void createSessions(final int sessionCount, boolean distributeOverChanne
...
+        numMultiplexedSessionsBeingCreated = false;
+        while (!multiplexedSessionWaiters.isEmpty()) {
+          multiplexedSessionWaiters.poll().put(multiplexedSession.get());
+        }
+      }
+    }
+
+    @Override
+    public void onSessionCreateFailure(Throwable t, int createFailureForSessionCount) {
+      numMultiplexedSessionsBeingCreated = false;
```
olavloite, line 3597: This should only be updated while holding the lock

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -3321,6 +3532,30 @@ private boolean canCreateSession() {
...
+          numMultiplexedSessionsBeingCreated = true;
+          sessionClient.createMultiplexedSession(multiplexedSessionConsumer);
+        } catch (Throwable ignore) {
+          // such an exception will never be thrown. the exception will be passed onto the consumer.
+        }
+      }
+    }
+  }
+
+  private void handleMultiplexedSessionsFailure(SpannerException e) {
```
olavloite, line 3549: nit: rename to `handleCreateMultiplexedSessionFailure()`

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -3343,6 +3578,27 @@ private void createSessions(final int sessionCount, boolean distributeOverChanne
...
+        while (!multiplexedSessionWaiters.isEmpty()) {
+          multiplexedSessionWaiters.poll().put(multiplexedSession.get());
+        }
+      }
+    }
+
+    @Override
+    public void onSessionCreateFailure(Throwable t, int createFailureForSessionCount) {
+      numMultiplexedSessionsBeingCreated = false;
+      handleMultiplexedSessionsFailure(newSpannerException(t));
```
olavloite, line 3598: What is the state of the session pool if this happens for the initial `CreateSession` RPC?

The curent implementation seems to get into a non-valid state, as the maintainer will not try to re-create the session, but the error that we get here is only being propagated to the current waiters. That means that threads that try to get a multiplexed session in the future will get a ... `NullPointerException`? Or something else?

I think that the following would be simpler, both from a code perspective and usability perspective:
1. Keep the current multiplexed session in a `SettableFuture`. 
2. Let this method call `SettableFuture.setException(..)`.
3. Let all waiters just wait on that future (so simply by calling `SettableFuture.get()`). That will fail automatically when this method calls `SettableFuture.setException(..)`.
4. All threads that in the future try to get a multiplexed session also automatically get the exception that was set here when they try to call `SettableFuture.get()`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Removed.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Removed it from here and added to `onSessionReady` method.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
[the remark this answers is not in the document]
arpan14: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Sure, removed the metric.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Done, simplified the follow by using `SettableFuture`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Done, simplified this flow by using `AtomicReference<SettableApiFuture<MultiplexedSessionFuture>>`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Done, simplified this flow by using `AtomicReference<SettableApiFuture<MultiplexedSessionFuture>>.`

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: `Done, simplified this flow by using AtomicReference<SettableApiFuture<MultiplexedSessionFuture>>.`

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2922,6 +2962,40 @@ boolean isValid() {
     }
   }
 
+  /**
+   * Returns a multiplexed session. We would always return the session which is at the front of the
+   * queue since this will be the most recently created session.
```
olavloite, line 2967: nit: this comment is a bit out of date, as this implementation no longer uses a queue

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2606,6 +2635,10 @@ enum Position {
 
   private AtomicLong numWaiterTimeouts = new AtomicLong();
 
+  @GuardedBy("lock")
+  private final AtomicReference<SettableApiFuture<MultiplexedSessionFuture>>
+      multiplexedSessionFuture = new AtomicReference<>(SettableApiFuture.create());
```
olavloite, line 2640: nit: change the variable name to `currentMultiplexedSessionReference`. The current name seems to indicate that the field is of type `Future`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2922,6 +2962,40 @@ boolean isValid() {
...
+   * Returns a multiplexed session. We would always return the session which is at the front of the
+   * queue since this will be the most recently created session.
+   */
+  SessionFuture getMultiplexedSessionWithFallback() throws SpannerException {
+    if (options.getUseMultiplexedSession()) {
+      ISpan span = tracer.getCurrentSpan();
+      span.addAnnotation("Acquiring multiplexed session");
+      synchronized (lock) {
+        try {
+          SessionFuture sessionFuture = getMultiplexedSession();
```
olavloite, line 2975: This method should not be called while holding the lock, as it is potentially blocking (it calls `Future.get()`. It is also not necessary to hold the lock to call this method, as both `AtomicReference` and `SettableApiFuture` are thread safe.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2606,6 +2635,10 @@ enum Position {
 
   private AtomicLong numWaiterTimeouts = new AtomicLong();
 
+  @GuardedBy("lock")
```
olavloite, line 2638: This field does not need to be guarded by a lock, as it is thread safe. Also the `SettableApiFuture` that it holds is largely thread safe. The only thing that you need to take into account is that only one thread should be writing to it. Not because it writing is not thread safe, but because writing to it twice is not supported.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2922,6 +2962,40 @@ boolean isValid() {
...
   }
 
+  /**
+   * Returns a multiplexed session. We would always return the session which is at the front of the
+   * queue since this will be the most recently created session.
+   */
+  SessionFuture getMultiplexedSessionWithFallback() throws SpannerException {
+    if (options.getUseMultiplexedSession()) {
+      ISpan span = tracer.getCurrentSpan();
+      span.addAnnotation("Acquiring multiplexed session");
```
olavloite, line 2972: I don't think we should add any annotations or do anything else special with the current span when getting a multiplexed session. In almost all cases, this will be a non-blocking, very quick operation, and adding an annotation will rather add noise (and a few microseconds of latency) than any real information.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -3321,6 +3399,26 @@ private boolean canCreateSession() {
...
+        } catch (Throwable ignore) {
+          // such an exception will never be thrown. the exception will be passed onto the consumer.
+        }
+      }
+    }
+  }
+
+  private void handleCreateMultiplexedSessionFailure(SpannerException e) {
+    synchronized (lock) {
+      multiplexedSessionFuture.get().setException(e);
```
olavloite, line 3418: I'm wondering what the best strategy is here:
1. For the initial session creation: This is the way we should do it.
2. For session refreshing, we have to options:
   2.1. Set the exception here. That means that all operations from this point on will fail, as there is no longer a valid multiplexed session, and this error will be propagated to the application.
   2.2. Only log the error, and let the maintainer retry creating a new session. The old multiplexed session will then continue to be used by the application, and this will work as long as it does not expire completely.

I lean slightly towards option 2.2 for refreshing the session, as it does not break the application right away if session refresh fails once.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -3343,6 +3441,43 @@ private void createSessions(final int sessionCount, boolean distributeOverChanne
...
+  class MultiplexedSessionConsumer implements SessionConsumer {
+
+    @Override
+    public void onSessionReady(SessionImpl sessionImpl) {
+      final SettableFuture settableFuture = SettableFuture.create();
+      final MultiplexedSession newSession = new MultiplexedSession(sessionImpl);
+      settableFuture.set(newSession);
+
+      synchronized (lock) {
+        if (multiplexedSessionFuture.get() != null && multiplexedSessionFuture.get().isDone()) {
```
olavloite, line 3453: I think that the first check `multiplexedSessionFuture.get() != null` will always be true and is redundant.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -3343,6 +3441,43 @@ private void createSessions(final int sessionCount, boolean distributeOverChanne
...
+          final MultiplexedSession oldSession = getMultiplexedSession().get();
+          logger.log(
+              Level.INFO,
+              String.format(
+                  "Removed Multiplexed Session => %s created at => %s and",
+                  oldSession.getName(), oldSession.getDelegate().getCreateTime()));
+          if (multiplexedSessionRemovedListener != null) {
+            multiplexedSessionRemovedListener.apply(oldSession);
+          }
```
olavloite, line 3462: nit: could the code block for logging and notifying about the removal of the old session to after it has actually been removed? The reason that I think that would be better, is that I think that the code block that hands out the multiplexed session should not take the `lock` (see comment further above). That means that getting a session could overlap with this code block that has taken the lock, and that you could (very theoretically) get a situation where the `removedListener` is called for a session first, and then that the session that was notified as 'removed' was still handed out by the pool.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -3343,6 +3441,43 @@ private void createSessions(final int sessionCount, boolean distributeOverChanne
...
+        }
+        SettableApiFuture settableApiFuture = SettableApiFuture.create();
+        settableApiFuture.set(
+            new MultiplexedSessionFuture(settableFuture, tracer.getCurrentSpan()));
```
olavloite, line 3466: Why is `tracer.getCurrentSpan()` passed in to this constructor? The current span is the span that has been set on the background thread that created the session. Is that a span that we want to be set on this shared multiplexed session?

olavloite (commented):
Thanks, I really like how the structure is now looking! This will significantly simplify the entire session pool once we can get rid of all 'normal' sessions. I left a couple of minor comments, but I think that structurally this is largely how it should be.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Wouldn't customers get confused about the missing spans?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: I was taking care of the broader `setSpan()` refactoring in a different PR - https://github.com/googleapis/java-spanner/pull/3016 . But now, will remove `span` from the constructor in this PR itself since its a small change.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Yes, this was actually causing a flaky behavior in unit test `MultiplexedSessionMaintainerTest`. I have fixed it.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Yes, removed.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Removed the annotation.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Removed

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Reworded this docs.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Yes, I have now separated two consumers `MultiplexedSessionInitializationConsumer` and `MultiplexedSessionMaintainerConsumer` and achieved this behavior. Also as discussed added a 10 minutes delay in the maintainer task to ensure we don't retry too often in case of such failures when a session is already available.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
olavloite: * These should all be sub-spans of the actual spans (e.g. for a read-only transactions). So they would still see the main spans, but just miss one sub-span.
* The sub-span would always be a few microseconds, except for the very first few queries. So it would then trigger the question 'why do we include a span for this, but not for (insert any random in-mem operation that takes a few microseconds)?'.

Would it maybe make more sense to only add a span in case the multiplexed session has not yet been created?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -3343,6 +3451,85 @@ private void createSessions(final int sessionCount, boolean distributeOverChanne
...
+  }
+
+  /**
+   * Callback interface which is invoked when a multiplexed session is getting initialised for the
+   * first time when a session is getting created.
+   */
+  class MultiplexedSessionInitializationConsumer implements SessionConsumer {
+    @Override
+    public void onSessionReady(SessionImpl sessionImpl) {
+      final SettableFuture settableFuture = SettableFuture.create();
```
olavloite, line 3492: 1. Could we split the logic of 'create the first multiplexed session' and 'replace the existing multiplexed session' into the two sub-classes that we actually have? This is confusing, as the Javadoc says that this is only used for the initial creation.
2. I think that the current implementation creates a problem for threads that wait for the initial creation of a multiplexed session. Assume that the following happens:
 2.1. A `DatabaseClient` is created (currently non-blocking). The initial value of `currentMultiplexedSessionReference` is set to a `SettableApiFuture.create()` in the constructor.
2.2. A thread tries to get a multiplexed session before it is done. It is then kept waiting on line 3001 (`currentMultiplexedSessionReference.get().get()`). That line will use the `SettableApiFuture` that was created in the constructor.
2.3. This implementation just swaps the `SettableApiFuture` in the `currentMultiplexedSessionReference` with a new instance, which means that the initial `SettableApiFuture` will never finish. That again means that those initial threads will be blocked forever.

olavloite: The fact that both this problem, but also the one mentioned further above regarding the `singleUseIsNonBlocking()` test, seem to introduce problems that are not caught by the current tests, indicate that we should enable multiplexed sessions by default for this PR, so we can actually test the implementation. We can always flip the flag back to false once we feel confident that it works before actually merging it.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -3343,6 +3451,85 @@ private void createSessions(final int sessionCount, boolean distributeOverChanne
...
+      final MultiplexedSession newSession = new MultiplexedSession(sessionImpl);
+      settableFuture.set(newSession);
+
+      synchronized (lock) {
+        MultiplexedSession oldSession = null;
+        if (currentMultiplexedSessionReference.get().isDone()) {
+          oldSession = getMultiplexedSession().get();
+        }
+        SettableApiFuture settableApiFuture = SettableApiFuture.create();
```
olavloite, line 3501: nit: `SettableApiFuture` is a generic class and accepts a type argument. Please use that to clarify what type of object is stores.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2922,6 +2974,38 @@ boolean isValid() {
...
+        throw SpannerExceptionFactory.asSpannerException(t.getCause());
+      }
+    } else {
+      return getSession();
+    }
+  }
+
+  MultiplexedSessionFuture getMultiplexedSession() {
+    try {
+      return currentMultiplexedSessionReference.get().get();
```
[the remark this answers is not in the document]
arpan14: * `DatabaseClientImpl` is not modified in this PR and also the unit tests at `DatabaseClientImplTest` was not parameterized in this PR to keep the diff small. Should we take them at  https://github.com/googleapis/java-spanner/pull/2994 because that is the PR which swaps the method?
* I'll try to mitigate the getSession() blocking/non-blocking issue in the current PR. There is one other major refactoring for spans - https://github.com/googleapis/java-spanner/pull/3016 and post this we will parameterize most test classes to cover multiplexed session test coverage.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Yes, sounds good. I'll include a span just for the case where a session could not be acquired.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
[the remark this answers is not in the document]
arpan14: As discussed, I'm preferring approach #2. Created a new wrapper interface `SessionFutureWrapper` and a couple of implementation which allow keeping the getSession() method non-blocking.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: * Done. Split the logic to two different implementations. For initialisation, I am doing a get() and then set() on the `SettableApiFuture` object. That should fix the use-case you mentioned.
* For maintainer threads, get() / set() wont work since the Future is already set. Hence, swapping with a new Future.
* Regarding tests, I will use this single PR - https://github.com/googleapis/java-spanner/pull/3016 for the remainder of refactoring and parameterizing the tests.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1654,17 +1693,9 @@ private MultiplexedSession getOrNull() {
...
-      } catch (Throwable e) {
-        // ignore the exception as it will be handled by the call to super.get() below.
-      }
-      if (res != null) {
-        res.markBusy(span);
-      }
-      try {
-        return super.get();
+        MultiplexedSession res = super.get();
+        return res;
```
olavloite, line 1698: nit: combine into one line with just `return super.get()`

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2501,6 +2540,44 @@ private void removeLongRunningSessions(
         }
       }
     }
+
+    void maintainMultiplexedSession(Instant currentTime) {
+      try {
+        if (options.getUseMultiplexedSession()) {
+          synchronized (lock) {
+            if (getMultiplexedSession().get() != null && isMultiplexedSessionStale(currentTime)) {
```
olavloite, line 2548: nit: add a check for `getMultiplexedSession().is done()` at the start of the if statement. This prevents us from blocking here in the theoretical case that the maintainer tries to maintain the multiplexed session before it has been created. We should never call any blocking methods when holding the `lock`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2922,6 +3014,40 @@ boolean isValid() {
     }
   }
 
+  /**
+   * Returns a multiplexed session. The method fallbacks to a regular session if {@link
+   * SessionPoolOptions#useMultiplexedSession} is not set.
+   */
+  SessionFutureWrapper getMultiplexedSessionWithFallback() throws SpannerException {
+    if (options.getUseMultiplexedSession()) {
+      ISpan span = tracer.getCurrentSpan();
```
olavloite, line 3023: nit: move this to the `catch` clause, as it is only used there

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2922,6 +3014,40 @@ boolean isValid() {
...
+        span.addAnnotation("No multiplexed session available.");
+        throw SpannerExceptionFactory.asSpannerException(t.getCause());
+      }
+    } else {
+      return new PooledSessionFutureWrapper(getSession());
+    }
+  }
+
+  SessionFutureWrapper getWrappedMultiplexedSessionFuture() {
+    return new MultiplexedSessionFutureWrapper(currentMultiplexedSessionReference.get());
```
olavloite, line 3038: We only need one instance of `MultiplexedSessionFutureWrapper` per instance of `SettableFuture` that we have for the multiplexed session, as the wrapper does not contain any additional state. So we should just create one instance of it per multiplexed session that we have (meaning one at startup, and one each time that we successfully refresh the multiplexed session). That will reduce the number of objects that we have to create each time a read-only operation is executed.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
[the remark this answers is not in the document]
olavloite: Ack

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -3343,6 +3493,99 @@ private void createSessions(final int sessionCount, boolean distributeOverChanne
...
+    public void onSessionReady(SessionImpl sessionImpl) {
+      final SettableFuture<MultiplexedSession> settableFuture = SettableFuture.create();
+      final MultiplexedSession newSession = new MultiplexedSession(sessionImpl);
+      settableFuture.set(newSession);
+
+      synchronized (lock) {
+        SettableApiFuture<MultiplexedSessionFuture> settableApiFuture =
+            currentMultiplexedSessionReference.get();
+        settableApiFuture.set(new MultiplexedSessionFuture(settableFuture));
+        currentMultiplexedSessionReference.set(settableApiFuture);
```
olavloite, line 3570: nit: this can be removed

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -3321,6 +3451,26 @@ private boolean canCreateSession() {
...
+          multiplexedSessionBeingCreated = true;
+          sessionClient.createMultiplexedSession(sessionConsumer);
+        } catch (Throwable ignore) {
+          // such an exception will never be thrown. the exception will be passed onto the consumer.
+        }
+      }
+    }
+  }
+
+  private void handleCreateMultiplexedSessionFailure(SpannerException e) {
```
olavloite, line 3468: nit: I think we should remove this method and just add the logic directly to the `onSessionCreateFailure` method of the initialization consumer.  As it is now, it might confuse future developers who believe that it is a bug that the maintainer consumer is not calling this method.

olavloite (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2922,6 +3014,40 @@ boolean isValid() {
...
+    } else {
+      return new PooledSessionFutureWrapper(getSession());
+    }
+  }
+
+  SessionFutureWrapper getWrappedMultiplexedSessionFuture() {
+    return new MultiplexedSessionFutureWrapper(currentMultiplexedSessionReference.get());
+  }
+
+  MultiplexedSessionFuture getMultiplexedSession() {
```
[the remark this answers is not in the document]
arpan14: Yes, this is called by all invocations which need a session to be available (for ex - maintainer). Modifying the logic in this method to 

```
MultiplexedSessionFuture getMultiplexedSession() {
    return (MultiplexedSessionFuture) getWrappedMultiplexedSessionFuture().get();
  }
```

This allows callers to not do a get().get() on wrapper as post initialization there will always be a instance available.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Sorry, I did not understand this. It's not clear what needs to be removed.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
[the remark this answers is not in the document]
arpan14: As discussed, i've added necessary docs to mention this method is blocking.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Done.

arpan14: Yes, for the time-being it is stateless. Hence, stored a local variable `MultiplexedSessionFutureWrapper` and managing this in the call back methods. 

I fore-see that we may need to store the state of `span` in the Future object. I'll change this back to a stateful object based on how that refactoring goes at - https://github.com/googleapis/java-spanner/pull/3016

PR #3005, opened 2024-04-04
feat: support client-side hints for tags and priority

olavloite:
Supports including a hint in the SQL string to set a statement tag or an RPC priority for a single SQL statement in the Connection API. This makes it easier to use these features from frameworks and tools that only support SQL statements.

Replaces #2978


google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/SimpleParser.java
```diff
@@ -0,0 +1,302 @@
...
+      return true;
+    }
+    return false;
+  }
+
+  /**
+   * Returns true if the given character is valid as the first character of an identifier. That
+   * means that it can be used as the first character of an unquoted identifier.
+   */
+  private boolean isValidIdentifierFirstChar(char c) {
```
arpan14, line 225: The methods `isValidIdentifierFirstChar` and `isValidIdentifierChar` already exist in `AbstractStatementParser`. Is there a way to re-use?

arpan14 (approved):

PR #3012, opened 2024-04-05
feat: disable affinity for grpc-gcp when there is no channel hint.

arpan14:
SpannerOptions#enableGrpcGcpExtension enabled GRPC GCP extension which is a separate library that handles channel management. By default this assumes session to channel affinity. 

With multiplexed session we no longer maintain any channel hint with the session. However, grpc-gcp would still assume session to channel affinity and that may cause higher latencies. This change disables affinity whenever there is no channel hint associated with the session. Effectively this will disable affinity when multiplexed sessions are launched.


nimf (approved):
LGTM

psinghbay1 (approved):

olavloite:

PR #3016, opened 2024-04-06
refactor: remove setSpan() method from SessionImpl class.

arpan14:
Since multiplexed sessions can run more than one transaction concurrently, it shouldn't store a single span object. A new span object is created and stored in the context of the public method which is invoked. There are two changes in this PR

* setSpan() method is removed from the `SessionTransaction` interface in `SessionImpl` class.
* Modify setActive() method within `SessionImpl` interface.

setActive() method within SessionImpl does the below things

1. Handles nested transaction use-case
2. Invalidates active transactions - Disabling this will allow multiple transactions per session.
3. Resets readyTransactionId - The value for readyTransactionId is only set in the prepareReadWriteTransaction method. Otherwise the value anyway remains null.
4. Sets the span per active transaction - A session no longer can maintain a single span. Hence, we will require to change the way spans are added for multiplexed session. This is doable since for each new RPC, a new span object is created and stored in the local thread context. Instead of storing the span state in `SessionImpl`, we can fetch the span from context and set it.

For multiplexed session we are disabling https://github.com/googleapis/java-spanner/issues/2, https://github.com/googleapis/java-spanner/pull/3 and https://github.com/googleapis/java-spanner/pull/4 for launching read-only transactions. The functionality of spans/traces for multiplexed session will be tested and added in a separate PR.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionImpl.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionImpl.java
@@ -467,12 +467,13 @@ class SessionImpl implements Session {
 
   <T extends SessionTransaction> T setActive(@Nullable T ctx) {
     throwIfTransactionsPending();
-
-    if (activeTransaction != null) {
-      activeTransaction.invalidate();
+    if (!this.isMultiplexed) {
+      if (activeTransaction != null) {
+        activeTransaction.invalidate();
+      }
+      activeTransaction = ctx;
+      readyTransactionId = null;
     }
-    activeTransaction = ctx;
-    readyTransactionId = null;
     return ctx;
   }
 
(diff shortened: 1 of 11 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionImpl.java
```diff
@@ -485,9 +473,6 @@ <T extends SessionTransaction> T setActive(@Nullable T ctx) {
     }
     activeTransaction = ctx;
     readyTransactionId = null;
-    if (activeTransaction != null) {
```
surbhigarg92, line 488: This might be an issue.

arpan14:
Closing in favor of - https://github.com/googleapis/java-spanner/pull/3027

PR #3023, opened 2024-04-09
refactor: remove code related to prepareReadWriteTransactions.

arpan14:
* prepareReadWriteTransactions() is deprecated and has no usages.
* the refactoring will help more refactoring that will be made in `SessionImpl` class for multiplexed session.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionImpl.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionImpl.java
@@ -100,7 +100,6 @@ class SessionImpl implements Session {
   private final String name;
   private final DatabaseId databaseId;
   private SessionTransaction activeTransaction;
-  ByteString readyTransactionId;
   private final Map<SpannerRpc.Option, ?> options;
   private volatile Instant lastUseTime;
   @Nullable private final Instant createTime;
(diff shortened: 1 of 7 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionImpl.java
```diff
@@ -463,7 +446,7 @@ TransactionContextImpl newTransaction(Options options) {
     return TransactionContextImpl.newBuilder()
         .setSession(this)
         .setOptions(options)
-        .setTransactionId(readyTransactionId)
+        .setTransactionId(null)
```
arpan14, line 449: `readyTransactionId` was only being set in prepareReadWriteTransaction() method. In absence of it, this will always remain null. @olavloite could you give this a closer look if this looks safe? If yes, we can get rid of `readyTransactionId` variable.

olavloite: It should certainly be safe to get rid of the `readyTransactionId`. And we can probably also remove the `setTransactionId(..)` method. Or is it being used for something else?

olavloite (approved):
LGTM (with a question on whether we can remove `setTransactionId(..)` as well)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionImpl.java
arpan14: `setTransactionId(...)` sets the `transactionId` field in `TransactionContextImpl` 's builder class. This value is further used within the constructor of `TransactionContextImpl` to set member `transactionId`. And I see that its further used by other classes.

So it will be best to retain the setter.

PR #3025, opened 2024-04-09
chore: support overlapping transactions for the emulator

olavloite:
Adds support for overlapping transactions on the emulator in the Connection API, even when these are executed on one thread. This is normally not supported, as it requires multiple transactions to run concurrently on the emulator.

This change works around that by keeping track of what the active transaction on the emulator is, and then abort that transaction when another transaction needs to make progress. The transaction that was aborted will be retried up to the point where it was when it again wants to make progress.

This step-by-step progress is implemented by automatically setting a savepoint in the transaction after each successful step. The transaction is then rolled back to that savepoint when another transaction needs to make progress.

The feature is only enabled if the autoConfigEmulator flag is enabled.


google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ReadWriteTransaction.java
```diff
@@ -105,9 +131,20 @@ class ReadWriteTransaction extends AbstractMultiUseTransaction {
...
+      this.savepoint = savepoint;
+    }
+
+    Savepoint getSavepoint() {
+      return this.savepoint;
+    }
+  }
 
   static class Builder extends AbstractMultiUseTransaction.Builder<Builder, ReadWriteTransaction> {
+    private boolean useAutoSavepointsForEmulator;
```
arpan14, line 147: should we really need to include `ForEmulator` in the naming? We are not restricting some other client (apart form emulator) from setting this field. Or do we want to annotate this `setUseAutoSavepointsForEmulator` with InternalApi/VisibleForTesting annotations?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ReadWriteTransaction.java
```diff
@@ -331,7 +383,10 @@ void checkAborted() {
 
   void checkRolledBackToSavepoint() {
     if (this.rolledBackToSavepointException != null) {
-      if (savepointSupport == SavepointSupport.FAIL_AFTER_ROLLBACK) {
+      if (savepointSupport == SavepointSupport.FAIL_AFTER_ROLLBACK
+          && !((RollbackToSavepointException) this.rolledBackToSavepointException.getCause())
+              .getSavepoint()
+              .isAutoSavepoint()) {
```
arpan14, line 389: Would this throw a NPE when getSavepoint() returns null? I believe it is only set for emulator cases.

arpan14 (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ReadWriteTransaction.java
olavloite: It is very much intentional, and currently the only way to enable this through any public API is by setting the `autoConfigEmulator=true` connection URL. The reason for this is that the feature is really only intended for the emulator. For real Spanner, this feature should not be used, and instead the client should just rely on the Spanner deciding which transaction to abort or not.
The only reason that we need to override this specifically for the emulator (and specifically in the Connection API), is that some frameworks that depend on JDBC (and similar drivers) require support for executing multiple transactions in parallel using one client thread.
(The `ReadWriteTransaction` class and related builder class are all package-private, and so already internal APIs)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ReadWriteTransaction.java
```diff
@@ -82,7 +83,32 @@ class ReadWriteTransaction extends AbstractMultiUseTransaction {
...
-  private static final int MAX_INTERNAL_RETRIES = 50;
+  private static final int DEFAULT_MAX_INTERNAL_RETRIES = 50;
+
+  /**
+   * A reference to the currently active transaction on the emulator that was started by the same
+   * thread. This reference is only used when running on the emulator, and enables the Connection
+   * API to manually abort the current transaction on the emulator, so other transactions can try to
+   * make progress.
+   */
+  private static final ThreadLocal<ReadWriteTransaction> CURRENT_ACTIVE_TRANSACTION =
```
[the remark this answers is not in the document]
olavloite: See also my comment above. It is really intentional that this should only be used with the emulator. I've updated the javadoc comment to better reflect that, so no one is tempted in the future to enable this for regular Spanner.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ReadWriteTransaction.java
olavloite: It would if it was null, but it is set for all cases (so not only for the emulator case). I've added a `Preconditions.checkNotNull(savepoint)` check to the constructor of `RollbackToSavepointException` to guarantee it there as well.

PR #3027, opened 2024-04-11
refactor: support multiple active transactions and traces for multipl…

arpan14:
`SessionImpl` class today has two responsibilities - a) POJO for holding session metadata and b) holding the current transaction context (transaction ID, span, etc.) . Given it has the dual responsibility, it is used both in control plane (during session creation) as well as data-plane (transaction execution). 

This approach used to work for regular sessions since regular sessions only had 1 transaction running concurrently. For multiplexed sessions, since there can be multiple transactions running per `SessionImpl` object, it can no longer hold the transaction context.

This PR does the refactoring in two steps -

1. Create a new class `SessionInstance` which is a stateless POJO representing 1 session. This does not hold any transaction context. Refactor all member variables which is specific to a session (session name, databaseID, etc.) from `SessionImpl` to `SessionInstance`.
2. For multiplexed sessions, while creating a session store the reference as `AtomicReference<SettableApiFuture<SessionInstance>>`. There is only object for this stored as a global variable. This is only replaced when a new session gets created by maintainer.
3. For multiplexed session, when a new request comes, it will first create a new span. This span is stored for all objects on the data-plane (`MultiplexedSessionFutureWrapper`, `MultiplexedSessionFuture` and `MultiplexedSession`). Finally, we invoke `markBusy` to set the span on the `SessionImpl` object. This works since we now have 1 reference of `SessionImpl` per new request 
```
SessionImpl sessionImpl =
            new SessionImpl(
                sessionClient.getSpanner(), this.multiplexedSessionSettableApiFuture.get());
```

------------------------------------------------------------------------------

Alternate Approach - An alternate approach was explored at - https://github.com/googleapis/java-spanner/pull/3016 where I was refactoring the `setSpan()` method and removing the method from `SessionTransaction` interface. This approach somewhat worked for RO transactions but it will be difficult to extend it for RW transactions in future. Also the approach involves a lot of refactoring/testing for all other classes (`AbstractReadContext`, `AsyncTransactionManagerImpl`, `TransactionRunnerImpl`) that implement `SessionTransaction` interface. So, it's best to not go by this approach to minimize the changes on the client library side.

------------------------------------------------------------------------------

Testing - Parameterized the `SpanTest` class and tested that this refactoring works for Multiplexed sessions. Also locally tested multiplexed session via integration tests, performance tests.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionClient.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionClient.java
@@ -198,7 +198,7 @@ class SessionClient implements AutoCloseable {
   }
 
   /** Create a single session. */
-  SessionImpl createSession() {
+  SessionInstance createSession() {
     // The sessionChannelCounter could overflow, but that will just flip it to Integer.MIN_VALUE,
     // which is also a valid channel hint.
     final Map<SpannerRpc.Option, ?> options;
@@ -215,8 +215,7 @@ class SessionClient implements AutoCloseable {
                   spanner.getOptions().getDatabaseRole(),
                   spanner.getOptions().getSessionLabels(),
                   options);
-      return new SessionImpl(
-          spanner, session.getName(), session.getCreateTime(), session.getMultiplexed(), options);
+      return new SessionInstance(session.getName(), options);
     } catch (RuntimeException e) {
       span.setStatus(e);
       throw e;
@@ -248,7 +247,9 @@ class SessionClient implements AutoCloseable {
                   true);
       SessionImpl sessionImpl =
           new SessionImpl(
-              spanner, session.getName(), session.getCreateTime(), session.getMultiplexed(), null);
+              spanner,
+              new SessionInstance(
+                  session.getName(), session.getCreateTime(), session.getMultiplexed(), null));
       consumer.onSessionReady(sessionImpl);
     } catch (Throwable t) {
       span.setStatus(t);
@@ -348,10 +349,11 @@ class SessionClient implements AutoCloseable {
         res.add(
             new SessionImpl(
                 spanner,
-                session.getName(),
-                session.getCreateTime(),
-                session.getMultiplexed(),
-                options));
+                new SessionInstance(
+                    session.getName(),
+                    session.getCreateTime(),
+                    session.getMultiplexed(),
+                    options)));
       }
       return res;
     } catch (RuntimeException e) {
@@ -367,6 +369,6 @@ class SessionClient implements AutoCloseable {
     synchronized (this) {
       options = optionMap(SessionOption.channelHint(sessionChannelCounter++));
     }
-    return new SessionImpl(spanner, name, options);
+    return new SessionImpl(spanner, new SessionInstance(name, options));
   }
 }
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionClient.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionClient.java
@@ -198,7 +198,7 @@ class SessionClient implements AutoCloseable {
   }
 
   /** Create a single session. */
-  SessionInstance createSession() {
+  SessionImpl createSession() {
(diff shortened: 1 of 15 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionClient.java
```diff
@@ -198,7 +198,7 @@ DatabaseId getDatabaseId() {
   }
 
   /** Create a single session. */
-  SessionImpl createSession() {
+  SessionInstance createSession() {
```
arpan14, line 201: @olavloite Presently I have not modified this for multiplexed session method. If you like the approach, I can do a bit more refactoring to ensure that the session creation flow (control plane) only uses the new `SessionInstance` class and we use `SessionImpl` on the data-plane side of things. Also I haven't modified the class docs as of now. Will do after we finalize the approach, naming conventions, etc.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpanTest.java
```diff
@@ -381,7 +426,13 @@ public void transactionRunnerWithError() {
     assertEquals(ErrorCode.INVALID_ARGUMENT, e.getErrorCode());
 
     Map<String, Boolean> spans = failOnOverkillTraceComponent.getSpans();
-    assertThat(spans.size()).isEqualTo(3);
+
+    if (useMultiplexed) {
+      assertThat(spans.size()).isEqualTo(4);
```
arpan14, line 431: I'm noticing a flaky behavior due to the new assertions for multiplexed sessions. This is because multiplexed session creation does not block on application startup. Probably we can plug some logic in existing `setWaitForMinSessions` method to also wait for multiplexed session creation? 

To start with this could be used for tests. @olavloite WDYT? This also goes back to our previous discussion that it's ok to block application startup if its an option that customers opt-into and is not switched on by default.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionInstance.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.Map;
+import javax.annotation.Nullable;
+import org.threeten.bp.Instant;
+
+/**
+ * A {@code Session} can be used to perform transactions that read and/or modify data in a Cloud
+ * Spanner database. Sessions are managed internally by the client library, and users need not be
+ * aware of the actual session management, pooling and handling.
+ */
+class SessionInstance {
```
olavloite, line 32: Maybe `SessionReference`? As in: This class is just a reference to a backend session.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1178,32 +1178,38 @@ public PooledSessionFuture get() {
...
-        // Creating a new reference where the request's span state can be stored.
-        SessionImpl sessionImpl =
-            new SessionImpl(
-                sessionClient.getSpanner(), this.multiplexedSessionSettableApiFuture.get());
-        MultiplexedSession multiplexedSession = new MultiplexedSession(sessionImpl);
-        SettableFuture<MultiplexedSession> settableFuture = SettableFuture.create();
-        settableFuture.set(multiplexedSession);
-        MultiplexedSessionFuture multiplexedSessionFuture =
-            new MultiplexedSessionFuture(settableFuture, span);
+      if (multiplexedSessionFuture != null) {
```
olavloite, line 1193: nit:
1. The field `multiplexedSessionFuture` should be `volatile`.
2. The lazy initialization should use a lock (the lock only needs to be taken during the actual initialization)

See this for a concrete example: https://www.baeldung.com/java-singleton-double-checked-locking#implementation

olavloite (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Done

PR #3030, opened 2024-04-12
chore(spanner): add multiplexed session flag in executor code

harshachinta:
This PR has the changes to accept a new flag `--multiplexed_session_operations_ratio` as input argument when running the executor artifact.
Based on the value of multiplexed_session_operations_ratio flag the client will either use regular sessions or multiplexed sessions

1.  If the value is > 0.0, then all the supported operations will use multiplexed sessions.
2. If the value is == 0.0, then all the operations will use regular sessions

**Note**: To start with, we are converting this ratio (double) into boolean property. In future we will use this ratio to perform a load splitter operation where some transactions use multiplexed sessions whereas the rest uses regular sessions.


google-cloud-spanner-executor/src/main/java/com/google/cloud/executor/spanner/CloudExecutorImpl.java
```diff
@@ -48,6 +57,17 @@ public StreamObserver<SpannerAsyncActionRequest> executeActionAsync(
       @Override
       public void onNext(SpannerAsyncActionRequest request) {
         LOGGER.log(Level.INFO, String.format("Receiving request: \n%s", request));
+
+        // Use Multiplexed sessions for all supported operations if the
+        // multiplexedSessionOperationsRatio from command line is > 0.0.
+        if (multiplexedSessionOperationsRatio > 0.0) {
+          SessionPoolOptions.Builder sessionPoolOptions =
+              SessionPoolOptions.newBuilder().setUseMultiplexed(true);
```
arpan14, line 65: Instead of `newBuilder` use `toBuilder`. That way if something is already set in the action input then it won't get over-written.

google-cloud-spanner-executor/src/main/java/com/google/cloud/executor/spanner/CloudExecutorImpl.java
```diff
@@ -34,8 +37,14 @@ public class CloudExecutorImpl extends SpannerExecutorProxyGrpc.SpannerExecutorP
   // Executors to proxy.
   private final CloudClientExecutor clientExecutor;
 
-  public CloudExecutorImpl(boolean enableGrpcFaultInjector) {
+  // Ratio of operations to use multiplexed sessions. Only applies to supported operations
```
arpan14, line 40:
```suggestion
  // Ratio of operations to use multiplexed sessions.
```

We don't need to detail out what operations are supported over here.

google-cloud-spanner-executor/src/main/java/com/google/cloud/executor/spanner/CloudExecutorImpl.java
```diff
@@ -34,8 +37,14 @@ public class CloudExecutorImpl extends SpannerExecutorProxyGrpc.SpannerExecutorP
...
+  // (currently read-only).
+  private final double multiplexedSessionOperationsRatio;
+
+  public CloudExecutorImpl(
+      boolean enableGrpcFaultInjector, double multiplexedSessionOperationsRatioOption) {
     clientExecutor = new CloudClientExecutor(enableGrpcFaultInjector);
+    multiplexedSessionOperationsRatio = multiplexedSessionOperationsRatioOption;
```
arpan14, line 47:
```suggestion
    this.multiplexedSessionOperationsRatio = multiplexedSessionOperationsRatioOption;
```

google-cloud-spanner-executor/src/main/java/com/google/cloud/executor/spanner/WorkerProxy.java
```diff
@@ -139,6 +154,11 @@ private static CommandLine buildOptions(String[] args) {
         OPTION_ENABLE_GRPC_FAULT_INJECTOR,
         false,
         "Enable grpc fault injector in cloud client executor.");
+    options.addOption(
+        null,
+        OPTION_MULTIPLEXED_SESSION_OPERATIONS_RATIO,
+        true,
+        "Ratio of operations to use multiplexed sessions. Only applies to supported operations (currently read-only).");
```
arpan14, line 161:
```suggestion
        "Ratio of operations to use multiplexed sessions.");
```

google-cloud-spanner-executor/src/main/java/com/google/cloud/executor/spanner/CloudExecutorImpl.java
harshachinta: Done

google-cloud-spanner-executor/src/main/java/com/google/cloud/executor/spanner/WorkerProxy.java
```diff
@@ -45,11 +45,14 @@ public class WorkerProxy {
...
   private static final String OPTION_USE_PLAIN_TEXT_CHANNEL = "use_plain_text_channel";
   private static final String OPTION_ENABLE_GRPC_FAULT_INJECTOR = "enable_grpc_fault_injector";
+  private static final String OPTION_MULTIPLEXED_SESSION_OPERATIONS_RATIO =
+      "multiplexed_session_operations_ratio";
 
   public static int spannerPort = 0;
   public static int proxyPort = 0;
   public static String cert = "";
   public static String serviceKeyFile = "";
+  public static double multiplexedSessionOperationsRatio = 0.0;
```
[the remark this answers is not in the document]
harshachinta: As discussed offline, we will start with boolean option. We can look at ratio option later.

harshachinta:
@gyang-google Can you help review this?

google-cloud-spanner-executor/src/main/java/com/google/cloud/executor/spanner/CloudExecutorImpl.java
```diff
@@ -48,6 +56,33 @@ public StreamObserver<SpannerAsyncActionRequest> executeActionAsync(
       @Override
       public void onNext(SpannerAsyncActionRequest request) {
         LOGGER.log(Level.INFO, String.format("Receiving request: \n%s", request));
+
+        // Use Multiplexed sessions for all supported operations if the
+        // multiplexedSessionOperationsRatio from command line is > 0.0
+        if (multiplexedSessionOperationsRatio > 0.0) {
+          SessionPoolOptions.Builder sessionPoolOptionsBuilder;
+          if (request.getAction().getSpannerOptions().hasSessionPoolOptions()) {
```
arpan14, line 64: Add a check like `request.getAction().hasSpannerOptions()` before doing `getSpannerOptions`. Else this will throw a NullPointerException.

google-cloud-spanner-executor/src/main/java/com/google/cloud/executor/spanner/CloudExecutorImpl.java
```diff
@@ -48,6 +56,33 @@ public StreamObserver<SpannerAsyncActionRequest> executeActionAsync(
...
+                    .toBuilder()
+                    .setUseMultiplexed(true);
+          } else {
+            sessionPoolOptionsBuilder = SessionPoolOptions.newBuilder().setUseMultiplexed(true);
+          }
+
+          SpannerOptions.Builder optionsBuilder =
+              request
+                  .getAction()
+                  .getSpannerOptions()
```
arpan14, line 79: I think `getSpannerOptions` will throw a `NullPointerException` since no one is passing it today. Add a handling similar to what you did for `SessionPoolOptions` above.

google-cloud-spanner-executor/src/main/java/com/google/cloud/executor/spanner/CloudExecutorImpl.java
```diff
@@ -48,6 +56,33 @@ public StreamObserver<SpannerAsyncActionRequest> executeActionAsync(
...
+                  .toBuilder()
+                  .setSessionPoolOptions(sessionPoolOptionsBuilder);
+          SpannerAction.Builder actionBuilder =
+              request.getAction().toBuilder().setSpannerOptions(optionsBuilder);
+          request = request.toBuilder().setAction(actionBuilder).build();
```
arpan14, line 84: Since we don't have unit tests, can you please link a before/after log object of this request object before and after your change? This is to ensure we are not mis-constructing the entire input object and avoid bugs like above.

google-cloud-spanner-executor/src/main/java/com/google/cloud/executor/spanner/WorkerProxy.java
```diff
@@ -95,10 +98,22 @@ public static void main(String[] args) throws Exception {
     usePlainTextChannel = commandLine.hasOption(OPTION_USE_PLAIN_TEXT_CHANNEL);
     enableGrpcFaultInjector = commandLine.hasOption(OPTION_ENABLE_GRPC_FAULT_INJECTOR);
 
+    if (commandLine.hasOption(OPTION_MULTIPLEXED_SESSION_OPERATIONS_RATIO)) {
```
arpan14, line 101: Have you made sure this works when the value is null? Do we have some manual logs to validate that this is working?

google-cloud-spanner-executor/src/main/java/com/google/cloud/executor/spanner/WorkerProxy.java
```diff
@@ -95,10 +98,22 @@ public static void main(String[] args) throws Exception {
...
+      multiplexedSessionOperationsRatio =
+          Double.parseDouble(
```
arpan14, line 103: Nit: If value is not null, we can add a validation that it is between 0 and 100. Similar validation is present for ports.

google-cloud-spanner-executor/src/main/java/com/google/cloud/executor/spanner/CloudExecutorImpl.java
harshachinta: The autogenerated methods from the proto always returns a default instance of the class and not `null`. In this case `getAction()`, `getSpannerOptions` etc will never be `null`. Hence, we are safe from getting an `NullPointerException `
Reference: https://github.com/googleapis/java-spanner/blob/6f522dd73123b3de8399902e28d755a5496829c3/proto-google-cloud-spanner-executor-v1/src/main/java/com/google/spanner/executor/v1/SpannerAsyncActionRequest.java#L112-L115

google-cloud-spanner-executor/src/main/java/com/google/cloud/executor/spanner/CloudExecutorImpl.java
harshachinta: The autogenerated methods from the proto always returns a default instance of the class and not `null`. In this case `getAction()`, `getSpannerOptions` etc will never be `null`. Hence, we are safe from not getting an `NullPointerException `
Reference: https://github.com/googleapis/java-spanner/blob/6f522dd73123b3de8399902e28d755a5496829c3/proto-google-cloud-spanner-executor-v1/src/main/java/com/google/spanner/executor/v1/SpannerAsyncActionRequest.java#L112-L115

google-cloud-spanner-executor/src/main/java/com/google/cloud/executor/spanner/CloudExecutorImpl.java
arpan14: Ok, got it.

arpan14 (approved):
LGTM, please take care of the testing before merging. And if possible, attach the test results.

google-cloud-spanner-executor/src/main/java/com/google/cloud/executor/spanner/CloudExecutorImpl.java
harshachinta: I have added a short log [here](https://paste.googleplex.com/6423074894774272).

**Request object Before:**
Screenshot 2024-04-23 at 3 19 16 PM

Request Object after adding multiplexed session
Screenshot 2024-04-23 at 3 24 38 PM

google-cloud-spanner-executor/src/main/java/com/google/cloud/executor/spanner/WorkerProxy.java
harshachinta: `OPTION_MULTIPLEXED_SESSION_OPERATIONS_RATIO` flag value will never be `null`. The default value passed from systest is `0.0`.

google-cloud-spanner-executor/src/main/java/com/google/cloud/executor/spanner/WorkerProxy.java
harshachinta: Added a validation for multiplexed session ratio flag value to be between `0.0` and `1.0`

harshachinta:
Tested the code changes by running systests locally. The runs were successful.
https://screenshot.googleplex.com/3AEDfgWqyRm6TkY

PR #3031, opened 2024-04-12
fix: the idle session remover could drop below MinSessions

olavloite:
Fixes a number of issues with the session pool maintainer:
1. The idle session remover could allow the number of sessions to drop below `MinSessions`.
2. The `keepAlive` method calculated the number of sessions to keep alive only based on how often it runs. This could cause sessions to be garbage collected by the backend if all sessions have approximately the same last-use-time, as no sessions would be kept alive for a long time. Then all of a sudden all sessions required a keep-alive, but the maintainer would only ping one session for each iteration. The new calculation checks the number of sessions that actually need to be pinged, and bases the number of sessions to ping on that.
3. The `retryOnSessionNotFound` method would close the session twice if the session had to be replaced. This would decrease the value of `numSessionsInUse` twice, causing it to show one number less than it should.


google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClientImpl.java
```diff
@@ -253,14 +253,16 @@ public long executePartitionedUpdate(final Statement stmt, final UpdateOption...
...
+    try {
+      while (true) {
+        try {
+          return callable.apply(session);
+        } catch (SessionNotFoundException e) {
+          session = pool.getPooledSessionReplacementHandler().replaceSession(e, session);
+        }
       }
+    } finally {
+      session.close();
```
olavloite, line 265: The responsibility to close the session at the end has been moved from the underlying method that is being called to this method. This prevents the `PooledSession` from being closed twice if the session was replaced. That would again cause the `numSessionsInUse` to be decreased by 1 more than it should, which could cause it to become negative.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1266,11 +1271,7 @@ public Timestamp write(Iterable<Mutation> mutations) throws SpannerException {
     @Override
     public CommitResponse writeWithOptions(
         Iterable<Mutation> mutations, TransactionOption... options) throws SpannerException {
-      try {
-        return get().writeWithOptions(mutations, options);
-      } finally {
-        close();
-      }
+      return get().writeWithOptions(mutations, options);
```
olavloite, line 1274: This method is no longer responsible for closing the session, instead that is done by the `runWithSessionRetry` method.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1922,12 +1911,15 @@ public ApiFuture<Empty> asyncClose() {
     public void close() {
       synchronized (lock) {
         numSessionsInUse--;
+        if (options.isFailIfSessionNotFound() && numSessionsInUse < 0) {
+          throw new IllegalStateException("Num sessions in use is negative");
```
olavloite, line 1915: This exception would be thrown without the changes in this pull request.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1922,12 +1911,15 @@ public ApiFuture<Empty> asyncClose() {
...
+        }
         numSessionsReleased++;
       }
       if ((lastException != null && isSessionNotFound(lastException)) || isRemovedFromPool) {
         invalidateSession(this);
       } else {
-        if (lastException != null && isDatabaseOrInstanceNotFound(lastException)) {
+        if (isDatabaseOrInstanceNotFound(lastException)) {
```
olavloite, line 1922: Removed redundant null check, the input argument to this function is null-safe.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1952,14 +1944,14 @@ public String getName() {
...
     private void keepAlive() {
-      markUsed();
       final ISpan previousSpan = delegate.getCurrentSpan();
       delegate.setCurrentSpan(tracer.getBlankSpan());
       try (ResultSet resultSet =
           delegate
               .singleUse(TimestampBound.ofMaxStaleness(60, TimeUnit.SECONDS))
               .executeQuery(Statement.newBuilder("SELECT 1").build())) {
         resultSet.next();
+        markUsed();
```
olavloite, line 1954: The session should only be marked as used if the ping actually succeeded.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2402,7 +2394,7 @@ private void removeIdleSessions(Instant currTime) {
         // below MinSessions.
         Instant minLastUseTime = currTime.minus(options.getRemoveInactiveSessionAfter());
         Iterator<PooledSession> iterator = sessions.descendingIterator();
-        while (iterator.hasNext()) {
+        while (iterator.hasNext() && allSessions.size() > options.getMinSessions()) {
```
olavloite, line 2397: Don't let the idle session remover remove so many sessions that we drop below `MinSessions`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2422,23 +2414,12 @@ private void removeIdleSessions(Instant currTime) {
...
-          return;
-        }
         // In each cycle only keep alive a subset of sessions to prevent burst of traffic.
-        numSessionsToKeepAlive =
-            (long)
-                Math.ceil(
-                    (double)
-                            ((options.getMinSessions() + options.getMaxIdleSessions())
-                                - numSessionsInUse)
-                        / numKeepAliveCycles);
```
olavloite, line 2437: This logic has been moved into a separate method

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2720,10 +2724,9 @@ enum Position {
   @VisibleForTesting Function<PooledSession, Void> longRunningSessionRemovedListener;
   @VisibleForTesting Function<SessionReference, Void> multiplexedSessionRemovedListener;
   private final CountDownLatch waitOnMinSessionsLatch;
-  private final SessionReplacementHandler pooledSessionReplacementHandler =
-      new PooledSessionReplacementHandler();
-  private static final SessionReplacementHandler multiplexedSessionReplacementHandler =
-      new MultiplexedSessionReplacementHandler();
+  private final SessionReplacementHandler<PooledSessionFuture> pooledSessionReplacementHandler;
+  private static final SessionReplacementHandler<MultiplexedSessionFuture>
+      multiplexedSessionReplacementHandler = new MultiplexedSessionReplacementHandler();
```
olavloite, line 2729: Added generic type arguments to remove some warnings

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2986,7 +2993,7 @@ private boolean isSessionNotFound(SpannerException e) {
     return e.getErrorCode() == ErrorCode.NOT_FOUND && e.getMessage().contains("Session not found");
   }
 
-  private boolean isDatabaseOrInstanceNotFound(SpannerException e) {
+  private boolean isDatabaseOrInstanceNotFound(@Nullable SpannerException e) {
```
olavloite, line 2996: `instanceof` checks are safe with nulls.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/RetryOnInvalidatedSessionTest.java
```diff
@@ -226,327 +233,327 @@ public void setUp() throws InterruptedException {
...
-      invalidateSessionPool(client, spanner.getOptions().getSessionPoolOptions().getMinSessions());
+      invalidateSessionPool();
     }
   }
 
-  private static void invalidateSessionPool(DatabaseClient client, int minSessions)
-      throws InterruptedException {
-    // Wait for all sessions to have been created, and then delete them.
-    Stopwatch watch = Stopwatch.createStarted();
-    while (((DatabaseClientImpl) client).pool.totalSessions() < minSessions) {
```
olavloite, line 237: This wait loop is no longer needed, as the session pool has a built-in option to wait for `MinSessions`.

PR #3040, opened 2024-04-17
chore: use a singleton session pool clock

olavloite:
* Changes the session pool maintainer clock to a singleton.
* Adds more tests for session pool maintenance.
* Makes sure the clock that is set in the options is used for all transactions.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
@@ -2952,6 +2952,13 @@ class SessionPool {
   }
 
   @VisibleForTesting
+  int getTotalSessionsPlusNumSessionsBeingCreated() {
+    synchronized (lock) {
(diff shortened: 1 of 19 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2349,8 +2349,13 @@ private PooledSession pollUninterruptiblyWithTimeout(
                           + acquireSessionTimeout.toMillis()
                           + "ms for acquiring session. To mitigate error SessionPoolOptions#setAcquireSessionTimeout(Duration) to set a higher timeout"
                           + " or increase the number of sessions in the session pool.");
-              waiter.setException(exception);
-              throw exception;
+              if (waiter.setException(exception)) {
+                // Only throw the exception if setting it on the waiter was successful. The
```
olavloite, line 2353: This fixes the flaky error in `SpanExceptionTest`

arpan14 (approved):

PR #3041, opened 2024-04-17
chore: fix all unit tests  with multiplexed sessions.

arpan14:
Tests have been parameterized or skipped when test runners pass the env variable `GOOGLE_CLOUD_SPANNER_ENABLE_MULTIPLEXED_SESSIONS`. We presently have created separate test runners (unit, emulator and integration tests) which run during pre-submits and test the multiplexed session code path.

**Pending Stuff**

1. `ChannelUsageTest` has been disabled (with a followup TODO) when multiplexed sessions is enabled. We are coordinating with GRPC team to see how this can be enabled. This is the only test which currently exudes a flaky behavior very frequently.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionClient.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionClient.java
@@ -263,6 +263,61 @@ class SessionClient implements AutoCloseable {
   }
 
   /**
+   * Create a multiplexed session asynchronously and returns it to the given {@link
+   * SessionConsumer}. A multiplexed session is not affiliated with any GRPC channel. The given
+   * {@link SessionConsumer} is guaranteed to eventually get exactly 1 multiplexed session unless an
+   * error occurs. In case of an error on the gRPC calls, the consumer will receive one {@link
+   * SessionConsumer#onSessionCreateFailure(Throwable, int)} calls with the error.
+   *
+   * @param consumer The {@link SessionConsumer} to use for callbacks when sessions are available.
+   */
+  void asyncCreateMultiplexedSession(SessionConsumer consumer) {
+    try {
+      executor.submit(new CreateMultiplexedSessionsRunnable(consumer));
+    } catch (Throwable t) {
+      consumer.onSessionCreateFailure(t, 1);
+    }
+  }
+
+  private final class CreateMultiplexedSessionsRunnable implements Runnable {
+    private final SessionConsumer consumer;
+
+    private CreateMultiplexedSessionsRunnable(SessionConsumer consumer) {
+      Preconditions.checkNotNull(consumer);
+      this.consumer = consumer;
+    }
+
+    @Override
+    public void run() {
+      ISpan span = spanner.getTracer().spanBuilder(SpannerImpl.CREATE_MULTIPLEXED_SESSION);
+      try (IScope s = spanner.getTracer().withSpan(span)) {
+        com.google.spanner.v1.Session session =
+            spanner
+                .getRpc()
+                .createSession(
+                    db.getName(),
+                    spanner.getOptions().getDatabaseRole(),
+                    spanner.getOptions().getSessionLabels(),
+                    null,
+                    true);
+        SessionImpl sessionImpl =
+            new SessionImpl(
+                spanner,
+                new SessionReference(
+                    session.getName(), session.getCreateTime(), session.getMultiplexed(), null));
+        span.addAnnotation(
+            String.format("Request for %d multiplexed session returned %d session", 1, 1));
+        consumer.onSessionReady(sessionImpl);
+      } catch (Throwable t) {
+        span.setStatus(t);
+        consumer.onSessionCreateFailure(t, 1);
+      } finally {
+        span.end();
+      }
+    }
+  }
+
+  /**
    * Asynchronously creates a batch of sessions and returns these to the given {@link
    * SessionConsumer}. This method may split the actual session creation over several gRPC calls in
    * order to distribute the sessions evenly over all available channels and to parallelize the
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
@@ -142,6 +142,14 @@ class SessionPool {
             ErrorCode.DEADLINE_EXCEEDED,
             "Timed out after waiting " + timeoutMillis + "ms for session pool creation");
       }
+
+      if (options.getUseMultiplexedSession()
+          && !waitOnMultiplexedSessionsLatch.await(timeoutNanos, TimeUnit.NANOSECONDS)) {
+        final long timeoutMillis = options.getWaitForMinSessions().toMillis();
+        throw SpannerExceptionFactory.newSpannerException(
+            ErrorCode.DEADLINE_EXCEEDED,
+            "Timed out after waiting " + timeoutMillis + "ms for multiplexed session creation");
+      }
     } catch (InterruptedException e) {
       throw SpannerExceptionFactory.propagateInterrupt(e);
     }
@@ -1159,10 +1167,86 @@ class SessionPool {
   }
 
   /** Wrapper class for the {@link SessionFuture} implementations. */
-  interface SessionFutureWrapper<T extends SessionFuture> {
+  interface SessionFutureWrapper<T extends SessionFuture> extends DatabaseClient {
 
     /** Method to resolve {@link SessionFuture} implementation for different use-cases. */
     T get();
+
+    default Dialect getDialect() {
+      return get().getDialect();
+    }
+
+    default String getDatabaseRole() {
+      return get().getDatabaseRole();
+    }
+
+    default Timestamp write(Iterable<Mutation> mutations) throws SpannerException {
+      return get().write(mutations);
+    }
+
+    default CommitResponse writeWithOptions(
+        Iterable<Mutation> mutations, TransactionOption... options) throws SpannerException {
+      return get().writeWithOptions(mutations, options);
+    }
+
+    default Timestamp writeAtLeastOnce(Iterable<Mutation> mutations) throws SpannerException {
+      return get().writeAtLeastOnce(mutations);
+    }
+
+    default CommitResponse writeAtLeastOnceWithOptions(
+        Iterable<Mutation> mutations, TransactionOption... options) throws SpannerException {
+      return get().writeAtLeastOnceWithOptions(mutations, options);
+    }
+
+    default ServerStream<BatchWriteResponse> batchWriteAtLeastOnce(
+        Iterable<MutationGroup> mutationGroups, TransactionOption... options)
+        throws SpannerException {
+      return get().batchWriteAtLeastOnce(mutationGroups, options);
+    }
+
+    default ReadContext singleUse() {
+      return get().singleUse();
+    }
+
+    default ReadContext singleUse(TimestampBound bound) {
+      return get().singleUse(bound);
+    }
+
+    default ReadOnlyTransaction singleUseReadOnlyTransaction() {
+      return get().singleUseReadOnlyTransaction();
+    }
+
+    default ReadOnlyTransaction singleUseReadOnlyTransaction(TimestampBound bound) {
+      return get().singleUseReadOnlyTransaction(bound);
+    }
+
+    default ReadOnlyTransaction readOnlyTransaction() {
+      return get().readOnlyTransaction();
+    }
+
+    default ReadOnlyTransaction readOnlyTransaction(TimestampBound bound) {
+      return get().readOnlyTransaction(bound);
+    }
+
+    default TransactionRunner readWriteTransaction(TransactionOption... options) {
+      return get().readWriteTransaction(options);
+    }
+
+    default TransactionManager transactionManager(TransactionOption... options) {
+      return get().transactionManager(options);
+    }
+
+    default AsyncRunner runAsync(TransactionOption... options) {
+      return get().runAsync(options);
+    }
+
+    default AsyncTransactionManager transactionManagerAsync(TransactionOption... options) {
+      return get().transactionManagerAsync(options);
+    }
+
+    default long executePartitionedUpdate(Statement stmt, UpdateOption... options) {
+      return get().executePartitionedUpdate(stmt, options);
+    }
   }
 
   class PooledSessionFutureWrapper implements SessionFutureWrapper<PooledSessionFuture> {
@@ -1179,38 +1263,31 @@ class SessionPool {
   }
 
   class MultiplexedSessionFutureWrapper implements SessionFutureWrapper<MultiplexedSessionFuture> {
-    private SettableApiFuture<SessionReference> multiplexedSessionSettableApiFuture;
     private ISpan span;
     private volatile MultiplexedSessionFuture multiplexedSessionFuture;
 
-    public MultiplexedSessionFutureWrapper(
-        SettableApiFuture<SessionReference> multiplexedSessionSettableApiFuture, ISpan span) {
-      this.multiplexedSessionSettableApiFuture = multiplexedSessionSettableApiFuture;
+    public MultiplexedSessionFutureWrapper(ISpan span) {
       this.span = span;
     }
 
     @Override
     public MultiplexedSessionFuture get() {
+      if (resourceNotFoundException != null) {
+        span.addAnnotation("Database has been deleted");
+        throw SpannerExceptionFactory.newSpannerException(
+            ErrorCode.NOT_FOUND,
+            String.format(
+                "The session pool has been invalidated because a previous RPC returned 'Database not found': %s",
+                resourceNotFoundException.getMessage()),
+            resourceNotFoundException);
+      }
       if (multiplexedSessionFuture == null) {
         synchronized (lock) {
           if (multiplexedSessionFuture == null) {
-            try {
-              // Creating a new reference where the request's span state can be stored.
-              SessionImpl sessionImpl =
-                  new SessionImpl(
-                      sessionClient.getSpanner(), this.multiplexedSessionSettableApiFuture.get());
-              MultiplexedSession multiplexedSession = new MultiplexedSession(sessionImpl);
-              SettableFuture<MultiplexedSession> settableFuture = SettableFuture.create();
-              settableFuture.set(multiplexedSession);
-              MultiplexedSessionFuture multiplexedSessionFuture =
-                  new MultiplexedSessionFuture(settableFuture, span);
-              this.multiplexedSessionFuture = multiplexedSessionFuture;
-              return multiplexedSessionFuture;
-            } catch (InterruptedException interruptedException) {
-              throw SpannerExceptionFactory.propagateInterrupt(interruptedException);
-            } catch (ExecutionException executionException) {
-              throw SpannerExceptionFactory.asSpannerException(executionException.getCause());
-            }
+            // Creating a new reference where the request's span state can be stored.
+            MultiplexedSessionFuture multiplexedSessionFuture = new MultiplexedSessionFuture(span);
+            this.multiplexedSessionFuture = multiplexedSessionFuture;
+            return multiplexedSessionFuture;
           }
         }
       }
@@ -1496,14 +1573,12 @@ class SessionPool {
     }
   }
 
-  class MultiplexedSessionFuture extends SimpleForwardingListenableFuture<MultiplexedSession>
-      implements SessionFuture {
+  class MultiplexedSessionFuture implements SessionFuture {
 
     private ISpan span;
+    private volatile MultiplexedSession multiplexedSession;
 
-    @VisibleForTesting
-    MultiplexedSessionFuture(ListenableFuture<MultiplexedSession> delegate, ISpan span) {
-      super(delegate);
+    MultiplexedSessionFuture(ISpan span) {
       this.span = span;
     }
 
@@ -1710,11 +1785,20 @@ class SessionPool {
     @Override
     public MultiplexedSession get() {
       try {
-        MultiplexedSession multiplexedSession = super.get();
-        if (multiplexedSession != null) {
-          multiplexedSession.markBusy(span);
-          span.addAnnotation("Using Session", "sessionId", multiplexedSession.getName());
-          incrementNumSessionsInUse(true);
+        if (multiplexedSession == null) {
+          // access the settableApiFuture object before acquiring lock to avoid deadlock
+          SessionImpl sessionImpl =
+              new SessionImpl(
+                  sessionClient.getSpanner(), currentMultiplexedSessionReference.get().get());
+          synchronized (lock) {
+            if (multiplexedSession == null) {
+              MultiplexedSession multiplexedSession = new MultiplexedSession(sessionImpl);
+              multiplexedSession.markBusy(span);
+              span.addAnnotation("Using Session", "sessionId", multiplexedSession.getName());
+              incrementNumSessionsInUse(true);
+              this.multiplexedSession = multiplexedSession;
+            }
+          }
         }
         return multiplexedSession;
       } catch (ExecutionException e) {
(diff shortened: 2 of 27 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1710,11 +1792,19 @@ private MultiplexedSession getOrNull() {
     @Override
     public MultiplexedSession get() {
       try {
-        MultiplexedSession multiplexedSession = super.get();
-        if (multiplexedSession != null) {
-          multiplexedSession.markBusy(span);
-          span.addAnnotation("Using Session", "sessionId", multiplexedSession.getName());
-          incrementNumSessionsInUse(true);
+        if (multiplexedSession == null) {
+          // access the settableApiFuture object before acquiring lock to avoid deadlock
```
arpan14, line 1796: @olavloite this is something that I have added for the time being. But I am still debugging some more race conditions which is causing flaky failures in `SpannerGaxRetryTest`. I think there is still a scope to improve synchronization of the `SettableApiFuture` without causing deadlocks.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -284,7 +284,7 @@ long getRandomizePositionQPSThreshold() {
     return randomizePositionQPSThreshold;
   }
 
-  boolean getUseMultiplexedSession() {
+  public boolean getUseMultiplexedSession() {
```
arpan14, line 287: This is made public since its needed for connection API tests.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1179,38 +1263,31 @@ public PooledSessionFuture get() {
   }
 
   class MultiplexedSessionFutureWrapper implements SessionFutureWrapper<MultiplexedSessionFuture> {
-    private SettableApiFuture<SessionReference> multiplexedSessionSettableApiFuture;
```
arpan14, line 1182: We don't really need to save `SettableApiFuture<SessionReference>` per `MultiplexedSessionFutureWrapper`. This can directly be obtained from the global reference.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -3587,6 +3678,10 @@ public void onSessionReady(SessionImpl sessionImpl) {
     public void onSessionCreateFailure(Throwable t, int createFailureForSessionCount) {
       synchronized (lock) {
         multiplexedSessionBeingCreated = false;
+        if (isDatabaseOrInstanceNotFound(newSpannerException(t))) {
+          setResourceNotFoundException((ResourceNotFoundException) t);
+          poolMaintainer.close();
+        }
         currentMultiplexedSessionReference.get().setException(newSpannerException(t));
```
arpan14, line 3685: @olavloite I think the order here does not matter. i.e it is safe to first do `poolMaintainer.close();` and then propagate the exceptions to all the waiting threads.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/ChannelUsageTest.java
```diff
@@ -220,6 +221,9 @@ private SpannerOptions createSpannerOptions() {
   @Test
   public void testCreatesNumChannels() {
     try (Spanner spanner = createSpannerOptions().getService()) {
+      assumeFalse(
+          "GRPC-GCP is currently not supported with multiplexed sessions",
+          isMultiplexedSessionsEnabled(spanner));
```
olavloite, line 226: This test works for the normal channel pool, it is only with Gcp pool that it fails.

```suggestion
          enableGcpPool && isMultiplexedSessionsEnabled(spanner));
```

google-cloud-spanner/src/test/java/com/google/cloud/spanner/ChannelUsageTest.java
```diff
@@ -231,6 +235,9 @@ public void testCreatesNumChannels() {
   @Test
   public void testUsesAllChannels() throws InterruptedException, ExecutionException {
     try (Spanner spanner = createSpannerOptions().getService()) {
+      assumeFalse(
+          "GRPC-GCP is currently not supported with multiplexed sessions",
+          isMultiplexedSessionsEnabled(spanner));
```
olavloite, line 240:
```suggestion
          enableGcpPool && isMultiplexedSessionsEnabled(spanner));
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionClient.java
```diff
@@ -262,6 +262,61 @@ void createMultiplexedSession(SessionConsumer consumer) {
     }
   }
 
+  /**
+   * Create a multiplexed session asynchronously and returns it to the given {@link
+   * SessionConsumer}. A multiplexed session is not affiliated with any GRPC channel. The given
+   * {@link SessionConsumer} is guaranteed to eventually get exactly 1 multiplexed session unless an
+   * error occurs. In case of an error on the gRPC calls, the consumer will receive one {@link
+   * SessionConsumer#onSessionCreateFailure(Throwable, int)} calls with the error.
```
olavloite, line 270: nit:
```suggestion
   * SessionConsumer#onSessionCreateFailure(Throwable, int)} call with the error.
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
olavloite: Is it possible that multiple threads will call `MultiplexedSessionFuture#get()` (so this method) before it has been initialized? If so, then this would need its own lock.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1710,11 +1785,20 @@ private MultiplexedSession getOrNull() {
...
-        if (multiplexedSession != null) {
-          multiplexedSession.markBusy(span);
-          span.addAnnotation("Using Session", "sessionId", multiplexedSession.getName());
-          incrementNumSessionsInUse(true);
+        if (multiplexedSession == null) {
+          // access the settableApiFuture object before acquiring lock to avoid deadlock
+          SessionImpl sessionImpl =
+              new SessionImpl(
+                  sessionClient.getSpanner(), currentMultiplexedSessionReference.get().get());
+          synchronized (lock) {
```
olavloite, line 1793: nit: I think that you only need to take the lock for the `incrementNumSessionsInuse(true);` call. The rest does not need the session pool lock.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -3587,6 +3678,10 @@ public void onSessionReady(SessionImpl sessionImpl) {
     public void onSessionCreateFailure(Throwable t, int createFailureForSessionCount) {
       synchronized (lock) {
         multiplexedSessionBeingCreated = false;
+        if (isDatabaseOrInstanceNotFound(newSpannerException(t))) {
```
olavloite, line 3681: nit: use `asSpannerException(t)` instead of `newSpannerException(t)`. That will keep the original exception if it was already a `SpannerException`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
olavloite: Yes, I think so too.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
olavloite: nit: can we mark it with `@InternalApi` until the setter has also been made public? Or maybe `@VisibleForTesting`?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
```diff
@@ -3009,6 +3013,10 @@ public void testDatabaseOrInstanceDoesNotExistOnCreate() {
               "Instance", SpannerExceptionFactory.INSTANCE_RESOURCE_TYPE, INSTANCE_NAME)
         };
     for (StatusRuntimeException exception : exceptions) {
+      if (isMultiplexedSessionsEnabled()) {
```
olavloite, line 3016: I don't think you need the `if` statement. `createSessions` is never used by the regular session pool, so you can just always add the exception for that as well.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
```diff
@@ -3662,6 +3677,34 @@ public void testBatchCreateSessionsPermissionDenied() {
       // Actually trying to get any results will cause an exception.
       SpannerException e = assertThrows(SpannerException.class, rs::next);
       assertEquals(ErrorCode.PERMISSION_DENIED, e.getErrorCode());
+    } finally {
+      mockSpanner.setBatchCreateSessionsExecutionTime(SimulatedExecutionTime.none());
+    }
+  }
+
+  @Test
+  public void testCreateSessionsPermissionDenied() {
```
olavloite, line 3686: I don't think you need separate tests for these conditions. `BatchCreateSessions` and `CreateSession` are both exclusively used by respectively non-multiplexed-sessions and multiplexed-sessions. That means that you can just register the same error for both, and depending on whether multiplexed sessions are enabled or not, one of the RPCs will be called, and the behavior should be the same in both cases.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/IntegrationTestWithClosedSessionsEnv.java
```diff
@@ -86,6 +87,20 @@ PooledSessionFuture getSession() {
       return session;
     }
 
+    @Override
+    SessionFutureWrapper getMultiplexedSession() {
```
olavloite, line 91: I don't think this is ever used, as the integration test that uses this test env is (correctly) skipped for multiplexed sessions. And if it is used, what does this then mean?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/MockSpannerServiceImpl.java
```diff
@@ -585,6 +585,8 @@ private static void checkStreamException(
   private ConcurrentMap<String, StatementResult> partialStatementResults =
       new ConcurrentHashMap<>();
   private ConcurrentMap<String, Session> sessions = new ConcurrentHashMap<>();
+  private ConcurrentMap<String, Session> multiplexedSessions = new ConcurrentHashMap<>();
```
olavloite, line 588: This does not appear to be used. The only place where we need to check whether a session is multiplexed, is when a transaction is being executed, and that point the code just checks whether the `isMultiplexed` flag is set on that session. So I think we can safely store the multiplexed sessions in the same map as the normal sessions. That also makes methods like `listSessions` and `deleteSession` work (I think they will exclude any multiplexed session in this implementation).

google-cloud-spanner/src/test/java/com/google/cloud/spanner/ReadAsyncTest.java
```diff
@@ -185,6 +185,10 @@ public void pointReadNotFound() throws Exception {
 
   @Test
   public void invalidDatabase() throws Exception {
+    if (isMultiplexedSessionsEnabled()) {
```
olavloite, line 188: Same here: I don't think you need the if statement.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/RetryOnInvalidatedSessionTest.java
```diff
@@ -217,17 +218,21 @@ public void setUp() throws InterruptedException {
...
               .setChannelProvider(channelProvider)
-              .setSessionPoolOption(builder.build())
+              .setSessionPoolOption(sessionPoolOptions)
               .setCredentials(NoCredentials.getInstance())
               .build()
               .getService();
       client = spanner.getDatabaseClient(DatabaseId.of("[PROJECT]", "[INSTANCE]", "[DATABASE]"));
       invalidateSessionPool(client, spanner.getOptions().getSessionPoolOptions().getMinSessions());
     }
+    assumeFalse(
```
olavloite, line 233: Note: I also have a large update (cleanup) of this test pending in https://github.com/googleapis/java-spanner/pull/3040.

Once this PR has been merged, I'll rebase my PR and update it so it executes this test for the cases that still use regular sessions, even when multiplexed sessions are enabled (i.e. for read/write transactions). That will ensure that the session retry continues to work for those cases as well after enabling multiplexed sessions.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolTest.java
```diff
@@ -272,6 +264,17 @@ private void setupMockSessionCreation() {
         .when(sessionClient)
         .asyncBatchCreateSessions(
             Mockito.anyInt(), Mockito.anyBoolean(), any(SessionConsumer.class));
+    if (options.getUseMultiplexedSession()) {
```
olavloite, line 267: here also: I think that you can remove the if statement, as it does not interfere with the non-multiplexed case

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolTest.java
```diff
@@ -2177,13 +2186,20 @@ public void testWaitOnMinSessionsWhenSessionsAreCreatedBeforeTimeout() {
                     }))
         .when(sessionClient)
         .asyncBatchCreateSessions(Mockito.eq(1), Mockito.anyBoolean(), any(SessionConsumer.class));
+    if (options.getUseMultiplexedSession()) {
```
olavloite, line 2189: remove if?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpanTest.java
```diff
@@ -201,7 +201,7 @@ public void setUp() throws Exception {
             .setSessionPoolOption(
                 SessionPoolOptions.newBuilder()
                     .setMinSessions(2)
-                    .setWaitForMinSessions(Duration.ofSeconds(5))
+                    .setWaitForMinSessions(Duration.ofSeconds(10))
```
olavloite, line 204: Is this needed in order to prevent it from being flaky? (It probably doesn't hurt, but it seems like a surprising change)

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerGaxRetryTest.java
```diff
@@ -206,6 +212,11 @@ private void warmUpSessionPool(DatabaseClient client) {
 
   @Test
   public void singleUseTimeout() {
+    if (isMultiplexedSessionsEnabled()) {
+      // for multiplexed sessions CreateSessions RPC is already completed during start-up
+      // hence, we are setting a strict delay with the next RPC
+      mockSpanner.setExecuteStreamingSqlExecutionTime(ONE_SECOND);
```
olavloite, line 218: In order to replicate the original behavior of this test, you should set `mockSpanner.setCreateSessionExecutionTime(ONE_SECOND)`.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerGaxRetryTest.java
```diff
@@ -130,7 +130,13 @@ public void setUp() throws Exception {
             .setChannelProvider(channelProvider)
             .setCredentials(NoCredentials.getInstance());
     // Make sure the session pool is empty by default.
-    builder.setSessionPoolOption(SessionPoolOptions.newBuilder().setMinSessions(0).build());
+    // Add a wait time for sessions to be initialized. In this case, since minSessions = 0, the
+    // wait time is for multiplexed sessions
+    builder.setSessionPoolOption(
+        SessionPoolOptions.newBuilder()
+            .setMinSessions(0)
+            .setWaitForMinSessions(Duration.ofSeconds(5))
```
olavloite, line 138: This test verifies the behavior of the client when a query is blocked while waiting for the session pool to have a session available. This changes that. Is there something in this test that does not work with multiplexed sessions if this wait time is not specified?

olavloite (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
arpan14: Done

google-cloud-spanner/src/test/java/com/google/cloud/spanner/ChannelUsageTest.java
arpan14: I had originally done this. But it still introduced a bit of flaky behavior. I will be following up to re-enabled this test pending some investigation.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/ChannelUsageTest.java
arpan14: I had originally done this. But it still introduced a bit of flaky behavior even without gcp-pool. I will be following up to re-enable this test pending some investigation.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Yeah, I removed the lock from here since we are acquiring the lock again within the `incrementNumSessionsInUse` method.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Done

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
arpan14: Right, modified to re-use the same test.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/IntegrationTestWithClosedSessionsEnv.java
arpan14: No, this is still used in the regular session call path. We have replaced `getSession()` method with `getMultiplexedSession ` in `DatabaseClientImpl` class for all RO transactions in this PR. So if we don't add this override behaviour, those integration tests will fail.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/MockSpannerServiceImpl.java
arpan14: Actually a couple of tests assert the size of `sessions` list. So if we add the multiplexed session in the same list, it will increase the size by 1 and then we would have to modify all those tests to work with a different list size (if-else) logic. I felt changing it here is a cleaner solution than modifying different tests.

* google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ConnectionTest.java
* google-cloud-spanner/src/test/java/com/google/cloud/spanner/BatchCreateSessionsTest.java (`testSpannerReturnsAllAvailableSessionsAndThenNoSessions`)

google-cloud-spanner/src/test/java/com/google/cloud/spanner/ReadAsyncTest.java
arpan14: Removed.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/RetryOnInvalidatedSessionTest.java
arpan14: Sure, sounds good.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolTest.java
arpan14: Done

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpanTest.java
arpan14: I think both `SpanTest` and `OpenTelemetrySpanTest` exude a flaky behavior today irrespective of multiplexed sessions. What I saw is increasing this time reduces the flakiness.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerGaxRetryTest.java
arpan14: The regular sessions path uses `setMinSessions=0` as a workaround to prevent `BatchCreateSessions` RPC during `@Setup`. This ensures that during the test, the `BatchCreateSessions` RPC are invoked and they timeout.

For multiplexed session, the creation only happens during app initialization. Hence during `@Setup` itself the multiplexed session creation call will go. So there is no point of setting `setCreateSessionExecutionTime` within the test as by then mock spanner already would have received the call to create session and returned a response.

So given multiplexed session creation needs to be guaranteed within these tests, we would have to set the delay on whatever the next RPC is and then verify the GAX retry on the same.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerGaxRetryTest.java
arpan14: Makes sense. I have now modified the code to 
```
if (sessionPoolOptions.getUseMultiplexedSession()) {
      sessionPoolOptions.toBuilder().setWaitForMinSessions(Duration.ofSeconds(5));
    }
```

The previous code actually had no impact on regular sessions test because minSessions = 0. And hence the blocking wait time did not actually block for the regular sessions.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
arpan14: Removed.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
arpan14: Yes, `getDatabaseClient(...).singleUse().read()` is completely non-blocking. So post app initialization multiple threads can get this object and issue a `next()` on the result set. When they do this there can be a race condition. Atleast some threads will get a `SettableApiFuture` which is not yet finished. And all of them will get blocked. 

In parallel another thread will issue the `CreateSession` call and will take a lock on `currentMultiplexedSessionReference` object in the `onSessesionReady()` callback function to initialize the session. This will unblock all the waiting threads. But I see the issue that in case the above thread are blocked on the uninitialized `SettableApiFuture` then that is already stale and these threads will eventually timeout.

However introducing a lock over here can introduce a deadlock? because some threads will acquire the lock on `currentMultiplexedSessionReference` and then `onSessesionReady` method will wait for this lock to be released. 

Do we need an additional waiter queue to mitigate this edge case?

olavloite: I modified the implementation a bit to use a double-checked locking mechanism to ensure that only one thread will initialize the object. That way, we effectively create a 'waiter' queue for the multiplexed session instance.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/ChannelUsageTest.java
olavloite: Are we using a random channel hint? If so, then I understand why it is flaky. The test iterates 2*numChannels and then checks that all channels have been used. That will often, but not always, be the case if we use random channel affiliation. The test will probably become acceptably non-flaky if we increase the number of iterations to something like 50*numChannels.

PR #3055, opened 2024-04-20
fix: ignored argument to String.format call

### LINKED ISSUE #3054: Extra format argument to String.format call in SessionPool
This code is passing an argument to `String.format` that is ignored, because the string doesn't have any format specifiers:

https://github.com/googleapis/java-spanner/blob/6f522dd73123b3de8399902e28d755a5496829c3/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java#L3664-L3669

The intent may have been:

```diff
 logger.log( 
     Level.WARNING, 
     "Failed to create multiplexed session. " 
         + "Pending replacing stale multiplexed session", 
     t); 
```

This bug was caught by https://errorprone.info/bugpattern/FormatString

cushon:
This change fixes a log statement where a `Throwable` is being passed to a `String.format` call without any format specifiers.

Fixes #3054 ☕️

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
@@ -3663,10 +3663,9 @@ class SessionPool {
       }
       logger.log(
           Level.WARNING,
-          String.format(
-              "Failed to create multiplexed session. "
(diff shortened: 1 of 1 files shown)


olavloite:
Closing, as this change is in a part of the code that has been removed entirely from the library by now.

PR #3058, opened 2024-04-22
chore: add a random hint for multi-use transactions when they are use…

arpan14:
Regular sessions maintain a channel hint per session at time of regular session creation. For each multi-use transaction, this hint is fetched from the session object and passed in the RPC. This hint helps to ensure, all RPCs within a multi-use transaction go to the same channel (and hence same SpanFE).

With multiplexed sessions, there is no channel hint being maintained with the session object. Hence, we need to store the hint with the transaction object and then pass it in the RPC methods.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
@@ -232,11 +232,8 @@ abstract class AbstractReadContext
     @GuardedBy("lock")
     private Timestamp timestamp;
 
-    private Long channelHint;
-
     private SingleUseReadOnlyTransaction(SingleReadContext.Builder builder) {
       super(builder);
-      this.channelHint = ThreadLocalRandom.current().nextLong(Long.MAX_VALUE);
     }
 
     @Override
@@ -269,11 +266,6 @@ abstract class AbstractReadContext
               ErrorCode.INTERNAL, "Bad value in transaction.read_timestamp metadata field", e);
         }
       }
-    }
-
-    @Override
-    public Long getTransactionChannelHint() {
-      return channelHint;
     }
   }
 
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
@@ -214,7 +214,7 @@ abstract class AbstractReadContext
     }
 
     @Override
-    Long getTransactionChannelHint() {
+    Map<SpannerRpc.Option, ?> getTransactionChannelHint() {
       // single use transaction have a single RPC and hence there is no need
       // of a channel hint. GAX will automatically choose a hint when used
       // with a multiplexed session.
@@ -312,7 +312,7 @@ abstract class AbstractReadContext
     @GuardedBy("txnLock")
     private ByteString transactionId;
 
-    private Long channelHint;
+    private final Map<SpannerRpc.Option, ?> channelHint;
 
     MultiUseReadOnlyTransaction(Builder builder) {
       super(builder);
@@ -332,11 +332,13 @@ abstract class AbstractReadContext
         this.timestamp = builder.timestamp;
         this.transactionId = builder.transactionId;
       }
-      this.channelHint = ThreadLocalRandom.current().nextLong(Long.MAX_VALUE);
+      this.channelHint =
+          getChannelHintOptions(
+              session.getOptions(), ThreadLocalRandom.current().nextLong(Long.MAX_VALUE));
     }
 
     @Override
-    public Long getTransactionChannelHint() {
+    public Map<SpannerRpc.Option, ?> getTransactionChannelHint() {
       return channelHint;
     }
 
@@ -400,10 +402,7 @@ abstract class AbstractReadContext
                   .setOptions(options)
                   .build();
           Transaction transaction =
-              rpc.beginTransaction(
-                  request,
-                  getChannelHintOptions(session.getOptions(), getTransactionChannelHint()),
-                  isRouteToLeader());
+              rpc.beginTransaction(request, getTransactionChannelHint(), isRouteToLeader());
           if (!transaction.hasReadTimestamp()) {
             throw SpannerExceptionFactory.newSpannerException(
                 ErrorCode.INTERNAL, "Missing expected transaction.read_timestamp metadata field");
@@ -752,7 +751,7 @@ abstract class AbstractReadContext
                 rpc.executeQuery(
                     request.build(),
                     stream.consumer(),
-                    getChannelHintOptions(session.getOptions(), getTransactionChannelHint()),
+                    getTransactionChannelHint(),
                     isRouteToLeader());
             session.markUsed(clock.instant());
             call.request(prefetchChunks);
@@ -769,11 +768,7 @@ abstract class AbstractReadContext
     if (channelHintForSession != null) {
       return channelHintForSession;
     } else if (channelHintForTransaction != null) {
-      final Map<SpannerRpc.Option, ?> options;
-      synchronized (this) {
-        options = optionMap(SessionOption.channelHint(channelHintForTransaction));
-      }
-      return options;
+      return optionMap(SessionOption.channelHint(channelHintForTransaction));
     }
     return null;
   }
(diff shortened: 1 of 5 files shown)


google-cloud-spanner/src/test/java/com/google/cloud/spanner/TransactionChannelHintTest.java
```diff
@@ -0,0 +1,335 @@
...
+
+/**
+ * Transaction always utilize a channel hint to ensure multiple RPCs that are part of the same
+ * transaction, they go via same channel. For regular session, the hint is stored per session. For
+ * multiplexed sessions this hint is stored per transaction.
+ *
+ * <p>The below tests assert this behavior for both kinds of sessions.
+ */
+@RunWith(JUnit4.class)
+public class TransactionChannelHintTest {
```
olavloite, line 69: Can we combine this test with `ChannelUsageTest`? It seems that much of this test is a copy-paste from that test.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/TransactionChannelHintTest.java
```diff
@@ -0,0 +1,335 @@
...
+                  public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(
+                      ServerCall<ReqT, RespT> call,
+                      Metadata headers,
+                      ServerCallHandler<ReqT, RespT> next) {
+                    // Verify that the compressor name header is set.
+                    assertEquals(
+                        "gzip",
+                        headers.get(
+                            Metadata.Key.of(
+                                "x-response-encoding", Metadata.ASCII_STRING_MARSHALLER)));
```
olavloite, line 128: If we decide to keep this test separate from `ChannelUsageTest`; can we remove the non-relevant parts for channel hints, such as this check?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/TransactionChannelHintTest.java
```diff
@@ -0,0 +1,335 @@
...
+
+  @Test
+  public void testSingleUse_usesSingleChannel() {
+    try (Spanner spanner = createSpannerOptions().getService()) {
+      DatabaseClient client = spanner.getDatabaseClient(DatabaseId.of("p", "i", "d"));
+      try (ResultSet resultSet = client.singleUse().executeQuery(SELECT1)) {
+        while (resultSet.next()) {}
+      }
+    }
+    assertEquals(1, executeSqlLocalIps.size());
```
olavloite, line 202: A single-use read-only transaction only executes 1 RPC, so it would in any case just use 1 channel. Could we instead extend this test to execute more than 1 single-use read-only transaction and verify that each new single-use read-only transaction uses a new channel? (Assuming that the latter is what we expect, but I think it is, as we want round-robin channel pooling in this case)

olavloite: The same also applies to the other single-use tests in this file.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/TransactionChannelHintTest.java
```diff
@@ -0,0 +1,335 @@
...
+  }
+
+  @Test
+  public void testReadOnlyTransaction_usesSingleChannel() {
+    try (Spanner spanner = createSpannerOptions().getService()) {
+      DatabaseClient client = spanner.getDatabaseClient(DatabaseId.of("p", "i", "d"));
+      try (ReadOnlyTransaction transaction = client.readOnlyTransaction()) {
```
olavloite, line 248: Read-only transactions (currently) use an explicit `BeginTransaction` RPC. Can we verify that not only the `ExecuteStreamingSql` calls all use the same channel, but that also the `BeginTransaction` RPC uses the same channel?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
```diff
@@ -198,13 +199,16 @@ public void removeListener(Runnable listener) {
     private CommitResponse commitResponse;
     private final Clock clock;
 
+    private Long channelHint;
```
olavloite, line 202: nit: This should be `final`

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
```diff
@@ -300,6 +312,8 @@ static Builder newBuilder() {
     @GuardedBy("txnLock")
     private ByteString transactionId;
 
+    private Long channelHint;
```
olavloite, line 315: nit: this should be final

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
```diff
@@ -738,6 +764,20 @@ CloseableIterator<PartialResultSet> startStream(@Nullable ByteString resumeToken
         stream, this, options.hasDecodeMode() ? options.decodeMode() : defaultDecodeMode);
   }
 
+  Map<SpannerRpc.Option, ?> getChannelHintOptions(
+      Map<SpannerRpc.Option, ?> channelHintForSession, Long channelHintForTransaction) {
+    if (channelHintForSession != null) {
+      return channelHintForSession;
+    } else if (channelHintForTransaction != null) {
+      final Map<SpannerRpc.Option, ?> options;
+      synchronized (this) {
```
olavloite, line 773: You don't need to take a lock here. `channelHintForTransaction` is read-only (and should therefore be declared final), and all the other variables involved are local variables.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
```diff
@@ -380,7 +400,10 @@ void initTransaction() {
                   .setOptions(options)
                   .build();
           Transaction transaction =
-              rpc.beginTransaction(request, session.getOptions(), isRouteToLeader());
+              rpc.beginTransaction(
+                  request,
+                  getChannelHintOptions(session.getOptions(), getTransactionChannelHint()),
```
olavloite, line 405: We should not call `getChannelHintOptions(...)` for each request that is executed. Instead, this call should be moved to the constructor and called only once there. All of the fields involved are final, and the outcome of this calculation can also be final, so we should only do that once.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/TransactionChannelHintTest.java
arpan14: I thought it will be better to keep it separate. This class is to test whether a hint was passed or not. `ChannelUsageTest` focusses on channel creation/utilization.

In this class, we can expand it to all kinds of methods. Wheres the creation/utilization is sufficient to be tested for any 1 method (today - singleUse())

google-cloud-spanner/src/test/java/com/google/cloud/spanner/TransactionChannelHintTest.java
arpan14: Sure, will do that

google-cloud-spanner/src/test/java/com/google/cloud/spanner/TransactionChannelHintTest.java
arpan14: Yeah, I think I included them as an afterthought. Can't recall why. Have removed the same.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/TransactionChannelHintTest.java
arpan14: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
arpan14: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
arpan14: Done, moved to constructor.

olavloite (approved):

PR #3063, opened 2024-04-25
refactor: move multiplexed session handling to separate class

olavloite:
Refactors the multiplexed session implementation to use its own class instead of being combined with the existing session pool. This simplifies the code path for multiplexed sessions, and makes it possible to entirely remove the `SessionPool` class when all operations can be executed on multiplexed sessions.


rahul2393 (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClient.java
```diff
@@ -0,0 +1,428 @@
...
+   * server, then we set this field to mark the client as invalid.
+   */
+  private final AtomicReference<ResourceNotFoundException> resourceNotFoundException =
+      new AtomicReference<>();
+
+  /**
+   * This flag is set to true if the server return UNIMPLEMENTED when we try to create a multiplexed
+   * session.
+   */
+  private final AtomicBoolean unimplemented = new AtomicBoolean(false);
```
arpan14, line 159: Nit: The formatting looks off. Also, should we add a TODO that this is a temporary measure and we can remove it in future when this has a bit of increased usage?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClient.java
```diff
@@ -0,0 +1,428 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.spanner;
+
+import static com.google.cloud.spanner.SessionImpl.NO_CHANNEL_HINT;
```
arpan14, line 19: Given we are initializing this to -1, can there be an edge case where the value gets decremented to -1? Would using a boxed integer and setting `null` be a better default than -1?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClient.java
```diff
@@ -0,0 +1,428 @@
...
+   * 'checked out session' of a pool, except as multiplexed sessions support multiple parallel
+   * transactions, we do not need to actually check out and exclusively reserve a single session for
+   * a transaction. This class therefore only contains context information about the current
+   * transaction, such as the current span, and a reference to the multiplexed session that is used
+   * for the transaction.
+   */
+  static class MultiplexedSessionTransaction extends SessionImpl {
+    private final MultiplexedSessionDatabaseClient client;
+
+    private final boolean singleUse;
```
arpan14, line 62: 1. How is the channel hint managed for multi-use transactions?
2. Given we will always set the `options` at `SessionImpl`, the random hint logic that was introduced in `AbstractReadContext` becomes dead code. Do we clean it up here itself? Or do that in a separate PR? I'm fine either ways.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClient.java
```diff
@@ -0,0 +1,428 @@
...
+  private int getSingleUseChannelHint() {
+    if (this.numCurrentSingleUseTransactions.incrementAndGet() > this.numChannels) {
+      return NO_CHANNEL_HINT;
+    }
+    synchronized (this.channelUsage) {
+      // Get the first unused channel.
+      int channel = this.channelUsage.nextClearBit(/* fromIndex = */ 0);
+      // BitSet returns an index larger than its original size if all the bits are set.
+      // This then means that all channels have already been assigned to single-use transactions,
+      // and that we should not use a specific channel, but rather pick a random one.
```
arpan14, line 322: `but rather pick a random one`

I am a bit confused by this comment. If we return NO_CHANNEL_HINT, we would pass an empty options map to GAX which will do round-robin? Do we need a correction to the docs here?

arpan14 (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClient.java
```diff
@@ -0,0 +1,428 @@
...
+  @Override
+  public ReadOnlyTransaction readOnlyTransaction(TimestampBound bound) {
+    return createMultiplexedSessionTransaction(false).readOnlyTransaction(bound);
+  }
+
+  /**
+   * It is enough with one executor to maintain the multiplexed sessions in all the clients, as they
+   * do not need to be updated often, and the maintenance task is light.
+   */
+  private static final ScheduledExecutorService MAINTAINER_SERVICE =
```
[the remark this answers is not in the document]
olavloite: It would have been slightly harder, but that was not my main reason for not using the same thread. My reasoning for creating a separate maintainer and executor was:
1. This separates multiplexed sessions completely from the session pool. That means that once all operations work on multiplexed sessions, we can just remove the entire session pool implementation.
2. This maintainer uses a `ScheduledThreadPoolExecutor` with zero core threads, and only runs the task once every 10 minutes. The additional resource usage is therefore minimal.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClient.java
olavloite: I added a TODO for removing.

I'm not sure what you mean with the formatting being off (?) The code is formatted using the normal formatter, and I don't see anything strange with this line or the ones directly above.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClient.java
olavloite: 1. Channel hint for multi-use read-only transactions are always random. In theory we could use the same strategy for them as for single-use read-only transactions, but I think we should not, because it would make the 'channel-hint-bitset' depend on read-only transactions always being closed. Failure to close read-only transactions has been a common source of session leaks in the past, and that is exactly what we want to get rid of with multiplexed sessions. Adding a new dependency on closing all read-only transactions here could in theory give us a bit better latency if someone is running 1qps multi-use read-only transactions, but I don't think the benefit of that outweighs the potential downside.
2. The code for assigning a random hint in `AbstractReadContext` is still used. We only create an `options` instance if the channel hint is __not__ the special value `NO_CHANNEL_HINT`. See https://github.com/googleapis/java-spanner/blob/5979f2de26b9ec0a4c6a5f2c91b27304dcd9404a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionImpl.java#L125. What happens in this case is the following:
  2.1. `MultiplexedSessionDatabaseClient` returns `NO_CHANNEL_HINT` if we should just use a random channel: https://github.com/googleapis/java-spanner/blob/5979f2de26b9ec0a4c6a5f2c91b27304dcd9404a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClient.java#L315
  2.2. That value is then passed to the constructor of `SessionImpl`.
  2.3. `SessionImpl` checks if the channel hint is `NO_CHANNEL_HINT`, and if so does not create an `options` instance, but instead calls `sessionReference.getOptions()`.
  2.4. `sessionReference.getOptions()` returns null for multiplexed sessions and the affiliated channel for regular sessions.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClient.java
olavloite: See also my comment above.

This is why the code in `AbstractReadContext` is not dead. This line will generate the random channel hint: https://github.com/googleapis/java-spanner/blob/5979f2de26b9ec0a4c6a5f2c91b27304dcd9404a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java#L195-L197

The reason for that is that by passing in `NO_CHANNEL_HINT` to the constructor of `SessionImpl`, then `SessionImpl` will use `sessionReference.getOptions()`. That returns `null` for multiplexed sessions, and the actually affiliated channel for pooled sessions. And if `options` is null, then the second argument that is passed in to `getChannelHintOptions` will be used (which is a random channel hint).

arpan14: Got it, thanks. I was confused by the `NO_CHANNEL_HINT`. What we mean is `NO_CHANNEL_HINT` returned from this layer. But ultimately it will always fallback to the random hint on `AbstractReadContext`.

PR #3065, opened 2024-04-26
chore: add a random channel hint with single use transactions while u…

arpan14:
Without the hint, GAX would do a round-robin on channels. In theory, a random hint would give a better performance than round robin at a high QPS workload. Also YCSB benchmarks suggest having the hint is better for higher QPS.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
@@ -190,9 +190,8 @@ abstract class AbstractReadContext
       this.bound = builder.bound;
       // single use transaction have a single RPC and hence there is no need
       // of a channel hint. GAX will automatically choose a hint when used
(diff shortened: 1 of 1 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
```diff
@@ -190,8 +190,12 @@ private SingleReadContext(Builder builder) {
       this.bound = builder.bound;
       // single use transaction have a single RPC and hence there is no need
       // of a channel hint. GAX will automatically choose a hint when used
-      // with a multiplexed session.
-      this.channelHint = getChannelHintOptions(session.getOptions(), null);
+      // with a multiplexed session. But GAX currently uses an AtomicLong to perform round-robin
+      // channel selection and that will cause contention issues at high-concurrency. So we are
+      // passing a random hint instead of doing GAX round-robin.
```
olavloite, line 195: Please remove this/replace this comment. Gax uses an `AtomicInteger` to implement round-robin, and that does not introduce any contention.

PR #3068, opened 2024-04-29
chore: adding a benchmark utility in client.

arpan14:
* The existing setup using JMH produces flaky results at time due to thread safe scope variables and also it lacks the ability to pass different parameters from command-line.
* This is a more light-weight setup which allows to pass params from command-line and keeps the setup much simpler.
* This util also allows us to export metrics/traces directly to cloud monitoring so that its easy to visualize and compare stats.


benchmarks/README.md
```diff
@@ -0,0 +1,54 @@
...
+version in `pom.xml` file if you wish to benchmark a different version of Java Client.
+
+
+* The below command uses only 1 thread and 1000 operations. So the total load would 
+be 1000 read operations. The test also uses multiplexed sessions.
+```shell
+mvn clean compile exec:java -Dexec.args="--clients=1 --operations=1000 --multiplexed=true"
+```
+
+* The below command uses only 1 thread and 1000 operations. So the total load would be
```
harshachinta, line 43: Should this line be removed?

benchmarks/README.md
```diff
@@ -0,0 +1,54 @@
...
+1000 read operations. The test uses 10 threads, so at any point in time there would be roughly
+10 concurrent requests. The total load of the benchmark would be 50000 read operations. 
+```shell
+mvn clean compile exec:java -Dexec.args="--clients=10 --operations=5000 --multiplexed=true"
+```
+
+* To run the same test without multiplexed sessions avoid passing `multiplexed` flag. This will 
+make sure that tests uses regular sessions.
+```shell
+mvn clean compile exec:java -Dexec.args="--clients=10 --operations=5000 --multiplexed=true"
```
harshachinta, line 53:
```suggestion
mvn clean compile exec:java -Dexec.args="--clients=10 --operations=5000"
```

benchmarks/README.md
arpan14: Yeah.

harshachinta (approved):

PR #3083, opened 2024-05-03
chore: add flag to force-disable multiplexed sessions

olavloite:
Adds an environment variable that can be used to force-disable the use of multiplexed sessions.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
@@ -105,7 +105,11 @@ public class SessionPoolOptions {
     this.randomizePositionQPSThreshold = builder.randomizePositionQPSThreshold;
     this.inactiveTransactionRemovalOptions = builder.inactiveTransactionRemovalOptions;
     this.poolMaintainerClock = builder.poolMaintainerClock;
-    this.useMultiplexedSession = builder.useMultiplexedSession;
+    // TODO: Remove when multiplexed sessions are guaranteed to be supported.
+    this.useMultiplexedSession =
+        builder.useMultiplexedSession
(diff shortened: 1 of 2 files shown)


rahul2393 (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -105,7 +105,11 @@ private SessionPoolOptions(Builder builder) {
     this.randomizePositionQPSThreshold = builder.randomizePositionQPSThreshold;
     this.inactiveTransactionRemovalOptions = builder.inactiveTransactionRemovalOptions;
     this.poolMaintainerClock = builder.poolMaintainerClock;
-    this.useMultiplexedSession = builder.useMultiplexedSession;
+    // TODO: Remove when multiplexed sessions are guaranteed to be supported.
+    this.useMultiplexedSession =
+        builder.useMultiplexedSession
+            && !Boolean.parseBoolean(
+                System.getenv("GOOGLE_CLOUD_SPANNER_FORCE_DISABLE_MULTIPLEXED_SESSIONS"));
```
rahul2393, line 112: In this case are we expecting customers to set this env to disable multiplexed sessions, if yes then why not customers just set `useMultiplexedSession=false` in Spanner.Options?

olavloite: We want an escape that they can very quickly set with zero code changes in case of calamities.

rahul2393: ok so bare minimum one need to restart the application after the env changes to disable multiplexed session.

Thanks for clarification.

arpan14 (approved):

PR #3088, opened 2024-05-03
fix: multiplexed session metrics were not included in refactor move

olavloite:
The metrics for multiplexed sessions were not included in the refactoring that moved multiplexed sessions out of the session pool. This change re-adds those metrics based on the new client for multiplexed sessions.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
@@ -68,7 +68,6 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Function;
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Preconditions;
-import com.google.common.base.Supplier;
 import com.google.common.collect.ImmutableList;
 import com.google.common.util.concurrent.ForwardingListenableFuture;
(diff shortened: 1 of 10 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -3143,7 +3153,7 @@ boolean isValid() {
 
   /**
    * Returns a multiplexed session. The method fallbacks to a regular session if {@link
-   * SessionPoolOptions#useMultiplexedSession} is not set.
+   * SessionPoolOptions#getUseMultiplexedSession} is not set.
```
olavloite, line 3156: This reference was wrong, and showed up as a warning in IntelliJ

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -68,7 +68,6 @@
 import com.google.common.base.Function;
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Preconditions;
-import com.google.common.base.Supplier;
```
olavloite, line 71: This is not directly related to this change, but we should start to replace the use of `com.google.common.base.Supplier` with `java.util.function.Supplier`.

rahul2393 (approved):

PR #3097, opened 2024-05-07
feat: add tracing for batchUpdate, executeUpdate, and connections

olavloite:
* Adds OpenTelemetry tracing to the Connection API. This again enables tracing in the JDBC driver and PGAdapter.
* Adds traces for `batchUpdate` calls in the Java client library (these calls were not traced).
* Adds traces for `executeUpdate` calls in the Java client library (these calls were not traced).
* Adds an option to include the actual SQL statement that was executed with the trace. This option is disabled by default. The option is named `enableExtendedTracing`, and might include additional features in the future.
* Adds the transaction tag and statement tag as attributes to the OpenTelemetry traces. This makes it easier to find traces for specific transactions and statements.


rahul2393 (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/DdlBatch.java
```diff
@@ -204,36 +212,48 @@ public ApiFuture<Void> writeAsync(CallType callType, Iterable<Mutation> mutation
...
+                Arrays.fill(updateCounts, 1L);
+                state = UnitOfWorkState.RAN;
+                return updateCounts;
+              } catch (SpannerException e) {
+                long[] updateCounts = extractUpdateCounts(operation);
+                throw SpannerExceptionFactory.newSpannerBatchUpdateException(
+                    e.getErrorCode(), e.getMessage(), updateCounts);
+              }
+            } catch (Throwable t) {
+              state = UnitOfWorkState.RUN_FAILED;
```
surbhigarg92, line 242: Do we want to capture span status here as failed ?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/OpenTelemetryTracingTest.java
```diff
@@ -0,0 +1,592 @@
...
+    SdkTracerProvider tracerProvider =
+        SdkTracerProvider.builder()
+            .addSpanProcessor(SimpleSpanProcessor.create(spanExporter))
+            .build();
+
+    openTelemetry =
+        OpenTelemetrySdk.builder()
+            .setPropagators(ContextPropagators.create(W3CTraceContextPropagator.getInstance()))
+            .setTracerProvider(tracerProvider)
+            .buildAndRegisterGlobal();
```
surbhigarg92, line 80: We can inject OT object here instead of using Global ?

surbhigarg92:
We should also update the Readme [OpenTelemetry](https://github.com/googleapis/java-spanner?tab=readme-ov-file#traces) about the new Environment Variable and Option

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1321,6 +1339,19 @@ protected Builder setUseVirtualThreads(boolean useVirtualThreads) {
...
 
+    /**
+     * Sets whether to enable extended OpenTelemetry tracing. Enabling this option will add the
+     * following additional attributes to the traces that are generated by the client:
+     *
+     * <ul>
+     *   <li>db.statement: Contains the SQL statement that is being executed.
+     * </ul>
+     */
+    public Builder setEnableExtendedTracing(boolean enableExtendedTracing) {
```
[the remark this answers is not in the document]
olavloite: I would very much prefer to keep options as much as possible as s instance variables, and not static. Static options makes it hard to predict what happens if you change the value halfway through the lifetime of the application. Does it affect existing clients? Or only new ones?
Static options are also very hard to use in combination with the JDBC driver, as JDBC assumes that all configuration goes through the JDBC connection URL (so basically a string). If one JDBC connection enables extended tracing, it would enable it for all connections (where 'all' could mean 'all existing and future ones' or 'all future ones').

Setting an environment variable does indeed set it for all clients, but it is something that (in Java) can only be set at startup. Changing an environment variable after an application has started is not possible, which prevents the previously mentioned problem around 'what happens to existing clients if you change the value during the lifetime of the application?'.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TraceWrapper.java
```diff
@@ -16,48 +16,76 @@
...
 
   private final Tracer openCensusTracer;
   private final io.opentelemetry.api.trace.Tracer openTelemetryTracer;
+  private final boolean includeSqlStatementInTrace;
 
-  TraceWrapper(Tracer openCensusTracer, io.opentelemetry.api.trace.Tracer openTelemetryTracer) {
+  TraceWrapper(
+      Tracer openCensusTracer,
+      io.opentelemetry.api.trace.Tracer openTelemetryTracer,
+      boolean includeSqlStatementInTrace) {
```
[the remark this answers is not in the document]
olavloite: Good point, done.

(Small note: In Java, the convention is to name the variable `enableExtendedTracing` and use the `is` prefix for the getter method. So you would then have `enableExtendedTracing` as a field, and `isEnableExtendedTracing()` as a getter method.)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/DdlBatch.java
olavloite: Good catch, done.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/OpenTelemetryTracingTest.java
olavloite: No, not really. The Connection API (and JDBC) requires that all connection options are set in the connection URL. JDBC does add one additional option to that: It allows you to pass in an instance of `java.util.Properties` to the `connect(..)` method. According to the specification, that `java.util.Properties` instance may only contain string key/value pairs, but technically it does allow other structured objects, like for example an `OpenTelemetry` instance to be passed in. That is something that we will add in the JDBC driver (so not in this PR, as it goes in the other repository).

olavloite:
> We should also update the Readme [OpenTelemetry](https://github.com/googleapis/java-spanner?tab=readme-ov-file#traces) about the new Environment Variable and Option

Done

surbhigarg92 (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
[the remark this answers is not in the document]
surbhigarg92: Makes sense, we faced enough problems earlier with static options. Let move forward with this instance option

PR #3123, opened 2024-05-23
feat(spanner): add support for Proto Columns in Connection API

harshachinta:
Adds support for Proto Columns feature in the Connection API.
This PR is taken from https://github.com/googleapis/java-spanner/pull/2495.


google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
```diff
@@ -1885,7 +1940,11 @@ private void popUnitOfWorkFromTransactionStack() {
   }
 
   private ApiFuture<Void> executeDdlAsync(CallType callType, ParsedStatement ddl) {
-    return getOrStartDdlUnitOfWork().executeDdlAsync(callType, ddl);
+    ApiFuture<Void> result = getOrStartDdlUnitOfWork().executeDdlAsync(callType, ddl);
+    // reset proto descriptors after executing a DDL statement
+    this.protoDescriptors = null;
+    this.protoDescriptorsFilePath = null;
```
olavloite, line 1946: Would it make more sense to move this to `setDefaultTransactionOptions()`? It keeps the code cleaner, as we have all the reset code in one method, but it does mean that executing for example a query after setting the proto descriptors also clears the fields. One option to prevent the latter could be to add an input argument to `setDefaultTransactionOptions()` that indicates the type of statement that was executed, and only reset these fields if the last statement was a DDL statement.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/DdlClient.java
```diff
@@ -35,11 +36,13 @@
  */
 class DdlClient {
   private final DatabaseAdminClient dbAdminClient;
+  private final String projectId;
   private final String instanceId;
   private final String databaseName;
 
   static class Builder {
     private DatabaseAdminClient dbAdminClient;
+    private String projectId;
```
olavloite, line 45: nit: Instead of adding this field, could we change the individual fields `projectId`, `instanceId`, `databaseName` into one field of type `DatabaseId`?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ProtoDescriptorsConverterTest.java
```diff
@@ -0,0 +1,64 @@
+/*
+ * Copyright 2023 Google LLC
```
olavloite, line 2: super nit: 2024

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ProtoDescriptorsFileConverterTest.java
```diff
@@ -0,0 +1,46 @@
+/*
+ * Copyright 2023 Google LLC
```
olavloite, line 2: nit: 2024

olavloite (approved):

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ProtoDescriptorsFileConverterTest.java
harshachinta: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
harshachinta: sure, will check if it makes sense to move that logic to `setDefaultTransactionOptions()` as a seperate PR. This PR is very big that the code review will be difficult. Will work on this immediately.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/DdlClient.java
harshachinta: sure, will do this along with the previous comment as a seperate PR.

harshachinta: This comment will be addressed in PR https://github.com/googleapis/java-spanner/pull/3126

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionImpl.java
harshachinta: This comment will be addressed in PR https://github.com/googleapis/java-spanner/pull/3126

PR #3125, opened 2024-05-24
chore: client metrics

surbhigarg92:
This PR adds built-in metrics to the Java client library. This will introduce below metrics for the consumers of this library . 

This feature is not currently available for customers to use. Once GA'ed consumers of this library will start getting these metrics in the [MetricsExplorer](https://console.cloud.google.com/monitoring/metrics-explorer?_ga=2.218375641.1367355517.1723789359-1822529888.1723789359)

<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-5283f587-7fff-966a-2e86-a09678ad4ca0"><br /><br /><div dir="ltr" style="margin-left:-117.75pt;" align="left">
Attempt Latency | The latency of an RPC attempt. | Useful for monitoring the general health of the applicationWith attributes like “Token Expiry”, “Status” etc, this metric can help us troubleshoot a wide range of issues.
-- | -- | --
Attempt Count | The count of RPC attempts | Useful for monitoring trafficUseful for deriving “Error Rate” with status attributes
Operation Latency | The total end-to-end latency across all request attempts, including retries. | Useful for monitoring the general health of the applicationThe combination of this metric and Attempt Latency can help us troubleshoot connectivity issues.
Operation Count | The count of operations | Useful for monitoring trafficUseful for deriving “Error Rate” with status labelThe combination of this metric and Attempt Count can be used for deriving Retry Count
</div></b>


google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITBuiltInMetricsTest.java
```diff
@@ -0,0 +1,117 @@
...
+            .setName(name.toString())
+            .setFilter(metricFilter)
+            .setInterval(interval)
+            .setView(ListTimeSeriesRequest.TimeSeriesView.FULL);
+
+    ListTimeSeriesRequest request = requestBuilder.build();
+
+    ListTimeSeriesResponse response = metricClient.listTimeSeriesCallable().call(request);
+    while (response.getTimeSeriesCount() == 0
+        && metricsPollingStopwatch.elapsed(TimeUnit.MINUTES) < 10) {
```
olavloite, line 107: It does not sound reasonable that a test like this should be allowed to run for 10 minutes while waiting for results. Can we fail earlier?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITBuiltInMetricsTest.java
```diff
@@ -0,0 +1,117 @@
...
+      // Call listTimeSeries every minute
+      Thread.sleep(Duration.ofMinutes(1).toMillis());
```
olavloite, line 109: Here also: Do we really need to wait 1 minute between each time we check? That sounds very long for a test.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BuiltInMetricsConstant.java
```diff
@@ -16,16 +16,24 @@
...
+import io.opentelemetry.sdk.metrics.InstrumentSelector;
+import io.opentelemetry.sdk.metrics.InstrumentType;
+import io.opentelemetry.sdk.metrics.View;
+import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 
 public class BuiltInMetricsConstant {
 
-  public static final String METER_NAME = "spanner.googleapis.com/internal/client";
```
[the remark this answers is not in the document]
surbhigarg92: Yes. I realized later that this need not to be public.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BuiltInOpenTelemetryMetricsProvider.java
```diff
@@ -0,0 +1,117 @@
...
+
+  OpenTelemetry getOpenTelemetry(String projectId, @Nullable Credentials credentials) {
+    try {
+      return getOpenTelemetry(SpannerCloudMonitoringExporter.create(projectId, credentials));
+    } catch (IOException ex) {
+      logger.log(
+          Level.WARNING,
+          "Unable to get OpenTelemetry object for client side metrics, will skip exporting client side metrics",
+          ex);
+      return null;
```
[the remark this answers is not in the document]
surbhigarg92: I have added a check in the SpannerOptions. If the OpenTelemetry is null then we are not registering `MetricsTracerFactory` . I can change this to `null` and the check as well. But I don't think it will make any difference.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BuiltInMetricsConstant.java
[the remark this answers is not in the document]
surbhigarg92: Reverted the change and marked the class as Internal

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BuiltInMetricsConstant.java
```diff
@@ -16,16 +16,24 @@
...
 import io.opentelemetry.api.common.AttributeKey;
+import io.opentelemetry.sdk.metrics.Aggregation;
+import io.opentelemetry.sdk.metrics.InstrumentSelector;
+import io.opentelemetry.sdk.metrics.InstrumentType;
+import io.opentelemetry.sdk.metrics.View;
+import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 
 public class BuiltInMetricsConstant {
```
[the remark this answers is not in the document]
surbhigarg92: Done. Marked the class as Internal

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BuiltInOpenTelemetryMetricsProvider.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+  }
+
+  @VisibleForTesting
+  OpenTelemetry getOpenTelemetry(MetricExporter metricExporter) {
```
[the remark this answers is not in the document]
surbhigarg92: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BuiltInOpenTelemetryMetricsProvider.java
```diff
@@ -0,0 +1,117 @@
...
+
+  /**
+   * In most cases this should look like ${UUID}@${hostname}. The hostname will be retrieved from
+   * the jvm name and fallback to the local hostname.
+   */
+  private String getDefaultTaskValue() {
+    // Something like '<pid>@<hostname>'
+    final String jvmName = ManagementFactory.getRuntimeMXBean().getName();
+    // If jvm doesn't have the expected format, fallback to the local hostname
+    if (jvmName.indexOf('@') < 1) {
```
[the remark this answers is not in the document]
surbhigarg92: I have re-written the method as recommended. PTAL

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BuiltInOpenTelemetryMetricsView.java
```diff
@@ -0,0 +1,33 @@
...
+ * limitations under the License.
+ */
+
+package com.google.cloud.spanner;
+
+import io.opentelemetry.sdk.metrics.SdkMeterProviderBuilder;
+import io.opentelemetry.sdk.metrics.export.MetricExporter;
+import io.opentelemetry.sdk.metrics.export.PeriodicMetricReader;
+
+public class BuiltInOpenTelemetryMetricsView {
```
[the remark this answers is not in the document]
surbhigarg92: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1377,6 +1396,19 @@ public Builder setEnableApiTracing(boolean enableApiTracing) {
       return this;
     }
 
+    /** Enabling this will enable built in metrics for each individual RPC execution. */
+    @VisibleForTesting
+    public Builder setEnableBuiltInMetrics(boolean enableBuiltInMetrics) {
```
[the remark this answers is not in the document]
surbhigarg92: This is required for Integration tests which is part of a different package.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1377,6 +1396,19 @@ public Builder setEnableApiTracing(boolean enableApiTracing) {
...
+      this.enableBuiltInMetrics = enableBuiltInMetrics;
+      return this;
+    }
+
+    @InternalApi
+    public Builder canUseDirectPath(boolean canUseDirectPath) {
```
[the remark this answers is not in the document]
surbhigarg92: This is required to set metric attribute `directpath_enabled` . Gax has exposed a method for the same https://github.com/googleapis/sdk-platform-java/blob/main/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java#L483 

The activation of DirectPath depends on multiple factors. In Spanner, we manage its enabling and disabling, but GAX also performs additional checks to determine whether the DirectPath channel is created. This method is consolidating all checks

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -360,9 +360,13 @@ public GapicSpannerRpc(final SpannerOptions options) {
...
 
+      InstantiatingGrpcChannelProvider defaultChannelProvider =
+          defaultChannelProviderBuilder.build();
+
       TransportChannelProvider channelProvider =
-          MoreObjects.firstNonNull(
-              options.getChannelProvider(), defaultChannelProviderBuilder.build());
+          MoreObjects.firstNonNull(options.getChannelProvider(), defaultChannelProvider);
+
+      options.toBuilder().canUseDirectPath(defaultChannelProvider.canUseDirectPath()).build();
```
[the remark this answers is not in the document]
surbhigarg92: This is used [here](https://github.com/googleapis/java-spanner/pull/3125/files#diff-e6869135bf0431d58f306cbc0acc2130dc85070456763884854b5aae9555853aR1671) in SpannerOptions to set the metric attribute.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/HeaderInterceptor.java
```diff
@@ -96,8 +99,10 @@ public void start(Listener<RespT> responseListener, Metadata headers) {
           DatabaseName databaseName = extractDatabaseName(headers);
           String key = databaseName + method.getFullMethodName();
           TagContext tagContext = getTagContext(key, method.getFullMethodName(), databaseName);
+          CompositeTracer compositeTracer = (CompositeTracer) callOptions.getOption(TRACER_KEY);
```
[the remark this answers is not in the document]
surbhigarg92: Done

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITBuiltInMetricsTest.java
surbhigarg92: I have changed the time from 10 minutes to 3 minutes. Rationale is that we are getting the timeseries data from the server and it can take some time for the data to reach there (delay in exporter, or delay in precomputation etc ) .

Note: I will optimize this time once I am able to test it thoroughly after backend changes are available.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITBuiltInMetricsTest.java
surbhigarg92: Exporter pushes the data every 1 min , also pre-computation max time duration is 1 min , hence it makes sense to wait for 1 min here.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BuiltInOpenTelemetryMetricsProvider.java
```diff
@@ -0,0 +1,117 @@
...
+import javax.annotation.Nullable;
+
+class BuiltInOpenTelemetryMetricsProvider {
+
+  private static final Logger logger =
+      Logger.getLogger(BuiltInOpenTelemetryMetricsProvider.class.getName());
+
+  private OpenTelemetry openTelemetry;
+
+  OpenTelemetry getOpenTelemetry(String projectId, @Nullable Credentials credentials) {
```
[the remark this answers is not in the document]
surbhigarg92: I'm not sure on why thread safety is requried here. This will be used when creating the SpannerClient object (during the registration of TracerFactory). If multiple threads are creating different Spanner objects, there should be multiple OT objects.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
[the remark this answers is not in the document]
olavloite: I don't understand what you mean.

What I meant was that the following line:

```
options.toBuilder().canUseDirectPath(defaultChannelProvider.canUseDirectPath()).build();
```

does not actually do anything useful. It takes the `options` object and creates a new builder from it. It then calls `canUseDirectPath(..)` on that builder and calls `build()`. But the result of the `build()` call is not assigned to anything, so the effect of it is nothing, as it does not modify an existing object.

Note: Calling `options.toBuilder().setWhateverOption(...).build()` does __not__ modify the existing `options` object. It creates a new one.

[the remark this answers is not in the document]
surbhigarg92: Removed this and passed the canUseDirectPath in apitracer

rahul2393 (approved):
nit: please check behaviour of this with emulator env set, ideally it should be noop with emulator

surbhigarg92:
> nit: please check behaviour of this with emulator env set, ideally it should be noop with emulator

Added a check for skipping enabling metrics for emulator

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BuiltInOpenTelemetryMetricsProvider.java
```diff
@@ -0,0 +1,149 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import javax.annotation.Nullable;
+
+final class BuiltInOpenTelemetryMetricsProvider {
+
+  public static BuiltInOpenTelemetryMetricsProvider INSTANCE =
```
olavloite, line 49: nit:

```suggestion
  static BuiltInOpenTelemetryMetricsProvider INSTANCE =
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BuiltInOpenTelemetryMetricsProvider.java
```diff
@@ -0,0 +1,149 @@
...
+    } catch (IOException ex) {
+      logger.log(
+          Level.WARNING,
+          "Unable to get OpenTelemetry object for client side metrics, will skip exporting client side metrics",
+          ex);
+      return null;
+    }
+  }
+
+  Map<String, String> getClientAttributes(
```
olavloite, line 80: nit: rename this method to `createClientAttributes(..)` so it is clear that it creates a new object every time it is invoked.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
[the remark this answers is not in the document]
olavloite: If it is only for testing, can we then use the following trick to make this package-private:
1. Make this package-private
2. Add a public static method to SpannerOptionsTestHelper that takes a `SpannerOptions.Builder` instance as an argument, and that calls this `setEnableBuiltInMetrics(...)` method.
3. Call the `SpannerOptionsTestHelper` method from the tests that need to set this value.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -366,9 +366,15 @@ public GapicSpannerRpc(final SpannerOptions options) {
...
+          defaultChannelProviderBuilder.build();
+
       TransportChannelProvider channelProvider =
-          MoreObjects.firstNonNull(
-              options.getChannelProvider(), defaultChannelProviderBuilder.build());
+          MoreObjects.firstNonNull(options.getChannelProvider(), defaultChannelProvider);
+
+      boolean isDirectPathChannelCreated =
+          defaultChannelProvider.canUseDirectPath()
+              && defaultChannelProvider.isDirectPathXdsEnabled();
```
olavloite, line 377: I think that this should be:

```suggestion
      boolean isDirectPathChannelCreated =
          options.getChannelProvider() == null && defaultChannelProvider.canUseDirectPath()
              && defaultChannelProvider.isDirectPathXdsEnabled();
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -394,7 +402,11 @@ public GapicSpannerRpc(final SpannerOptions options) {
                     .setTransportChannelProvider(channelProvider)
                     .setCredentialsProvider(credentialsProvider)
                     .setStreamWatchdogProvider(watchdogProvider)
-                    .setTracerFactory(options.getApiTracerFactory())
+                    .setTracerFactory(
+                        options.getApiTracerFactory(
+                            isDirectPathChannelCreated,
+                            false,
+                            isEmulatorEnabled(options, emulatorHost)))
```
olavloite, line 409: nit, here and for the other calls to this method:
```suggestion
                    .setTracerFactory(
                        options.getApiTracerFactory(
                            isDirectPathChannelCreated,
                            /* isAdminClient = */ false,
                            isEmulatorEnabled(options, emulatorHost)))
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -366,9 +366,15 @@ public GapicSpannerRpc(final SpannerOptions options) {
       // If it is enabled in options uses the channel pool provided by the gRPC-GCP extension.
       maybeEnableGrpcGcpExtension(defaultChannelProviderBuilder, options);
 
+      InstantiatingGrpcChannelProvider defaultChannelProvider =
+          defaultChannelProviderBuilder.build();
```
olavloite, line 370: Can we make this conditional on whether `options.getChannelProvider()` returns null? That makes sure that we don't get any unwanted side-effects of creating a channel provider (that might not work) when the user has specified a custom channel provider.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1654,6 +1699,21 @@ private ApiTracerFactory getDefaultApiTracerFactory() {
     return BaseApiTracerFactory.getInstance();
   }
 
+  private ApiTracerFactory getMetricsApiTracerFactory(boolean isDirectPathChannelCreated) {
```
olavloite, line 1702: Rename to `createMetricsApiTracerFactory` to make clear that it creates a new instance every time it is called.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1630,11 +1655,31 @@ public OpenTelemetry getOpenTelemetry() {
 
   @Override
   public ApiTracerFactory getApiTracerFactory() {
-    List<ApiTracerFactory> apiTracerFactories = new ArrayList();
+    return createApiTracerFactory(false, false, false);
```
olavloite, line 1658: A get-method should preferably not call something that creates a new instance every time it is called. The original implementation of this did create a new list every time it was called, but it did not create any of the elements in the list. Can we make sure that this method still does not create a new instance of ApiTracerFactory each time it is called?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/HeaderInterceptor.java
```diff
@@ -197,4 +205,13 @@ private Attributes getMetricAttributes(String key, String method, DatabaseName d
           return attributesBuilder.build();
         });
   }
+
+  private void addBuiltInMetricAttributes(
+      CompositeTracer compositeTracer, DatabaseName databaseName) {
+    // Built in metrics Attributes.
+    compositeTracer.addAttributes(
```
olavloite, line 212: Strictly speaking, we shouldn't add these attributes to the `compositeTracer`, but only to the underlying built-in metrics tracer. Now the attributes will also be added to any other `MetricsTracer` that is part of the composite tracer, but it only happens if you have built-in tracing enabled.
Meaning that:
1. Enabling built-in metrics will obviously enable built-in metrics.
2. But it will also have the side-effect that any other metrics-tracer that you had enabled will suddenly get two additional attributes.

Is there any way that we can avoid that side-effect?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/OpenTelemetryBuiltInMetricsTracerTest.java
```diff
@@ -0,0 +1,293 @@
...
+
+      try {
+        Thread.sleep(100);
+      } catch (InterruptedException interruptedException) {
+        Thread.currentThread().interrupt();
+        throw new RuntimeException(interruptedException);
+      }
+    }
+
+    assertFalse(String.format("MetricData is missing for metric {0}", fullMetricName), false);
```
olavloite, line 267: This does not do anything, as it just asserts that `false==false`

google-cloud-spanner/src/test/java/com/google/cloud/spanner/OpenTelemetryBuiltInMetricsTracerTest.java
```diff
@@ -0,0 +1,293 @@
...
+              fullMetricName, matchingMetadata)
+          .that(matchingMetadata.size())
+          .isAtMost(1);
+
+      if (!matchingMetadata.isEmpty()) {
+        return matchingMetadata.get(0);
+      }
+
+      try {
+        Thread.sleep(100);
```
olavloite, line 260: Could we reduce this to 1 millisecond and increase the number of attempts to 1000? That should reduce the overall waiting time.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITBuiltInMetricsTest.java
```diff
@@ -0,0 +1,122 @@
...
+            .setEndTime(Timestamps.fromMillis(end.toEpochMilli()))
+            .build();
+    String ddl =
+        "CREATE TABLE FOO ("
+            + "  K    STRING(MAX) NOT NULL,"
+            + "  V    INT64,"
+            + ") PRIMARY KEY (K)";
+    OperationFuture<Void, UpdateDatabaseDdlMetadata> op =
+        db.updateDdl(Collections.singletonList(ddl), null);
+    op.get();
```
olavloite, line 88: We don't seem to use this table. Can we skip the creation of it?

olavloite (approved):

olavloite:
Can we add a more descriptive pull request title than `client metrics`, and also add a description to the pull request? That will improve the release notes that are generated from this change.

If it is something that is not yet GA, but still needs to be merged now, then we should rename the pull request title to  `chore:...`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
surbhigarg92: Yes, even the metricstracerfactory getting added is only created once. We are using the static instance of `BuiltInOpenTelemetryMetricsProvider` to call `getOrCreateOpenTelemetry` . We have a check in this if `openTelemetry` is already created, then we do not create it again. 

It will create a new `MetricsTracerFactory` object with the same `openTelemetry` object similar to `OpenTelemetryApiTracerFactory`

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/HeaderInterceptor.java
surbhigarg92: I didn't quite get you. We have a check in the `CompositeTracer` to pass these attributes further only for `MetricsTracer`.  We only have one MetricsTracer

```
public void addAttributes(String key, String value) {
    for (ApiTracer child : children) {
      if (child instanceof MetricsTracer) {
        MetricsTracer metricsTracer = (MetricsTracer) child;
        metricsTracer.addAttributes(key, value);
      }
    }
  };
  ```

olavloite: Cannot a user create and add a `MetricsTracer` themselves?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
surbhigarg92: Added this check, please verify once.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITBuiltInMetricsTest.java
surbhigarg92: removed. I was in the assumption that we need to have a table to run Select 1 query.

surbhigarg92:
>> If it is only for testing, can we then use the following trick to make this package-private:

Make this package-private
Add a public static method to SpannerOptionsTestHelper that takes a SpannerOptions.Builder instance as an argument, and that calls this setEnableBuiltInMetrics(...) method.
Call the SpannerOptionsTestHelper method from the tests that need to set this value.

I have actually removed `public` from `setEnableBuiltInMetrics` . It was only required for `ITBuiltinMetricsTest` which we are not enabling right now anyway. Lets not give an option to customers to enable it till the time we actually GA our feature.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/HeaderInterceptor.java
surbhigarg92: This class is decorated with @InternalApi annotation . So we don't recommend customers to directly use this class. But even if they use this class to register metricstracerfactory, then yes they would get extra attributes.

PR #3130, opened 2024-05-29
fix: make SessionPoolOptions#setUseMultiplexedSession(boolean) package private

harshachinta:
The `SessionPoolOptions#setUseMultiplexedSessions(boolean)` method was made public by accident. This method is not intended for public use.


olavloite (approved):

olavloite (changes requested):
I think that we should change the PR title

olavloite (approved):
I took the liberty of changing the PR title and description, so that it shows up in the release notes and triggers a new patch version (`chore:` is not a releasable unit).

PR #3141, opened 2024-05-30
feat: support multiplexed sessions for RO transactions

harshachinta:
With multiplexed sessions, the client optimizes and runs multiple applicable requests concurrently on a single session. A single multiplexed session is sufficient to handle all concurrent read-only traffic. Read/write transactions always use regular sessions.
This can be enabled through an environment variable `GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS`.

Session management behavior in Java client,
1. `GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS` is set to `true` - Client uses multiplexed sessions.
2. `GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS` is set to `false` - Client uses regular session cached in the session pool for running 1 concurrent transaction per session.
3. `GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS` is not set - Client uses default session management mechanism which is using regular session cached in the session pool.

**Note: Multiplexed sessions is unsupported when you use SpannerOption#enableGrpcGcpExtension. Enable multiplexed sessions when gRPC-GCP is unset.**


harshachinta:
@harshachinta 
Check if https://github.com/googleapis/java-spanner/pull/3156 is also a blocker for releasing mux.
Reason: https://github.com/googleapis/java-spanner/blob/main/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java#L2971-L2975

olavloite:
> Check if #3156 is also a blocker for releasing mux.
> Reason: https://github.com/googleapis/java-spanner/blob/main/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java#L2971-L2975

@harshachinta No, it's not a blocker. All of it is dead code, due to the static `return false` in https://github.com/googleapis/java-spanner/blob/main/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java#L2971-L2975

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -105,11 +104,12 @@ private SessionPoolOptions(Builder builder) {
     this.randomizePositionQPSThreshold = builder.randomizePositionQPSThreshold;
     this.inactiveTransactionRemovalOptions = builder.inactiveTransactionRemovalOptions;
     this.poolMaintainerClock = builder.poolMaintainerClock;
-    // TODO: Remove when multiplexed sessions are guaranteed to be supported.
+    // useMultiplexedSession priority => Environment var > private setter > client default
```
olavloite, line 107: We should change this priority to `private setter -> Env var -> client default`. Otherwise, it will be impossible to use this feature in for example JDBC and PGAdapter, as it is not possible to set an environment variable in Java after the JVM has started. So instead, JDBC and PGAdapter will be using the private setter to override the default. This is also the reason that we still need the escape hatch in the form of an environment variable that __disables__ multiplexed sessions.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -105,11 +104,12 @@ private SessionPoolOptions(Builder builder) {
...
+    Boolean useMultiplexedSessionFromEnvVariable = getUseMultiplexedSessionFromEnvVariable();
     this.useMultiplexedSession =
-        builder.useMultiplexedSession
-            && !Boolean.parseBoolean(
-                System.getenv("GOOGLE_CLOUD_SPANNER_FORCE_DISABLE_MULTIPLEXED_SESSIONS"));
```
olavloite, line 112: See also above: We should not remove the `GOOGLE_CLOUD_SPANNER_FORCE_DISABLE_MULTIPLEXED_SESSIONS` env var yet, as it is needed as an escape hatch for clients that use the private setter to enable multiplexed sessions.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
olavloite: So the behavior should be:
1. If `GOOGLE_CLOUD_SPANNER_FORCE_DISABLE_MULTIPLEXED_SESSIONS=true` => Use regular sessions.
2. If the private setter has been called => Use the value from the private setter.
3. If `GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS` has been set => Use the value from that env var.
4. Use the default in the client library (currently `false`)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1266,14 +1266,20 @@ public Builder setHost(String host) {
       return this;
     }
 
-    /** Enables gRPC-GCP extension with the default settings. */
+    /**
+     * Enables gRPC-GCP extension with the default settings. Do not set
+     * GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS to true, since Multiplexed sessions is not
```
olavloite, line 1271:
```suggestion
     * GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS to true in combination with this option, as Multiplexed sessions are not
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1266,14 +1266,20 @@ public Builder setHost(String host) {
...
+     */
     public Builder enableGrpcGcpExtension() {
       return this.enableGrpcGcpExtension(null);
     }
 
     /**
      * Enables gRPC-GCP extension and uses provided options for configuration. The metric registry
-     * and default Spanner metric labels will be added automatically.
+     * and default Spanner metric labels will be added automatically. Do not set
+     * GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS to true, since Multiplexed sessions is not
```
olavloite, line 1281:
```suggestion
     * GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS to true in combination with this option, as Multiplexed sessions are not
```

olavloite (changes requested):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
harshachinta: @olavloite 
1. The value in environment variable overrides the private setter only when the env `GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS` is **set**. In case of JDBC and PGAdapter, the env will not be set and the private setter will override the default behavior.
2. For escape hatch they can set `GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS` to `false` which will disable multiplexed sessions. I don't understand why we need to maintain a seperate environment variable `GOOGLE_CLOUD_SPANNER_FORCE_DISABLE_MULTIPLEXED_SESSIONS` for disabling?

olavloite: Yeah, I think you are right. So the working will be:
1. PGAdapter / JDBC use the private setter to enable mux sessions by default.
2. As long as no env var has been set, the private setter value is respected.
3. Users can disable mux sessions by setting the env var to `false`.

Setting the env var to `false` will also work for the client library itself in the future, even after the default in the client lib has become true.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
olavloite: No longer applicable

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -518,7 +528,9 @@ public static class Builder {
      */
     private long randomizePositionQPSThreshold = 0L;
 
-    private boolean useMultiplexedSession = getUseMultiplexedSessionFromEnvVariable();
+    // This field controls the default behavior of session management in Java client.
+    // Set useMultiplexedSession to true to make multiplexed session as default.
```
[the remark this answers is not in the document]
harshachinta: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -312,6 +312,16 @@ public boolean getUseMultiplexedSession() {
     return useMultiplexedSession;
   }
 
+  private static Boolean getUseMultiplexedSessionFromEnvVariable() {
+    String useMultiplexedSessionFromEnvVariable =
+        System.getenv("GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS");
+    if (useMultiplexedSessionFromEnvVariable != null
+        && useMultiplexedSessionFromEnvVariable.length() > 0) {
+      return Boolean.parseBoolean(useMultiplexedSessionFromEnvVariable);
+    }
```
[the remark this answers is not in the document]
harshachinta: Thus throwing an exception makes sense in that case? I have added the code changes can you PTAL?
or should we ignore the env in that case with out throwing errors?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
harshachinta: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
harshachinta: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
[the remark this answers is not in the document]
harshachinta: @olavloite 
I have added a test for this
https://github.com/googleapis/java-spanner/pull/3141/files#diff-da528327933e25a61b783f588836f20df8e46e91f98b42191236fb3527867a52R210

olavloite (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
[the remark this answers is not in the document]
olavloite: Yes, I think that throwing an exception in a case like this is the best. It is easier to debug for a user than silently ignoring an invalid value.

PR #3161, opened 2024-06-13
build: make executor deps test scoped

olavloite:
The executor dependencies are not needed for the normal build, but are only used for tests, and should therefore be test scoped.

### DIFF
--- a/google-cloud-spanner/pom.xml
+++ b/google-cloud-spanner/pom.xml
@@ -209,16 +209,7 @@
     </dependency>
     <dependency>
       <groupId>com.google.api.grpc</groupId>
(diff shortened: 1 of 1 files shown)


olavloite:
@gyang-google Can these dependencies be test-scoped? Or do we need them for a specific reason as a compile dependency, which again also means that they are included in the .jar that we ship to users?

gyang-google:
Actually we need to build them to .jar targets, but as long as we have a
specific way to build them with the client libraries, we are good. Do you
have any ideas here?

Best,
Guang

gyang-google (approved):

gyang-google:
@olavloite please let me know once this check is merged, I need to update the build command accordingly.

rahul2393 (approved):

PR #3163, opened 2024-06-18
chore: Composite Tracer

surbhigarg92:


google-cloud-spanner/src/main/java/com/google/cloud/spanner/CompositeTracer.java
```diff
@@ -0,0 +1,172 @@
...
+
+import com.google.api.gax.tracing.ApiTracer;
+import com.google.api.gax.tracing.BaseApiTracer;
+import com.google.api.gax.tracing.MetricsTracer;
+import com.google.common.collect.ImmutableList;
+import java.util.ArrayList;
+import java.util.List;
+import org.threeten.bp.Duration;
+
+public class CompositeTracer extends BaseApiTracer {
```
olavloite, line 27: Can this class be package-private? (I.e. remove the `public` qualifier)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/CompositeTracer.java
```diff
@@ -0,0 +1,172 @@
...
+  private final List<ApiTracer> children;
+
+  public CompositeTracer(List<ApiTracer> children) {
```
olavloite, line 30: Same also here: Can we make this package-private?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/CompositeTracer.java
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  @Override
+  public void batchRequestSent(long elementCount, long requestSize) {
+    for (ApiTracer child : children) {
+      child.batchRequestSent(elementCount, requestSize);
+    }
+  }
+
+  public void addAttributes(String key, String value) {
```
olavloite, line 164: Where/when is this method called? It does not override any methods (at least it is not marked as such), and I can't find anything in this PR that seems to call it.

olavloite (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/CompositeTracer.java
surbhigarg92: This class needs to be accessed from [HeaderInterceptor](https://github.com/googleapis/java-spanner/blob/main/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/HeaderInterceptor.java) which is in a different package. 

Should I add @InternalApi annotation ?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/CompositeTracer.java
surbhigarg92: This will be required for [MetricsTracer](https://github.com/googleapis/sdk-platform-java/blob/f952e3fc06a0920cd8bacc7fe030bd629433baae/gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsTracer.java#L232) 

Added in advance

PR #3164, opened 2024-06-19
chore: custom exporter for Client Built-in Metrics

surbhigarg92:


google-cloud-spanner/pom.xml
```diff
@@ -435,6 +433,16 @@
...
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-monitoring</artifactId>
+      <version>3.38.0</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.api.grpc</groupId>
+      <artifactId>proto-google-cloud-monitoring-v3</artifactId>
+      <version>3.38.0</version>
+    </dependency>
```
olavloite, line 445: nit: Move this further up in the pom.xml. The convention is to have the test dependencies last. This also applies to the opentelemetry dependencies above that are changed from test to compile scope.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
```diff
@@ -0,0 +1,246 @@
...
+  private CompletableResultCode lastExportCode;
+  private final MetricServiceClient client;
+  private final String spannerProjectId;
+  private final AtomicBoolean isShutdown = new AtomicBoolean(false);
+
+  static SpannerCloudMonitoringExporter create(String projectId, @Nullable Credentials credentials)
+      throws IOException {
+    MetricServiceSettings.Builder settingsBuilder = MetricServiceSettings.newBuilder();
+    CredentialsProvider credentialsProvider;
+    if (credentials == null) {
```
olavloite, line 87: Normally, when a user has not set any credentials, we will use the default credentials found in the environment. Is that what will normally happen here as well?
Put another way: Will we normally call `SpannerCloudMonitoringExporter.create(..)` with the default credentials when the user has not set any specific credentials?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
```diff
@@ -0,0 +1,246 @@
...
+ * instrumentation scope and upload it via the Google Cloud Monitoring API.
+ */
+class SpannerCloudMonitoringExporter implements MetricExporter {
+
+  private static final Logger logger =
+      Logger.getLogger(SpannerCloudMonitoringExporter.class.getName());
+
+  // This system property can be used to override the monitoring endpoint
+  // to a different environment. It's meant for internal testing only.
+  private static final String MONITORING_ENDPOINT =
```
olavloite, line 69: System properties are allowed to change during the lifetime of a JVM. This will only read the value when this class is loaded. Is that intentional? Or would it maybe be better to read the value dynamically every time a new SpannerCloudMonitoringExporter is created?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
```diff
@@ -0,0 +1,246 @@
...
+      throws IOException {
+    MetricServiceSettings.Builder settingsBuilder = MetricServiceSettings.newBuilder();
+    CredentialsProvider credentialsProvider;
+    if (credentials == null) {
+      credentialsProvider = NoCredentialsProvider.create();
+    } else {
+      credentialsProvider = FixedCredentialsProvider.create(credentials);
+    }
+    settingsBuilder.setCredentialsProvider(credentialsProvider);
+    settingsBuilder.setEndpoint(MONITORING_ENDPOINT);
```
olavloite, line 93: See comment above: We could also read the system property here.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
```diff
@@ -0,0 +1,246 @@
...
+  @VisibleForTesting
+  SpannerCloudMonitoringExporter(String projectId, MetricServiceClient client) {
+    this.client = client;
+    this.spannerProjectId = projectId;
+  }
+
+  @Override
+  public CompletableResultCode export(Collection<MetricData> collection) {
+    if (isShutdown.get()) {
+      logger.log(Level.WARNING, "Exporter is shutting down");
```
olavloite, line 113:
```suggestion
      logger.log(Level.WARNING, "Exporter is shut down");
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
```diff
@@ -0,0 +1,246 @@
...
+      return CompletableResultCode.ofFailure();
+    }
+
+    lastExportCode = exportSpannerClientMetrics(collection);
```
olavloite, line 117: nit (this makes it a bit easier to quickly see that we are setting an instance field):
```suggestion
    this.lastExportCode = exportSpannerClientMetrics(collection);
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
```diff
@@ -0,0 +1,246 @@
...
+        collection.stream()
+            .filter(md -> SPANNER_METRICS.contains(md.getName()))
+            .collect(Collectors.toList());
+
+    // Skips exporting if there's none
+    if (spannerMetricData.isEmpty()) {
+      return CompletableResultCode.ofSuccess();
+    }
+
+    // Verifies metrics project id are the same as the spanner project id set on this client
```
olavloite, line 134:
```suggestion
    // Verifies metrics project id is the same as the spanner project id set on this client
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
```diff
@@ -0,0 +1,246 @@
...
+    if (!spannerMetricData.stream()
+        .flatMap(metricData -> metricData.getData().getPoints().stream())
+        .allMatch(
+            pd -> spannerProjectId.equals(SpannerCloudMonitoringExporterUtils.getProjectId(pd)))) {
+      logger.log(Level.WARNING, "Metric data has different a projectId. Skip exporting.");
```
olavloite, line 139:
```suggestion
      logger.log(Level.WARNING, "Metric data has a different projectId. Skipping export.");
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
```diff
@@ -0,0 +1,246 @@
...
+
+    CompletableResultCode spannerExportCode = new CompletableResultCode();
+    ApiFutures.addCallback(
+        futureList,
+        new ApiFutureCallback<List<Empty>>() {
+          @Override
+          public void onFailure(Throwable throwable) {
+            if (spannerExportFailureLogged.compareAndSet(false, true)) {
+              String msg = "createServiceTimeSeries request failed for spanner metrics.";
+              if (throwable instanceof PermissionDeniedException) {
```
olavloite, line 167: Do we want to continue to retry exporting after this error? I could see this both ways:
1. A `PermissionDenied` error is unlikely to go away, unless the user takes action, so continously retrying is not likely to work, and the client is shut down after this error. The user needs to re-start the client if they fix the permission issue.
2. We want the client to automatically start working again if permission issue is fixed.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
```diff
@@ -0,0 +1,246 @@
...
+    lastExportCode = exportSpannerClientMetrics(collection);
+    return lastExportCode;
+  }
+
+  /** Export client built in metrics */
+  private CompletableResultCode exportSpannerClientMetrics(Collection<MetricData> collection) {
+    // Filter spanner metrics
+    List<MetricData> spannerMetricData =
+        collection.stream()
+            .filter(md -> SPANNER_METRICS.contains(md.getName()))
```
olavloite, line 126: How does this work with any other MetricExporter that might also be active in the same JVM? This filters the collection so it only exports the metrics that this exporter cares about. But the collection that it was given remains unmodified. Will any other MetricExporter also have to filter out metrics that it does not care about to avoid getting the Spanner-internal metrics? Or are we fine with the Spanner-internal metrics also being exported by other exporters?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
```diff
@@ -0,0 +1,246 @@
...
+    List<ApiFuture<Empty>> batchResults = new ArrayList<>();
+
+    for (List<TimeSeries> batch : Iterables.partition(timeSeries, EXPORT_BATCH_SIZE_LIMIT)) {
+      CreateTimeSeriesRequest req =
+          CreateTimeSeriesRequest.newBuilder()
+              .setName(projectName.toString())
+              .addAllTimeSeries(batch)
+              .build();
+      ApiFuture<Empty> f = this.client.createServiceTimeSeriesCallable().futureCall(req);
+      batchResults.add(f);
```
olavloite, line 203: nit:
```suggestion
      batchResults.add(this.client.createServiceTimeSeriesCallable().futureCall(req));
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporterUtils.java
```diff
@@ -0,0 +1,209 @@
...
+import io.opentelemetry.sdk.metrics.data.MetricData;
+import io.opentelemetry.sdk.metrics.data.MetricDataType;
+import io.opentelemetry.sdk.metrics.data.PointData;
+import io.opentelemetry.sdk.metrics.data.SumData;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+class SpannerCloudMonitoringExporterUtils {
```
olavloite, line 58: nit: add a private no-arg constructor, as this class is not intended to be instantiated

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
```diff
@@ -0,0 +1,246 @@
...
+    return ApiFutures.allAsList(batchResults);
+  }
+
+  @Override
+  public CompletableResultCode flush() {
+    if (lastExportCode != null) {
+      return lastExportCode;
+    }
+    return CompletableResultCode.ofSuccess();
+  }
```
aabmass, line 215: The usual use of this would be if the exporter does additional asynchronous queueing or retries, `flush()` would drain the queue or wait for retries to finish. I imagine this could just be an empty method here, which is what we do for the user facing exporter https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/11c85b9e3809a60a6348982ee9ea38d28cbe49db/exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/InternalMetricExporter.java#L261-L269

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
```diff
@@ -0,0 +1,246 @@
...
+
+  @Override
+  public CompletableResultCode shutdown() {
+    if (!isShutdown.compareAndSet(false, true)) {
```
aabmass, line 219: @psx95 do we do this in our user facing exporter?

Usually the user would not call this directly but calling `shutdown()` on `OpenTelemetrySdk` or `MeterProvider` would run this for you, handling the duplicate calls.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
```diff
@@ -0,0 +1,246 @@
...
+      logger.log(Level.WARNING, "shutdown is called multiple times");
+      return CompletableResultCode.ofSuccess();
+    }
+    CompletableResultCode flushResult = flush();
```
aabmass, line 223: The SDK usually do this for you before calling this method. Check out our implementation https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/11c85b9e3809a60a6348982ee9ea38d28cbe49db/exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/InternalMetricExporter.java#L271-L276

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
surbhigarg92: My plan was to use [getCredentials](https://github.com/googleapis/java-spanner/blob/main/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java) when I call `SpannerCloudMonitoringExporter.create(..)`

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
surbhigarg92: >> If these are internal metrics that the customer doesn't know anything about, what "incentive" does the customer have to fix the permission issue?

These metrics will be visible to customers as well which will help them troubleshoot their latency problems. 

>> A PermissionDenied error is unlikely to go away, unless the user takes action, so continously retrying is not likely to work, and the client is shut down after this error. The user needs to re-start the client if they fix the permission issue.

We have not added any retry logic for this error. But `export` method will be called every 60 sec and customer would see this log once per minute . This check is to only inform customers that what steps they can take to get desired permissions. 

>> We want the client to automatically start working again if permission issue is fixed.
Yes, if the permissions are added, then the next `export` call will be success.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
surbhigarg92: This exporter will only be added for our Built-in metrics. Customer will use their own exporter and their own set of metrics, so this will not have any impact on other metrics. 

> OpenTelemetrySdk, in which case these metrics won't be created in the global SDK.
Yes you are right , this will not be in the global SDK. 

This check ensures that only our built-in metrics are pushed to Monarch. For example, today all the metrics from the GAX layer are included in our built-in metrics. If the GAX layer introduces a new metric in the future that we don't want to add, this check will reject any additional metrics.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
surbhigarg92: User facing exporter also returns [CompletableResultCode.ofSuccess()](https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/11c85b9e3809a60a6348982ee9ea38d28cbe49db/exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/InternalMetricExporter.java#L261-L269) . We have just added an extra check if the `lastExportCode` has any code set

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
surbhigarg92: There is not much difference in what is done in user facing exporter and what is done here except some extra checks. 

Here as well we are calling `metricServiceClient.shutdown()` , its just we are having extra try catch block here to make sure that any error in these exporter does not propagate to customer. 

Also we are explicitly calling `flush` in the `shutdown` .

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
olavloite: Do you mean this method: https://github.com/googleapis/java-spanner/blob/bee79af389de2a9ace74a3952744311e756ae4bd/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java#L1439

That method is only usable in case the user uses the Connection API (which no one uses directly, meaning it is only usable in combination with the JDBC driver). But you could use the same type of logic and/or move that logic to a place where it is usable directly in the client library.

olavloite (approved):
LGTM, with some minor questions/assumptions.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
surbhigarg92: Got it. Currently we don't have any plan for such tests. We can modify later if there is a need.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
surbhigarg92: Yes, that is true. We will create a new OT object for built in metrics and customer would create their own OT object.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
surbhigarg92: I assume that so. But I will double check if the export fails with error then metrics data should be discarded and the memory should not keep on growing.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
aabmass: Yup. I just was concerned about semantics. If the previous export failed say 30 seconds ago, why would we indicate the force flush failed as well?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
aabmass: OK after digging a bit more, I agree flush() makes sense.

One question though, if an exception is thrown in the `whenComplete`, does it actually propagate to the user? I would imagine the error is surfaced to the caller of `OpenTelemetrySdk.shutdown()`

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporterUtils.java
```diff
@@ -0,0 +1,211 @@
...
+            .setMetricKind(convertMetricKind(metricData))
+            .setValueType(convertValueType(metricData.getType()));
+    Metric.Builder metricBuilder = Metric.newBuilder().setType(metricData.getName());
+
+    Attributes attributes = pointData.getAttributes();
+    MonitoredResource.Builder monitoredResourceBuilder =
+        MonitoredResource.newBuilder().setType(SPANNER_RESOURCE_TYPE);
+
+    for (AttributeKey<?> key : attributes.asMap().keySet()) {
+      if (SPANNER_PROMOTED_RESOURCE_LABELS.contains(key)) {
```
aabmass, line 99: It's interesting all the MonitoredResource labels are coming from metric level attributes instead of OTel Resource attributes. None of these could be set in the OTel resource?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BuiltInMetricsConstant.java
```diff
@@ -0,0 +1,82 @@
...
+
+  public static final AttributeKey<String> PROJECT_ID_KEY = AttributeKey.stringKey("project_id");
+  public static final AttributeKey<String> INSTANCE_ID_KEY = AttributeKey.stringKey("instance_id");
+  public static final AttributeKey<String> LOCATION_ID_KEY = AttributeKey.stringKey("location");
+  public static final AttributeKey<String> INSTANCE_CONFIG_ID_KEY =
+      AttributeKey.stringKey("instance_config");
+
+  // These metric labels will be promoted to the spanner monitored resource fields
+  public static final Set<AttributeKey<String>> SPANNER_PROMOTED_RESOURCE_LABELS =
+      ImmutableSet.of(PROJECT_ID_KEY, INSTANCE_ID_KEY, INSTANCE_CONFIG_ID_KEY, LOCATION_ID_KEY);
```
aabmass, line 57: nit, iirc the `project_id` MonitoredResource label must always match the project the metrics are being written to, and so it is unnecessary to send it. We've taken this approach in our exporters since there is no benefit in sending it

aabmass (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
psx95: Could you explain the purpose of that check ?

The [`flush()`](https://github.com/open-telemetry/opentelemetry-java/blob/0aacc55d1e3f5cc6dbb4f8fa26bcb657b01a7bc9/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/MetricExporter.java#L64) method is an hint to the exporter to complete export for any metrics previously collected - and return the result of the flush operation.

This check would cause to violate this.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
psx95: I think its ok to surround in try-catch to prevent any error leaking to customer.

However, `flush()` does not actually perform any actual export for metrics, what is the reason behind the explicit call in shutdown?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
psx95: No, we do not have this check. 

This is because we rely on the `shutdown()` method of the [MetricServiceClient](https://github.com/googleapis/google-cloud-java/blob/5a2e219452e1f57e2c26c8bd4edd2d8311d0a849/java-monitoring/google-cloud-monitoring/src/main/java/com/google/cloud/monitoring/v3/MetricServiceClient.java#L2097). This class implements [`BackgroundResource`](https://github.com/googleapis/sdk-platform-java/blob/502f80101dec191befb660a1aba6d0c354758c18/gax-java/gax/src/main/java/com/google/api/gax/core/BackgroundResource.java#L38) and the documentation suggests invocation has no effect if already shutdown, so we don't care if its called multiple times.

IIUC, a similar thing can be done in this exporter.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
```diff
@@ -0,0 +1,246 @@
...
+          MetricServiceSettings.getDefaultEndpoint());
+
+  // This the quota limit from Cloud Monitoring. More details in
+  // https://cloud.google.com/monitoring/quotas#custom_metrics_quotas.
+  private static final int EXPORT_BATCH_SIZE_LIMIT = 200;
+  private final AtomicBoolean spannerExportFailureLogged = new AtomicBoolean(false);
+  private CompletableResultCode lastExportCode;
+  private final MetricServiceClient client;
+  private final String spannerProjectId;
+  private final AtomicBoolean isShutdown = new AtomicBoolean(false);
```
psx95, line 81: If you absolutely need to track the shutdown state, I see that this exporter relies on `MetricServiceClient` so perhaps [`isShutdown()`](https://github.com/googleapis/google-cloud-java/blob/5a2e219452e1f57e2c26c8bd4edd2d8311d0a849/java-monitoring/google-cloud-monitoring/src/main/java/com/google/cloud/monitoring/v3/MetricServiceClient.java#L2102) would be well-suited to replace this AtomicBoolean.

psx95 (commented):
Left some thoughts about handling the shutdown mechanism of the exporter.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporterUtils.java
surbhigarg92: @aabmass It can be set, but does it really make a difference if these Monitored resource labels comes from metrics attributes or OTel resource attributes ?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
surbhigarg92: Verified , metric data will not grow

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
surbhigarg92: Removed this flag and used the [isShutdown()](https://github.com/googleapis/google-cloud-java/blob/5a2e219452e1f57e2c26c8bd4edd2d8311d0a849/java-monitoring/google-cloud-monitoring/src/main/java/com/google/cloud/monitoring/v3/MetricServiceClient.java#L2102) as suggested

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
surbhigarg92: Removed the flush() call from shutdown as flush is not doing anything

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
surbhigarg92: Removed this check

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
```diff
@@ -0,0 +1,236 @@
...
+  @VisibleForTesting
+  SpannerCloudMonitoringExporter(String projectId, MetricServiceClient client) {
+    this.client = client;
+    this.spannerProjectId = projectId;
+  }
+
+  @Override
+  public CompletableResultCode export(Collection<MetricData> collection) {
+    if (client.isShutdown()) {
+      logger.log(Level.WARNING, "Exporter is shut down");
```
odeke-em, line 111: Nit: "Exporter was already shut down"

odeke-em: Also I think it is an error to try to use an already shutdown exporter, hence perhaps change the log level to Error?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
```diff
@@ -0,0 +1,236 @@
...
+  /** Export client built in metrics */
+  private CompletableResultCode exportSpannerClientMetrics(Collection<MetricData> collection) {
+    // Filter spanner metrics
+    List<MetricData> spannerMetricData =
+        collection.stream()
+            .filter(md -> SPANNER_METRICS.contains(md.getName()))
+            .collect(Collectors.toList());
+
+    // Skips exporting if there's none
+    if (spannerMetricData.isEmpty()) {
```
odeke-em, line 128: Useful to perform an info log so that debugging will reveal truly that .isEmpty() and that data wasn't exported, instead of one scratching their head wondering why metrics weren't exported.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporterUtils.java
```diff
@@ -0,0 +1,211 @@
...
+            .build();
+
+    builder.addPoints(createPoint(metricData.getType(), pointData, timeInterval));
+
+    return builder.build();
+  }
+
+  private static MetricKind convertMetricKind(MetricData metricData) {
+    switch (metricData.getType()) {
+      case HISTOGRAM:
```
odeke-em, line 122: I'd suggest avoiding the fall through and instead explicitly add common cases on the same line so: `case HISTOGRAM, EXPONENTIAL_HISTOGRAM` for clarity and to help avoid any future unexpected issues.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporterUtils.java
```diff
@@ -0,0 +1,211 @@
...
+      case EXPONENTIAL_HISTOGRAM:
+        return convertHistogramType(metricData.getHistogramData());
+      case LONG_GAUGE:
```
odeke-em, line 125: Ditto about not allowing fall through and instead putting similar case values on the same line.

odeke-em (commented):
Thanks @surbhigarg92, just some drive by nits/comments.

PR #3180, opened 2024-06-27
feat: support Read RPC OrderBy

mayurkale22:
Adds support for setting Read OrderBy.

Customers can set this option only for the Read API. This option will allow users to specify whether they want the results to be returned in primary key order (PRIMARY_KEY) or in an out of order (NO_ORDER). The NO_ORDER is significantly faster than KEY_ORDER, because spanner uses higher parallelism for non-ordered reads.

Go counterpart: https://github.com/googleapis/google-cloud-go/pull/10289

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
@@ -902,6 +902,9 @@ abstract class AbstractReadContext
     if (readOptions.hasDataBoostEnabled()) {
       builder.setDataBoostEnabled(readOptions.dataBoostEnabled());
     }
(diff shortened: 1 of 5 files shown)


harshachinta:
@mayurkale22 
Looks like the sample added in the PR description is for Golang. Can we move this sample to the Go PR - https://github.com/googleapis/google-cloud-go/pull/10289?

mayurkale22:
> @mayurkale22 Looks like the sample added in the PR description is for Golang. Can we move this sample to the Go PR - [googleapis/google-cloud-go#10289](https://github.com/googleapis/google-cloud-go/pull/10289)?

Done

mayurkale22:
@harshachinta Can you pls review this PR?

rahul2393 (approved):

PR #3198, opened 2024-07-04
build: run graalvm integration tests on emulator

rahul2393:
Run the GraalVM integration tests on the emulator instead of production to speed up builds.


.kokoro/build.sh
```diff
@@ -48,6 +48,14 @@ if [[ ! -z "${GOOGLE_APPLICATION_CREDENTIALS}" && "${GOOGLE_APPLICATION_CREDENTI
     export GOOGLE_APPLICATION_CREDENTIALS=$(realpath ${KOKORO_GFILE_DIR}/${GOOGLE_APPLICATION_CREDENTIALS})
 fi
 
+# Start the Spanner emulator if the job type is graalvm or graalvm17
+if [[ "$JOB_TYPE" == "graalvm" ]] || [[ "$JOB_TYPE" == "graalvm17" ]]; then
+  echo "Starting emulator"
+  export SPANNER_EMULATOR_HOST=localhost:9010
+  docker pull gcr.io/cloud-spanner-emulator/emulator
+  docker run -d --rm --name spanner-emulator -p 9010:9010 -p 9020:9020 gcr.io/cloud-spanner-emulator/emulator
```
harshachinta, line 56: nit: How about moving this with in the switch case below?
https://github.com/googleapis/java-spanner/blob/ade770221cc325cc4aee52881b4b07e569ef75df/.kokoro/build.sh#L175-L176

I understand that there will be code duplication. But just thinking that this will give clarity to someone looking at graalvm that it runs on emulator and not real backend.

harshachinta (approved):

PR #3215, opened 2024-07-17
feat(spanner): set manual affinity incase of gRPC-GCP extenstion

harshachinta:
Earlier, gRPC-GCP maintains affinity using the [ApiConfig](https://github.com/googleapis/java-spanner/blob/main/google-cloud-spanner/src/main/resources/com/google/cloud/spanner/spi/v1/grpc-gcp-apiconfig.json) by looking into the gRPC messages to get an affinity key.

This PR sets the affinity manually to gRPC-GCP via call options. Reason - With multiplexed sessions we need to maintain transaction to channel affinity, and with regular sessions we need to maintain session to channel affinity.
Setting manual affinity will override the API config setting.

References:
1. Java client design and benchmarking - https://docs.google.com/document/d/1iQwAGCwvatdoGxvfucd0eFAysx6roGMxJsoyygX79NY/edit?resourcekey=0-feEwEwbAPxsA-QydfaG53A&tab=t.0
2. gRPC-GCP manual affinity design - https://docs.google.com/document/d/1NfnaCHpE6EzePU_IBTZHvevGizYTBtjuMfHlm9SZ70w/edit?tab=t.0

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
@@ -59,7 +59,6 @@ import com.google.cloud.RetryHelper.RetryHelperException;
 import com.google.cloud.grpc.GcpManagedChannel;
 import com.google.cloud.grpc.GcpManagedChannelBuilder;
 import com.google.cloud.grpc.GcpManagedChannelOptions;
-import com.google.cloud.grpc.GcpManagedChannelOptions.GcpChannelPoolOptions;
 import com.google.cloud.grpc.GcpManagedChannelOptions.GcpMetricsOptions;
 import com.google.cloud.grpc.GrpcTransportOptions;
 import com.google.cloud.spanner.AdminRequestsPerMinuteExceededException;
@@ -268,6 +267,7 @@ public class GapicSpannerRpc implements SpannerRpc {
   private static final ConcurrentMap<String, RateLimiter> ADMINISTRATIVE_REQUESTS_RATE_LIMITERS =
       new ConcurrentHashMap<>();
   private final boolean leaderAwareRoutingEnabled;
+  private final int numChannels;
 
   public static GapicSpannerRpc create(SpannerOptions options) {
     return new GapicSpannerRpc(options);
@@ -319,6 +319,7 @@ public class GapicSpannerRpc implements SpannerRpc {
     this.callCredentialsProvider = options.getCallCredentialsProvider();
     this.compressorName = options.getCompressorName();
     this.leaderAwareRoutingEnabled = options.isLeaderAwareRoutingEnabled();
+    this.numChannels = options.getNumChannels();
 
     if (initializeStubs) {
       // First check if SpannerOptions provides a TransportChannelProvider. Create one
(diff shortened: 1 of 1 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -529,6 +531,10 @@ private static String parseGrpcGcpApiConfig() {
   private static GcpManagedChannelOptions grpcGcpOptionsWithMetrics(SpannerOptions options) {
     GcpManagedChannelOptions grpcGcpOptions =
         MoreObjects.firstNonNull(options.getGrpcGcpOptions(), new GcpManagedChannelOptions());
+    GcpChannelPoolOptions gcpChanelPoolOptions =
```
olavloite, line 534:
```suggestion
    GcpChannelPoolOptions gcpChannelPoolOptions =
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -1950,7 +1957,17 @@ <ReqT, RespT> GrpcCallContext newCallContext(
...
       context = context.withChannelAffinity(Option.CHANNEL_HINT.getLong(options).intValue());
+
+      // Set channel affinity in gRPC-GCP. This is a no-op for GAX.
+      context =
+          context.withCallOptions(
+              context
+                  .getCallOptions()
+                  .withOption(
+                      GcpManagedChannel.AFFINITY_KEY,
+                      Option.CHANNEL_HINT.getLong(options).toString()));
```
olavloite, line 1970: This channel hint can be any random Long, meaning that we might be using a very large range of different channel hints. That again means that we might be adding an unbounded set of channel hints to the map in grpc-gcp. We should add a computation here that makes sure that the set of hints is limited (e.g. do something like `hint % options.getNumChannels()`).

See https://github.com/googleapis/java-spanner/blob/e859b29ccf4e68b1ab62cffdd4cf197011ba9878/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java#L342

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -1950,7 +1953,17 @@ <ReqT, RespT> GrpcCallContext newCallContext(
       boolean routeToLeader) {
     GrpcCallContext context = GrpcCallContext.createDefault();
     if (options != null) {
+      // Set channel affinity in GAX. This is a no-op for gRPC-GCP.
       context = context.withChannelAffinity(Option.CHANNEL_HINT.getLong(options).intValue());
+
+      // Set channel affinity in gRPC-GCP. This is a no-op for GAX.
```
olavloite, line 1959: nit: is it possible to only do this if grpc-gcp is being used? I know that it is a no-op for gax, and so does not affect anything in that way, but `context.withCallOptions(..)` always creates a new `CallContext` instance. That means that we reserve a small bit of additional memory for every RPC that we invoke that then needs to be cleaned up again afterwards. (And yes, that's a bit of a pre-emptive optimization, so if it is not possible or hard, then please ignore this comment, but if there's a simple if-condition that we can use, then let's do that.)

olavloite (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
harshachinta: I have made changes to do it only when grpcgcp is enabled. Can you please take a look?

olavloite: LGTM

PR #3226, opened 2024-07-26
[test]: protobuf to 3.25.4 for v6.66.0

zhumin8:
test only, for lts7


zhumin8:
kokoro test: integration on cloud-devel fails with auth failure:
```
Caused by: com.google.auth.oauth2.GoogleAuthException: Error getting access token for service account: 400 Bad Request
POST https://oauth2.googleapis.com/token
{"error":"invalid_grant","error_description":"Invalid JWT Signature."}
```

PR #3229, opened 2024-07-30
feat: support multiplexed session for blind write with single use transaction

pratickchokhani:
With multiplexed sessions, the client optimizes and runs multiple applicable requests concurrently on a single session. A single multiplexed session is sufficient to handle all concurrent writeAtLeastOnce requests.

Normal read/write transactions always use regular sessions.

The use of multiplexed sessions can be enabled through the environment variable `GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS`.


google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClientImpl.java
```diff
@@ -112,6 +126,16 @@ public Timestamp writeAtLeastOnce(final Iterable<Mutation> mutations) throws Spa
...
       final Iterable<Mutation> mutations, final TransactionOption... options)
       throws SpannerException {
+    if (useMultiplexedSessionForBlindWrite) {
+      return getMultiplexedSession().writeAtLeastOnceWithOptions(mutations, options);
+    } else {
+      return writeAtLeastOnceWithSession(mutations, options);
+    }
+  }
+
+  public CommitResponse writeAtLeastOnceWithSession(
```
olavloite, line 136: Can this be private? Or at least package-private?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClientImpl.java
```diff
@@ -37,24 +37,38 @@ class DatabaseClientImpl implements DatabaseClient {
   @VisibleForTesting final SessionPool pool;
   @VisibleForTesting final MultiplexedSessionDatabaseClient multiplexedSessionDatabaseClient;
 
+  final boolean useMultiplexedSessionForBlindWrite;
```
olavloite, line 40: Do we need this to be a separate option? Or could we just say 'if multiplexed sessions are enabled, then we use it for both read-only transactions and single-write transactions'?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClientImpl.java
```diff
@@ -112,6 +126,16 @@ public Timestamp writeAtLeastOnce(final Iterable<Mutation> mutations) throws Spa
   public CommitResponse writeAtLeastOnceWithOptions(
       final Iterable<Mutation> mutations, final TransactionOption... options)
       throws SpannerException {
+    if (useMultiplexedSessionForBlindWrite) {
+      return getMultiplexedSession().writeAtLeastOnceWithOptions(mutations, options);
```
olavloite, line 130: I think that this will also use regular sessions if multiplexed sessions for read-only has not been enabled. Meaning that in the current setup:
1. It is impossible to enable multiplexed sessions for only blind writes.
2. You have to enable multiplexed sessions for read-only transactions in order to be able to enable if for blind writes.

See also my comment above; If possible, I would prefer just one setting that enables multiplexed sessions for everything that supports it.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DelayedMultiplexedSessionTransaction.java
```diff
@@ -119,4 +121,28 @@ public ReadOnlyTransaction readOnlyTransaction(TimestampBound bound) {
                     .readOnlyTransaction(bound),
             MoreExecutors.directExecutor()));
   }
+
+  @Override
+  public CommitResponse writeAtLeastOnceWithOptions(
```
olavloite, line 126: This method seems to be blocking. The methods in this class are intended to be non-blocking.

EDIT: When I come to think of it, we don't really need this one to be non-blocking, as the only way that this is invoked is through `writeAtLeastOnce(..)`. And that method again is blocking.

So would you mind:
1. Adding a comment to the method explaining why it is OK to implement this as a blocking method.
2. Maybe simplify the method a bit. We don't need the `ApiFutures.transform(...)` construction. Just call `this.sessionFuture.get()` at the start of the method, and then return `new MultiplexedSessionTransaction(..).writeAtLeastOnceWithOptions(...)` using the session reference that you get from the `this.sessionFuture.get()` call.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -307,6 +317,28 @@ public boolean getUseMultiplexedSession() {
     return useMultiplexedSession;
   }
 
+  @VisibleForTesting
+  @InternalApi
+  public boolean getUseMultiplexedSessionForBlindWrite() {
+    return useMultiplexedSessionForBlindWrite;
+  }
+
+  private static Boolean getUseMultiplexedSessionBlindWriteFromEnvVariable() {
```
olavloite, line 326: If we choose to keep this as a separate setting; Can we re-use the logic for getting a env-var that __must__ be either `true` or `false`?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClientMockServerTest.java
```diff
@@ -309,6 +317,118 @@ public void testMaintainerInvalidatesMultiplexedSessionClientIfUnimplemented() {
...
+        SimulatedExecutionTime.ofException(
+            mockSpanner.createAbortedException(ByteString.copyFromUtf8("test"))));
+    Timestamp timestamp =
+        client.writeAtLeastOnce(
+            Collections.singletonList(
+                Mutation.newInsertBuilder("FOO").set("ID").to(1L).set("NAME").to("Bar").build()));
+    assertNotNull(timestamp);
+
+    List<CommitRequest> commitRequests = mockSpanner.getRequestsOfType(CommitRequest.class);
+    assertEquals(2, commitRequests.size());
```
olavloite, line 335: These tests do not verify whether a multiplexed session was used or not. Please add that. See the other test cases for how you can check that.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DelayedMultiplexedSessionTransaction.java
pratickchokhani: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClientImpl.java
pratickchokhani: Have removed the flag.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClientImpl.java
pratickchokhani: Have removed this flag.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
pratickchokhani: Have removed this funtion.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClientImpl.java
pratickchokhani: Done.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClientMockServerTest.java
pratickchokhani: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -789,6 +809,22 @@ Builder setUseMultiplexedSession(boolean useMultiplexedSession) {
...
+     * Sets whether the client should use multiplexed session for blind write or not. If set to
+     * true, the client optimises and runs multiple applicable requests concurrently on a single
+     * session. A single multiplexed session is sufficient to handle all concurrent traffic.
+     *
+     * <p>When set to false, the client uses the regular session cached in the session pool for
+     * running 1 concurrent transaction per session. We require to provision sufficient sessions by
+     * making use of {@link SessionPoolOptions#minSessions} and {@link
+     * SessionPoolOptions#maxSessions} based on the traffic load. Failing to do so will result in
+     * higher latencies.
+     */
```
olavloite, line 822: I think that it is better to remove this whole block of documentation, and instead just mark this method as `@InternalApi` and `@VisibleForTesting`, and maybe also add one bit of Javadoc that says something like this:

```
This method enables multiplexed sessions for blind writes. This method will be removed in the future when multiplexed sessions has been made the default for all operations.
```

(The above to prevent users from asking why it is not public / how they can enable it / etc.)

google-cloud-spanner-executor/src/main/java/com/google/cloud/spanner/SessionPoolOptionsHelper.java
```diff
@@ -30,4 +30,12 @@ public static SessionPoolOptions.Builder setUseMultiplexedSession(
       SessionPoolOptions.Builder sessionPoolOptionsBuilder, boolean useMultiplexedSession) {
     return sessionPoolOptionsBuilder.setUseMultiplexedSession(useMultiplexedSession);
   }
+
+  // TODO: Remove when Builder.setUseMultiplexedSession(..) has been made public.
```
olavloite, line 34: nit: I don't think that we will ever make that method public. Instead, it is something that will be made the default in the future, and the whole method will be removed.

olavloite (approved):
LGTM, with some minor nits on documentation.

PR #3231, opened 2024-08-01
feat(spanner): support multiplexed session for Partitioned operations

pratickchokhani:
With multiplexed sessions, the client optimises and runs multiple applicable requests concurrently on a single session. A single multiplexed session is sufficient to handle all concurrent partitioned read, query and dml operations.

This can be enabled through an environment variable `GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS_PARTITIONED_OPS`.


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionClient.java
```diff
@@ -423,10 +414,16 @@ private List<SessionImpl> internalBatchCreateSessions(
 
   /** Returns a {@link SessionImpl} that references the existing session with the given name. */
   SessionImpl sessionWithId(String name) {
+    return sessionWithId(name, /*isMultiplexedSession= */ false);
+  }
+
+  /** Returns a {@link SessionImpl} that references the existing session with the given name. */
+  SessionImpl sessionWithId(String name, boolean isMultiplexedSession) {
```
olavloite, line 421: We should not add a boolean that indicates whether the session is a multiplexed session or not here, as the caller does not know. The session might have been created by a  different client, and this client does not know whether the other client created a regular or a multiplexed session.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BatchClientImpl.java
```diff
@@ -72,7 +79,9 @@ public BatchReadOnlyTransaction batchReadOnlyTransaction(TimestampBound bound) {
   @Override
   public BatchReadOnlyTransaction batchReadOnlyTransaction(BatchTransactionId batchTransactionId) {
     SessionImpl session =
-        sessionClient.sessionWithId(checkNotNull(batchTransactionId).getSessionId());
+        sessionClient.sessionWithId(
+            checkNotNull(batchTransactionId).getSessionId(),
+            batchTransactionId.isMultiplexedSession());
```
olavloite, line 84: We should not pass in multiplexed true/false here, because this client does not know anything about this session. The session might have been created by a different client and shared out-of-band with this client. The fact that this client uses multiplexed sessions does not mean that the other client uses multiplexed sessions (and vice-versa).

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BatchTransactionId.java
```diff
@@ -32,12 +32,18 @@ public class BatchTransactionId implements Serializable {
   private final ByteString transactionId;
   private final String sessionId;
   private final Timestamp timestamp;
+  private final boolean isMultiplexedSession;
```
olavloite, line 35: Do we need this flag for anything? Otherwise, I think we should remove it. See also my comment above; If a client receives a session ID from another client, then this client does not know whether that session ID is for a regular or a multiplexed session. If such an ID is then used to create a `BatchTransactionId`, then this could lead to a `BatchTransactionId` that contains 'false' information.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClientImpl.java
```diff
@@ -309,6 +314,14 @@ public AsyncTransactionManager transactionManagerAsync(TransactionOption... opti
 
   @Override
   public long executePartitionedUpdate(final Statement stmt, final UpdateOption... options) {
+    if (useMultiplexedSessionPartitionedOps) {
+      return getMultiplexedSession().executePartitionedUpdate(stmt, options);
+    }
+    return executePartitionedUpdateSession(stmt, options);
+  }
+
+  private long executePartitionedUpdateSession(
```
olavloite, line 323: nit:
```suggestion
  private long executePartitionedUpdateWithPooledSession(
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DelayedMultiplexedSessionTransaction.java
```diff
@@ -224,4 +225,22 @@ private SessionReference getSessionReference() {
...
+  /**
+   * Execute `stmt` within PARTITIONED_DML transaction using multiplexed session. This method is a
+   * blocking call as the interface expects to return the output of the `stmt`.
+   */
+  @Override
+  public long executePartitionedUpdate(Statement stmt, UpdateOption... options) {
+    try {
+      SessionReference sessionReference = this.sessionFuture.get();
+      return new MultiplexedSessionTransaction(
+              client, span, sessionReference, NO_CHANNEL_HINT, true)
```
olavloite, line 238:
```suggestion
              client, span, sessionReference, NO_CHANNEL_HINT, /* singleUse = */ true)
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DelayedMultiplexedSessionTransaction.java
```diff
@@ -224,4 +225,22 @@ private SessionReference getSessionReference() {
...
   }
+
+  /**
+   * Execute `stmt` within PARTITIONED_DML transaction using multiplexed session. This method is a
+   * blocking call as the interface expects to return the output of the `stmt`.
+   */
+  @Override
+  public long executePartitionedUpdate(Statement stmt, UpdateOption... options) {
+    try {
+      SessionReference sessionReference = this.sessionFuture.get();
```
olavloite, line 236: Use:

```suggestion
      SessionReference sessionReference = getSessionReference();
```

This automatically handles the exceptions, so you can also get rid of the throw-catch block.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClient.java
```diff
@@ -553,6 +554,11 @@ public AsyncTransactionManager transactionManagerAsync(TransactionOption... opti
         .transactionManagerAsync(options);
   }
 
+  @Override
+  public long executePartitionedUpdate(Statement stmt, UpdateOption... options) {
+    return createMultiplexedSessionTransaction(true).executePartitionedUpdate(stmt, options);
```
olavloite, line 559:
```suggestion
    return createMultiplexedSessionTransaction(/* singleUse = */ true).executePartitionedUpdate(stmt, options);
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionClient.java
```diff
@@ -237,6 +237,19 @@ SessionImpl createSession() {
...
+      SessionImpl sessionImpl = createMultiplexedSession();
+      consumer.onSessionReady(sessionImpl);
+    } catch (Throwable t) {
+      consumer.onSessionCreateFailure(t, 1);
+    }
+  }
+
+  /**
+   * Create a multiplexed session and returns it. A multiplexed session is not affiliated with any
+   * GRPC channel. In case of an error on the gRPC calls, the error will be returned.
```
olavloite, line 250:
```suggestion
   * Creates a multiplexed session and returns it. A multiplexed session is not affiliated with any
   * GRPC channel. In case of an error during the gRPC calls, an exception will be thrown.
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BatchClientImpl.java
```diff
@@ -30,16 +30,53 @@
...
+  private final Duration sessionExpirationDuration;
+
+  /** The expiration date/time of the current multiplexed session. */
+  @GuardedBy("multiplexedSessionLock")
+  private final AtomicReference<Instant> expirationDate;
+
+  @GuardedBy("multiplexedSessionLock")
+  private final AtomicReference<SessionImpl> multiplexedSessionReference;
+
+  private final Clock clock;
```
olavloite, line 62: nit: you can remove this, as this always uses the system clock.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BatchClientImpl.java
```diff
@@ -92,15 +134,31 @@ public BatchReadOnlyTransaction batchReadOnlyTransaction(BatchTransactionId batc
...
+      return this.multiplexedSessionReference.get();
+    } finally {
+      this.multiplexedSessionLock.unlock();
+    }
+  }
+
   private static class BatchReadOnlyTransactionImpl extends MultiUseReadOnlyTransaction
       implements BatchReadOnlyTransaction {
     private final String sessionName;
+    private final boolean isMultiplexedSession;
```
olavloite, line 154: nit: Remove this. It is not used for anything (and you could otherwise get it by just calling `session.getIsMultiplexed()`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BatchTransactionId.java
```diff
@@ -34,7 +34,11 @@ public class BatchTransactionId implements Serializable {
   private final Timestamp timestamp;
   private static final long serialVersionUID = 8067099123096783939L;
 
-  BatchTransactionId(String sessionId, ByteString transactionId, Timestamp timestamp) {
+  BatchTransactionId(
+      String sessionId,
+      ByteString transactionId,
+      Timestamp timestamp,
+      boolean isMultiplexedSession) {
```
olavloite, line 41: nit: remove, it is not used

olavloite (approved):

PR #3236, opened 2024-08-06
[DO NOT MERGE | POC] feat: Adding support for UUID type

gauravpurohit06:
feat: Adding support for uuid type.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/GrpcStruct.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/GrpcStruct.java
@@ -94,6 +94,9 @@ class GrpcStruct extends Struct implements Serializable {
         case PG_NUMERIC:
           builder.set(fieldName).to((String) value);
           break;
+        case UUID:
+          builder.set(fieldName).to((String) value);
+          break;
         case STRING:
           builder.set(fieldName).to((String) value);
           break;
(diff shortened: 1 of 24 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/GrpcStruct.java
```diff
@@ -94,6 +94,9 @@ private Object writeReplace() {
         case PG_NUMERIC:
           builder.set(fieldName).to((String) value);
           break;
+        case UUID:
+          builder.set(fieldName).to((String) value);
```
harshachinta, line 98: I think this `value` is from backend as string. Should we be using `fromString` to convert string value to java.util.UUID value? I don't think this is the right place to do this conversion. May be when customer reads it as UUID then we should use the fromString method.
https://docs.oracle.com/javase/8/docs/api/java/util/UUID.html

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Type.java
```diff
@@ -287,6 +296,7 @@ public enum Code {
     FLOAT64(TypeCode.FLOAT64, "double precision"),
     FLOAT32(TypeCode.FLOAT32, "real"),
     STRING(TypeCode.STRING, "character varying"),
+    UUID(TypeCode.STRING, "uuid"),
```
harshachinta, line 299: As per design doc a new TypeCode `UUID` will be added. This will recognize the input serialized string value as UUID in backend to store it appropriately.
```suggestion
    UUID(TypeCode.UUID, "uuid"),
```

olavloite:

PR #3263, opened 2024-08-07
chore: Re-run renovate-config-validator

lqiu96:


lqiu96:
This probably should be run in synthtool

PR #3278, opened 2024-08-18
fix: getRowAsStruct should always return a decoded struct

olavloite:
`getRowAsStruct` could return a (partially) undecoded struct, which would have an internally mutable data structure. The returned row should however preferably be an immutable object. To achieve this, the struct should be guaranteed to be decoded before returning it to the caller.


olavloite:
Closing as this is not necessary once the changes in #3267 have been merged.

PR #3287, opened 2024-08-21
docs: fix tracing sample to exit when completed, and use custom monitored resource for export

rahul2393:

### DIFF
--- a/samples/snippets/src/main/java/com/example/spanner/TracingSample.java
+++ b/samples/snippets/src/main/java/com/example/spanner/TracingSample.java
@@ -16,6 +16,8 @@
 
 package com.example.spanner;
 
+import com.google.api.MonitoredResource;
+import com.google.cloud.MetadataConfig;
 import com.google.cloud.spanner.DatabaseClient;
 import com.google.cloud.spanner.DatabaseId;
 import com.google.cloud.spanner.ResultSet;
@@ -23,6 +25,7 @@ import com.google.cloud.spanner.Spanner;
 import com.google.cloud.spanner.SpannerOptions;
 import com.google.cloud.spanner.Statement;
 import com.google.cloud.spanner.spi.v1.SpannerRpcViews;
+import io.opencensus.common.Duration;
 import io.opencensus.common.Scope;
 import io.opencensus.contrib.grpc.metrics.RpcViews;
 import io.opencensus.contrib.zpages.ZPageHandlers;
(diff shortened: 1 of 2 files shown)


rahul2393:
TracingSample was not using project_id set via GOOGLE_CLOUD_PROJECT and will use metadata API request to set project  on default MetricResource, which causes exceotion in exporting stats.
Reference
a) https://github.com/census-instrumentation/opencensus-java/blob/master/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExportUtils.java#L495-L497
b) https://github.com/googleapis/sdk-platform-java/blob/main/java-core/google-cloud-core/src/main/java/com/google/cloud/MetadataConfig.java#L42
c) https://github.com/googleapis/sdk-platform-java/blob/main/java-core/google-cloud-core/src/main/java/com/google/cloud/MetadataConfig.java#L71-L82

samples/snippets/src/main/java/com/example/spanner/TracingSample.java
```diff
@@ -87,6 +97,9 @@ public static void main(String[] args) throws Exception {
     } finally {
       // Closes the client which will free up the resources used
       spanner.close();
+      StackdriverExporter.unregister();
+      StackdriverStatsExporter.unregister();
+      System.exit(0);
```
olavloite, line 102: This seems overly harsh. What is the reason that we need to call `System.exit(0)` to stop the application here? Is the Stackdriver exporter using non-daemon threads that are not stopped? If so, does the `StackdriverExporter` have a close/shutdown method?

rahul2393: StackdriverExporter does not export any close/shutdown method they only have unregister, without system.exit I see exceptions and sample getting stuck https://pastebin.com/drh6qubH

samples/snippets/src/main/java/com/example/spanner/TracingSample.java
```diff
@@ -58,7 +64,13 @@ public static void main(String[] args) throws Exception {
         .registerSpanNamesForCollection(Arrays.asList(SAMPLE_SPAN));
 
     // Installs an exporter for stack driver stats.
-    StackdriverStatsExporter.createAndRegister();
+    MonitoredResource.Builder builder = MonitoredResource.newBuilder();
```
rahul2393, line 67: This is being added because without this an exception is thrown

```
Caused by: io.grpc.StatusRuntimeException: INVALID_ARGUMENT: Field timeSeries[0].resource.labels.project_id had an invalid value of "google.com:xxxxx: if present, must be the project number or ID in the request name.                                                      
        at io.grpc.Status.asRuntimeException(Status.java:533) 
        ```

samples/snippets/src/main/java/com/example/spanner/TracingSample.java
olavloite: The 'getting stuck' problem is actually caused by `ZPageHandlers.startHttpServerAndRegisterAll(8080);`. That starts an HTTP server that can be accessed on `http://localhost:8080/tracez` to view the traces that were collected. That component does not have a public `stop()` method, and keeps the JVM alive, as it uses non-daemon threads. Can we update the comment here to reflect this?

The errors regarding closed channels could in theory still happen. I think that there is a race condition in the OpenCensus stats exporter that tries to flush any remaining stats, while at the same time also closing the underlying exporter client. Those can be ignored.

olavloite (approved):

PR #3327, opened 2024-09-10
chore(spanner): add package protected setter to enable mux for RW

harshachinta:
This PR adds a package protected setter that can be used to enable/disable multiplexed sessions for R/W operations. This method will be used internally for testing purposes.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
@@ -73,6 +73,8 @@ public class SessionPoolOptions {
 
   private final boolean useMultiplexedSession;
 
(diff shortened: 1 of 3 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -757,6 +789,15 @@ Builder setUseMultiplexedSession(boolean useMultiplexedSession) {
       return this;
     }
 
+    /**
+     * Sets whether the client should use multiplexed session for R/W operations or not. This method
+     * is package-protected and used internally.
```
olavloite, line 794:
```suggestion
     * Sets whether the client should use multiplexed session for R/W operations or not. This method
     * is intentionally package-private and intended for internal use.
```

olavloite (approved):

PR #3348, opened 2024-09-25
chore(spanner): preserving lock order - R/W mux session

harshachinta:
This PR introduces changes to support the lock order preservation protocol for multiplexed sessions in read/write transactions. According to this protocol, when a read/write transaction on a multiplexed session is retried, the transaction ID from the previous abort must be passed when creating a new transaction. This ensures that the retried transaction is recognized as older, rather than being treated as a new one.

The transaction context object is structured as follows,
```
txn {
    transactionId: The transaction ID.
    previousTransactionId: The transaction ID of the most recently failed transaction.
}
```


google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
```diff
@@ -1079,7 +1092,7 @@ public TransactionRunner allowNestedTransaction() {
   TransactionRunnerImpl(SessionImpl session, TransactionOption... options) {
     this.session = session;
     this.options = Options.fromTransactionOptions(options);
-    this.txn = session.newTransaction(this.options);
+    this.txn = session.newTransaction(this.options, null);
```
olavloite, line 1095: nit (here and elsewhere): It is good practice to add a `/* prevTransactionId = */` style comment when you are passing in an argument like this that is otherwise hard to understand what it represents.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
```diff
@@ -1117,7 +1130,17 @@ private <T> T runInternal(final TransactionCallable<T> txCallable) {
             // Do not inline the BeginTransaction during a retry if the initial attempt did not
             // actually start a transaction.
             useInlinedBegin = txn.transactionId != null;
-            txn = session.newTransaction(options);
+
+            // Determine the latest transactionId when using a multiplexed session.
+            ByteString multiplexedSessionPreviousTransactionId = null;
+            if (session.getIsMultiplexed()) {
+              // Use the current transactionId if available, otherwise fallback to the previous
+              // transactionId.
```
olavloite, line 1138: Could you elaborate a bit further on this in the comment. If I understand it correctly, this is a fallback that if there is no current transactionId, then we'll use the prevTransactionId of the prev transaction. Basically meaning something like 'the previous transaction never got a transaction ID, so we'll use the one from before that'. Does this mean that this condition will kick in in the case that:
1. Transaction 1 is started and the aborted.
2. An attempt to retry with Transaction 2 is initiated, but somehow the BeginTransaction RPC or the first statement with an inline BeginTransaction option fails, meaning that Transaction 2 never gets an ID. It is then aborted/retried.
3. An attempt to retry with Transaction 3 is initiated. As Transaction 2 never got an ID, we use the prevTransactionId of Transaction 2, which is the ID of Transaction 1.

Is the above correct? And do we have a test that covers this scenario?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClientMockServerTest.java
```diff
@@ -467,6 +476,106 @@ public void testWriteAtLeastOnceWithExcludeTxnFromChangeStreams() {
...
+        assertTrue(mockSpanner.getSession(request.getSession()).getMultiplexed());
+      }
+
+      // Verify that the first request uses inline begin, and the previous transaction ID is set to ByteString.EMPTY
+      assertTrue(executeSqlRequests.get(0).hasTransaction());
+      assertTrue(executeSqlRequests.get(0).getTransaction().hasBegin());
+      assertTrue(executeSqlRequests.get(0).getTransaction().getBegin().hasReadWrite());
+      assertNotNull(executeSqlRequests.get(0).getTransaction().getBegin().getReadWrite()
+          .getMultiplexedSessionPreviousTransactionId());
+      assertThat(executeSqlRequests.get(0).getTransaction().getBegin().getReadWrite().getMultiplexedSessionPreviousTransactionId()).isEqualTo(ByteString.EMPTY);
```
olavloite, line 515: nit: Use `assertEquals`

google-cloud-spanner/src/test/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClientMockServerTest.java
```diff
@@ -467,6 +476,106 @@ public void testWriteAtLeastOnceWithExcludeTxnFromChangeStreams() {
...
+
+      // Verify that the second request uses inline begin, and the previous transaction ID is set appropriately
+      assertTrue(executeSqlRequests.get(1).hasTransaction());
+      assertTrue(executeSqlRequests.get(1).getTransaction().hasBegin());
+      assertTrue(executeSqlRequests.get(1).getTransaction().getBegin().hasReadWrite());
+      assertNotNull(executeSqlRequests.get(1).getTransaction().getBegin().getReadWrite()
+          .getMultiplexedSessionPreviousTransactionId());
+      assertThat(executeSqlRequests.get(1).getTransaction().getBegin().getReadWrite().getMultiplexedSessionPreviousTransactionId()).isNotEqualTo(ByteString.EMPTY);
```
olavloite, line 523: Can we modify this test so that it not only checks that it is not empty, but that it has exactly the prevTransactionId that we expect (so equal to the transaction ID of the previous attempt).

google-cloud-spanner/src/test/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClientMockServerTest.java
```diff
@@ -467,6 +478,188 @@ public void testWriteAtLeastOnceWithExcludeTxnFromChangeStreams() {
...
+            .getBegin()
+            .getReadWrite()
+            .getMultiplexedSessionPreviousTransactionId());
+  }
+*/
+  
+  // TODO(sriharshach): Uncomment test once Lock order preservation proto is published
+  /*
+  @Test
+  public void testAbortedReadWriteTxnUsesPreviousTxnIdOnRetryWithExplicitBegin()
```
harshachinta, line 575: Test Scenario

1. Transaction 1 is started and the **inline begin** fails due to invalid update statement. This leads to ABORTED exception and transaction is retried.
2. Transaction 2 is initiated with **explicit begin** transaction which succeeds and creates a transactionId (**txn2**). however transaction2 gets aborted during commit (The test forces the commit rpc to return ABORTED the first time it is called).
3. Transaction 3 is started but again **inline begin** fails. Here txn id is null, but prevTransactionId will be txn2.
4. Transaction 4 is started with **explicit begin**. As Transaction 3 never got an ID, the transaction id of Transaction 2 (txn2) is used in TransactionOptions.
5. 
This behavior is verified in test by checking that second BeginTransaction RPC carries the **txn2** as previous transaction id.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
harshachinta: Yes and this scenario is covered in the below mock spanner test. I have linked to the comment which explains the test scenario
https://github.com/googleapis/java-spanner/pull/3348/files#r1785744949

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionImpl.java
```diff
@@ -78,6 +79,11 @@ static TransactionOptions createReadWriteTransactionOptions(Options options) {
     if (options.withOptimisticLock() == Boolean.TRUE) {
       readWrite.setReadLockMode(TransactionOptions.ReadWrite.ReadLockMode.OPTIMISTIC);
     }
+    if (previousTransactionId != null
```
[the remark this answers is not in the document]
harshachinta: Updated prevTransactionId to have ByteString.Empty as default instead of null.
However, I have this null check in this if condition alone to ensure that we don't see NullPointerExceptions.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
harshachinta: Done. Thanks for the suggestion

google-cloud-spanner/src/test/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClientMockServerTest.java
harshachinta: Done

google-cloud-spanner/src/test/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClientMockServerTest.java
harshachinta: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncTransactionManagerImpl.java
```diff
@@ -77,7 +78,17 @@ public TransactionContextFutureImpl beginAsync() {
 
   private ApiFuture<TransactionContext> internalBeginAsync(boolean firstAttempt) {
     txnState = TransactionState.STARTED;
-    txn = session.newTransaction(options);
+
+    // Determine the latest transactionId when using a multiplexed session.
+    ByteString multiplexedSessionPreviousTransactionId = null;
+    if (session.getIsMultiplexed() && !firstAttempt) {
```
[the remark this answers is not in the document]
harshachinta: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
```diff
@@ -201,6 +209,8 @@ public void removeListener(Runnable listener) {
 
     volatile ByteString transactionId;
 
+    ByteString previousTransactionId;
```
olavloite, line 212: nit: can this be `final`?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
olavloite: Cool, thanks 👍

google-cloud-spanner/src/test/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClientMockServerTest.java
```diff
@@ -467,6 +479,183 @@ public void testWriteAtLeastOnceWithExcludeTxnFromChangeStreams() {
...
+  public void testAbortedReadWriteTxnUsesPreviousTxnIdOnRetryWithInlineBegin()
+      throws InterruptedException {
+    DatabaseClientImpl client =
+        (DatabaseClientImpl) spanner.getDatabaseClient(DatabaseId.of("p", "i", "d"));
+    // Force the Commit RPC to return Aborted the first time it is called. The exception is cleared
+    // after the first call, so the retry should succeed.
+    mockSpanner.setCommitExecutionTime(
+        SimulatedExecutionTime.ofException(
+            mockSpanner.createAbortedException(ByteString.copyFromUtf8("test"))));
+    Thread.sleep(10000);
```
olavloite, line 492: Please remove :-)

google-cloud-spanner/src/test/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClientMockServerTest.java
```diff
@@ -467,6 +479,183 @@ public void testWriteAtLeastOnceWithExcludeTxnFromChangeStreams() {
...
+  public void testAbortedReadWriteTxnUsesPreviousTxnIdOnRetryWithExplicitBegin()
+      throws InterruptedException {
+    DatabaseClientImpl client =
+        (DatabaseClientImpl) spanner.getDatabaseClient(DatabaseId.of("p", "i", "d"));
+    // Force the Commit RPC to return Aborted the first time it is called. The exception is cleared
+    // after the first call, so the retry should succeed.
+    mockSpanner.setCommitExecutionTime(
+        SimulatedExecutionTime.ofException(
+            mockSpanner.createAbortedException(ByteString.copyFromUtf8("test"))));
+    Thread.sleep(10000);
```
olavloite, line 580: Remove

olavloite (commented):
Looks generally good to me. Waiting with approving until we can run the tests and verify that those are all green.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
harshachinta: This field is modified in unit tests, so it cannot be declared as final

google-cloud-spanner/src/test/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClientMockServerTest.java
harshachinta: Done

google-cloud-spanner/src/test/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClientMockServerTest.java
harshachinta: Done

harshachinta:
> Looks generally good to me. Waiting with approving until we can run the tests and verify that those are all green.
@olavloite The checks are green. Can you please help review?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
olavloite: Fair point. But those unit tests are both mocking `TransactionContextImpl`. So could we then solve that like this:
1. Make this field final.
2. Add a package-private `getPreviousTransactionId()` that returns this field. Use this getter everywhere in this class where the field would otherwise be accessed.
3. Return a mocked result from that getter in the unit tests that need to have it return something else.

olavloite (approved):
LGTM, but with a minor request on making `previousTransactionId` final.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
harshachinta: Updated `previousTransactionId` field to be `final`.

PR #3380, opened 2024-10-04
chore: add directpath_enabled attribute

surbhigarg92:
As part of this [bug](https://github.com/googleapis/sdk-platform-java/pull/3170) , modification were made in ClientContext which can be used to fetch `isDirectPath()` attribute. 

Recommendation was to pass clientContext in the `GrpcSpannerStub.create` instead of SpannerStubSetting. 
```
SpannerStubSettings stubSettings = options
        .getSpannerStubSettings()
        .toBuilder()
        .setTransportChannelProvider(channelProvider)
        .setCredentialsProvider(credentialsProvider)
        .setStreamWatchdogProvider(watchdogProvider)
        .setTracerFactory(options.getApiTracerFactory())
        .build();
ClientContext clientContext = ClientContext.create(stubSettings);
// create GrpcSpannerStub from clientContext instead. 
// this is equivelant to GrpcSpannerStub.create(stubSettings)
this.spannerStub =
    GrpcSpannerStub.create(clientContext);
// access this property from transport channel.
((GrpcTransportChannel) clientContext.getTransportChannel()).isDirectPath();
```
It was working for the regular calls, however, all the test cases using timeouts were failing. 

Ideally this should have worked the similar way. However , the code [here](https://github.com/googleapis/java-spanner/blob/main/google-cloud-spanner/src/main/java/com/google/cloud/spanner/v1/stub/GrpcSpannerStub.java#L247) is not using the spannerStubSettings from clientContext, instead creating a new `SpannerStubSettings.newBuilder().build()` 

Added a new method in the `GrpcSpannerStub` file which accepts both stubSettings and clientContext.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
@@ -29,6 +29,7 @@ import com.google.api.gax.grpc.GaxGrpcProperties;
 import com.google.api.gax.grpc.GrpcCallContext;
 import com.google.api.gax.grpc.GrpcCallSettings;
 import com.google.api.gax.grpc.GrpcStubCallableFactory;
+import com.google.api.gax.grpc.GrpcTransportChannel;
 import com.google.api.gax.grpc.InstantiatingGrpcChannelProvider;
 import com.google.api.gax.longrunning.OperationFuture;
 import com.google.api.gax.retrying.ResultRetryAlgorithm;
@@ -63,6 +64,7 @@ import com.google.cloud.grpc.GcpManagedChannelOptions.GcpMetricsOptions;
 import com.google.cloud.grpc.GrpcTransportOptions;
 import com.google.cloud.spanner.AdminRequestsPerMinuteExceededException;
 import com.google.cloud.spanner.BackupId;
+import com.google.cloud.spanner.BuiltInMetricsConstant;
 import com.google.cloud.spanner.ErrorCode;
 import com.google.cloud.spanner.Restore;
 import com.google.cloud.spanner.SpannerException;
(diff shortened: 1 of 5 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/v1/stub/GrpcSpannerStub.java
```diff
@@ -247,6 +247,11 @@ public static final GrpcSpannerStub create(ClientContext clientContext) throws I
     return new GrpcSpannerStub(SpannerStubSettings.newBuilder().build(), clientContext);
   }
 
+  public static final GrpcSpannerStub create(
```
olavloite, line 250: This is a generated file, so you can't manually update it. (That is: you can update it, but the next round of code generation will overwrite your changes)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -396,18 +398,22 @@ public GapicSpannerRpc(final SpannerOptions options) {
...
+                .toBuilder()
+                .setTransportChannelProvider(channelProvider)
+                .setCredentialsProvider(credentialsProvider)
+                .setStreamWatchdogProvider(watchdogProvider)
+                .setTracerFactory(
+                    options.getApiTracerFactory(
+                        /* isAdminClient = */ false, isEmulatorEnabled(options, emulatorHost)))
+                .build();
+        ClientContext clientContext = ClientContext.create(spannerStubSettings);
+        this.spannerStub = GrpcSpannerStub.create(spannerStubSettings, clientContext);
```
olavloite, line 413: See below: This calls a method in a generated file, and this method is manually added to that generated file. That will be overwritten the next time the code generation is run, so this won't work.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -396,18 +398,22 @@ public GapicSpannerRpc(final SpannerOptions options) {
...
+        BuiltInMetricsConstant.DIRECT_PATH_ENABLED =
```
olavloite, line 414: A user could create multiple Spanner instances. Then this code will be invoked multiple times. There is no guarantee that `isDirectPath()` will return the same value each time, as it depends on (AFAIK) the credentials that are used. `DIRECT_PATH_ENABLED` is a public static variable that should preferably only be set once (or changed to an instance field of the `SpannerImpl` class or one of the other underlying classes)

PR #3411, opened 2024-10-15
chore(spanner): track precommit token for R/W multiplexed session

harshachinta:
When a read-write transaction is executed on a multiplexed session, the RPC responses of that transaction return a `MultiplexedSessionPrecommitToken`. In client library, the precommit token with the highest sequence number is tracked at the transaction context level. During the commit, this latest precommit token is fetched and set in the CommitRequest. If the precommit token is not set during the commit, the backend will throw an `INVALID_ARGUMENT` error.

Including the latest token in the CommitRequest is essential to prevent latency regression, though it does not impact the correctness of the transaction.

This PR tracks the precommit token from the following RPC responses,
1. ResultSet
2. PartialResultSet
3. ExecuteBatchDmlResponse


google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
```diff
@@ -625,6 +635,24 @@ public void onTransactionMetadata(Transaction transaction, boolean shouldInclude
...
     }
 
+    /**
+     * In read-write transactions, the precommit token with the highest sequence number from this
+     * transaction attempt will be tracked and included in the
+     * [Commit][google.spanner.v1.Spanner.Commit] request for the transaction.
+     */
+    @Override
+    public void onPrecommitToken(MultiplexedSessionPrecommitToken token) {
+      if (token == null) return;
```
olavloite, line 645: nit: always use curly braces for `if` statements. See https://google.github.io/styleguide/javaguide.html#s4.1-braces

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
```diff
@@ -625,6 +635,24 @@ public void onTransactionMetadata(Transaction transaction, boolean shouldInclude
...
+      synchronized (precommitTokenLock) {
+        if (this.latestPrecommitToken == null
+            || token.getSeqNum() > this.latestPrecommitToken.getSeqNum()) {
+          this.latestPrecommitToken = token;
+          System.out.println("Updating precommit token to " + this.latestPrecommitToken);
```
olavloite, line 650: Remove

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
```diff
@@ -625,6 +635,24 @@ public void onTransactionMetadata(Transaction transaction, boolean shouldInclude
...
+          txnLogger.log(Level.ALL, "Updating precommit token to " + this.latestPrecommitToken);
```
olavloite, line 651: I think that you want log level FINE for this.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
```diff
@@ -940,6 +981,11 @@ public long[] batchUpdate(Iterable<Statement> statements, UpdateOption... update
             }
           }
 
+          // TODO(sriharshach): check if we need to get precommit_token from response.getResultSets
```
olavloite, line 984: Is this TODO still pending?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/MockSpannerServiceImpl.java
```diff
@@ -1080,6 +1094,9 @@ private void returnResultSet(
       metadata = metadata.toBuilder().setTransaction(transaction).build();
     }
     resultSet = resultSet.toBuilder().setMetadata(metadata).build();
+    if (session.getMultiplexed()) {
+      resultSet = resultSet.toBuilder().setPrecommitToken(getResultSetPrecommitToken()).build();
+    }
```
olavloite, line 1099: nit: could we maybe combine these so that we don't execute `toBuilder().build()` too many times?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/MockSpannerServiceImpl.java
```diff
@@ -2447,4 +2471,23 @@ Session getSession(String name) {
     }
     return null;
   }
+
+  static MultiplexedSessionPrecommitToken getResultSetPrecommitToken() {
+    return getPrecommitToken("ResultSetPrecommitToken", 1);
```
olavloite, line 2476: The sequence numbers should be more intelligent than this. This just assumes that RPCs will be called in a given order.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
harshachinta: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
harshachinta: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
harshachinta: Yeah. Thanks for pointing this

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
harshachinta: No. This is resolved.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/MockSpannerServiceImpl.java
harshachinta: Done

google-cloud-spanner/src/test/java/com/google/cloud/spanner/MockSpannerServiceImpl.java
harshachinta: The reason for hardcoding the sequence number is to maintain greater control in our tests and to minimize flakiness. This approach does not assume that RPCs will be called in a strict order. For instance, consider the following RPC response order in a read-write transaction:
```
ResultSet -> sequence number 1
PartialResultSet -> sequence number 3
ExecuteBatchDml -> sequence number 2
```
Even though the ExecuteBatchDml RPC was executed later, we know that the PartialResultSet, with sequence number 3, will have the latest precommit token. This ensures that we are properly validating the sequence number logic

However, I agree that there is a flaw in this approach. If the same order of RPCs is executed on the backend, the ExecuteBatchDml would have the latest sequence number since it was the last RPC executed in the transaction. In contrast, our mock Spanner does not adhere to this behavior, which causes the discrepancy.

**Other options**.
1. **Incremental counter** - We can use AtomicInteger to generate incrementing sequence numbers in a thread-safe manner. This ensures that the seq number is assigned based on when the request reaches mock spanner. This also ensure seq number is always incremental and will not lead to flakiness.
2. **Time-Based Increments** - We can use `Instant.now()` as sequence number. However the seq number in this case will not always be incremental (2 RPCs can have same seq number) leading to test flakiness.

**I have modified code to use incremental counter.**

However, please let me know if there are other better ways to solve this.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/MockSpannerServiceImpl.java
```diff
@@ -2447,4 +2489,32 @@ Session getSession(String name) {
...
+  }
+
+  static MultiplexedSessionPrecommitToken getPrecommitToken(
+      String value, ByteString transactionId) {
+    if (!transactionSequenceNo.containsKey(transactionId)) {
+      transactionSequenceNo.put(transactionId, new AtomicInteger(0));
+    }
+
+    // Generates an incrementing sequence number
+    int seqNum = transactionSequenceNo.get(transactionId).addAndGet(1);
```
olavloite, line 2514: nit: you can simplify this by calling `transactionSequenceNo.putIfAbsent(..)`

olavloite: another tiny nit: use `AtomicInteger#incrementAndGet()` instead of `AtomicInteger#addAndGet(int)`

olavloite (approved):

google-cloud-spanner/src/test/java/com/google/cloud/spanner/MockSpannerServiceImpl.java
harshachinta: Thanks. Done

PR #3416, opened 2024-10-18
feat: Adds support for Interval datatype in Java client

sagarwaal:
Feat: Adds support for the INTERVAL datatype.


google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITIntervalTest.java
```diff
@@ -0,0 +1,214 @@
...
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+@Category(ParallelIntegrationTest.class)
+@RunWith(Parameterized.class)
+public class ITIntervalTest {
```
rahul2393, line 46: Looks good, Can we add tests using comparison operators with Intervals 
e.g crated_time > time+ Interval $1?

sagarwaal: added

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Interval.java
```diff
@@ -0,0 +1,316 @@
...
+  public static final BigInteger NANOS_PER_MINUTE =
+      BigInteger.valueOf(MICROS_PER_MINUTE * NANOS_PER_MICRO);
+  public static final BigInteger NANOS_PER_HOUR =
+      BigInteger.valueOf(MICROS_PER_HOUR * NANOS_PER_MICRO);
+
+  private static final Pattern INTERVAL_PATTERN =
+      Pattern.compile(
+          "^P(?!$)(-?\\d+Y)?(-?\\d+M)?(-?\\d+D)?(T(?=-?.?\\d)(-?\\d+H)?(-?\\d+M)?(-?((\\d+(\\.\\d{1,9})?)|(\\.\\d{1,9}))S)?)?$");
+
+  /** Returns the months component of the interval. */
```
olavloite, line 56: Would you mind adding the range of possible values that Spanner could return for each of these fields?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Interval.java
```diff
@@ -0,0 +1,316 @@
...
+  public abstract int months();
+
+  /** Returns the days component of the interval. */
+  public abstract int days();
+
+  /** Returns the microseconds component of the interval. */
+  public abstract long micros();
+
+  /** Returns the nanoFractions component of the interval. */
```
olavloite, line 65: I assume that this is the nano fractions of the microseconds part? So it has a value in the range [0, 1000).

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Interval.java
```diff
@@ -0,0 +1,316 @@
...
+  /** Returns the nanoseconds component of the interval. */
+  public BigInteger nanos() {
+    return BigInteger.valueOf(micros())
+        .multiply(BigInteger.valueOf(NANOS_PER_MICRO))
+        .add(BigInteger.valueOf(nanoFractions()));
+  }
+
+  /** Returns the total microseconds represented by the interval. */
+  public long getAsMicros() {
+    return months() * MICROS_PER_MONTH + days() * MICROS_PER_DAY + micros();
```
olavloite, line 81: Is this guaranteed not to overflow for any value that is returned by Spanner? If not, could we either:
1. Document what the behavior is (e.g. it silently overflows and returns gibberish)
2. OR check whether it would overflow and then return an error?

(Note: Java just silently overflows/underflows if you do nothing)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Interval.java
```diff
@@ -0,0 +1,316 @@
...
+  /** Returns the total nanoseconds represented by the interval. */
+  public BigInteger getAsNanos() {
+    return BigInteger.valueOf(getAsMicros())
+        .multiply(BigInteger.valueOf(NANOS_PER_MICRO))
+        .add(BigInteger.valueOf(nanoFractions()));
+  }
+
+  /** Creates an interval with specified number of months. */
+  public static Interval ofMonths(int months) {
+    return builder().setMonths(months).setDays(0).setMicros(0).setNanoFractions((short) 0).build();
```
olavloite, line 93: nit (here and elsewhere): just omit all the `setDays(0)`, `setMicros(0)` etc. Those are the defaults anyway.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Interval.java
```diff
@@ -0,0 +1,316 @@
...
+  public static Interval ofDays(int days) {
+    return builder().setMonths(0).setDays(days).setMicros(0).setNanoFractions((short) 0).build();
+  }
+
+  /** Creates an interval with specified number of seconds. */
+  public static Interval ofSeconds(long seconds) {
+    return builder()
+        .setMonths(0)
+        .setDays(0)
+        .setMicros(seconds * MICROS_PER_SECOND)
```
olavloite, line 106: We should either document what happens if you specify a value for seconds that will make this overflow, or check it and return an error.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Interval.java
```diff
@@ -0,0 +1,316 @@
...
+      BigDecimal seconds = new BigDecimal(nanos).movePointLeft(9);
+
+      if (seconds.compareTo(new BigDecimal(zero)) != 0) {
+        result.append(String.format("%sS", seconds.stripTrailingZeros()));
+      }
+    }
+
+    if (result.length() == 1) {
+      result.append("0Y");
+    }
```
olavloite, line 250: nit: would it not be more readable to have a simple `if (this.equals(ZERO))` at the top of this method, and a `public static final Interval ZERO = Interval.builder().build()` constant at the top of this file?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Interval.java
```diff
@@ -0,0 +1,316 @@
...
+
+    return result.toString();
+  }
+
+  /** Creates an interval which representing 0-duration. */
+  public static Interval zeroInterval() {
+    return builder().setMonths(0).setDays(0).setMicros(0).setNanoFractions((short) 0).build();
```
olavloite, line 257: nit: simplify to `public static final Interval ZERO = Interval.builder().build()`. Intervals are immutable, so it is perfectly safe to have a public constant for the zero value.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Interval.java
```diff
@@ -0,0 +1,316 @@
...
+  @Override
+  public boolean equals(Object rhs) {
+    if (!(rhs instanceof Interval)) {
+      return false;
+    }
+
+    Interval anotherInterval = (Interval) rhs;
+    return months() == anotherInterval.months()
+        && days() == anotherInterval.days()
+        && nanos().equals(anotherInterval.nanos());
```
olavloite, line 269: This appears to be missing a `micros` comparison. Which again also automatically leads me to believe that there must be some tests missing, as that would otherwise also have come up as test failures.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Interval.java
```diff
@@ -0,0 +1,316 @@
...
+  }
+
+  @Override
+  public int hashCode() {
+    int result = 17;
+    result = 31 * result + Long.valueOf(months()).hashCode();
+    result = 31 * result + Long.valueOf(days()).hashCode();
+    result = 31 * result + nanos().hashCode();
```
olavloite, line 277: Same here: missing micros

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Interval.java
```diff
@@ -0,0 +1,316 @@
...
+    return Interval.builder()
+        .setMonths(Math.toIntExact(totalMonths))
+        .setDays(Math.toIntExact(days))
+        .setMicros(totalMicros.longValueExact())
+        .setNanoFractions(nanoFractions.shortValueExact())
+        .build();
+  }
+
+  /** Converts Interval to ISO8601 Duration Formatted String. */
+  public String ToISO8601() {
```
olavloite, line 204: nit: 
```suggestion
  public String toISO8601() {
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Interval.java
```diff
@@ -0,0 +1,316 @@
...
+      }
+    }
+
+    if (result.length() == 1) {
+      result.append("0Y");
+    }
+
+    return result.toString();
+  }
+
```
olavloite, line 254: nit: also add a `toString()` method:

```suggestion
@Override
public String toString() {
  return toISO8601();
}
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Value.java
```diff
@@ -1795,6 +1839,29 @@ void valueToString(StringBuilder b) {
...
+    private IntervalImpl(boolean isNull, Interval value) {
+      super(isNull, Type.interval(), value);
+    }
+
+    @Override
+    public Interval getInterval() {
+      checkNotNull();
+      return value;
+    }
+
```
olavloite, line 1853: What is the default implementation for `getAsString()`? Does that return what we would want for `Interval`? Otherwise, we should add an implementation for that as well here.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Value.java
```diff
@@ -2797,6 +2864,29 @@ void appendElement(StringBuilder b, Date element) {
...
+    @Override
+    void appendElement(StringBuilder b, Interval element) {
+      b.append(element.ToISO8601());
+    }
+
+    @Override
+    String elementToString(Interval element) {
+      return element.ToISO8601();
```
olavloite, line 2886: Is this guaranteed to be called only for non-null elements? Otherwise, this will throw a `NullPointerException` if it is called on an array element that is null.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/IntervalTest.java
```diff
@@ -0,0 +1,284 @@
+/*
+ * Copyright 2021 Google LLC
```
olavloite, line 2:
```suggestion
 * Copyright 2024 Google LLC
```

google-cloud-spanner/src/test/java/com/google/cloud/spanner/IntervalTest.java
```diff
@@ -0,0 +1,284 @@
...
+    Interval interval2 = Interval.fromMonthsDaysMicros(10, 20, 30);
+    assertEquals(interval1, interval2);
+  }
+
+  @Test
+  public void testHashCode() {
+    Interval interval1 = Interval.fromMonthsDaysMicros(10, 20, 30);
+    Interval interval2 = Interval.fromMonthsDaysMicros(10, 20, 30);
+    assertEquals(interval1.hashCode(), interval2.hashCode());
+  }
```
olavloite, line 186: These should cover more variations, and also include verifications for non-equality

google-cloud-spanner/src/test/java/com/google/cloud/spanner/RandomResultSetGenerator.java
```diff
@@ -142,6 +147,16 @@ private void setRandomValue(Value.Builder builder, Type type) {
...
+        case INTERVAL:
+          Interval interval =
+              Interval.builder()
+                  .setMonths(10)
+                  .setDays(45)
+                  .setMicros(-89)
+                  .setNanoFractions((short) 45)
+                  .build();
+          builder.setStringValue(interval.ToISO8601());
+          break;
```
olavloite, line 159: This is supposed to generate a random value that varies per test run.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ChecksumResultSetTest.java
```diff
@@ -25,18 +25,11 @@
 import com.google.cloud.ByteArray;
 import com.google.cloud.Date;
 import com.google.cloud.Timestamp;
-import com.google.cloud.spanner.AbortedDueToConcurrentModificationException;
-import com.google.cloud.spanner.AbortedException;
-import com.google.cloud.spanner.ResultSet;
-import com.google.cloud.spanner.ResultSets;
+import com.google.cloud.spanner.*;
```
olavloite, line 28: nit: Google code style forbids the use of wildcard imports (https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports)

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITQueryTest.java
```diff
@@ -30,22 +30,9 @@
...
 import com.google.cloud.Date;
 import com.google.cloud.Timestamp;
-import com.google.cloud.spanner.Database;
-import com.google.cloud.spanner.DatabaseClient;
-import com.google.cloud.spanner.Dialect;
-import com.google.cloud.spanner.ErrorCode;
-import com.google.cloud.spanner.IntegrationTestEnv;
-import com.google.cloud.spanner.Mutation;
-import com.google.cloud.spanner.ParallelIntegrationTest;
+import com.google.cloud.spanner.*;
```
olavloite, line 33: Here also: no wildcard imports

google-cloud-spanner/pom.xml
```diff
@@ -455,6 +460,12 @@
       <artifactId>opentelemetry-sdk-testing</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.jetbrains</groupId>
+      <artifactId>annotations</artifactId>
+      <version>24.1.0</version>
+      <scope>compile</scope>
```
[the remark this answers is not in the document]
sagarwaal: local builds are failing without this dependency.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Interval.java
```diff
@@ -0,0 +1,316 @@
...
+  public static Builder builder() {
+    return new AutoValue_Interval.Builder();
+  }
+
+  /** Returns the nanoseconds component of the interval. */
+  public BigInteger nanos() {
+    return BigInteger.valueOf(micros())
+        .multiply(BigInteger.valueOf(NANOS_PER_MICRO))
+        .add(BigInteger.valueOf(nanoFractions()));
+  }
```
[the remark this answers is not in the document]
sagarwaal: nanoFractions() function returns fractional part of microseconds which is in range [-999, 999].
nanos() returns microsecond and nanoFractions() combined.
getAsNanos() returns all the components of Interval - months, days and nanos() combined as nanoseconds.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Interval.java
sagarwaal: that was because of AutoValue, it will throw illegal state exception as all the params of the class are not provided. I'll refactor code to not use AutoValue dependency.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Interval.java
sagarwaal: nanos() here convert microseconds and nanoFractions to nanoseconds and comparing them.

I am doing this because I am allowing nanoFractions to be in range of [-999, 999].

google-cloud-spanner/src/main/java/com/google/cloud/spanner/GrpcStruct.java
```diff
@@ -503,6 +513,11 @@ protected Date getDateInternal(int columnIndex) {
     return (Date) rowData.get(columnIndex);
   }
 
+  protected Interval getIntervalInternal(int columnIndex) {
```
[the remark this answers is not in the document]
sagarwaal: done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Interval.java
sagarwaal: added

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Interval.java
sagarwaal: I am keeping the range of nanoFractions from [-999, 999]. Let me know your thoughts on this.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Interval.java
sagarwaal: added.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Value.java
sagarwaal: added default implementation for `getAsString`, it also returns ISO8601 formatted Interval

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Interval.java
```diff
@@ -0,0 +1,316 @@
+/*
+ * Copyright 2021 Google LLC
```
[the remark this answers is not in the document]
sagarwaal: done.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/IntervalTest.java
sagarwaal: done.

google-cloud-spanner/pom.xml
```diff
@@ -364,6 +364,11 @@
       <version>${graal-sdk.version}</version>
       <scope>provided</scope>
     </dependency>
+    <dependency>
```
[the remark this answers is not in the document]
sagarwaal: Removed dependency of auto-value.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/IntervalTest.java
sagarwaal: Updated tests for Equality and HashCode.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/RandomResultSetGenerator.java
sagarwaal: changed.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITQueryTest.java
sagarwaal: fixed.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ChecksumResultSetTest.java
sagarwaal: fixed.

rahul2393 (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Interval.java
```diff
@@ -0,0 +1,277 @@
...
+  /** Creates an interval with specified number of microseconds. */
+  public static Interval ofMicroseconds(long micros) {
+    return builder()
+        .setNanoseconds(
+            BigInteger.valueOf(micros).multiply(BigInteger.valueOf(NANOS_PER_MICROSECOND)))
+        .build();
+  }
+
+  /** Creates an interval with specified number of nanoseconds. */
+  public static Interval ofNanoseconds(@NotNull BigInteger nanos) {
```
sakthivelmanii, line 122: Can we use one naming everywhere? For example, either NanoSeconds or Nanos? I can see both the usages. 

It would be preferred to use Nanos, Micros, Millis everywere since java.time.Duration also uses same naming convention. 

@olavloite FYI

sagarwaal: changed it to nanos, and made similar changes at other places.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Interval.java
```diff
@@ -0,0 +1,277 @@
...
+  /** Converts Interval to ISO8601 Duration Formatted String. */
+  public String toISO8601() {
+    if (this.equals(ZERO)) {
+      return "P0Y";
+    }
+
+    StringBuilder result = new StringBuilder();
+    result.append("P");
+
+    long months_part = this.getMonths();
```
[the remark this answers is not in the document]
sagarwaal: fixed.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Interval.java
```diff
@@ -0,0 +1,275 @@
...
+  /** Creates an interval with specified number of microseconds. */
+  public static Interval ofMicros(long micros) {
+    return builder()
+        .setNanos(BigInteger.valueOf(micros).multiply(BigInteger.valueOf(NANOS_PER_MICRO)))
+        .build();
+  }
+
+  /** Creates an interval with specified number of nanoseconds. */
+  public static Interval ofNanos(@NotNull BigInteger nanos) {
+    return builder().setNanos(nanos).build();
```
sakthivelmanii, line 120: Can't we do `Preconditions.checkNotNull(nanos);`?

google-cloud-spanner/pom.xml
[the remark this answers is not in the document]
sakthivelmanii: Can't we use predicate instead of @NotNull annotation?


https://github.com/googleapis/java-spanner/pull/3416/files#r2017843792

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Interval.java
sagarwaal: moved not null check to Interval builder, and used `Preconditions.checkNotNull(nanos)`

sakthivelmanii (approved):

PR #3423, opened 2024-10-23
chore(spanner): support mutation only operation for read-write mux

harshachinta:
The following describes the case when a read-write transaction is executed on a multiplexed session and this PR handles this scenario:

If a read-write transaction contains only mutations, a random mutation is selected from the mutation list and sent with the BeginTransactionRequest. The resulting Transaction response includes a precommit token, which is tracked by the client library and used in the subsequent CommitRequest.


google-cloud-spanner/src/main/java/com/google/cloud/spanner/Mutation.java
```diff
@@ -402,20 +404,41 @@ private boolean isFloat32NaN(Value value) {
...
   }
 
-  static void toProto(Iterable<Mutation> mutations, List<com.google.spanner.v1.Mutation> out) {
+  // Converts Mutation and returns a random Mutation from the available mutation list based on the
+  // following heuristics:
+  // 1. Prefer mutations other than INSERT, since INSERT mutations may contain autogenerated columns
+  //    whose information is unavailable on the client.
+  // 2. If the list only contains INSERT mutations, select the one with the highest number of
+  // values.
+  static com.google.spanner.v1.Mutation toProtoGetRandomMutation(
```
olavloite, line 413: nit:
```suggestion
  static com.google.spanner.v1.Mutation toProtoAndReturnRandomMutation(
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Mutation.java
```diff
@@ -402,20 +404,41 @@ private boolean isFloat32NaN(Value value) {
     return value.getType().equals(Type.float32()) && Float.isNaN(value.getFloat32());
   }
 
-  static void toProto(Iterable<Mutation> mutations, List<com.google.spanner.v1.Mutation> out) {
+  // Converts Mutation and returns a random Mutation from the available mutation list based on the
```
olavloite, line 407:
```suggestion
  // Converts the list of mutations to the corresponding protobuf mutations and returns a random mutation from the available mutation list based on the
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Mutation.java
```diff
@@ -402,20 +404,41 @@ private boolean isFloat32NaN(Value value) {
...
+  static com.google.spanner.v1.Mutation toProtoGetRandomMutation(
+      Iterable<Mutation> mutations, List<com.google.spanner.v1.Mutation> out) {
     Mutation last = null;
     // The mutation currently being built.
     com.google.spanner.v1.Mutation.Builder proto = null;
     // The "write" (!= DELETE) or "keySet" (==DELETE) for the last mutation encoded, for coalescing.
     com.google.spanner.v1.Mutation.Write.Builder write = null;
     com.google.spanner.v1.KeySet.Builder keySet = null;
+
+    // Store all the mutations excluding INSERT.
```
olavloite, line 422:
```suggestion
    // Stores all the mutations excluding INSERT mutations.
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Mutation.java
```diff
@@ -402,20 +404,41 @@ private boolean isFloat32NaN(Value value) {
...
+    List<com.google.spanner.v1.Mutation> allMutationsExcludingInsert = new ArrayList<>();
+    // Stores INSERT mutation with large number of values.
```
olavloite, line 424:
```suggestion
    // Stores the INSERT mutation with largest number of values.
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Mutation.java
```diff
@@ -402,20 +404,41 @@ private boolean isFloat32NaN(Value value) {
...
+    com.google.spanner.v1.Mutation largeInsertMutation =
```
olavloite, line 425:
```suggestion
    com.google.spanner.v1.Mutation largestInsertMutation =
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Mutation.java
```diff
@@ -402,20 +404,41 @@ private boolean isFloat32NaN(Value value) {
...
           if (proto != null) {
-            out.add(proto.build());
+            com.google.spanner.v1.Mutation builtMutation = proto.build();
+            out.add(builtMutation);
+            if (checkIfInsertMutationWithLargeValue(builtMutation, largeInsertMutation)) {
+              largeInsertMutation = builtMutation;
+            }
+            if (!builtMutation.hasInsert()) {
+              allMutationsExcludingInsert.add(builtMutation);
+            }
```
olavloite, line 441: You can skip most of this, as the if statement on line 429 already guarantees that this is a DELETE mutation. So you can just directly add it to allMutationsExcludingInsert.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Mutation.java
```diff
@@ -464,7 +494,32 @@ static void toProto(Iterable<Mutation> mutations, List<com.google.spanner.v1.Mut
...
     // Flush last item.
     if (proto != null) {
+      com.google.spanner.v1.Mutation builtMutation = proto.build();
       out.add(proto.build());
+      if (checkIfInsertMutationWithLargeValue(builtMutation, largeInsertMutation)) {
+        largeInsertMutation = builtMutation;
+      }
+      if (!builtMutation.hasInsert()) {
+        allMutationsExcludingInsert.add(builtMutation);
+      }
```
olavloite, line 504: This code block is repeated a couple of times. Could we extract that to a separate method?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Mutation.java
```diff
@@ -464,7 +494,32 @@ static void toProto(Iterable<Mutation> mutations, List<com.google.spanner.v1.Mut
...
     }
+
+    // Select a random mutation based on the heuristic.
+    if (allMutationsExcludingInsert.size() > 0) {
```
olavloite, line 508:
```suggestion
    if (!allMutationsExcludingInsert.isEmpty()) {
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Mutation.java
```diff
@@ -464,7 +494,32 @@ static void toProto(Iterable<Mutation> mutations, List<com.google.spanner.v1.Mut
...
+      int randomIndex = ThreadLocalRandom.current().nextInt(allMutationsExcludingInsert.size());
+      return allMutationsExcludingInsert.get(randomIndex);
```
olavloite, line 510: nit: combine this into one line (meaning; skip the declaration of `int randomIndex`)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Mutation.java
olavloite: Also: Method comments should be Javadoc, so change this to something like this:

```
/**
 * Converts the list of mutations to ...
*/
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Mutation.java
olavloite: One very peculiar, but not impossible, case is that the list of mutations only contains empty insert mutations. This could for example be possible for a table that has a default value for the primary key (or allows null for the primary key), and also either has a default value for all other columns, or allows them to be null. Such a case would lead to this method returning `Mutation.getDefaultInstance()` as the largest insert mutation.

The heuristics should therefore also be changed to prefer an insert mutation with zero columns and a table name, over an insert mutation with zero columns and no table name. (Any other non-empty property of a real mutation can also be used instead of the table name for this check).

Can we also add a test for this specific scenario? It is typically something that could also backslide in the future.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Mutation.java
harshachinta: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Mutation.java
harshachinta: done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Mutation.java
harshachinta: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Mutation.java
harshachinta: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Mutation.java
harshachinta: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Mutation.java
harshachinta: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Mutation.java
harshachinta: Thanks Knut for pointing out to this edge case.

I have updated the code to handle this case by adding below lines, and added a test for empty insert mutation scenario.
```
    // If largestInsertMutation is a default instance of Mutation, replace it with the current
    // INSERT mutation, even if it contains zero values.
    if (!largestInsertMutation.hasInsert()) {
      return true;
    }
```

Talking out loud of all the edge cases,
1. If mutations are empty, then this logic never gets called.
2. Table name in a mutation can not be empty.
3. If there are mutations other than INSERT operation, then once of them is chosen at random.
4. If only INSERT mutations, then one with largest value is chosen
5. If only INSERT mutation, with no columns/values set, then this mutation will be chosen as random mutation

Please let me know if there is anything missing.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Mutation.java
harshachinta: We have a coalescing mechanism to group similar consecutive mutations together.
As a result of this logic, the if statement ensures that the current mutation selected from the list is a DELETE mutation. However, the proto being built is based on prior iterations and will not necessarily be a DELETE mutation.

For example, consider the following mutation list:
```
Mutation.newInsertBuilder("T1").set("C").to("V1").build(),
Mutation.delete("T1", Key.of("k1")),
Mutation.newInsertBuilder("T2").set("C").to("V1").build()
```
In this case, the if condition will match the DELETE operation on line 2, but the proto will be an INSERT mutation from line 1.
We have a test case `toProtoCoalescingDeleteChanges` to verify this behavior.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Mutation.java
harshachinta: I have extracted them to seperate methods. Please let me know if there are any other better ways.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Mutation.java
```diff
@@ -402,20 +404,43 @@ private boolean isFloat32NaN(Value value) {
     return value.getType().equals(Type.float32()) && Float.isNaN(value.getFloat32());
   }
 
-  static void toProto(Iterable<Mutation> mutations, List<com.google.spanner.v1.Mutation> out) {
+  /**
+   * Converts the list of mutations to the corresponding protobuf mutations and returns a random
+   * mutation from the available list based on the following heuristics:
+   *
+   * <p>1. Prefer mutations other than INSERT, as INSERT mutations may contain autogenerated columns
```
olavloite, line 411: nit: (feel free to leave as is, as this is not a public method, but for future reference) when making a summation in Javadoc, then you should do this:

```
/**
 * <ol>
 *   <li>Item 1</li>
 *   <li>Item 2</li>
 * </ol>
 */
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Mutation.java
```diff
@@ -402,20 +404,43 @@ private boolean isFloat32NaN(Value value) {
...
           mutation.keySet.appendToProto(keySet);
         } else {
           if (proto != null) {
-            out.add(proto.build());
+            com.google.spanner.v1.Mutation builtMutation = proto.build();
+            out.add(builtMutation);
+            if (checkIfInsertMutationWithLargeValue(builtMutation, largestInsertMutation)) {
+              largestInsertMutation = builtMutation;
+            }
+            addMutationsExcludingInsert(builtMutation, allMutationsExcludingInsert);
```
olavloite, line 443: nit: name this method `maybeAddMutationToListExludingInserts(...)` (or something similar). The current naming seems to indicate that it is added unconditionally.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Mutation.java
```diff
@@ -464,7 +494,44 @@ static void toProto(Iterable<Mutation> mutations, List<com.google.spanner.v1.Mut
...
+  }
+
+  // Returns true if the input mutation is of type INSERT and has more values than the current
+  // largest insert mutation.
+  private static boolean checkIfInsertMutationWithLargeValue(
+      com.google.spanner.v1.Mutation mutation,
+      com.google.spanner.v1.Mutation largestInsertMutation) {
+    // If largestInsertMutation is a default instance of Mutation, replace it with the current
+    // INSERT mutation, even if it contains zero values.
+    if (!largestInsertMutation.hasInsert()) {
```
olavloite, line 521: I think that this should be:

```suggestion
    if (mutation.hasInsert() && !largestInsertMutation.hasInsert()) {
```

Please add a test that fails with the current implementation and succeeds with the above change (unless I missed something here, and the current implementation is correct).

google-cloud-spanner/src/test/java/com/google/cloud/spanner/MutationTest.java
```diff
@@ -493,6 +521,23 @@ public void toProtoCoalescingDeleteChanges() {
...
+  @Test
+  public void toProtoWithEmptyInsertMutations() {
+    List<Mutation> mutations =
+        Arrays.asList(
+            Mutation.newInsertBuilder("T").build(), Mutation.newInsertBuilder("A").build());
+
+    List<com.google.spanner.v1.Mutation> proto = new ArrayList<>();
+    com.google.spanner.v1.Mutation mutation =
+        Mutation.toProtoAndReturnRandomMutation(mutations, proto);
+    System.out.println(mutation);
```
olavloite, line 533: remove

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Mutation.java
olavloite: I just realized that this won't fail, as there are only two possibilities:
1. There are only insert mutations. Then the largest insert mutation will be used.
2. There is at least one non-insert mutation. Although it is possible that that mutation is set as 'the largest insert mutation' with the current if statement, it would not really make any difference, as the largest insert mutation won't be used anyways.

We should still fix the above if condition, though, as the current implementation is a bit confusing.

There is however another (small) optimization possible here; Once `allMutationsExcludingInserts.isEmpty()` returns false, you can skip the tracking of the largest insert mutation all-together, as it won't be used anyways.

olavloite (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Mutation.java
harshachinta: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Mutation.java
harshachinta: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Mutation.java
harshachinta: 1. Thanks for pointing this out. It does not affect the result, but it is indeed a flaw. Updated the code.
2. Added the optimization.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/MutationTest.java
harshachinta: Done

PR #3441, opened 2024-10-30
chore(spanner): handle server side kill switch of R/W multiplexed sessions

harshachinta:
The following information apply only if:
1. The application using the Java client library enables read-write multiplexed sessions via an environment variable.
2. The JDBC driver enables multiplexed sessions for read-write transactions by default.

On the backend, a server-side switch controls whether read-write transactions on multiplexed sessions are enabled. If this switch is disabled, all read-write transactions on multiplexed sessions will fail with UNIMPLEMENTED errors. This scenario is handled in the Java Spanner client as follows:

1. **When the server-side switch is OFF before application start:**
a. When the application starts, the Spanner client library asynchronously creates a multiplexed session.
b. After creating the multiplexed session, an asynchronous BeginTransaction RPC is explicitly invoked on a read-write transaction. If this call fails with an UNIMPLEMENTED error, the client disables read-write transactions on multiplexed sessions.
c. From this point, all new read-write transactions will fall back to regular sessions. However, any read-write transactions that started before this process completes will fail with an UNIMPLEMENTED error.

2. **When the server-side switch is turned OFF at time t1 after the application has started at t0 (where t0 < t1):**
a. If a read-write transaction fails with an UNIMPLEMENTED error, the Spanner client records this information and fails the transaction.
b. All subsequent read-write transactions then automatically fall back to regular sessions.

**Note: This does not impact the behavior of read-only transactions on multiplexed sessions.**

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClient.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClient.java
@@ -30,6 +30,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.util.concurrent.MoreExecutors;
 import com.google.spanner.v1.BeginTransactionRequest;
+import com.google.spanner.v1.RequestOptions;
 import com.google.spanner.v1.Transaction;
 import java.time.Clock;
 import java.time.Duration;
@@ -313,7 +314,9 @@ final class MultiplexedSessionDatabaseClient extends AbstractMultiplexedSessionD
             .setSession(sessionName)
             .setOptions(
                 SessionImpl.createReadWriteTransactionOptions(
-                    Options.fromTransactionOptions(), /* previousTransactionId = */ null));
+                    Options.fromTransactionOptions(), /* previousTransactionId = */ null))
+            .setRequestOptions(
+                RequestOptions.newBuilder().setTransactionTag("multiplexed-rw-background-begin-txn").build());
     final BeginTransactionRequest request = requestBuilder.build();
     final ApiFuture<Transaction> requestFuture;
     requestFuture =
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClient.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClient.java
@@ -17,6 +17,7 @@
 package com.google.cloud.spanner;
 
 import static com.google.cloud.spanner.SessionImpl.NO_CHANNEL_HINT;
+import static com.google.cloud.spanner.SpannerExceptionFactory.newSpannerException;
 
 import com.google.api.core.ApiFuture;
 import com.google.api.core.ApiFutures;
@@ -27,6 +28,9 @@ import com.google.cloud.spanner.SessionClient.SessionConsumer;
 import com.google.cloud.spanner.SpannerException.ResourceNotFoundException;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
+import com.google.common.util.concurrent.MoreExecutors;
+import com.google.spanner.v1.BeginTransactionRequest;
+import com.google.spanner.v1.Transaction;
 import java.time.Clock;
 import java.time.Duration;
 import java.time.Instant;
@@ -92,6 +96,9 @@ final class MultiplexedSessionDatabaseClient extends AbstractMultiplexedSessionD
         // synchronizing, as it does not really matter exactly which error is set.
         this.client.resourceNotFoundException.set((ResourceNotFoundException) spannerException);
       }
+      // Mark multiplexed sessions for RW as unimplemented and fall back to regular sessions if
+      // UNIMPLEMENTED is returned.
+      this.client.maybeMarkUnimplementedForRW(spannerException);
     }
 
     @Override
@@ -164,6 +171,13 @@ final class MultiplexedSessionDatabaseClient extends AbstractMultiplexedSessionD
   /** The current multiplexed session that is used by this client. */
   private final AtomicReference<ApiFuture<SessionReference>> multiplexedSessionReference;
 
+  @VisibleForTesting
+  /**
+   * The Transaction response returned by the BeginTransaction request with read-write when a
+   * multiplexed session is created during client initialization.
+   */
+  private final SettableApiFuture<Transaction> readWriteBeginTransactionReferenceFuture;
+
   /** The expiration date/time of the current multiplexed session. */
   private final AtomicReference<Instant> expirationDate;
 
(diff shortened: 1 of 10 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClient.java
```diff
@@ -164,6 +172,13 @@ public void close() {
   /** The current multiplexed session that is used by this client. */
   private final AtomicReference<ApiFuture<SessionReference>> multiplexedSessionReference;
 
+  @VisibleForTesting
```
olavloite, line 175: nit: the variable is `private`, so it is not visible (also not for testing), so this annotation should be removed

google-cloud-spanner/src/main/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClient.java
```diff
@@ -267,6 +299,59 @@ private void maybeMarkUnimplemented(Throwable t) {
...
+  private void maybeMarkUnimplementedForRW(SpannerException spannerException) {
+    if (spannerException.getErrorCode() == ErrorCode.UNIMPLEMENTED
+    // && spannerException.getCause().getMessage().contains("Transaction type read_write not
+    // supported with multiplexed sessions")
+    ) {
+      unimplementedForRW.set(true);
+    }
+  }
+
+  private void verifyBeginTransactionWithRWOnMultiplexedSession(String sessionName) {
```
olavloite, line 311: nit: rename this to the following to indicate that it is non-blocking

```suggestion
  private ApiFuture<Transaction> verifyBeginTransactionWithRWOnMultiplexedSessionAsync(String sessionName) {
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClient.java
```diff
@@ -267,6 +299,59 @@ private void maybeMarkUnimplemented(Throwable t) {
     }
   }
 
+  private void maybeMarkUnimplementedForRW(SpannerException spannerException) {
+    if (spannerException.getErrorCode() == ErrorCode.UNIMPLEMENTED
+    // && spannerException.getCause().getMessage().contains("Transaction type read_write not
+    // supported with multiplexed sessions")
```
olavloite, line 305: nit: either remove commented code or re-enable it (see also comment above)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClient.java
```diff
@@ -92,6 +97,9 @@ void onError(SpannerException spannerException) {
         // synchronizing, as it does not really matter exactly which error is set.
         this.client.resourceNotFoundException.set((ResourceNotFoundException) spannerException);
       }
+      // Mark multiplexed sessions for RW as unimplemented and fall back to regular sessions if
+      // UNIMPLEMENTED is returned.
+      this.client.maybeMarkUnimplementedForRW(spannerException);
```
olavloite, line 102: I think that this will cause any UNIMPLEMENTED error to disable multiplexed sessions for read/write transactions, as the implementation of the method just checks for an `UNIMPLEMENTED` error code, and this `onError(..)` method is used for all types of transactions.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/MockSpannerServiceImpl.java
```diff
@@ -1872,7 +1872,16 @@ private Transaction getTemporaryTransactionOrNull(TransactionSelector tx) {
   @Override
   public void beginTransaction(
       BeginTransactionRequest request, StreamObserver<Transaction> responseObserver) {
-    requests.add(request);
+    // Skip storing the explicit BeginTransactionRequest used to verify read-write transaction
```
olavloite, line 1875: nit: add a TODO here so it is easier to find

google-cloud-spanner/src/test/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClientMockServerTest.java
```diff
@@ -1290,6 +1291,231 @@ public void testMutationOnlyUsingAsyncTransactionManager() {
...
+    assertNotNull(session2);
+    assertFalse(session2.getMultiplexed());
+
+    assertNotNull(client.multiplexedSessionDatabaseClient);
+    assertEquals(1L, client.multiplexedSessionDatabaseClient.getNumSessionsAcquired().get());
+    assertEquals(1L, client.multiplexedSessionDatabaseClient.getNumSessionsReleased().get());
+  }
+
+  @Test
+  public void testReadWriteUnimplemented_firstReceivesError_secondFallsBackToRegularSessions() {
```
olavloite, line 1439: Can we add a test that also verifies that multiplexed sessions continues to work for read/write transactions after some other operation (e.g. a query using a single-use read-only transaction) returned UNIMPLEMENTED?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClient.java
harshachinta: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClient.java
harshachinta: Done, I have updated the method name.
However, I have still kept the return type to be `void` since it is not being used.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClient.java
harshachinta: Added error message verification.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClient.java
harshachinta: Yes this is indeed a problem. Added error message verification to avoid this situation

google-cloud-spanner/src/test/java/com/google/cloud/spanner/MockSpannerServiceImpl.java
harshachinta: Done

google-cloud-spanner/src/test/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClientMockServerTest.java
harshachinta: Added

olavloite (approved):

PR #3446, opened 2024-11-03
fix(spanner): Avoid blocking thread in AsyncResultSet

sakthivelmanii:

**Fixes** [b/372282080](https://b.corp.google.com/issues/372282080)☕️

**How are we solving?**
1. We noticed that unless we receive the chunk with resume token, we won't call callback method
2. Based on that assumption, we will initiate the stream as soon as callback is registered and release the thread back to the pool
3. Once we received resume token or reached end of stream or reached maximum buffer(depends on prefetchChunks configuration), we will produce rows by calling callback method
4. With this approach, thread will not be waiting till we receive N chunks(N - prefetch chunks configured) so that it can be used by another ResultSet

**Performance Testing Results:**  [result](https://docs.google.com/spreadsheets/d/1WTSZXOE75mI6_s1tp4YuYongHZgAEaxpoQEScZ95DdQ/edit?gid=247862237#gid=247862237)

### DIFF
--- a/google-cloud-spanner/clirr-ignored-differences.xml
+++ b/google-cloud-spanner/clirr-ignored-differences.xml
@@ -758,7 +758,7 @@
     <className>com/google/cloud/spanner/connection/Connection</className>
     <method>boolean isKeepTransactionAlive()</method>
   </difference>
-  
+
   <!-- Automatic DML batching -->
   <difference>
     <differenceType>7012</differenceType>
@@ -790,9 +790,5 @@
     <className>com/google/cloud/spanner/connection/Connection</className>
     <method>boolean isAutoBatchDmlUpdateCountVerification()</method>
   </difference>
-  <difference>
-    <differenceType>7012</differenceType>
-    <className>com/google/cloud/spanner/ResultSet</className>
-    <method>boolean initiateStreaming(com.google.cloud.spanner.AsyncResultSet$StreamMessageListener)</method>
-  </difference>
+
 </differences>
--- a/google-cloud-spanner/clirr-ignored-differences.xml
+++ b/google-cloud-spanner/clirr-ignored-differences.xml
@@ -758,7 +758,7 @@
     <className>com/google/cloud/spanner/connection/Connection</className>
     <method>boolean isKeepTransactionAlive()</method>
   </difference>
-
+  
   <!-- Automatic DML batching -->
   <difference>
     <differenceType>7012</differenceType>
@@ -790,5 +790,5 @@
     <className>com/google/cloud/spanner/connection/Connection</className>
     <method>boolean isAutoBatchDmlUpdateCountVerification()</method>
   </difference>
-
+  
 </differences>
--- a/google-cloud-spanner/clirr-ignored-differences.xml
+++ b/google-cloud-spanner/clirr-ignored-differences.xml
@@ -790,5 +790,9 @@
     <className>com/google/cloud/spanner/connection/Connection</className>
     <method>boolean isAutoBatchDmlUpdateCountVerification()</method>
   </difference>
-  
+  <difference>
+    <differenceType>7012</differenceType>
+    <className>com/google/cloud/spanner/ResultSet</className>
+    <method>boolean initiateStreaming(com.google.cloud.spanner.AsyncResultSet$StreamMessageListener)</method>
+  </difference>
 </differences>
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
@@ -150,6 +150,10 @@ abstract class AbstractResultSet<R> extends AbstractStructReader implements Resu
     void close(@Nullable String message);
 
     boolean isWithBeginTransaction();
+
+    default boolean initiateStreaming(AsyncResultSet.StreamMessageListener streamMessageListener) {
+      return false;
+    }
   }
 
   static double valueProtoToFloat64(com.google.protobuf.Value proto) {
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
@@ -151,6 +151,10 @@ abstract class AbstractResultSet<R> extends AbstractStructReader implements Resu
 
     boolean isWithBeginTransaction();
 
+    /**
+     * @param streamMessageListener A class object which implements StreamMessageListener
+     * @return true if streaming is supported by the iterator, otherwise false
+     */
     default boolean initiateStreaming(AsyncResultSet.StreamMessageListener streamMessageListener) {
       return false;
     }
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java
@@ -34,7 +34,12 @@ import com.google.spanner.v1.ResultSetStats;
 import java.util.Collection;
 import java.util.LinkedList;
 import java.util.List;
-import java.util.concurrent.*;
+import java.util.concurrent.BlockingDeque;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Executor;
+import java.util.concurrent.Future;
+import java.util.concurrent.LinkedBlockingDeque;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
@@ -46,7 +51,7 @@ class AsyncResultSetImpl extends ForwardingStructReader
   /** State of an {@link AsyncResultSetImpl}. */
   private enum State {
     INITIALIZED,
-    IN_PROGRESS,
+    STREAMING_INITIALIZED,
     /** SYNC indicates that the {@link ResultSet} is used in sync pattern. */
     SYNC,
     CONSUMING,
@@ -400,7 +405,7 @@ class AsyncResultSetImpl extends ForwardingStructReader
             synchronized (monitor) {
               stop = cursorReturnedDoneOrException;
             }
-          } catch (InterruptedException e) {
+          } catch (Throwable e) {
             result.setException(e);
             return;
           }
@@ -415,15 +420,13 @@ class AsyncResultSetImpl extends ForwardingStructReader
         synchronized (monitor) {
           if (executionException != null) {
             result.setException(executionException);
-            throw executionException;
-          }
-          if (state == State.CANCELLED) {
+          } else if (state == State.CANCELLED) {
             result.setException(CANCELLED_EXCEPTION);
-            throw CANCELLED_EXCEPTION;
+          } else {
+            result.set(null);
           }
         }
       }
-      result.set(null);
     }
 
     private void waitIfPaused() throws InterruptedException {
@@ -460,6 +463,12 @@ class AsyncResultSetImpl extends ForwardingStructReader
     @Override
     public void run() {
       try {
+        // This method returns true if the underlying result set is a streaming result set (e.g. a
+        // GrpcResultSet).
+        // Those result sets will trigger initiateProduceRows() when the first results are received.
+        // Non-streaming result sets do not trigger this callback, and for those result sets, we
+        // need to eagerly
+        // start the ProduceRowsRunnable.
         if (!initiateStreaming(AsyncResultSetImpl.this)) {
           initiateProduceRows();
         }
@@ -480,7 +489,7 @@ class AsyncResultSetImpl extends ForwardingStructReader
 
       // Start to fetch data and buffer these.
       this.result = SettableApiFuture.create();
-      this.state = State.IN_PROGRESS;
+      this.state = State.STREAMING_INITIALIZED;
       this.service.execute(new InitiateStreamingRunnable());
       this.executor = MoreExecutors.newSequentialExecutor(Preconditions.checkNotNull(exec));
       this.callback = Preconditions.checkNotNull(cb);
@@ -491,7 +500,7 @@ class AsyncResultSetImpl extends ForwardingStructReader
 
   private void initiateProduceRows() {
     this.service.execute(new ProduceRowsRunnable());
-    if (this.state == State.IN_PROGRESS) {
+    if (this.state == State.STREAMING_INITIALIZED) {
       this.state = State.RUNNING;
     }
     produceRowsInitiated = true;
@@ -607,9 +616,8 @@ class AsyncResultSetImpl extends ForwardingStructReader
     return delegateResultSet.get().getMetadata();
   }
 
-  @Override
   public boolean initiateStreaming(StreamMessageListener streamMessageListener) {
-    return delegateResultSet.get().initiateStreaming(streamMessageListener);
+    return StreamingUtil.initiateStreaming(delegateResultSet.get(), streamMessageListener);
   }
 
   @Override
@@ -631,20 +639,19 @@ class AsyncResultSetImpl extends ForwardingStructReader
   @Override
   public void onStreamMessage(
       PartialResultSet partialResultSet,
-      int prefetchChunks,
-      int currentBufferSize,
+      boolean bufferIsFull,
       StreamMessageRequestor streamMessageRequestor) {
     synchronized (monitor) {
       if (produceRowsInitiated) {
         return;
       }
-      // if PartialResultSet contains resume token or buffer size is more than configured size or
+      // if PartialResultSet contains resume token or buffer size is ful or
       // we have reached end of stream, we can start the thread
       boolean startJobThread =
           !partialResultSet.getResumeToken().isEmpty()
-              || currentBufferSize >= prefetchChunks
+              || bufferIsFull
               || partialResultSet == GrpcStreamIterator.END_OF_STREAM;
-      if (startJobThread || state != State.IN_PROGRESS) {
+      if (startJobThread || state != State.STREAMING_INITIALIZED) {
         initiateProduceRows();
       } else {
         streamMessageRequestor.requestMessages(1);
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java
@@ -18,7 +18,6 @@ package com.google.cloud.spanner;
 
 import com.google.api.core.ApiFuture;
 import com.google.api.core.ApiFutures;
-import com.google.api.core.ListenableFutureToApiFuture;
 import com.google.api.core.SettableApiFuture;
 import com.google.api.gax.core.ExecutorProvider;
 import com.google.cloud.spanner.AbstractReadContext.ListenableAsyncResultSet;
@@ -29,28 +28,25 @@ import com.google.common.base.Suppliers;
 import com.google.common.collect.ImmutableList;
 import com.google.common.util.concurrent.ListeningScheduledExecutorService;
 import com.google.common.util.concurrent.MoreExecutors;
+import com.google.spanner.v1.PartialResultSet;
 import com.google.spanner.v1.ResultSetMetadata;
 import com.google.spanner.v1.ResultSetStats;
 import java.util.Collection;
 import java.util.LinkedList;
 import java.util.List;
-import java.util.concurrent.BlockingDeque;
-import java.util.concurrent.Callable;
-import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.Executor;
-import java.util.concurrent.Future;
-import java.util.concurrent.LinkedBlockingDeque;
+import java.util.concurrent.*;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
 /** Default implementation for {@link AsyncResultSet}. */
-class AsyncResultSetImpl extends ForwardingStructReader implements ListenableAsyncResultSet {
+class AsyncResultSetImpl extends ForwardingStructReader
+    implements ListenableAsyncResultSet, AsyncResultSet.StreamMessageListener {
   private static final Logger log = Logger.getLogger(AsyncResultSetImpl.class.getName());
 
   /** State of an {@link AsyncResultSetImpl}. */
   private enum State {
     INITIALIZED,
+    IN_PROGRESS,
     /** SYNC indicates that the {@link ResultSet} is used in sync pattern. */
     SYNC,
     CONSUMING,
@@ -115,12 +111,15 @@ class AsyncResultSetImpl extends ForwardingStructReader implements ListenableAsy
 
   private State state = State.INITIALIZED;
 
+  /** This variable indicates that produce rows thread is initiated */
+  private volatile boolean produceRowsInitiated;
+
   /**
    * This variable indicates whether all the results from the underlying result set have been read.
    */
   private volatile boolean finished;
 
-  private volatile ApiFuture<Void> result;
+  private volatile SettableApiFuture<Void> result;
 
   /**
    * This variable indicates whether {@link #tryNext()} has returned {@link CursorState#DONE} or a
@@ -329,12 +328,12 @@ class AsyncResultSetImpl extends ForwardingStructReader implements ListenableAsy
   private final CallbackRunnable callbackRunnable = new CallbackRunnable();
 
   /**
-   * {@link ProduceRowsCallable} reads data from the underlying {@link ResultSet}, places these in
+   * {@link ProduceRowsRunnable} reads data from the underlying {@link ResultSet}, places these in
    * the buffer and dispatches the {@link CallbackRunnable} when data is ready to be consumed.
    */
-  private class ProduceRowsCallable implements Callable<Void> {
+  private class ProduceRowsRunnable implements Runnable {
     @Override
-    public Void call() throws Exception {
+    public void run() {
       boolean stop = false;
       boolean hasNext = false;
       try {
@@ -393,12 +392,17 @@ class AsyncResultSetImpl extends ForwardingStructReader implements ListenableAsy
         }
         // Call the callback if there are still rows in the buffer that need to be processed.
         while (!stop) {
-          waitIfPaused();
-          startCallbackIfNecessary();
-          // Make sure we wait until the callback runner has actually finished.
-          consumingLatch.await();
-          synchronized (monitor) {
-            stop = cursorReturnedDoneOrException;
+          try {
+            waitIfPaused();
+            startCallbackIfNecessary();
+            // Make sure we wait until the callback runner has actually finished.
+            consumingLatch.await();
+            synchronized (monitor) {
+              stop = cursorReturnedDoneOrException;
+            }
+          } catch (InterruptedException e) {
+            result.setException(e);
+            return;
           }
         }
       } finally {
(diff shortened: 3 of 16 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java
```diff
@@ -593,4 +627,28 @@ public Struct getCurrentRowAsStruct() {
...
+      // if PartialResultSet contains resume token or buffer size is more than configured size or
+      // we have reached end of stream, we can start the thread
+      boolean startJobThread =
+          !partialResultSet.getResumeToken().isEmpty()
+              || currentBufferSize >= prefetchChunks
+              || partialResultSet == GrpcStreamIterator.END_OF_STREAM;
+      if (startJobThread || state != State.IN_PROGRESS) {
+        initiateProduceRows();
+      } else {
+        streamMessageRequestor.requestMessages(1);
```
olavloite, line 650: Should we really do this? If I understand the code correctly, then the following will happen in most cases:
1. The stream is started with prefetchChunks=4. This means that we have already requested 4 messages.
2. The first time that this callback is triggered, the chunk is likely to have an empty resume token. In addition, the buffer is not full, and it is not the END_OF_STREAM. So this means that we request one more message (although we already had requested 4, and the buffer size is 4)
3. Step 2 repeats for chunk 2 and 3. Meaning that we have now requested in total 3 more chunks in addition to the first 4.
4. The fourth chunk contains a resume token and the ProduceRowsRunnable starts. This again will start removing data from the buffer, and each time a chunk is removed from the gRPC buffer, another message will be requested. Meaning that we continuously have requested 3 chunks more than that the buffer can contain.

Or am I misunderstanding how this will behave?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/GrpcStreamIterator.java
```diff
@@ -57,15 +60,20 @@ class GrpcStreamIterator extends AbstractIterator<PartialResultSet>
   GrpcStreamIterator(
       Statement statement, int prefetchChunks, boolean cancelQueryWhenClientIsClosed) {
     this.statement = statement;
+    this.prefetchChunks = prefetchChunks;
     this.consumer = new ConsumerImpl(cancelQueryWhenClientIsClosed);
     // One extra to allow for END_OF_STREAM message.
-    this.stream = new LinkedBlockingQueue<>(prefetchChunks + 1);
+    this.stream = new LinkedBlockingQueue<>((prefetchChunks * 2) + 1);
```
olavloite, line 66: I don't think we should increase the buffer size for this change (although now that I see this, I also better understand why the above implementation of `onStreamMessage` works)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -288,6 +288,35 @@ public boolean next() throws SpannerException {
...
 
+        @Override
+        public boolean initiateStreaming(
+            AsyncResultSet.StreamMessageListener streamMessageListener) {
+          try {
+            boolean ret = super.initiateStreaming(streamMessageListener);
+            if (beforeFirst) {
+              synchronized (lock) {
+                session.get().markUsed();
+                beforeFirst = false;
```
olavloite, line 299: I don't think that this is correct. `beforeFirst` indicates whether the current cursor of the result set is positioned before the first row. Starting the stream would not move the cursor to the first row.

olavloite: Also, `session.get().markUsed()` and `sessionUsedForQuery` are set to true here, also if `super.initiateStreaming()` returned false. Is that correct? If it returned false, then that is an indication that the result set was not a streaming result set, which again indicated that the call was a no-op, right?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/AsyncResultSetImplTest.java
```diff
@@ -22,8 +22,8 @@
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertThrows;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.when;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.*;
```
olavloite, line 26: nit: no wildcard imports

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java
```diff
@@ -458,16 +479,24 @@ public ApiFuture<Void> setCallback(Executor exec, ReadyCallback cb) {
...
       this.callback = Preconditions.checkNotNull(cb);
-      this.state = State.RUNNING;
       pausedLatch.countDown();
       return result;
     }
   }
 
+  private void initiateProduceRows() {
+    this.service.execute(new ProduceRowsRunnable());
+    if (this.state == State.IN_PROGRESS) {
```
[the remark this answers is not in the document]
sakthivelmanii: This is to avoid to overriding of state. For example, resultset.cancel can be called before `initiateProduceRows` called. In such cases, this will not override.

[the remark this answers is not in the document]
sakthivelmanii: We don't need it as `initiateProduceRows` is being called from inside of `synchronized (monitor)` in `onStreamMessage`

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java
sakthivelmanii: You are right. This is just an improvement to have some extra chunks in the buffer.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/ResultSet.java
```diff
@@ -82,4 +82,12 @@ public interface ResultSet extends AutoCloseable, StructReader {
   default ResultSetMetadata getMetadata() {
     throw new UnsupportedOperationException("Method should be overridden");
   }
+
+  /**
+   * Returns the {@link boolean} for this {@link ResultSet}. This method will be used by
+   * AsyncResultSet to initiate gRPC streaming
+   */
+  default boolean initiateStreaming(AsyncResultSet.StreamMessageListener streamMessageListener) {
```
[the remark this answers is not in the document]
sakthivelmanii: Fixed.

google-cloud-spanner/clirr-ignored-differences.xml
```diff
@@ -790,5 +790,9 @@
     <className>com/google/cloud/spanner/connection/Connection</className>
     <method>boolean isAutoBatchDmlUpdateCountVerification()</method>
   </difference>
-  
+  <difference>
```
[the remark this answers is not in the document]
sakthivelmanii: Fixed.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java
```diff
@@ -29,28 +28,25 @@
...
 import java.util.LinkedList;
 import java.util.List;
-import java.util.concurrent.BlockingDeque;
-import java.util.concurrent.Callable;
-import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.Executor;
-import java.util.concurrent.Future;
-import java.util.concurrent.LinkedBlockingDeque;
+import java.util.concurrent.*;
```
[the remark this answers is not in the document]
sakthivelmanii: Fixed.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
sakthivelmanii: Fixed,

google-cloud-spanner/src/test/java/com/google/cloud/spanner/AsyncResultSetImplTest.java
sakthivelmanii: Fixed

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java
```diff
@@ -449,6 +455,21 @@ private void startCallbackWithBufferLatchIfNecessary(int bufferLatch) {
     }
   }
 
+  private class InitiateStreamingRunnable implements Runnable {
+
+    @Override
+    public void run() {
+      try {
+        if (!initiateStreaming(AsyncResultSetImpl.this)) {
```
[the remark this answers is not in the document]
sakthivelmanii: Fixed

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java
```diff
@@ -29,28 +28,25 @@
...
 /** Default implementation for {@link AsyncResultSet}. */
-class AsyncResultSetImpl extends ForwardingStructReader implements ListenableAsyncResultSet {
+class AsyncResultSetImpl extends ForwardingStructReader
+    implements ListenableAsyncResultSet, AsyncResultSet.StreamMessageListener {
   private static final Logger log = Logger.getLogger(AsyncResultSetImpl.class.getName());
 
   /** State of an {@link AsyncResultSetImpl}. */
   private enum State {
     INITIALIZED,
+    IN_PROGRESS,
```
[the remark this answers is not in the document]
sakthivelmanii: Fixed

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java
```diff
@@ -449,6 +458,27 @@ private void startCallbackWithBufferLatchIfNecessary(int bufferLatch) {
...
+
+    @Override
+    public void run() {
+      try {
+        // This method returns true if the underlying result set is a streaming result set (e.g. a
+        // GrpcResultSet).
+        // Those result sets will trigger initiateProduceRows() when the first results are received.
+        // Non-streaming result sets do not trigger this callback, and for those result sets, we
+        // need to eagerly
+        // start the ProduceRowsRunnable.
```
olavloite, line 471: nit: move this line up to the previous line (this was probably automatically mangled by the code formatter)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java
```diff
@@ -449,6 +455,21 @@ private void startCallbackWithBufferLatchIfNecessary(int bufferLatch) {
...
 
+  private class InitiateStreamingRunnable implements Runnable {
+
+    @Override
+    public void run() {
+      try {
+        if (!initiateStreaming(AsyncResultSetImpl.this)) {
+          initiateProduceRows();
+        }
+      } catch (SpannerException e) {
```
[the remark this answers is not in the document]
olavloite: This was marked as resolved without any changes or comments. Do you not agree with the suggestion?

My worry here is that other errors than `SpannerException` will be silently ignored, as `Runnable`s that are given to an async executor will not cause any upstream failures if the `Runnable` finishes with an uncaught exception. That again would probably cause the `AsyncResultSet` to just hang.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java
```diff
@@ -458,16 +488,24 @@ public ApiFuture<Void> setCallback(Executor exec, ReadyCallback cb) {
...
       this.executor = MoreExecutors.newSequentialExecutor(Preconditions.checkNotNull(exec));
       this.callback = Preconditions.checkNotNull(cb);
-      this.state = State.RUNNING;
       pausedLatch.countDown();
       return result;
     }
   }
 
+  private void initiateProduceRows() {
+    this.service.execute(new ProduceRowsRunnable());
```
olavloite, line 502: I think that you should move this line to the end of this method to be absolutely safe against weird race conditions. This method is called both from `InitiateStreamingRunnable` and from `onStreamMessage`. That theoretically means that:
1. `InitiateStreamingRunnable` calls this method and then triggers `ProduceRowsRunnable` here.
2. Before this thread continues with the next line, `ProduceRowsRunnable` has already been started and receives a `PartialResultSet`, and then calls `initiateProduceRows()`. (This is highly unlikely, due to the fact that this depends on a network call, but I still think it would be safer to change the status flags before starting `ProduceRowsRunnable`)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java
```diff
@@ -593,4 +635,22 @@ public Struct getCurrentRowAsStruct() {
...
   }
+
+  @Override
+  public void onStreamMessage(PartialResultSet partialResultSet, boolean bufferIsFull) {
+    synchronized (monitor) {
+      if (produceRowsInitiated) {
+        return;
+      }
+      // if PartialResultSet contains resume token or buffer size is ful or
+      // we have reached end of stream, we can start the thread
```
olavloite, line 646:
```suggestion
      // If PartialResultSet contains a resume token, the buffer is full, or
      // we have reached the end of the stream, we can start the thread.
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/GrpcStreamIterator.java
```diff
@@ -36,11 +37,13 @@
 class GrpcStreamIterator extends AbstractIterator<PartialResultSet>
     implements CloseableIterator<PartialResultSet> {
   private static final Logger logger = Logger.getLogger(GrpcStreamIterator.class.getName());
-  private static final PartialResultSet END_OF_STREAM = PartialResultSet.newBuilder().build();
+  public static final PartialResultSet END_OF_STREAM = PartialResultSet.newBuilder().build();
```
olavloite, line 40: nit: Does it need to be `public`, or would `package private` also be enough?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/GrpcStreamIterator.java
olavloite: Do we still need to increase the buffer size?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/GrpcStreamIterator.java
```diff
@@ -182,4 +191,9 @@ public boolean cancelQueryWhenClientIsClosed() {
       return this.cancelQueryWhenClientIsClosed;
     }
   }
+
+  private void onStreamMessage(PartialResultSet partialResultSet) {
+    Optional.ofNullable(streamMessageListener)
+        .ifPresent(sl -> sl.onStreamMessage(partialResultSet, prefetchChunks >= stream.size()));
```
olavloite, line 197: Is `prefetchChunks >= stream.size()` correct for `bufferIsFull`? Would it not make more sense to use `stream.remainingCapacity() <= 1`? (Assuming that we do not double the buffer size)

Then we could also remove `this.prefetchChunks` as a field.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -288,6 +288,34 @@ public boolean next() throws SpannerException {
...
         }
 
+        @Override
+        public boolean initiateStreaming(
+            AsyncResultSet.StreamMessageListener streamMessageListener) {
+          try {
+            boolean streamInitiated = super.initiateStreaming(streamMessageListener);
+            if (!streamInitiated) {
+              synchronized (lock) {
+                session.get().markUsed();
```
olavloite, line 298: I don't feel that this is correct. If `super.initiateStreaming()` returned false, then that means that no RPC was involved. That again means that the session was not used.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -288,6 +288,34 @@ public boolean next() throws SpannerException {
...
+                sessionUsedForQuery = true;
+              }
+            }
+            if (!streamInitiated && isSingleUse) {
+              close();
```
olavloite, line 303: I struggle to understand why we do this. If no streaming was initiated, that means that the result set does not support streaming. But I would not expect that we should then just close it.

It might be a bit of a theoretical discussion, as at the moment, the only result set implementation that we have is the streaming gRPC result set, but in theory we could in the future also add support for running queries with the `ExecuteSql` RPC (instead of `ExecuteStreamingSql`). That would then not support streaming, and would then end up being closed here.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/StreamingResultSet.java
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.spanner;
+
+/** Streaming implementation of ResultSet that supports streaming of chunks */
+public interface StreamingResultSet extends ResultSet {
```
olavloite, line 20: Does this have to be public? Or could we make it package private?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/StreamingResultSet.java
```diff
@@ -0,0 +1,27 @@
...
+
+  /**
+   * Returns the {@link boolean} for this {@link ResultSet}. This method will be used by
+   * AsyncResultSet to initiate gRPC streaming
+   */
+  boolean initiateStreaming(AsyncResultSet.StreamMessageListener streamMessageListener);
```
olavloite, line 26: If we cannot make the interface package-private, then we should mark both the interface and this method with `@InternalApi`, and also add a Javadoc that clearly states that users should not call this method.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/StreamingUtil.java
```diff
@@ -0,0 +1,27 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.spanner;
+
+public class StreamingUtil {
```
olavloite, line 19: Would this be possible?

```suggestion
class StreamingUtil {
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/StreamingUtil.java
```diff
@@ -0,0 +1,27 @@
...
+  public static boolean initiateStreaming(
```
olavloite, line 20: Same as above: Either make all of this package-private, and otherwise mark it with `@InternalApi` and Javadoc that indicates that users should not use this.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java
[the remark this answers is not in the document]
sakthivelmanii: Sorry my bad. This was missed somehow. will fix it.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/StreamingResultSet.java
sakthivelmanii: we can make it as package-private

google-cloud-spanner/src/main/java/com/google/cloud/spanner/StreamingUtil.java
sakthivelmanii: yes. will change it

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java
sakthivelmanii: Agreed. its very rate. but it should be safe to do it at last.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/StreamingUtil.java
sakthivelmanii: done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java
sakthivelmanii: done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/GrpcStreamIterator.java
sakthivelmanii: done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
sakthivelmanii: yes. I removed this whole function. it's not relevant for initiate streaming.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
```diff
@@ -150,6 +150,10 @@ interface CloseableIterator<T> extends Iterator<T> {
     void close(@Nullable String message);
 
     boolean isWithBeginTransaction();
+
+    default boolean initiateStreaming(AsyncResultSet.StreamMessageListener streamMessageListener) {
```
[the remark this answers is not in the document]
sakthivelmanii: done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java
```diff
@@ -578,6 +615,10 @@ public ResultSetMetadata getMetadata() {
     return delegateResultSet.get().getMetadata();
   }
 
+  public boolean initiateStreaming(StreamMessageListener streamMessageListener) {
```
olavloite, line 618: nit: add an `@Override` and `@InternalApi` or make the method package-private (if possible).

google-cloud-spanner/src/main/java/com/google/cloud/spanner/ForwardingResultSet.java
```diff
@@ -102,4 +103,9 @@ public ResultSetStats getStats() {
   public ResultSetMetadata getMetadata() {
     return delegate.get().getMetadata();
   }
+
+  @Override
+  public boolean initiateStreaming(AsyncResultSet.StreamMessageListener streamMessageListener) {
```
olavloite, line 108: nit: make it package-private (if possible), and otherwise mark with `@InternalApi`

google-cloud-spanner/src/main/java/com/google/cloud/spanner/GrpcResultSet.java
```diff
@@ -123,6 +124,11 @@ public ResultSetMetadata getMetadata() {
     return metadata;
   }
 
+  @Override
+  public boolean initiateStreaming(AsyncResultSet.StreamMessageListener streamMessageListener) {
```
olavloite, line 128: nit: package-private or `@InternalApi`

google-cloud-spanner/src/main/java/com/google/cloud/spanner/GrpcStreamIterator.java
```diff
@@ -66,6 +68,10 @@ protected final SpannerRpc.ResultStreamConsumer consumer() {
     return consumer;
   }
 
+  public void registerListener(AsyncResultSet.StreamMessageListener streamMessageListener) {
```
olavloite, line 71: nit: can this be package-private?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/GrpcStreamIterator.java
```diff
@@ -66,6 +68,10 @@ protected final SpannerRpc.ResultStreamConsumer consumer() {
...
+    this.streamMessageListener = streamMessageListener;
```
olavloite, line 72: nit:
```suggestion
    Preconditions.checkState(this.streamMessageListener == null);
    this.streamMessageListener = Preconditions.checkNotNull(streamMessageListener);
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/ResumableStreamIterator.java
```diff
@@ -220,23 +222,20 @@ public boolean isWithBeginTransaction() {
     return stream != null && stream.isWithBeginTransaction();
   }
 
+  @Override
+  public boolean initiateStreaming(AsyncResultSet.StreamMessageListener streamMessageListener) {
```
olavloite, line 226: nit: can this be package-private?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/StreamingUtil.java
```diff
@@ -0,0 +1,27 @@
...
+
+class StreamingUtil {
+  static boolean initiateStreaming(
+      ResultSet resultSet, AsyncResultSet.StreamMessageListener streamMessageListener) {
+    if (resultSet instanceof StreamingResultSet) {
+      return ((StreamingResultSet) resultSet).initiateStreaming(streamMessageListener);
+    }
+    return false;
+  }
+}
```
olavloite, line 27: nit: add a private constructor to prevent initialization

olavloite (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/GrpcStreamIterator.java
sakthivelmanii: this.streamMessageListener = Preconditions.checkNotNull(streamMessageListener); this alone is enough right?

sakthivelmanii: done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/GrpcStreamIterator.java
sakthivelmanii: done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/GrpcResultSet.java
sakthivelmanii: done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/ForwardingResultSet.java
sakthivelmanii: done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncResultSetImpl.java
sakthivelmanii: done

PR #3447, opened 2024-11-05
fix: Client built in metrics. Skip export if instance id is null

surbhigarg92:
This PR does two things 
1) Skip export if instance id is null
2) Update the IAM error message to add public documentation link

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
@@ -141,7 +141,7 @@ class SpannerCloudMonitoringExporter implements MetricExporter {
     // Verifies if metrics data has missing instance id.
     if (!spannerMetricData.stream()
         .flatMap(metricData -> metricData.getData().getPoints().stream())
-        .noneMatch(pd -> SpannerCloudMonitoringExporterUtils.getInstanceId(pd) == null)) {
+        .anyMatch(pd -> SpannerCloudMonitoringExporterUtils.getInstanceId(pd) == null)) {
       logger.log(Level.WARNING, "Metric data has missing instanceId. Skipping export.");
       return CompletableResultCode.ofFailure();
     }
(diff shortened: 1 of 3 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
```diff
@@ -138,6 +138,14 @@ private CompletableResultCode exportSpannerClientMetrics(Collection<MetricData>
       return CompletableResultCode.ofFailure();
     }
 
+    // Verifies if metrics data has missing instance id.
+    if (!spannerMetricData.stream()
+        .flatMap(metricData -> metricData.getData().getPoints().stream())
+        .noneMatch(pd -> SpannerCloudMonitoringExporterUtils.getInstanceId(pd) == null)) {
```
sakthivelmanii, line 144: you can use `anyMatch` instead of negating the result.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
```diff
@@ -138,6 +138,14 @@ private CompletableResultCode exportSpannerClientMetrics(Collection<MetricData>
       return CompletableResultCode.ofFailure();
     }
 
+    // Verifies if metrics data has missing instance id.
+    if (!spannerMetricData.stream()
```
sakthivelmanii, line 142: You have to remove this negation if you are using `anyMatch`

surbhigarg92: Why do we want to use anyMatch ? noneMatch works fine with empty streams

sakthivelmanii: Yes. both works fine. I think code should be easily readable. We have to avoid using negation in if statement if possible.

For example, `if(a < 3)` is easily readable than `if(!(a >= 3))`

sakthivelmanii (approved):

PR #3449, opened 2024-11-07
chore(spanner): handle commit retry protocol extension for mux rw

harshachinta:
In a read-write transaction using a multiplexed session with (read/query + mutation) operations, the CommitResponse from the backend during the commit RPC may include a `MultiplexedSessionRetry` field (indicated by a precommit token). This field signals that the commit RPC should be retried once using the new precommit token. During this retry, mutations should not be resent, as they were already buffered in spanFE during the initial commit RPC call.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
@@ -501,8 +501,6 @@ class TransactionRunnerImpl implements SessionTransaction, TransactionRunner {
                     onPrecommitToken(proto.getPrecommitToken());
                     span.addAnnotation(
                         "Commit operation will be retried with new precommit token as the CommitResponse includes a MultiplexedSessionRetry field");
-                    opSpan.addAnnotation(
-                        "Commit operation will be retried with new precommit token as the CommitResponse includes a MultiplexedSessionRetry field");
                     opSpan.end();
 
                     // Retry the commit RPC with the latest precommit token from CommitResponse.
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
@@ -409,7 +409,9 @@ class TransactionRunnerImpl implements SessionTransaction, TransactionRunner {
       }
       builder.addAllMutations(mutationsProto);
       finishOps.addListener(
-          new CommitRunnable(res, finishOps, builder), MoreExecutors.directExecutor());
+          new CommitRunnable(
+              res, finishOps, builder, /* retryAttemptDueToCommitProtocolExtension = */ false),
+          MoreExecutors.directExecutor());
       return res;
     }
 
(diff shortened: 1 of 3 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
```diff
@@ -479,6 +491,32 @@ public void run() {
...
+                    opSpan.end();
+
+                    // Retry the commit RPC with the latest precommit token from CommitResponse.
+                    MoreExecutors.directExecutor()
+                        .execute(
+                            new CommitRunnable(
+                                res,
+                                prev,
+                                requestBuilder,
+                                /* retryAttemptDueToCommitProtocolExtension = */ true));
```
olavloite, line 515: We should preferably not call this here:
1. This is just a complicated way of executing `new CommitRunnable(...).run()`.
2. That is a blocking call that will execute a new RPC.
3. We should not do that in the listener of the previous commit, as this will then be executed by a gRPC worker thread. That thread will then be blocked while waiting for this second commit to finish.

I'm not entirely sure what the best way is to bring this retry out of this listener, but I think that you need to return a specific error here, and then have the caller of this method handle that error. That handling would then be 'retry the commit with the new pre-commit token'.

olavloite: Per offline discussion: `CommitRunnable.run()` seems to be non-blocking. That would mean that this is safe to do.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
```diff
@@ -479,6 +491,31 @@ public void run() {
...
+                  // commit RPC once with the new token and clear any existing mutations.
+                  // This case is applicable only when the read-write transaction uses multiplexed
+                  // session.
+                  if (proto.hasPrecommitToken() && !retryAttemptDueToCommitProtocolExtension) {
+                    // track the latest pre commit token
+                    onPrecommitToken(proto.getPrecommitToken());
+                    span.addAnnotation(
+                        "Commit operation will be retried with new precommit token as the CommitResponse includes a MultiplexedSessionRetry field");
+                    opSpan.addAnnotation(
+                        "Commit operation will be retried with new precommit token as the CommitResponse includes a MultiplexedSessionRetry field");
```
olavloite, line 505: nit: This annotation is added twice

google-cloud-spanner/src/test/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClientMockServerTest.java
```diff
@@ -1575,6 +1575,62 @@ public void testOtherUnimplementedError_ReadWriteTransactionStillUsesMultiplexed
...
+            });
+
+    List<ExecuteSqlRequest> executeSqlRequests =
+        mockSpanner.getRequestsOfType(ExecuteSqlRequest.class);
+    assertEquals(1, executeSqlRequests.size());
+    // Verify the request is executed using multiplexed sessions
+    assertTrue(mockSpanner.getSession(executeSqlRequests.get(0).getSession()).getMultiplexed());
+
+    List<CommitRequest> commitRequests = mockSpanner.getRequestsOfType(CommitRequest.class);
+    assertEquals(2, commitRequests.size());
```
olavloite, line 1616: nit: also add a verification that only the first `CommitRequest` contains mutations.

olavloite (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
harshachinta: Done

google-cloud-spanner/src/test/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClientMockServerTest.java
harshachinta: Done

PR #3473, opened 2024-11-14
fix(spanner): support transaction tags in partition DML

rahul2393:
Fixes: https://github.com/googleapis/java-spanner/issues/3347


google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractMultiplexedSessionDatabaseClient.java
```diff
@@ -56,4 +57,10 @@ public ServerStream<BatchWriteResponse> batchWriteAtLeastOnce(
   public long executePartitionedUpdate(Statement stmt, UpdateOption... options) {
     throw new UnsupportedOperationException();
   }
+
+  @Override
+  public long executePartitionedUpdate(
+      Statement stmt, @Nullable String transactionTag, UpdateOption... options) {
```
sakthivelmanii, line 63: Should we create TransactionOption instead of String so that it can be extendable in future?

rahul2393: For all other options we already have `UpdateOption`, only reason we are adding transaction tag here because it collides with TagOption in UpdateOption otherwise we would have used UpdateOption iteself.

sakthivelmanii (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
```diff
@@ -197,6 +197,10 @@ public static ReadQueryUpdateTransactionOption tag(String name) {
     return new TagOption(name);
   }
 
+  public static ReadQueryUpdateTransactionOption transactionTag(String name) {
```
harshachinta, line 200: I feel this would bring in confusion to customers, mainly because transaction tags are supported in read-write transactions by using `Options.tag`. Customers might try using this Options.transactionTag for read-write and will not work.
https://cloud.google.com/spanner/docs/introspection/troubleshooting-with-tags#java_1

Should we rename this to be more specific to PDML or is this somehow restricted to be used only for pdml transaction?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/PartitionedDmlTransaction.java
```diff
@@ -194,22 +194,29 @@ ExecuteSqlRequest newTransactionRequestFrom(final Statement statement, final Opt
       if (options.hasTag()) {
         requestOptionsBuilder.setRequestTag(options.tag());
       }
+      if (options.hasTransactionTag()) {
+        requestOptionsBuilder.setTransactionTag(options.transactionTag());
```
harshachinta, line 198: The [official documentation](https://cloud.google.com/spanner/docs/introspection/troubleshooting-with-tags#mapping_between_api_methods_and_requesttransaction_tag) states that transaction tags not supported for read-only transactions and are supported for read-write, but there’s no mention of partitioned DML transactions.
can you please verify if transaction tags are indeed supported for pdml txn?

rahul2393:
Closing this because sending transactionTag in partitionDML is no-op

PR #3474, opened 2024-11-14
feat(spanner): add support for external hosts

sagnghos:

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseId.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseId.java
@@ -96,7 +96,11 @@ public final class DatabaseId {
 
   /** Creates a {@code DatabaseId} with "default" as project id, given instance and database IDs. */
   public static DatabaseId of(String instance, String database) {
-    return new DatabaseId(new InstanceId("default", instance), database);
+    String projectId = SpannerOptions.getDefaultProjectId();
+    if (projectId == null) {
+      projectId = "default";
+    }
+    return new DatabaseId(new InstanceId(projectId, instance), database);
   }
 
   /** Creates a {@code DatabaseId} given the instance identity and database id. */
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseId.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseId.java
@@ -94,7 +94,7 @@ public final class DatabaseId {
     return new DatabaseId(new InstanceId(project, instance), database);
   }
 
-  /** Creates a {@code DatabaseId} given instance and database IDs. */
+  /** Creates a {@code DatabaseId} with "default" project, given instance and database IDs. */
   public static DatabaseId of(String instance, String database) {
     return new DatabaseId(new InstanceId("default", instance), database);
   }
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseId.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseId.java
@@ -94,7 +94,7 @@ public final class DatabaseId {
     return new DatabaseId(new InstanceId(project, instance), database);
   }
 
-  /** Creates a {@code DatabaseId} with "default" project, given instance and database IDs. */
+  /** Creates a {@code DatabaseId} with "default" as project id, given instance and database IDs. */
   public static DatabaseId of(String instance, String database) {
     return new DatabaseId(new InstanceId("default", instance), database);
   }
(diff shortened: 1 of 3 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionClient.java
```diff
@@ -49,10 +51,16 @@ private SessionId(DatabaseId db, String name) {
     static SessionId of(String name) {
       Preconditions.checkNotNull(name);
       Map<String, String> parts = NAME_TEMPLATE.match(name);
+      if (parts == null) {
+        parts = EXTERNAL_HOST_NAME_TEMPLATE.match(name);
```
krishnamoorthy-r, line 55: Can you add unit tests to cover this pattern and ensure the functionality works as intended?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseId.java
```diff
@@ -94,6 +94,11 @@ public static DatabaseId of(String project, String instance, String database) {
     return new DatabaseId(new InstanceId(project, instance), database);
   }
 
+  /** Creates a {@code DatabaseId} given instance and database IDs. */
+  public static DatabaseId of(String instance, String database) {
+    return new DatabaseId(new InstanceId("default", instance), database);
```
krishnamoorthy-r, line 99: I think we should document to users that this method assumes instance Id to be "default".

ko3a4ok: maybe not instance Id, but rather project Id?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionClient.java
sagnghos: Added unit test to cover existing pattern (for regression) as well as new pattern

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseId.java
```diff
@@ -94,6 +94,11 @@ public static DatabaseId of(String project, String instance, String database) {
     return new DatabaseId(new InstanceId(project, instance), database);
   }
 
+  /** Creates a {@code DatabaseId} with "default" project, given instance and database IDs. */
```
krishnamoorthy-r, line 97: nit nit pick: "default" as project id.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionClientTests.java
```diff
@@ -503,4 +505,21 @@ public void onSessionCreateFailure(Throwable t, int createFailureForSessionCount
...
+  public void testSessionNamePatterns() {
+    // Valid pattern for host session name
+    String host =
+        "projects/spanner-project/instances/spanner-instance/databases/test-db/sessions/abcd1234";
+    // Valid pattern for external host session name
+    String externalHost = "instances/default/databases/test-db/sessions/abcd1234";
+    try {
+      SessionId.of(host);
+      SessionId.of(externalHost);
+      // If no exceptions are thrown, the test will pass
```
krishnamoorthy-r, line 520: Is it possible to add assertEquas here to validate all the attributes instanceid, projectid and databaseid as set in both the scenarios?

krishnamoorthy-r (changes requested):

krishnamoorthy-r (approved):

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionClientTests.java
sagnghos: added assert for session name.

rahul2393 (approved):

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionClientTests.java
```diff
@@ -503,4 +505,23 @@ public void onSessionCreateFailure(Throwable t, int createFailureForSessionCount
...
+        "projects/spanner-project/instances/spanner-instance/databases/test-db/sessions/abcd1234";
+    // Valid pattern for external host session name
+    String externalHost = "instances/default/databases/test-db/sessions/abcd1234";
+    try {
+      SessionId hostSession = SessionId.of(host);
+      assertEquals("abcd1234", hostSession.getName());
+      SessionId externalHostSession = SessionId.of(externalHost);
+      assertEquals("abcd1234", externalHostSession.getName());
+      // If no exceptions are thrown, the test will pass
+    } catch (IllegalArgumentException e) {
```
olavloite, line 523: nit: you can remove the entire try-catch block. If a (runtime) exception occurs during a test, then this test will be marked as failed with the error message. I.e. you will get the exact same behavior as you have now with the try-catch block, but with a bit less code and indentation.

olavloite (approved):
LGTM, with a couple of tiny nits.

PR #3477, opened 2024-11-15
fix: directpath enabled attribute

surbhigarg92:
This fix adds the correct value for "directPath_enabled" attribute for Client Built-in Metrics. 

As part of this [fix](https://github.com/googleapis/sdk-platform-java/pull/3170) in gax-grpc , modification were made in ClientContext which can be used to fetch isDirectPath() attribute. However due to a [bug](https://github.com/googleapis/sdk-platform-java/issues/3360) in the gapic generate code , we are creating a wrapper for GrpcSapnnerStub.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
@@ -278,10 +278,7 @@ public class GapicSpannerRpc implements SpannerRpc {
   private final int numChannels;
   private final boolean isGrpcGcpExtensionEnabled;
 
-  private Supplier<Boolean> directPathEnabledSupplier =
-      () -> {
-        return false;
-      };;
+  private Supplier<Boolean> directPathEnabledSupplier = () -> false;
 
   public static GapicSpannerRpc create(SpannerOptions options) {
     return new GapicSpannerRpc(options);
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
@@ -80,7 +80,7 @@ import com.google.cloud.spanner.admin.instance.v1.stub.GrpcInstanceAdminStub;
 import com.google.cloud.spanner.admin.instance.v1.stub.InstanceAdminStub;
 import com.google.cloud.spanner.admin.instance.v1.stub.InstanceAdminStubSettings;
 import com.google.cloud.spanner.encryption.EncryptionConfigProtoMapper;
-import com.google.cloud.spanner.v1.stub.GrpcSpannerStub;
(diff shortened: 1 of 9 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -276,6 +278,11 @@ public class GapicSpannerRpc implements SpannerRpc {
   private final int numChannels;
   private final boolean isGrpcGcpExtensionEnabled;
 
+  private Supplier<Boolean> directPathEnabledSupplier =
```
sakthivelmanii, line 281: why do we need to create this as `Supplier`. I think goal here to identify if direct path is enabled or not right?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GrpcSpannerStubWithStubSettingsAndClientContext.java
```diff
@@ -0,0 +1,40 @@
...
+import java.io.IOException;
+
+/**
+ * Wrapper around {@link GrpcSpannerStub} to make the constructor available inside this package.
+ * This makes it possible to create a {@link GrpcSpannerStub} with a {@link SpannerStubSettings} and
+ * a {@link ClientContext}.
+ */
+class GrpcSpannerStubWithStubSettingsAndClientContext extends GrpcSpannerStub {
+
+  public static final GrpcSpannerStubWithStubSettingsAndClientContext create(
```
sakthivelmanii, line 31: Do we need this as public? can't we keep it as package-private?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
surbhigarg92: Direct Path attribute can only be added from `HeaderInterceptor` class. And `HeaderInterceptor` class in initialized before the `spannerStub` from which we can fetch this value. So that is why added a supplier to fetch the direct path data.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GrpcSpannerStubWithStubSettingsAndClientContext.java
surbhigarg92: Yes, this was a miss. Thanks for identifying it.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
sakthivelmanii: nit: we can rewrite this in single line `private Supplier<Boolean> directPathEnabledSupplier = () -> false;`

surbhigarg92: done

rahul2393 (approved):

PR #3495, opened 2024-11-21
feat: introduce java.time variables and methods

diegomarquezp:
This PR introduces `java.time` alternatives to existing `org.threeten.bp.*` methods, as well as switching internal variables (if any) to `java.time`

The main constraint is to keep the changes backwards compatible, so for each existing threeten method "`method1(org.threeten.bp.Duration)`" we will add an alternative with a _Duration_ (or _Timestamp_ when applicable) suffix: "`method1Duration(java.time.Duration)`".

For most cases, the implementation will be held in the `java.time` method and the old threeten method will just delegate the call to it. However, for the case of abstract classes, the implementation will be kept in the threeten method to avoid breaking changes (i.e. users that already overloaded the method in their user code).


google-cloud-spanner/src/main/java/com/google/cloud/spanner/CompositeTracer.java
```diff
@@ -109,7 +108,7 @@ public void attemptCancelled() {
   }
 
   @Override
-  public void attemptFailed(Throwable error, Duration delay) {
+  public void attemptFailed(Throwable error, org.threeten.bp.Duration delay) {
```
lqiu96, line 111: qq, what changed here?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -689,7 +697,16 @@ Builder setInactiveTransactionRemovalOptions(
...
+    /**
+     * This method is obsolete. Use {@link
+     * #setRemoveInactiveSessionAfterDuration(java.time.Duration)} instead.
+     */
+    @ObsoleteApi("Use setRemoveInactiveSessionAfterDuration(java.time.Duration) instead")
+    public Builder setRemoveInactiveSessionAfter(org.threeten.bp.Duration duration) {
+      return setRemoveInactiveSessionAfterDuration(toJavaTimeDuration(duration));
+    }
+
+    public Builder setRemoveInactiveSessionAfterDuration(java.time.Duration duration) {
```
lqiu96, line 709: qq, do we need this if it's a public method in a package-private class?

lqiu96: Ok nvm. I just tried this locally and I can access this via `newBuilder()`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerRpc.java
```diff
@@ -56,11 +56,11 @@
 import com.google.spanner.admin.instance.v1.UpdateInstanceConfigMetadata;
 import com.google.spanner.admin.instance.v1.UpdateInstanceMetadata;
 import com.google.spanner.v1.*;
+import java.time.Duration;
```
lqiu96, line 59: This may impact
```
 ServerStream<PartialResultSet> executeStreamingPartitionedDml(
      ExecuteSqlRequest request, @Nullable Map<Option, ?> options, Duration timeout);
 ```

File: google-cloud-spanner/src/main/java/com/google/cloud/spanner/CompositeTracer.java (hunk 1 above)
Comment (diegomarquezp) [reply google-cloud-spanner/src/main/java/com/google/cloud/spanner/CompositeTracer.java:111]: https://github.com/googleapis/java-spanner/pull/3495/files#diff-504b19ab966dfdfb231d51ebf483b78fc52af98e608f3b081e094c96a80b38f1R117-R122 shows that the duration method was already added. I just modified the signature to make the package name explicit

File: google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerRpc.java (hunk 3 above)
Comment (diegomarquezp) [reply google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerRpc.java:59]: https://github.com/googleapis/java-spanner/blob/439e105d05107f7220ee2a19aeb55af4c074ee55/google-cloud-spanner/src/main/java/com/google/cloud/spanner/PartitionedDmlTransaction.java#L92 is the only production usage of method. Moreover, the interface is marked as `@InternalApi`. I relied on the tests to tell whether this may break something.

File: google-cloud-spanner/src/main/java/com/google/cloud/spanner/CompositeTracer.java (hunk 1 above)
Comment (lqiu96) [reply google-cloud-spanner/src/main/java/com/google/cloud/spanner/CompositeTracer.java:111]: Ok gotcha. Do you think we can migrate this to `attemptFailedDuration`? https://github.com/googleapis/sdk-platform-java/blob/cf32e72d83c06d1d36e1eea01afc37bee5766455/gax-java/gax/src/main/java/com/google/api/gax/tracing/ApiTracer.java#L126?

File: google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerRpc.java (hunk 3 above)
Comment (diegomarquezp) [reply google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerRpc.java:59]: @rahul2393 does it make sense to treat this code as internal?

File: google-cloud-spanner/src/main/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClient.java [hunk 4]
```diff
@@ -275,7 +275,7 @@ public void onSessionCreateFailure(Throwable t, int createFailureForSessionCount
 
   private static void maybeWaitForSessionCreation(
       SessionPoolOptions sessionPoolOptions, ApiFuture<SessionReference> future) {
-    org.threeten.bp.Duration waitDuration = sessionPoolOptions.getWaitForMinSessions();
+    java.time.Duration waitDuration = sessionPoolOptions.getWaitForMinSessions();
```
sakthivelmanii, line 278: nit: `java.time.Duration` is already imported in this class. Can't we just remove `org.threeten.bp.`

google-cloud-spanner/src/main/java/com/google/cloud/spanner/CompositeTracer.java
sakthivelmanii, line 111: nit: Can we import `java.time.Duration` so that we can remove `java.time` in `attemptFailedDuration` method?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/OpenTelemetryApiTracer.java
```diff
@@ -163,8 +165,18 @@ public void attemptCancelled() {
...
+   */
+  @Override
+  @ObsoleteApi("Use attemptFailedDuration(Throwable, java.time.Duration) instead")
+  public void attemptFailed(Throwable error, org.threeten.bp.Duration delay) {
+    attemptFailedDuration(error, toJavaTimeDuration(delay));
+  }
+
   @Override
-  public void attemptFailed(Throwable error, Duration delay) {
+  public void attemptFailedDuration(Throwable error, java.time.Duration delay) {
```
sakthivelmanii, line 179: nit: Can we import `java.time.Duration`?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -63,8 +66,8 @@ public class SessionPoolOptions {
   @Deprecated private final long initialWaitForSessionTimeoutMillis;
 
   private final boolean autoDetectDialect;
-  private final Duration waitForMinSessions;
-  private final Duration acquireSessionTimeout;
+  private final java.time.Duration waitForMinSessions;
```
sakthivelmanii, line 69: nit: We can import `java.time.Duration` so that in future we don't have a confusion on which class to use

sakthivelmanii: and also it avoids so many changes in the code which is not really necessary.

sakthivelmanii:
@diegomarquezp Can we make sure following in all the files?

`import java.time.Duration` in all the files and use `org.threeten.bp.Duration` in obsolete methods. This avoids lot of code changes which is unnecessary. This also provides the standard that `java.time.Duration` should be used.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerRpc.java
sakthivelmanii: yes @diegomarquezp you can treat this as an internal class.

diegomarquezp:
> @diegomarquezp Can we make sure following in all the files?
> 
> `import java.time.Duration` in all the files(whereever is required) and use `org.threeten.bp.Duration` directly in obsolete methods. This avoids lot of code changes which is unnecessary. This also provides the standard that `java.time.Duration` should be used.

@sakthivelmanii Thanks for the review! Before switching to `import java.time.Duration` I'd like to double check if our rationale for not `import`ing `java.time.*` seems sound to you: this is mainly to make it clear to the reader that there are two `Duration`/`Instant` classes coexisting in the same file and attention should be paid. We deliberately added these changes just to make this differentiation as explicit as possible. We've been doing it this way in other repos (e.g. https://github.com/googleapis/java-pubsub/pull/2271#discussion_r1852395166) (doesn't mean it _should_ be done as in other repos). 

Promoting `java.time` as the standard also makes sense, and it's ultimately the Spanner team's call, but wanted to make sure our approach doesn't come up as a mistake. I'm happy to move forward either way.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/CompositeTracer.java
diegomarquezp: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerRpc.java
diegomarquezp: Thank you

google-cloud-spanner/src/main/java/com/google/cloud/spanner/CompositeTracer.java
diegomarquezp: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/OpenTelemetryApiTracer.java
diegomarquezp: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
diegomarquezp: Done

lqiu96 (approved):
LGTM. Spanner team for approval.

rahul2393 (approved):

PR #3518, opened 2024-11-29
fix: shutdown built in metrics meter provider

surbhigarg92:
This PR calls Meter provider shutdown method when the client is shutdown to clean up the resources and export the remaining metrics data.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/BuiltInOpenTelemetryMetricsProvider.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/BuiltInOpenTelemetryMetricsProvider.java
@@ -42,7 +42,6 @@ import java.net.UnknownHostException;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.UUID;
-import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.annotation.Nullable;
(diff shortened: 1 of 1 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/BuiltInOpenTelemetryMetricsProvider.java
```diff
@@ -65,8 +66,14 @@ OpenTelemetry getOrCreateOpenTelemetry(String projectId, @Nullable Credentials c
...
             SpannerCloudMonitoringExporter.create(projectId, credentials), sdkMeterProviderBuilder);
-        this.openTelemetry =
-            OpenTelemetrySdk.builder().setMeterProvider(sdkMeterProviderBuilder.build()).build();
+        SdkMeterProvider sdkMeterProvider = sdkMeterProviderBuilder.build();
+        this.openTelemetry = OpenTelemetrySdk.builder().setMeterProvider(sdkMeterProvider).build();
+        Runtime.getRuntime()
+            .addShutdownHook(
+                new Thread(
+                    () -> {
+                      sdkMeterProvider.shutdown().join(5, TimeUnit.SECONDS);
```
sakthivelmanii, line 75: quick question: Is there any specific reason to put 5 seconds?

surbhigarg92: Not really. I was just trying to give some upper limit for the shutdown and export to complete.

sakthivelmanii: @surbhigarg92 got it. can we use close method? close method which is provided by `SdkMeterProvider` has 10 seconds wait time. We can reuse that method.

`Runtime.getRuntime().addShutdownHook(new Thread(sdkMeterProvider::close));`

Sharing the internal implementation of `SdkMeterProvider`

```
public void close() {
    this.shutdown().join(10L, TimeUnit.SECONDS);
  }
```

Ref: https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeterProvider.java#L166

sakthivelmanii (approved):

PR #3519, opened 2024-11-29
feat: Add Metrics host for built in metrics

surbhigarg92:
This PR allows users to set the custom monitoring host for built in metrics.


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -763,6 +770,11 @@ public boolean isEnableBuiltInMetrics() {
     public boolean isEnableEndToEndTracing() {
       return Boolean.parseBoolean(System.getenv(SPANNER_ENABLE_END_TO_END_TRACING));
     }
+
+    @Override
+    public String getMetricsHost() {
+      return System.getenv(SPANNER_METRICS_HOST);
```
sakthivelmanii, line 776: Can we make sure it returns DEFAULT MONITORING HOST if metrics host is empty? otherwise it's a breaking change since you are always asking customer for monitoring host.

rahul2393 (approved):

PR #3522, opened 2024-12-03
ci(spanner): Fix nightly job issues

sakthivelmanii:

### DIFF
--- a/google-cloud-spanner-executor/pom.xml
+++ b/google-cloud-spanner-executor/pom.xml
@@ -193,6 +193,7 @@
       <groupId>org.apache.maven.surefire</groupId>
       <artifactId>surefire-junit4</artifactId>
       <version>3.5.2</version>
+      <scope>test</scope>
     </dependency>
 
   </dependencies>
--- a/google-cloud-spanner-executor/pom.xml
+++ b/google-cloud-spanner-executor/pom.xml
@@ -189,6 +189,12 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
+      <groupId>org.apache.maven.surefire</groupId>
+      <artifactId>surefire-junit4</artifactId>
+      <version>3.5.2</version>
+    </dependency>
+
   </dependencies>
   <build>
     <plugins>
@@ -259,7 +265,7 @@
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-dependency-plugin</artifactId>
           <configuration>
-            <ignoredDependencies> com.google.api:gax</ignoredDependencies>
+            <ignoredDependencies> com.google.api:gax,org.apache.maven.surefire:surefire-junit4 </ignoredDependencies>
           </configuration>
         </plugin>
       </plugins>
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/slow/ITBackupTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/slow/ITBackupTest.java
@@ -717,7 +717,7 @@ public class ITBackupTest {
     Page<Backup> page = dbAdminClient.listBackups(instanceId, Options.pageSize(1));
     assertEquals(1, Iterables.size(page.getValues()));
     numBackups++;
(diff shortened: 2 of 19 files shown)


google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/slow/ITBackupTest.java
```diff
@@ -717,7 +717,7 @@ private void testPagination() {
     Page<Backup> page = dbAdminClient.listBackups(instanceId, Options.pageSize(1));
     assertEquals(1, Iterables.size(page.getValues()));
     numBackups++;
-    assertTrue(page.hasNextPage());
+    assertFalse(page.hasNextPage());
```
olavloite, line 720: This is a bit surprising. Did this always fail before? Or is there some other reason that this needs to be changed now?

google-cloud-spanner-executor/pom.xml
```diff
@@ -189,6 +189,12 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
+      <groupId>org.apache.maven.surefire</groupId>
+      <artifactId>surefire-junit4</artifactId>
+      <version>3.5.2</version>
+    </dependency>
```
olavloite, line 196: Could this have `<scope>test</scope>`?

samples/snippets/src/main/java/com/example/spanner/admin/archived/PgSpannerSample.java
```diff
@@ -203,7 +203,6 @@ static void createPostgreSqlDatabase(DatabaseAdminClient dbAdminClient, Database
       // Initiate the request which returns an OperationFuture.
       Database db = op.get();
       System.out.println("Created database [" + db.getId() + "]");
-      createTableUsingDdl(dbAdminClient, id);
```
olavloite, line 206: Why is this not needed anymore?

samples/snippets/src/test/java/com/example/spanner/admin/archived/PgSpannerSampleIT.java
```diff
@@ -130,7 +130,7 @@ public void testSample() throws Exception {
     assertThat(databaseId).isNotNull();
 
     System.out.println("Create Database ...");
-    String out = runSample("createpgdatabase");
+    String out = runSample("createdatabase");
```
olavloite, line 133: Are you sure? This sample is for creating a PG database, so I think that this was not a typo.

google-cloud-spanner-executor/pom.xml
sakthivelmanii: sure

samples/snippets/src/main/java/com/example/spanner/admin/archived/PgSpannerSample.java
sakthivelmanii: https://github.com/googleapis/java-spanner/blob/main/samples/snippets/src/test/java/com/example/spanner/admin/archived/PgSpannerSampleIT.java#L138 This is already being called here. Its failing because duplicate table creation.

olavloite: Ack

olavloite (approved):

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/slow/ITBackupTest.java
olavloite: No, that is not what that means. That value is for how many items per page that the backend should return. So it basically works like this:
1. If you have a query that returns 10 results, and you set page size to 1, then you get 10 pages with each 1 item in each page.
2. However, if you execute the same query with page size 5, then you get 2 pages with each 5 items.

sakthivelmanii: Yeah I meant the same. In this example, there's only one backup so it will not have next page.

PR #3536, opened 2024-12-09
feat(spanner): add jdbc support for external hosts

sagnghos:


google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
```diff
@@ -947,12 +965,18 @@ && getInitialConnectionPropertyValue(OAUTH_TOKEN) == null
       this.sessionPoolOptions = SessionPoolOptions.newBuilder().setAutoDetectDialect(true).build();
     }
 
-    String projectId = matcher.group(Builder.PROJECT_GROUP);
+    String projectId = "default";
+    String instanceId = matcher.group(Builder.INSTANCE_GROUP);
+    if (!isExternalHost) {
+      projectId = matcher.group(Builder.PROJECT_GROUP);
+    } else if (instanceId == null) {
```
olavloite, line 972: nit:
```suggestion
    } else if (isExternalHost && instanceId == null) {
```

Instance is actually an optional part of the connection URL. This was done to keep the possibility open to allow connections to only specify a project, so that it can also be used for commands like `CREATE INSTANCE ...` (for the same reason, the database is also optional).

olavloite (approved):

PR #3538, opened 2024-12-09
ci(spanner): Upgrade dependencies which are missing in Airlock

sakthivelmanii:
Updates the version of `maven-checkstyle-plugin` and `flatten-maven-plugin` to versions that are available in Airlock (or that will be available in Airlock?)


suztomo (commented):
I believe the flatten plugin version is only need to be declared once at the root pom.xml

sakthivelmanii:
it's used in all the submodule except one. probably that's why it's included in each module separately. I will take an AI to move it to root pom.xml

olavloite (approved):

PR #3544, opened 2024-12-11
feat: add support for ARRAY<STRUCT> to CloudCilentExecutor

larkee:
There are SPANNER_SYS tables that contain ARRAY<STRUCT> columns. Adding support for this in the CloudClientExecutor so that queries involving these tables do not throw an error.

### DIFF
--- a/google-cloud-spanner-executor/src/main/java/com/google/cloud/executor/spanner/CloudClientExecutor.java
+++ b/google-cloud-spanner-executor/src/main/java/com/google/cloud/executor/spanner/CloudClientExecutor.java
@@ -2842,61 +2842,75 @@ public class CloudClientExecutor extends CloudExecutor {
   /** Convert a result row to a row proto(value list) for sending back to the client. */
   private com.google.spanner.executor.v1.ValueList buildRow(
       StructReader result, OutcomeSender sender) throws SpannerException {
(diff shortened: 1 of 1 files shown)


google-cloud-spanner-executor/src/main/java/com/google/cloud/executor/spanner/CloudClientExecutor.java
```diff
@@ -2840,61 +2840,75 @@ private Status processResults(
...
             break;
           case NUMERIC:
-            String ascii = result.getBigDecimal(i).toPlainString();
+            String ascii = struct.getBigDecimal(i).toPlainString();
             value.setStringValue(ascii);
             break;
           case JSON:
-            value.setStringValue(result.getJson(i));
+            value.setStringValue(struct.getJson(i));
             break;
```
thiagotnunes, line 2904: Do we need to handle the case of STRUCT (non-array)?

larkee: A non-array STRUCT cannot be returned in a result set.

https://cloud.google.com/spanner/docs/reference/standard-sql/query-syntax#using_structs_with_select

thiagotnunes (approved):

rahul2393 (approved):

PR #3558, opened 2024-12-16
ci(spanner): improve performance of samples tests

sakthivelmanii:

### DIFF
--- a/samples/install-without-bom/pom.xml
+++ b/samples/install-without-bom/pom.xml
@@ -147,6 +147,8 @@
         <artifactId>maven-failsafe-plugin</artifactId>
         <version>3.5.2</version>
         <configuration>
+          <forkCount>10</forkCount>
+          <reuseForks>false</reuseForks>
           <systemPropertyVariables>
             <spanner.test.instance>java-sample-integration-tests</spanner.test.instance>
             <spanner.test.instance.mr>java-client-mr-integration-tests</spanner.test.instance.mr>
@@ -157,6 +159,9 @@
             <spanner.sample.database>mysample</spanner.sample.database>
             <spanner.quickstart.database>quick-db</spanner.quickstart.database>
           </systemPropertyVariables>
+          <excludes>
+            <exclude>**/SpannerSampleIT.java</exclude>
+          </excludes>
         </configuration>
       </plugin>
     </plugins>
--- a/samples/snapshot/pom.xml
+++ b/samples/snapshot/pom.xml
@@ -146,6 +146,8 @@
         <artifactId>maven-failsafe-plugin</artifactId>
         <version>3.5.2</version>
         <configuration>
+          <forkCount>10</forkCount>
+          <reuseForks>false</reuseForks>
           <systemPropertyVariables>
             <spanner.test.instance>java-sample-integration-tests</spanner.test.instance>
             <spanner.test.instance.mr>java-client-mr-integration-tests</spanner.test.instance.mr>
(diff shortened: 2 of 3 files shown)


samples/snippets/src/test/java/com/example/spanner/SpannerSampleIT.java
```diff
@@ -51,12 +52,14 @@
...
 import org.junit.BeforeClass;
 import org.junit.Test;
+import org.junit.experimental.categories.Category;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
 
 /** Unit tests for {@code SpannerSample} */
 @RunWith(JUnit4.class)
 @SuppressWarnings("checkstyle:abbreviationaswordinname")
+@Category(SlowTest.class)
```
sakthivelmanii, line 62: These two tests were taking almost an hour. Plan is to introduce a job for slow sample integration tests(Java 11) to speed up the performance.

rahul2393: Should we mark all the tests in SpannerSample as slow, or particualr backup/restore should be tagged?

sakthivelmanii: For now, I have marked both the files to not to run given that it's taking an almost an hour. and we are running these tests for each module. Goal is to run only once either with [install-without-bom] or [snapshot] or [snippets]

samples/install-without-bom/pom.xml
```diff
@@ -147,6 +147,8 @@
         <artifactId>maven-failsafe-plugin</artifactId>
         <version>3.5.2</version>
         <configuration>
+          <forkCount>10</forkCount>
+          <reuseForks>false</reuseForks>
```
olavloite, line 151: I don't think this does anything at the moment, as there are no integration tests defined in this project. Are we planning on adding tests here?

samples/snapshot/pom.xml
```diff
@@ -146,6 +146,8 @@
         <artifactId>maven-failsafe-plugin</artifactId>
         <version>3.5.2</version>
         <configuration>
+          <forkCount>10</forkCount>
+          <reuseForks>false</reuseForks>
```
olavloite, line 150: Same as for `install-without-bom`: There are no tests in this project at the moment, so I don't think this does anything

samples/snippets/pom.xml
```diff
@@ -187,6 +189,9 @@
             <spanner.sample.database>mysample</spanner.sample.database>
             <spanner.quickstart.database>quick-db</spanner.quickstart.database>
           </systemPropertyVariables>
+          <excludes>
+            <exclude>**/SpannerSampleIT.java</exclude>
+          </excludes>
```
olavloite, line 194: I understand that this one is slow, but do we really want to skip it all-together? And/or will we have some other nightly job that overrides this configuration and does include this test?

samples/install-without-bom/pom.xml
sakthivelmanii: It is. Currently we are runnings tests three times in all three modules.

samples/snapshot/pom.xml
sakthivelmanii: It is. Currently we are runnings tests three times in all three modules.

samples/snippets/pom.xml
sakthivelmanii: I am introducing a new job to run this test alone in parallel so that it can be completed in an hour.

olavloite (approved):

PR #3565, opened 2024-12-18
fix: retry specific internal errors

olavloite:
Some specific internal errors should be retrid. Instead of adding INTERNAL
as a standard retryable error code, we use an interceptor to catch and
translate those specific errors.

See also b/375684610

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerErrorInterceptor.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerErrorInterceptor.java
@@ -72,8 +72,7 @@ public final class SpannerErrorInterceptor implements ClientInterceptor {
               public void onClose(Status status, Metadata trailers) {
                 try {
                   // Translate INTERNAL errors that should be retried to a retryable error code.
-                  if (status.getCode() == Code.INTERNAL
-                      && IsRetryableInternalError.INSTANCE.isRetryableInternalError(status)) {
+                  if (IsRetryableInternalError.INSTANCE.isRetryableInternalError(status)) {
                     status =
                         Status.fromCode(Code.UNAVAILABLE).withDescription(status.getDescription());
                   }
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerErrorInterceptor.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerErrorInterceptor.java
@@ -16,6 +16,7 @@
 
 package com.google.cloud.spanner.spi.v1;
 
+import com.google.cloud.spanner.IsRetryableInternalError;
 import com.google.rpc.BadRequest;
 import com.google.rpc.Help;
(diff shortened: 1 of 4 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerErrorInterceptor.java
```diff
@@ -69,6 +71,12 @@ public void start(Listener<RespT> responseListener, Metadata headers) {
               @Override
               public void onClose(Status status, Metadata trailers) {
                 try {
+                  // Translate INTERNAL errors that should be retried to a retryable error code.
+                  if (status.getCode() == Code.INTERNAL
```
sakthivelmanii, line 75: I can see the same check in `isRetryableInternalError` as well. Can we remove it from here?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/RetryableInternalErrorTest.java
```diff
@@ -0,0 +1,80 @@
...
+      }
+      assertEquals(
+          Long.valueOf(1L),
+          client
+              .readWriteTransaction()
+              .run(transaction -> transaction.executeUpdate(INSERT_STATEMENT)));
+
+      // Verify that all the RPCs were retried.
+      assertEquals(2, mockSpanner.countRequestsOfType(BatchCreateSessionsRequest.class));
+      assertEquals(4, mockSpanner.countRequestsOfType(ExecuteSqlRequest.class));
```
sakthivelmanii, line 77: Do you know why `ExecuteSqlRequest` is sent 4 times? We have setup exception only once for ExecuteSql right?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerErrorInterceptor.java
olavloite: yeah, good point. I initially had this set up in a way where only the error message was checked in the RetryableInternalError class, and forgot to remove this after that refactor. Removed now.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/RetryableInternalErrorTest.java
```diff
@@ -0,0 +1,80 @@
...
+              Status.INTERNAL
+                  .withDescription("Authentication backend internal server error. Please retry.")
+                  .asRuntimeException()));
+      mockSpanner.setExecuteSqlExecutionTime(
+          SimulatedExecutionTime.ofException(
+              Status.INTERNAL
+                  .withDescription("Authentication backend internal server error. Please retry.")
+                  .asRuntimeException()));
+
+      try (ResultSet resultSet = client.singleUse().executeQuery(SELECT1_STATEMENT)) {
```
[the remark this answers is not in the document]
olavloite: I've re-organized the test a bit to show that both `ExecuteStreamingSql` and `ExecuteSql` are retried.

sakthivelmanii (approved):

PR #3567, opened 2024-12-18
chore: disable native metrics when there are no credentials

olavloite:
Native metrics were automatically disabled when the emulator is used, but some clients (e.g. PGAdapter) set up the connection to the emulator manually instead of setting the environment variable. Also, when using in-mem mock servers, native metrics should be disabled, as they cannot be exported.

This PR therefore adds an additional check that disables native metrics when the client uses a NoCredentials instance.


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1791,6 +1792,10 @@ public CallCredentialsProvider getCallCredentialsProvider() {
     return callCredentialsProvider;
   }
 
+  public boolean usesNoCredentials() {
```
sakthivelmanii, line 1795: Does it really have to be public? one usage is within this file itself. Can we make it private?

PR #3570, opened 2024-12-26
fix: update max_in_use_session at 10 mins interval

surbhigarg92:
This PR fixes the max_in_use_session value. Currently this value is only reset at client initialization.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
@@ -2048,7 +2048,7 @@ class SessionPool {
 
     // Does various pool maintenance activities.
     void maintainPool() {
-      Instant currTime = clock.instant();
+      Instant currTime;
       synchronized (lock) {
         if (SessionPool.this.isClosed()) {
           return;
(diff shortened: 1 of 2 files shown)


rahul2393 (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -2048,6 +2048,7 @@ boolean isClosed() {
 
     // Does various pool maintenance activities.
     void maintainPool() {
+      Instant currTime = clock.instant();
```
sakthivelmanii, line 2051: nit: Do we really have to initialize here? Can we initialize before if statement inside synchronized block so that we will get exact current time. It would be okay to initialize here if we don't have synchronized block.

sakthivelmanii: I just noticed that go does the same way where they are assigning right before if statement. https://github.com/googleapis/google-cloud-go/blob/main/spanner/session.go#L1775

surbhigarg92: this `currTime` is used at multiple places. Inside synchronized block and outside as well. It will be an overkill if we initialize at multiple times. Inside block and outside. WDYT ?

PR #3571, opened 2024-12-26
chore(spanner): fix mutation only case in rw mux with aborted error

harshachinta:
This PR fixes the following scenario.

In a mutation-only scenario for a read-write transaction using multiplexed sessions, an explicit BeginTransaction RPC is triggered [after the user operation completes](https://github.com/googleapis/java-spanner/blob/8d295c4a4030b4e97b1d653cc3baf412864f3042/google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java#L405).

However, if an ABORT occurs during this BeginTransaction RPC (this can happen if backend is unable to refresh the schema when a mutation key is passed), the client retry mechanism performs an explicit BeginTransaction [before the user operation begins](https://github.com/googleapis/java-spanner/blob/8d295c4a4030b4e97b1d653cc3baf412864f3042/google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java#L1254). This behavior causes issues in multiplexed sessions because, in mutation-only cases, the BeginTransaction requires a mutation key to generate a precommit token. This precommit token is essential for committing the transaction.

Currently, since the BeginTransaction during a retry is executed before the user operation, no mutation key is available, and consequently, no precommit token is included in the transaction response. This leads to commit failures with an error as shown below
```
com.google.cloud.spanner.SpannerException: INVALID_ARGUMENT: io.grpc.StatusRuntimeException: INVALID_ARGUMENT: A Commit request for a read-write transaction on a multiplexed session must specify a precommit token.
 ```

**Fix: The fix involves retrying the BeginTransaction RPC for mutation-only cases with multiplexed sessions when it fails with ABORTED error.**

### CODE REVIEW

**Comment by olavloite** [comment]
> However, if an ABORT occurs during this BeginTransaction RPC

Is this something that can actually happen?

File: google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java [hunk 1]
```diff
@@ -1229,7 +1237,16 @@ private <T> T runInternal(final TransactionCallable<T> txCallable) {
...
+            /*
+             If the transaction contains only mutations and is using a multiplexed session, perform a
+             `BeginTransaction` after the user operation completes during a retry.
+
+             This ensures that a random mutation from the mutations list is chosen when invoking
+             `BeginTransaction`. If `BeginTransaction` is performed before the user operation,
+             the mutations are not sent, and the precommit token is not received, resulting in
+             an INVALID_ARGUMENT error (missing precommit token) during commit.
+            */
+            useInlinedBegin = txn.mutationsOnlyTransaction || txn.transactionId != null;
```
olavloite, line 1249: nit: maybe we should rename this variable to something like `shouldEnsureTxnAtStart`? Now this variable is being set to true for a case that never uses inlinedBegin, which could cause confusion to new engineers who read this code for the first time.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -83,6 +83,7 @@ public class SessionPoolOptions {
 
   // TODO: Change to use java.time.Duration.
   private final Duration multiplexedSessionMaintenanceDuration;
+  private final boolean skipVerifyingBeginTransactionForMuxRW;
```
olavloite, line 86: Is this temporary just to be able to test with/without this fix?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClientMockServerTest.java
```diff
@@ -1291,6 +1292,280 @@ public void testMutationOnlyUsingAsyncTransactionManager() {
...
   }
 
+  // The following 4 tests validate mutation-only cases where the BeginTransaction RPC fails with an
+  // ABORTED or retryable error
+  @Test
+  public void testMutationOnlyCaseAbortedDuringBeginTransaction() {
+    // This test verifies that in the case of mutations-only, when a transaction is retried after an
+    // ABORT in BeginTransaction RPC, the mutation key is correctly set in the BeginTransaction
+    // request
+    // and precommit token is set in Commit request.
```
olavloite, line 1302: super-nit: fix formatting

google-cloud-spanner/src/test/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClientMockServerTest.java
```diff
@@ -1291,6 +1292,280 @@ public void testMutationOnlyUsingAsyncTransactionManager() {
...
+                    .setSkipVerifyingBeginTransactionForMuxRW(true)
+                    .build())
+            .build()
+            .getService();
+    DatabaseClientImpl client =
+        (DatabaseClientImpl) spanner1.getDatabaseClient(DatabaseId.of("p", "i", "d"));
+
+    // Force the BeginTransaction RPC to return Aborted the first time it is called. The exception
+    // is cleared
+    // after the first call, so the retry should succeed.
```
olavloite, line 1322: nit: fix formatting

google-cloud-spanner/src/test/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClientMockServerTest.java
```diff
@@ -1291,6 +1292,280 @@ public void testMutationOnlyUsingAsyncTransactionManager() {
...
+            .setChannelProvider(channelProvider)
+            .setCredentials(NoCredentials.getInstance())
+            .setSessionPoolOption(
+                SessionPoolOptions.newBuilder()
+                    .setUseMultiplexedSession(true)
+                    .setUseMultiplexedSessionForRW(true)
+                    .setSkipVerifyingBeginTransactionForMuxRW(true)
+                    .build())
+            .build()
+            .getService();
```
olavloite, line 1380: nit: it seems like a very large part of these tests are equal (e.g. setting up Spanner and verifying the requests). Can we combine those parts into shared methods that are used by the various test methods?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
harshachinta: I’ve captured its state in a separate variable named explicitBeginBeforeUserOperation to avoid confusion. Using the same variable could be misleading, as one represents inline or explicit begin, while the other differentiates between explicit begin occurring before or after the user operation.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClientMockServerTest.java
harshachinta: Done

harshachinta:
> this can happen if backend is unable to refresh the schema when a mutation key is passed

Yes, this can happen if backend is unable to refresh the schema when a mutation key is passed with a BeginTransaction RPC.

harshachinta:
Thanks @olavloite. Updated the approach to solve this problem by following the commit 
https://github.com/googleapis/java-spanner/commit/cafefb68dbb1542d493afc2d89e5775402344de5

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
harshachinta: Thanks @olavloite. Updated the approach as per our offline discussion

olavloite (approved):

PR #3582, opened 2025-01-03
chore: Replace channel provider for fixed channel provider.

nimf:
Currently, when creating a stub, the channel provider creates a new channel pool every time, thus we end up with four channel pools to the same target instead of one.

In this PR we create a single channel pool using the existing channel provider and then wrap it into FixedTransportChannelProvider so that any stub using the fixed channel provider uses the created pool instead of creating a new one.

@surbhigarg92 do I also need to change anything in the metrics according to this change?

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
@@ -41,6 +41,7 @@ import com.google.api.gax.rpc.ApiClientHeaderProvider;
 import com.google.api.gax.rpc.ApiException;
 import com.google.api.gax.rpc.ClientContext;
 import com.google.api.gax.rpc.FixedHeaderProvider;
+import com.google.api.gax.rpc.FixedTransportChannelProvider;
 import com.google.api.gax.rpc.HeaderProvider;
 import com.google.api.gax.rpc.InstantiatingWatchdogProvider;
 import com.google.api.gax.rpc.OperationCallable;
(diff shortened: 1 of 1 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -402,11 +403,14 @@ public GapicSpannerRpc(final SpannerOptions options) {
       final String emulatorHost = System.getenv("SPANNER_EMULATOR_HOST");
 
       try {
+        FixedTransportChannelProvider fixedChannelProvider =
+            FixedTransportChannelProvider.create(channelProvider.getTransportChannel());
```
olavloite, line 407: I think that we need a bit more management and logic around this:
1. We should only use a `FixedTransportChannelProvider` if the user has not specified a channel provider themselves. That is; We should only do this `if (options.getChannelProvider() == null)`. The reason is that for the default channel provider, we know that it produces a channel pool, which can safely be wrapped in a `FixedChannelProvider`. We don't know what any custom channel provider that has been set might produce, and whether it is a good idea to just get one channel from that provider and use that for all requests.
2. We also need to keep a reference to the channel that is returned by `defaultChannelProvider.build().getTransportChannel()`, and then close that channel when this `GapicSpannerRpc` instance is shut down. A `FixedTransportChannelProvider` is not closed when a stub is closed. This will leave the channels open after this instance has been closed.

olavloite (changes requested):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
nimf: Yes, I agree on 2. Will update.

For 1, as a user I would expect that the channel provider that I specified will only be executed once per client, not four times per client. However the current implementation is different and for the sake of not breaking anyone who may rely on that we may keep the existing behaviour for the user-specified channel provider. What do you prefer?

Also, about the test failures. Are those flakes?

Most errors look like
```
Error:    RetryOnInvalidatedSessionTest....
Spanner FAILED_PRECONDITION: io.grpc.StatusRuntimeException: FAILED_PRECONDITION: This transaction has been invalidated by a later transaction in the same session.
Transaction id: 1
Expected: 2
```

PR #3595, opened 2025-01-10
feat: add gcp client attributes in OpenTelemetry traces

surbhigarg92:


rahul2393 (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TraceWrapper.java
```diff
@@ -204,6 +211,9 @@ Attributes createCommonAttributes(DatabaseId db) {
     AttributesBuilder builder = Attributes.builder();
     builder.put(DB_NAME_KEY, db.getDatabase());
     builder.put(INSTANCE_NAME_KEY, db.getInstanceId().getInstance());
+    builder.put(GCP_CLIENT_SERVICE_KEY, "Spanner");
```
sakthivelmanii, line 214: Should we use spanner or Spanner? Any standards here?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/OpenTelemetrySpanTest.java
```diff
@@ -871,6 +872,13 @@ private static void verifySpans(List<String> actualSpanItems, List<String> expec
   private static void verifyCommonAttributes(SpanData span) {
     assertEquals(span.getAttributes().get(AttributeKey.stringKey("instance.name")), "my-instance");
     assertEquals(span.getAttributes().get(AttributeKey.stringKey("db.name")), "my-database");
+    assertEquals(span.getAttributes().get(AttributeKey.stringKey("gcp.service.name")), "Spanner");
+    assertEquals(
+        span.getAttributes().get(AttributeKey.stringKey("gcp.client.repo")),
+        "googleapis/java-spanner");
+    assertEquals(
+        span.getAttributes().get(AttributeKey.stringKey("gcp.client.version")),
+        GaxProperties.getLibraryVersion(TraceWrapper.class));
```
sakthivelmanii, line 881: Can we check the format as well? We are exactly putting the same code in tests as well

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TraceWrapper.java
```diff
@@ -204,6 +211,9 @@ Attributes createCommonAttributes(DatabaseId db) {
...
+    builder.put(GCP_CLIENT_REPO_KEY, "googleapis/java-spanner");
+    builder.put(GCP_CLIENT_VERSION_KEY, GaxProperties.getLibraryVersion(TraceWrapper.class));
     return builder.build();
```
sakthivelmanii, line 217: What about this attribute? `gcp.client.artifact`

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TraceWrapper.java
surbhigarg92: There is no standard mentioned . We can decide what we want to use. This will be used to filter traces in our internal dashboards

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TraceWrapper.java
surbhigarg92: This is not a mandatory artifact. IMO, we don't need this attribute, as we don't have multiple artifacts on which we want to filter our traces.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/OpenTelemetrySpanTest.java
surbhigarg92: When running these unit tests on local, this comes as blank. I have tested the actual value while running the sample. We have used the same way to fetch the version at multiple places in our code

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TraceWrapper.java
sakthivelmanii: They used firestore as an example. That's why I was wondering if there's any standard such as lowercase

google-cloud-spanner/src/test/java/com/google/cloud/spanner/OpenTelemetrySpanTest.java
sakthivelmanii: Got it. For me I felt this is assertion is redundant because we are comparing like 1==1 which is always true. If there is no other way, we can leave this as it is

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TraceWrapper.java
sakthivelmanii: I see thanks

sakthivelmanii (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TraceWrapper.java
surbhigarg92: modified to lowercase

PR #3604, opened 2025-01-20
fix(spanner): fix unit tests for rw mux

harshachinta:


harshachinta:
This will be fixed in https://github.com/googleapis/java-spanner/pull/3598

PR #3605, opened 2025-01-22
fix(spanner): moved mTLSContext configurator from builder to construtor

sagnghos:


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -741,7 +741,17 @@ protected SpannerOptions(Builder builder) {
 
     transportChannelExecutorThreadNameFormat = builder.transportChannelExecutorThreadNameFormat;
     channelProvider = builder.channelProvider;
-    channelConfigurator = builder.channelConfigurator;
+    if (builder.mTLSContext != null) {
```
olavloite, line 744: nit: I think it would be better if we first check if both `channelConfigurator` and `mTLSContext` are non-null, and if so, throw an error. Now setting both (implicitly) means that the channelConfigurator is ignored. (Or alternatively: Make sure that both are used)

sagnghos: change code to use the channelConfigurator as well when not null.

PR #3628, opened 2025-02-10
chore(spanner): fix tests cases for mux rw

harshachinta:


google-cloud-spanner/src/test/java/com/google/cloud/spanner/MockSpannerServiceImpl.java
```diff
@@ -2024,14 +2024,6 @@ public void commit(CommitRequest request, StreamObserver<CommitResponse> respons
       return;
     }
     sessionLastUsed.put(session.getName(), Instant.now());
-    if (session.getMultiplexed()
```
harshachinta, line 2027: Removing this since this no longer will be the case in backend.

PR #3631, opened 2025-02-10
chore(spanner): Implement fallback for multiplexed session on Partitioned Operations

pratickchokhani:
- handle server side kill switch for multiplexed sessions with Partitioned Ops.


google-cloud-spanner/src/main/java/com/google/cloud/spanner/BatchClientImpl.java
```diff
@@ -256,15 +288,30 @@ public List<Partition> partitionQuery(
...
       final PartitionQueryRequest request = builder.build();
-      PartitionResponse response = rpc.partitionQuery(request, options);
-      ImmutableList.Builder<Partition> partitions = ImmutableList.builder();
-      for (com.google.spanner.v1.Partition p : response.getPartitionsList()) {
-        Partition partition =
-            Partition.createQueryPartition(
-                p.getPartitionToken(), partitionOptions, statement, queryOptions);
-        partitions.add(partition);
+      try {
+
```
olavloite, line 292: nit: remove empty line

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BatchClientImpl.java
```diff
@@ -256,15 +288,30 @@ public List<Partition> partitionQuery(
...
+        }
+        return partitions.build();
+      } catch (SpannerException e) {
+        maybeMarkUnimplementedForPartitionedOps(e);
+        throw e;
+      }
+    }
+
+    void maybeMarkUnimplementedForPartitionedOps(SpannerException spannerException) {
+      if (spannerException.getErrorCode() == ErrorCode.INVALID_ARGUMENT
```
olavloite, line 309: Are you sure that the error code will be `INVALID_ARGUMENT` (and not `UNIMPLEMENTED`)?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BatchClientImpl.java
```diff
@@ -59,6 +61,12 @@ public class BatchClientImpl implements BatchClient {
   @GuardedBy("multiplexedSessionLock")
   private final AtomicReference<SessionImpl> multiplexedSessionReference;
 
+  /**
+   * This flag is set to true if the server return UNIMPLEMENTED when partitioned transaction is
+   * executed on a multiplexed session. TODO: Remove once this is guaranteed to be available.
+   */
+  @VisibleForTesting final AtomicBoolean unimplementedForPartitionedOps = new AtomicBoolean(false);
```
olavloite, line 68: Can/should we make this a static field, so that it is shared across all batch clients? I assume that once this flag is set, then we should just fall back to regular sessions for all clients, and not first try multiplexed sessions for each new batch client.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BatchClientImpl.java
pratickchokhani: Yes, for Partitioned Read or Query, we are returning INVALID_ARGUMENT.

olavloite (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/BatchClientImpl.java
harshachinta: @pratickchokhani 
As discussed in the meeting, can we temporarily remove the error code check?

Context:
We plan to update the backend to return the UNIMPLEMENTED error for partitioned operations. However, this change will not be part of the current release and will be introduced in a follow-up update.

If we proceed with this error code check now, it will break backward compatibility when the backend starts returning UNIMPLEMENTED. To avoid this issue, we are removing the check for now and will reintroduce it once the backend update is complete.

cc: @rahul2393 to do the same for golang.

harshachinta:
@pratickchokhani 
Please fix the error code discussion before merging this commit.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClient.java
```diff
@@ -316,7 +327,18 @@ && verifyErrorMessage(
     }
   }
 
-  private boolean verifyErrorMessage(SpannerException spannerException, String message) {
+  boolean maybeMarkUnimplementedForPartitionedOps(SpannerException spannerException) {
+    if (spannerException.getErrorCode() == ErrorCode.UNIMPLEMENTED
```
harshachinta, line 331: The backend does not throw UNIMPLMENETED now right? I think it now returns INVALID_ARGUMENT.
If that is correct, then we should remove error code check now and add it back once backend throws UNIMPLEMENTED.
With this change we are not really falling back since this check will always fail.
Correct me if i am missing something

harshachinta (approved):

PR #3636, opened 2025-02-11
chore(spanner): fix code and tests

harshachinta:


google-cloud-spanner/src/test/java/com/google/cloud/spanner/AsyncTransactionManagerImplTest.java
```diff
@@ -60,67 +60,4 @@ public void testCommitReturnsCommitStats() {
       verify(transaction).commitAsync();
     }
   }
-
-  @Test
-  public void testRetryUsesPreviousTransactionIdOnMultiplexedSession() {
```
harshachinta, line 65: Removing this unit test since it is not possible to change the txnstate of the TransactionManager using unit tests.
Will look at this later if there is a way to fix this test
We have same tests covered in mock spanner https://github.com/googleapis/java-spanner/blob/c8d1e63cb79ee18221dcdad91518842593cc8ded/google-cloud-spanner/src/test/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClientMockServerTest.java#L75

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITBatchReadTest.java
```diff
@@ -181,7 +181,9 @@ public static void setUpDatabase() throws Exception {
           totalSize = 0;
         }
       }
-      dbClient.write(mutations);
+      if (mutations.size() > 0) {
```
olavloite, line 184: nit: do we really need this check? (and if so, could we change is to `!mutations.isEmpty()`)

olavloite (approved):

PR #3638, opened 2025-02-12
fix(spanner): release resources in TransactionManager

harshachinta:
The `close()` method in `TransactionManager` was only being called when used within a `try-with-resources` block, as `TransactionManager` implements `AutoCloseable`.  

However, in cases where `try-with-resources` was not used or `.close()` was not explicitly called, `close()` was never invoked, leading to resource leaks—such as spans not being closed properly. This is particularly an issue in JDBC, where `try-with-resources` cannot be used.  

This PR ensures that `close()` is implicitly invoked whenever the transaction reaches an end state, preventing resource leaks.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionManagerImpl.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionManagerImpl.java
@@ -81,8 +81,8 @@ final class TransactionManagerImpl implements TransactionManager, SessionTransac
       txnState = TransactionState.COMMIT_FAILED;
       throw e2;
     } finally {
-      // At this point, if the TransactionState is not ABORTED, then the transaction has reached a
-      // terminal state.
+      // At this point, if the TransactionState is not ABORTED, then the transaction has reached an
+      // end state.
       // We can safely call close() to release resources.
       if (getState() != TransactionState.ABORTED) {
         close();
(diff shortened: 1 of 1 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionManagerImpl.java
```diff
@@ -80,6 +80,13 @@ public void commit() {
     } catch (SpannerException e2) {
       txnState = TransactionState.COMMIT_FAILED;
       throw e2;
+    } finally {
+      // At this point, if the TransactionState is not ABORTED, then the transaction has reached a
+      // terminal state.
```
olavloite, line 85: super-nit: I would recommend using `end state` instead of `terminal state`.

olavloite (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionManagerImpl.java
harshachinta: sure, I have used this term from existing code. But end state is more clear in terms of readability.
https://github.com/googleapis/java-spanner/blob/52915cdb829ed20259ba34e2b69fa163d51cab1a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionManager.java#L49

PR #3643, opened 2025-02-12
chore(x-goog-spanner-request-id): commit foundation

odeke-em:
This change adds in the bases to bring in the functionality for the "x-goog-spanner-request-id" that'll be used for debugging without the limits of trace continuity and sampling.

Updates #3537

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
@@ -25,7 +25,8 @@ import java.util.Objects;
 @InternalApi
 public class XGoogSpannerRequestId {
   // 1. Generate the random process Id singleton.
-  @VisibleForTesting static String RAND_PROCESS_ID = XGoogSpannerRequestId.generateRandProcessId();
+  @VisibleForTesting
+  static final String RAND_PROCESS_ID = XGoogSpannerRequestId.generateRandProcessId();
 
   @VisibleForTesting
   static final long VERSION = 1; // The version of the specification being implemented.
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
@@ -47,7 +47,8 @@ public class XGoogSpannerRequestId {
   }
 
   private static String generateRandProcessId() {
-    return String.format("%08x", new SecureRandom().nextInt());
+    // Expecting to use 64-bits of randomness to avoid clashes.
+    return String.format("%016x", new SecureRandom().nextInt(64));
   }
 
   @Override
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
@@ -18,6 +18,7 @@ package com.google.cloud.spanner;
 
 import com.google.api.core.InternalApi;
 import com.google.common.annotations.VisibleForTesting;
+import java.math.BigInteger;
 import java.security.SecureRandom;
 import java.util.Objects;
 
@@ -48,7 +49,8 @@ public class XGoogSpannerRequestId {
 
   private static String generateRandProcessId() {
     // Expecting to use 64-bits of randomness to avoid clashes.
-    return String.format("%016x", new SecureRandom().nextInt(64));
+    BigInteger bigInt = new BigInteger(64, new SecureRandom());
+    return String.format("%016x", bigInt);
   }
 
   @Override
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2025 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.spanner;
+
+import com.google.api.core.InternalApi;
+import com.google.common.annotations.VisibleForTesting;
+import java.security.SecureRandom;
+import java.util.Objects;
+
+@InternalApi
+public class XGoogSpannerRequestId {
+  // 1. Generate the random process Id singleton.
+  @VisibleForTesting static String RAND_PROCESS_ID = XGoogSpannerRequestId.generateRandProcessId();
+
+  @VisibleForTesting
+  static final long VERSION = 1; // The version of the specification being implemented.
+
+  private final long nthClientId;
+  private final long nthChannelId;
+  private final long nthRequest;
+  private long attempt;
+
+  XGoogSpannerRequestId(long nthClientId, long nthChannelId, long nthRequest, long attempt) {
+    this.nthClientId = nthClientId;
+    this.nthChannelId = nthChannelId;
+    this.nthRequest = nthRequest;
+    this.attempt = attempt;
+  }
+
+  public static XGoogSpannerRequestId of(
+      long nthClientId, long nthChannelId, long nthRequest, long attempt) {
+    return new XGoogSpannerRequestId(nthClientId, nthChannelId, nthRequest, attempt);
+  }
+
+  private static String generateRandProcessId() {
+    return String.format("%08x", new SecureRandom().nextInt());
+  }
+
+  @Override
+  public String toString() {
+    return String.format(
+        "%d.%s.%d.%d.%d.%d",
+        XGoogSpannerRequestId.VERSION,
+        XGoogSpannerRequestId.RAND_PROCESS_ID,
+        this.nthClientId,
+        this.nthChannelId,
+        this.nthRequest,
+        this.attempt);
+  }
+
+  @Override
+  public boolean equals(Object other) {
+    // instanceof for a null object returns false.
+    if (!(other instanceof XGoogSpannerRequestId)) {
+      return false;
+    }
+
+    XGoogSpannerRequestId otherReqId = (XGoogSpannerRequestId) (other);
+
+    return Objects.equals(this.nthClientId, otherReqId.nthClientId)
+        && Objects.equals(this.nthChannelId, otherReqId.nthChannelId)
+        && Objects.equals(this.nthRequest, otherReqId.nthRequest)
+        && Objects.equals(this.attempt, otherReqId.attempt);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(this.nthClientId, this.nthChannelId, this.nthRequest, this.attempt);
+  }
+}
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/XGoogSpannerRequestIdTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/XGoogSpannerRequestIdTest.java
@@ -29,7 +29,7 @@ import org.junit.runners.JUnit4;
 @RunWith(JUnit4.class)
 public class XGoogSpannerRequestIdTest {
   private static final Pattern REGEX_RAND_PROCESS_ID =
-      Pattern.compile("1.([0-9a-z]{8})(\\.\\d+){3}\\.(\\d+)$");
+      Pattern.compile("1.([0-9a-z]{16})(\\.\\d+){3}\\.(\\d+)$");
 
   @Test
   public void testEquals() {
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/XGoogSpannerRequestIdTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/XGoogSpannerRequestIdTest.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2025 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.spanner;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertTrue;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public class XGoogSpannerRequestIdTest {
+  private static final Pattern REGEX_RAND_PROCESS_ID =
+      Pattern.compile("1.([0-9a-z]{8})(\\.\\d+){3}\\.(\\d+)$");
+
+  @Test
+  public void testEquals() {
+    XGoogSpannerRequestId reqID1 = XGoogSpannerRequestId.of(1, 1, 1, 1);
+    XGoogSpannerRequestId reqID2 = XGoogSpannerRequestId.of(1, 1, 1, 1);
+    assertEquals(reqID1, reqID2);
+    assertEquals(reqID1, reqID1);
+    assertEquals(reqID2, reqID2);
+
+    XGoogSpannerRequestId reqID3 = XGoogSpannerRequestId.of(1, 1, 1, 2);
+    assertNotEquals(reqID1, reqID3);
+    assertNotEquals(reqID3, reqID1);
+    assertEquals(reqID3, reqID3);
+  }
+
+  @Test
+  public void testEnsureHexadecimalFormatForRandProcessID() {
+    String str = XGoogSpannerRequestId.of(1, 2, 3, 4).toString();
+    Matcher m = XGoogSpannerRequestIdTest.REGEX_RAND_PROCESS_ID.matcher(str);
+    assertTrue(m.matches());
+  }
+}


google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
```diff
@@ -0,0 +1,79 @@
...
+ * limitations under the License.
+ */
+
+package com.google.cloud.spanner;
+
+import java.security.SecureRandom;
+
+public class XGoogSpannerRequestId {
+  // 1. Generate the random process Id singleton.
+  public static String RAND_PROCESS_ID = XGoogSpannerRequestId.generateRandProcessId();
```
sakthivelmanii, line 23: why does it have to be `public`

google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
```diff
@@ -0,0 +1,79 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.spanner;
+
+import java.security.SecureRandom;
+
+public class XGoogSpannerRequestId {
```
sakthivelmanii, line 21: Does it have to public?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
```diff
@@ -0,0 +1,79 @@
...
+  // 1. Generate the random process Id singleton.
+  public static String RAND_PROCESS_ID = XGoogSpannerRequestId.generateRandProcessId();
+  public static long VERSION = 1; // The version of the specification being implemented.
```
sakthivelmanii, line 24: Does this also needs to be public?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
sakthivelmanii: if it needs for testing, then you can use `@VisibleForTesting`

google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
odeke-em: Yup! It is going to be used all around within other packages.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
odeke-em: For testing indeed.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
odeke-em: Added visible for testing @sakthivelmanii, thanks for the reviewin.

odeke-em:
@sakthivelmanii thanks for the first round of comments! I've addressed them, please take a look and also kindly help me run the bots.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
sakthivelmanii: Isn't this used only in the interceptor? If not, if you require at multiple places, please annotate it with `@InternalApi`

google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
```diff
@@ -0,0 +1,84 @@
...
+
+package com.google.cloud.spanner;
+
+import com.google.common.annotations.VisibleForTesting;
+import java.security.SecureRandom;
+
+public class XGoogSpannerRequestId {
+  // 1. Generate the random process Id singleton.
+  @VisibleForTesting
+  public static String RAND_PROCESS_ID = XGoogSpannerRequestId.generateRandProcessId();
```
sakthivelmanii, line 25: You have to make it as package-private.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
sakthivelmanii: You have to make it as package-private

google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
```diff
@@ -0,0 +1,84 @@
...
+
+  @VisibleForTesting
+  public static long VERSION = 1; // The version of the specification being implemented.
+
+  private long nthClientId;
+  private long nthChannelId;
+  private long nthRequest;
+  private long attempt;
+
+  public XGoogSpannerRequestId(long nthClientId, long nthChannelId, long nthRequest, long attempt) {
```
sakthivelmanii, line 35: if you have `of` static method why do you need constructor to be public?

odeke-em: 2 ways of creating it: there are instances in which I'll need the director constructor; your case works great with singletons but RequestId will be used all around. It is much more natural to have that constructor public, and also going to be later updated with the docs for easier following with.

sakthivelmanii: Sorry I still don't understand. why can't you follow one single way everywhere either through constructor or static method?

sakthivelmanii:
@odeke-em mark this PR as chore. feat creates a new minor version. This change really does nothing. In upcoming PRs where you are using this class, you can define it as feat.

odeke-em:
thank you @sakthivelmanii I didn't know that about automatically creating releases: today I learnt!

google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
odeke-em: It feels a little pedantic to be worrying about that right now given that deep in the code I use both, but sure I'll send an update.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
```diff
@@ -0,0 +1,84 @@
...
+
+  private static String generateRandProcessId() {
+    byte[] rBytes = new byte[8];
+    SecureRandom srng = new SecureRandom();
+    srng.nextBytes(rBytes);
+    int result =
+        Byte.toUnsignedInt(rBytes[0])
+            | Byte.toUnsignedInt(rBytes[1]) << 8
+            | Byte.toUnsignedInt(rBytes[2]) << 16
+            | Byte.toUnsignedInt(rBytes[3]) << 24;
```
olavloite, line 55: This won't generate an unsigned int (Java does not have an unsigned int type), so there's no need to make it this complicated. Also, the `toHexString(..)` method does not prefix the string with zeros, meaning that the current test case (which does assume that the length is 8) will be flaky.

It's probably the easiest to just do the following:

```java
return String.format("%08x", new SecureRandom().nextInt());
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
```diff
@@ -0,0 +1,84 @@
...
+
+    return Integer.toHexString(result);
+  }
+
+  public String toString() {
+    return String.format(
+        "%d.%s.%d.%d.%d.%d",
+        this.VERSION,
+        this.RAND_PROCESS_ID,
```
olavloite, line 64: Use static reference:

```suggestion
        XGoogSpannerRequestId.VERSION,
        XGoogSpannerRequestId.RAND_PROCESS_ID,
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
```diff
@@ -0,0 +1,84 @@
...
+        this.nthChannelId,
+        this.nthRequest,
+        this.attempt);
+  }
+
+  @Override
+  public boolean equals(Object other) {
+    if (other == null) {
+      return false;
+    }
```
olavloite, line 75: nit: this is redundant, as the `instanceof` check below also returns false for null.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
```diff
@@ -0,0 +1,84 @@
...
+        this.VERSION,
+        this.RAND_PROCESS_ID,
+        this.nthClientId,
+        this.nthChannelId,
+        this.nthRequest,
+        this.attempt);
+  }
+
+  @Override
+  public boolean equals(Object other) {
```
olavloite, line 72: The Java spec requires that if you implement `equals` for a class, then you *must* also implement `hashCode`, and you must guarantee that if `equals` returns true for two objects, then they must also return the same `hashCode`. For value classes like this, a simple way to implement hashCode is like this:

```java
  @Override
  public int hashCode() {
    return java.util.Objects.hash(this.nthClientId, this.nthChannelId, this.nthRequest, this.attempt);
  }
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
```diff
@@ -0,0 +1,84 @@
...
+    if (other == null) {
+      return false;
+    }
+
+    if (!(other instanceof XGoogSpannerRequestId)) {
+      return false;
+    }
+
+    XGoogSpannerRequestId otherReqId = (XGoogSpannerRequestId) (other);
+    return otherReqId.toString().equals(this.toString());
```
olavloite, line 82: Using `toString()` for equality can be dangerous, as `toString()` is intended for human consumption. Use something like this instead:

```java
    return Objects.equals(this.nthClientId, otherReqId.nthClientId)
        && Objects.equals(this.nthChannelId, otherReqId.nthChannelId)
        && Objects.equals(this.nthRequest, otherReqId.nthRequest)
        && Objects.equals(this.attempt, otherReqId.attempt);
```

google-cloud-spanner/src/test/java/com/google/cloud/spanner/XGoogSpannerRequestIdTest.java
```diff
@@ -0,0 +1,53 @@
...
+public class XGoogSpannerRequestIdTest {
+  private static String RAND_PROCESS_ID = XGoogSpannerRequestId.RAND_PROCESS_ID;
+  private static Pattern REGEX_RAND_PROCESS_ID =
+      Pattern.compile("1.([0-9a-z]{8})(\\.\\d+){3}\\.(\\d+)$");
+
+  @Test
+  public void testEquals() {
+    XGoogSpannerRequestId reqID1 = XGoogSpannerRequestId.of(1, 1, 1, 1);
+    XGoogSpannerRequestId reqID2 = XGoogSpannerRequestId.of(1, 1, 1, 1);
+    assertThat(reqID1).isEqualTo(reqID2);
```
olavloite, line 37: nit: prefer the use of `org.junit.Assert.assertEquals` over `assertThat(..).isEqualTo(..)`

google-cloud-spanner/src/test/java/com/google/cloud/spanner/XGoogSpannerRequestIdTest.java
```diff
@@ -0,0 +1,53 @@
...
+    assertThat(reqID1).isNotEqualTo(reqID3);
+    assertThat(reqID3).isNotEqualTo(reqID1);
+    assertThat(reqID3).isEqualTo(reqID3);
+  }
+
+  @Test
+  public void testEnsureHexadecimalFormatForRandProcessID() {
+    String str = XGoogSpannerRequestId.of(1, 2, 3, 4).toString();
+    Matcher m = REGEX_RAND_PROCESS_ID.matcher(str);
+    assertThat(m.matches()).isEqualTo(true);
```
olavloite, line 51: nit: prefer the use of `assertTrue(..)` over `assertThat(..).isEqualTo(..)`

Note also that this is currently flaky, as the process ID string is not guaranteed to always contain 8 characters if the random function happens to return a lower number.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/XGoogSpannerRequestIdTest.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public class XGoogSpannerRequestIdTest {
+  private static String RAND_PROCESS_ID = XGoogSpannerRequestId.RAND_PROCESS_ID;
```
olavloite, line 29:
```suggestion
  private static String final RAND_PROCESS_ID = XGoogSpannerRequestId.RAND_PROCESS_ID;
```

google-cloud-spanner/src/test/java/com/google/cloud/spanner/XGoogSpannerRequestIdTest.java
```diff
@@ -0,0 +1,53 @@
...
+  private static Pattern REGEX_RAND_PROCESS_ID =
```
olavloite, line 30:
```suggestion
  private static final Pattern REGEX_RAND_PROCESS_ID =
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
odeke-em: Thank you @sakthivelmanii, I added that annotation.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
```diff
@@ -0,0 +1,84 @@
...
+    int result =
+        Byte.toUnsignedInt(rBytes[0])
+            | Byte.toUnsignedInt(rBytes[1]) << 8
+            | Byte.toUnsignedInt(rBytes[2]) << 16
+            | Byte.toUnsignedInt(rBytes[3]) << 24;
+
+    return Integer.toHexString(result);
+  }
+
+  public String toString() {
```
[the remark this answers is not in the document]
odeke-em: Done, thanks!

google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
odeke-em: Got it and thank you @olavloite! Done, thanks.

odeke-em:
@sakthivelmanii @olavloite thanks for the reviews, please take another round.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
```diff
@@ -0,0 +1,84 @@
...
+  @VisibleForTesting
+  static String RAND_PROCESS_ID = XGoogSpannerRequestId.generateRandProcessId();
+
+  @VisibleForTesting
+  static long VERSION = 1; // The version of the specification being implemented.
+
+  private long nthClientId;
+  private long nthChannelId;
+  private long nthRequest;
+  private long attempt;
```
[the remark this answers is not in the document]
odeke-em: Done, thanks.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
```diff
@@ -0,0 +1,84 @@
...
+    this.attempt = attempt;
+  }
+
+  public static XGoogSpannerRequestId of(
+      long nthClientId, long nthChannelId, long nthRequest, long attempt) {
+    return new XGoogSpannerRequestId(nthClientId, nthChannelId, nthRequest, attempt);
+  }
+
+  private static String generateRandProcessId() {
+    byte[] rBytes = new byte[8];
```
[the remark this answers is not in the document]
odeke-em: Done, thanks.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
odeke-em: Done, thanks.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
odeke-em: Done, thanks.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
odeke-em: Done, thanks.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/XGoogSpannerRequestIdTest.java
odeke-em: Done, thanks.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/XGoogSpannerRequestIdTest.java
odeke-em: Done, thanks

google-cloud-spanner/src/test/java/com/google/cloud/spanner/XGoogSpannerRequestIdTest.java
odeke-em: Done, thanks

google-cloud-spanner/src/test/java/com/google/cloud/spanner/XGoogSpannerRequestIdTest.java
odeke-em: Done, thanks.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
odeke-em: Done, thanks! It was an odd one to learn about in Java.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
```diff
@@ -0,0 +1,85 @@
...
+
+import com.google.api.core.InternalApi;
+import com.google.common.annotations.VisibleForTesting;
+import java.security.SecureRandom;
+import java.util.Objects;
+
+@InternalApi
+public class XGoogSpannerRequestId {
+  // 1. Generate the random process Id singleton.
+  @VisibleForTesting static String RAND_PROCESS_ID = XGoogSpannerRequestId.generateRandProcessId();
```
olavloite, line 27: nit: I assume that this can also be `final`

google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  public static XGoogSpannerRequestId of(
+      long nthClientId, long nthChannelId, long nthRequest, long attempt) {
+    return new XGoogSpannerRequestId(nthClientId, nthChannelId, nthRequest, attempt);
+  }
+
+  private static String generateRandProcessId() {
+    // Expecting to use 64-bits of randomness to avoid clashes.
+    return String.format("%016x", new SecureRandom().nextInt(64));
```
olavloite, line 51: The `bound` argument to `nextInt(int)` does not indicate the number of bits to use, it indicates the highest number that should be generated. Meaning: This call will return a random value in the range `[0, 64)`. Also, an `int` in Java is by definition 32-bit, so the length of the hexadecimal representation of it will never be more than 8 characters (and the spec for the random process ID is that it should be a 32-bit unsigned int). So this should just be:

```suggestion
    return String.format("%08x", new SecureRandom().nextInt());
```

This will:
1. Generate a random 32-bit integer from the full 32-bit range (so including negative numbers)
2. The `%08x` format will convert it to the corresponding __unsigned__ hex format (so `-1` is formatted as `ffffffff`)

google-cloud-spanner/src/test/java/com/google/cloud/spanner/XGoogSpannerRequestIdTest.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public class XGoogSpannerRequestIdTest {
+  private static final Pattern REGEX_RAND_PROCESS_ID =
+      Pattern.compile("1.([0-9a-z]{16})(\\.\\d+){3}\\.(\\d+)$");
```
olavloite, line 32: The process ID should be a 32-bit unsigned number, so length 8 in hex format:

```suggestion
      Pattern.compile("1.([0-9a-z]{8})(\\.\\d+){3}\\.(\\d+)$");
```

odeke-em: @olavloite we had a long discussion with Simil and team in the same group that you are, about why it should be 64-bits due to higher collisions if it were 32-bits per https://orijtech.notion.site/x-goog-spanner-request-id-always-on-gRPC-header-to-aid-in-quick-debugging-of-errors-14aba6bc91348091a58fca7a505c9827#14aba6bc9134807f9f17d3ef58e26128

google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
odeke-em: Hey @olavloite thanks for the discourse and for pointing that out. I instead went with the use of BigInteger plus SecureRandom to generate a 64-bit integer. Also as per https://orijtech.notion.site/x-goog-spanner-request-id-always-on-gRPC-header-to-aid-in-quick-debugging-of-errors-14aba6bc91348091a58fca7a505c9827#14aba6bc9134807f9f17d3ef58e26128 I discussed the problems of using a 32-bit random integer and potential clashes, meanwhile 64-bit random integer is the right balance and also in our Google group.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/XGoogSpannerRequestId.java
odeke-em: Done and thank you @olavloite, PTAL!

olavloite (approved):

PR #3647, opened 2025-02-14
feat: add option to indicate that a statement is the last in a transaction

olavloite:
- Adds an option to indicate that a statement is the last in a read/write transaction. This option should only be set for the last statement before committing the transaction. Trying to execute any other statements after this statement will result in an error.
- The `last_statement` option is automatically set for statements that are executed in autocommit mode in the Connection API. This option will also automatically be set for statements executed on a JDBC connection in autocommit mode.

cc @sgorse123

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/SingleUseTransaction.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/SingleUseTransaction.java
@@ -34,6 +34,7 @@ import com.google.cloud.spanner.ErrorCode;
 import com.google.cloud.spanner.Mutation;
 import com.google.cloud.spanner.Options;
 import com.google.cloud.spanner.Options.QueryOption;
+import com.google.cloud.spanner.Options.QueryUpdateOption;
 import com.google.cloud.spanner.Options.UpdateOption;
 import com.google.cloud.spanner.PartitionOptions;
 import com.google.cloud.spanner.ReadOnlyTransaction;
@@ -298,7 +299,8 @@ class SingleUseTransaction extends AbstractBaseUnitOfWork {
                 writeTransaction.run(
                     transaction ->
                         DirectExecuteResultSet.ofResultSet(
-                            transaction.executeQuery(update.getStatement(), options)));
+                            transaction.executeQuery(
+                                update.getStatement(), appendLastStatement(options))));
             state = UnitOfWorkState.COMMITTED;
             return resultSet;
           } catch (Throwable t) {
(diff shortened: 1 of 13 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/SingleUseTransaction.java
```diff
@@ -582,6 +588,29 @@ private ApiFuture<Tuple<Long, ResultSet>> executeTransactionalUpdateAsync(
...
+
+  private static QueryOption[] appendLastStatement(QueryOption[] options) {
+    if (options.length == 0) {
+      return LAST_STATEMENT_OPTIONS;
+    }
+    QueryOption[] result = new QueryOption[options.length + 1];
+    System.arraycopy(options, 0, result, 0, options.length);
+    result[result.length - 1] = LAST_STATEMENT_OPTIONS[0];
+    return result;
+  }
```
olavloite, line 612: Note: We cannot generalize these two methods into one, due to the large number of different possible interfaces for Query/Update/Read/QueryOrUpdate/QueryOrUpdateOrTransaction/Options.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/SingleUseTransaction.java
```diff
@@ -298,7 +299,8 @@ private ApiFuture<ResultSet> executeDmlReturningAsync(
                 writeTransaction.run(
                     transaction ->
                         DirectExecuteResultSet.ofResultSet(
-                            transaction.executeQuery(update.getStatement(), options)));
+                            transaction.executeQuery(
+                                update.getStatement(), appendLastStatement(options))));
```
harshachinta, line 303: Quick clarification: Since a SingleUseTransaction can only execute one query, do we implicitly mark that statement as the last statement? The changes in this file seem to be related to this, correct?

What happens if we don’t mark it as the last statement in SingleUseTransaction?
Since the customer cannot run additional queries anyway, I’m trying to understand the benefit of marking it as the last statement.

harshachinta (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/SingleUseTransaction.java
olavloite: Note: This is only relevant in read/write transactions, so in this case only for DML statements with a THEN RETURN clause, which are also executed using `executeQuery(..)`.

Your observation that we automatically mark a statement in a SingleUseTransaction as the last_statement is correct. We do this, because we know that this will be the only statement in this transaction. By giving Spanner this hint, Spanner can internally optimise the execution of this statement by skipping some verifications, that are also executed during the Commit call.

And more in general: By supplying this hint to Spanner, Spanner could in the future also consider other possible optimizations that are possible when it knows that the current statement will be the last in a read/write transaction.

PR #3651, opened 2025-02-18
deps: stop declaring own exporter-trace version

suztomo:
I think the exporter-trace version is already declared in the google-cloud-shared-config.

This should fix the upper-bound failure in https://github.com/googleapis/java-shared-config/pull/985#issuecomment-2666226982.

CC: @lqiu96


suztomo:
It failed:

```
Error:  'dependencies.dependency.version' for com.google.cloud.opentelemetry:exporter-trace:jar is missing. @ line 45, column 17
 @ 
```

PR #3661, opened 2025-02-26
perf: Skip gRPC trailers for StreamingRead & ExecuteStreamingSql

sakthivelmanii:

Fixes: b/394475581

**Description:**
gRPC trailers are headers which is usually sent after all the messages are sent by the server in the stream. Trailers usually causes additional latency of 250-300 μs after last message is sent in the stream. From the server, we will receiving last field which tells us about last PartialResultSet and we no longer will be waiting for trailers to assume that all the messages are received.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/GrpcStreamIterator.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/GrpcStreamIterator.java
@@ -52,6 +52,7 @@ class GrpcStreamIterator extends AbstractIterator<PartialResultSet>
   private TimeUnit streamWaitTimeoutUnit;
   private long streamWaitTimeoutValue;
   private SpannerException error;
+  private boolean done;
 
   @VisibleForTesting
   GrpcStreamIterator(int prefetchChunks, boolean cancelQueryWhenClientIsClosed) {
@@ -166,11 +167,17 @@ class GrpcStreamIterator extends AbstractIterator<PartialResultSet>
     @Override
     public void onPartialResultSet(PartialResultSet results) {
       addToStream(results);
+      if (results.getLast()) {
+        done = true;
+        addToStream(END_OF_STREAM);
+      }
     }
 
     @Override
     public void onCompleted() {
-      addToStream(END_OF_STREAM);
+      if (!done) {
+        addToStream(END_OF_STREAM);
+      }
     }
 
     @Override
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/GrpcResultSetTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/GrpcResultSetTest.java
@@ -19,6 +19,7 @@ package com.google.cloud.spanner;
 import static com.google.common.testing.SerializableTester.reserialize;
 import static com.google.common.truth.Truth.assertThat;
 import static org.junit.Assert.assertEquals;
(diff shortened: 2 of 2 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/GrpcStreamIterator.java
```diff
@@ -52,6 +52,7 @@ class GrpcStreamIterator extends AbstractIterator<PartialResultSet>
   private TimeUnit streamWaitTimeoutUnit;
   private long streamWaitTimeoutValue;
   private SpannerException error;
+  private volatile boolean done;
```
olavloite, line 55: I don't think this needs to be `volatile`. It will only be used by the gRPC thread that handles this stream. At least, that is if my assumption that `onPartialResultSet` and `onCompleted` are both only called by that thread.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/GrpcResultSetTest.java
```diff
@@ -1115,4 +1115,58 @@ public void getProtoEnumList() {
...
+    long[] longArray = {111, 333, 444, 0, -1, -2234, Long.MAX_VALUE, Long.MIN_VALUE};
+
+    consumer.onPartialResultSet(
+        PartialResultSet.newBuilder()
+            .setMetadata(
+                makeMetadata(Type.struct(Type.StructField.of("f", Type.array(Type.int64())))))
+            .addValues(Value.int64Array(longArray).toProto())
+            .setLast(false)
+            .build());
+    assertThat(resultSet.next()).isTrue();
```
olavloite, line 1130: nit: prefer the use of `assertTrue` and other junit assertions instead of `com.google.truth`.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/GrpcResultSetTest.java
```diff
@@ -1115,4 +1115,58 @@ public void getProtoEnumList() {
...
+    SpannerException spannerException =
+        assertThrows(
+            SpannerException.class,
+            () -> {
+              assertThat(resultSet.next()).isFalse();
```
olavloite, line 1167: How long does this take to execute?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/GrpcStreamIterator.java
sakthivelmanii: I agree. atleast grpc threads will be accessing it sequentially not in parallel.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/GrpcResultSetTest.java
sakthivelmanii: Done.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/GrpcResultSetTest.java
sakthivelmanii: 1 nano second https://github.com/googleapis/java-spanner/blob/main/google-cloud-spanner/src/test/java/com/google/cloud/spanner/GrpcResultSetTest.java#L67

olavloite (approved):

PR #3670, opened 2025-03-03
feat: support isolation level REPEATABLE_READ for R/W transactions

shobhitsg:
Isolation level support added in various write APIs:
```
write
writeWithOptions
writeAtleastOnceWithOptions
readWriteTransaction
transactionManager
transactionManagerAsync
runAsync
```
Note: Samples and integration tests will be added in separate PRs.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java
@@ -111,6 +111,10 @@ public interface DatabaseClient {
    *       applied to any other requests on the transaction.
    *   <li>{@link Options#commitStats()}: Request that the server includes commit statistics in the
    *       {@link CommitResponse}.
+   *   <li>{@link Options#repeatableReadIsolationLevel()}: Request Repeatable Read Isolation Level
+   *       from the backend.
+   *   <li>{@link Options#serializableIsolationLevel()}: Request Serializable Isolation Level from
+   *       the backend.
    * </ul>
    *
    * @return a response with the timestamp at which the write was committed
@@ -186,6 +190,10 @@ public interface DatabaseClient {
    *       applied to any other requests on the transaction.
    *   <li>{@link Options#commitStats()}: Request that the server includes commit statistics in the
    *       {@link CommitResponse}.
+   *   <li>{@link Options#repeatableReadIsolationLevel()}: Request Repeatable Read Isolation Level
+   *       from the backend.
+   *   <li>{@link Options#serializableIsolationLevel()}: Request Serializable Isolation Level from
+   *       the backend.
    * </ul>
    *
    * @return a response with the timestamp at which the write was committed
@@ -414,6 +422,10 @@ public interface DatabaseClient {
    *       applied to any other requests on the transaction.
    *   <li>{@link Options#commitStats()}: Request that the server includes commit statistics in the
    *       {@link CommitResponse}.
+   *   <li>{@link Options#repeatableReadIsolationLevel()}: Request Repeatable Read Isolation Level
+   *       from the backend.
+   *   <li>{@link Options#serializableIsolationLevel()}: Request Serializable Isolation Level from
+   *       the backend.
    * </ul>
    */
   TransactionRunner readWriteTransaction(TransactionOption... options);
@@ -454,6 +466,10 @@ public interface DatabaseClient {
    *       applied to any other requests on the transaction.
    *   <li>{@link Options#commitStats()}: Request that the server includes commit statistics in the
    *       {@link CommitResponse}.
+   *   <li>{@link Options#repeatableReadIsolationLevel()}: Request Repeatable Read Isolation Level
+   *       from the backend.
+   *   <li>{@link Options#serializableIsolationLevel()}: Request Serializable Isolation Level from
+   *       the backend.
    * </ul>
    */
   TransactionManager transactionManager(TransactionOption... options);
@@ -494,6 +510,10 @@ public interface DatabaseClient {
    *       applied to any other requests on the transaction.
    *   <li>{@link Options#commitStats()}: Request that the server includes commit statistics in the
    *       {@link CommitResponse}.
+   *   <li>{@link Options#repeatableReadIsolationLevel()}: Request Repeatable Read Isolation Level
+   *       from the backend.
+   *   <li>{@link Options#serializableIsolationLevel()}: Request Serializable Isolation Level from
+   *       the backend.
    * </ul>
    */
   AsyncRunner runAsync(TransactionOption... options);
@@ -548,6 +568,10 @@ public interface DatabaseClient {
    *       applied to any other requests on the transaction.
    *   <li>{@link Options#commitStats()}: Request that the server includes commit statistics in the
    *       {@link CommitResponse}.
+   *   <li>{@link Options#repeatableReadIsolationLevel()}: Request Repeatable Read Isolation Level
+   *       from the backend.
+   *   <li>{@link Options#serializableIsolationLevel()}: Request Serializable Isolation Level from
+   *       the backend.
    * </ul>
    */
   AsyncTransactionManager transactionManagerAsync(TransactionOption... options);
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java
@@ -22,6 +22,7 @@ import com.google.cloud.spanner.Options.RpcPriority;
 import com.google.cloud.spanner.Options.TransactionOption;
 import com.google.cloud.spanner.Options.UpdateOption;
 import com.google.spanner.v1.BatchWriteResponse;
+import com.google.spanner.v1.TransactionOptions.IsolationLevel;
 
 /**
  * Interface for all the APIs that are used to read/write data into a Cloud Spanner database. An
@@ -111,10 +112,6 @@ public interface DatabaseClient {
    *       applied to any other requests on the transaction.
    *   <li>{@link Options#commitStats()}: Request that the server includes commit statistics in the
    *       {@link CommitResponse}.
-   *   <li>{@link Options#repeatableReadIsolationLevel()}: Request Repeatable Read Isolation Level
-   *       from the backend.
-   *   <li>{@link Options#serializableIsolationLevel()}: Request Serializable Isolation Level from
-   *       the backend.
    * </ul>
    *
    * @return a response with the timestamp at which the write was committed
@@ -190,10 +187,6 @@ public interface DatabaseClient {
    *       applied to any other requests on the transaction.
    *   <li>{@link Options#commitStats()}: Request that the server includes commit statistics in the
    *       {@link CommitResponse}.
-   *   <li>{@link Options#repeatableReadIsolationLevel()}: Request Repeatable Read Isolation Level
-   *       from the backend.
-   *   <li>{@link Options#serializableIsolationLevel()}: Request Serializable Isolation Level from
-   *       the backend.
    * </ul>
    *
    * @return a response with the timestamp at which the write was committed
@@ -422,10 +415,8 @@ public interface DatabaseClient {
    *       applied to any other requests on the transaction.
    *   <li>{@link Options#commitStats()}: Request that the server includes commit statistics in the
    *       {@link CommitResponse}.
-   *   <li>{@link Options#repeatableReadIsolationLevel()}: Request Repeatable Read Isolation Level
-   *       from the backend.
-   *   <li>{@link Options#serializableIsolationLevel()}: Request Serializable Isolation Level from
-   *       the backend.
+   *   <li>{@link Options#isolationLevelOption(IsolationLevel)}: The isolation level for the
+   *       transaction
    * </ul>
    */
   TransactionRunner readWriteTransaction(TransactionOption... options);
@@ -466,10 +457,8 @@ public interface DatabaseClient {
    *       applied to any other requests on the transaction.
    *   <li>{@link Options#commitStats()}: Request that the server includes commit statistics in the
    *       {@link CommitResponse}.
-   *   <li>{@link Options#repeatableReadIsolationLevel()}: Request Repeatable Read Isolation Level
-   *       from the backend.
-   *   <li>{@link Options#serializableIsolationLevel()}: Request Serializable Isolation Level from
-   *       the backend.
+   *   <li>{@link Options#isolationLevelOption(IsolationLevel)}: The isolation level for the
+   *       transaction
    * </ul>
    */
   TransactionManager transactionManager(TransactionOption... options);
@@ -510,10 +499,8 @@ public interface DatabaseClient {
    *       applied to any other requests on the transaction.
    *   <li>{@link Options#commitStats()}: Request that the server includes commit statistics in the
    *       {@link CommitResponse}.
-   *   <li>{@link Options#repeatableReadIsolationLevel()}: Request Repeatable Read Isolation Level
-   *       from the backend.
-   *   <li>{@link Options#serializableIsolationLevel()}: Request Serializable Isolation Level from
-   *       the backend.
+   *   <li>{@link Options#isolationLevelOption(IsolationLevel)}: The isolation level for the
+   *       transaction
    * </ul>
    */
   AsyncRunner runAsync(TransactionOption... options);
@@ -568,10 +555,8 @@ public interface DatabaseClient {
    *       applied to any other requests on the transaction.
    *   <li>{@link Options#commitStats()}: Request that the server includes commit statistics in the
    *       {@link CommitResponse}.
-   *   <li>{@link Options#repeatableReadIsolationLevel()}: Request Repeatable Read Isolation Level
-   *       from the backend.
-   *   <li>{@link Options#serializableIsolationLevel()}: Request Serializable Isolation Level from
-   *       the backend.
+   *   <li>{@link Options#isolationLevelOption(IsolationLevel)}: The isolation level for the
+   *       transaction
    * </ul>
    */
   AsyncTransactionManager transactionManagerAsync(TransactionOption... options);
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java
@@ -415,7 +415,7 @@ public interface DatabaseClient {
    *       applied to any other requests on the transaction.
    *   <li>{@link Options#commitStats()}: Request that the server includes commit statistics in the
    *       {@link CommitResponse}.
-   *   <li>{@link Options#isolationLevelOption(IsolationLevel)}: The isolation level for the
+   *   <li>{@link Options#isolationLevel(IsolationLevel)}: The isolation level for the
    *       transaction
    * </ul>
    */
@@ -457,7 +457,7 @@ public interface DatabaseClient {
    *       applied to any other requests on the transaction.
    *   <li>{@link Options#commitStats()}: Request that the server includes commit statistics in the
    *       {@link CommitResponse}.
-   *   <li>{@link Options#isolationLevelOption(IsolationLevel)}: The isolation level for the
+   *   <li>{@link Options#isolationLevel(IsolationLevel)}: The isolation level for the
    *       transaction
    * </ul>
    */
@@ -499,7 +499,7 @@ public interface DatabaseClient {
    *       applied to any other requests on the transaction.
    *   <li>{@link Options#commitStats()}: Request that the server includes commit statistics in the
    *       {@link CommitResponse}.
-   *   <li>{@link Options#isolationLevelOption(IsolationLevel)}: The isolation level for the
+   *   <li>{@link Options#isolationLevel(IsolationLevel)}: The isolation level for the
    *       transaction
    * </ul>
    */
@@ -555,7 +555,7 @@ public interface DatabaseClient {
    *       applied to any other requests on the transaction.
    *   <li>{@link Options#commitStats()}: Request that the server includes commit statistics in the
    *       {@link CommitResponse}.
-   *   <li>{@link Options#isolationLevelOption(IsolationLevel)}: The isolation level for the
+   *   <li>{@link Options#isolationLevel(IsolationLevel)}: The isolation level for the
    *       transaction
    * </ul>
    */
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java
@@ -415,8 +415,7 @@ public interface DatabaseClient {
    *       applied to any other requests on the transaction.
    *   <li>{@link Options#commitStats()}: Request that the server includes commit statistics in the
    *       {@link CommitResponse}.
-   *   <li>{@link Options#isolationLevel(IsolationLevel)}: The isolation level for the
-   *       transaction
+   *   <li>{@link Options#isolationLevel(IsolationLevel)}: The isolation level for the transaction
    * </ul>
    */
   TransactionRunner readWriteTransaction(TransactionOption... options);
@@ -457,8 +456,7 @@ public interface DatabaseClient {
    *       applied to any other requests on the transaction.
    *   <li>{@link Options#commitStats()}: Request that the server includes commit statistics in the
    *       {@link CommitResponse}.
-   *   <li>{@link Options#isolationLevel(IsolationLevel)}: The isolation level for the
-   *       transaction
+   *   <li>{@link Options#isolationLevel(IsolationLevel)}: The isolation level for the transaction
    * </ul>
    */
   TransactionManager transactionManager(TransactionOption... options);
@@ -499,8 +497,7 @@ public interface DatabaseClient {
    *       applied to any other requests on the transaction.
    *   <li>{@link Options#commitStats()}: Request that the server includes commit statistics in the
    *       {@link CommitResponse}.
-   *   <li>{@link Options#isolationLevel(IsolationLevel)}: The isolation level for the
-   *       transaction
+   *   <li>{@link Options#isolationLevel(IsolationLevel)}: The isolation level for the transaction
    * </ul>
    */
   AsyncRunner runAsync(TransactionOption... options);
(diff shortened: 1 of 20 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
```diff
@@ -159,6 +185,22 @@ public static TransactionOption optimisticLock() {
...
+    return REPEATABLE_READ_OPTION;
+  }
+
+  /**
+   * Specifying this instructs the transaction to request Serializable Isolation Level from the
+   * backend.
+   */
+  public static TransactionOption serializableIsolationLevel() {
+    return SERIALIZABLE_OPTION;
+  }
```
olavloite, line 202: Instead of creating two methods, one for each isolation level, this should be one method that takes an IsolationLevel input argument. This means that any future isolation levels that we add can be used without any changes.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
```diff
@@ -490,6 +532,26 @@ void appendToOptions(Options options) {
...
+
+  /** Option to request SERIALIZABLE isolation level for read/write transactions. */
+  static final class SerializableOption extends InternalOption implements TransactionOption {
+    @Override
+    void appendToOptions(Options options) {
+      options.isolationLevel = IsolationLevel.SERIALIZABLE;
+    }
+  }
+
+  static final SerializableOption SERIALIZABLE_OPTION = new SerializableOption();
```
olavloite, line 553: This should also be just one class that contains a field with the isolation level.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
```diff
@@ -131,7 +135,29 @@ public interface UpdateAdminApiOption extends AdminApiOption {}
   public interface QueryOption {}
 
   /** Marker interface to mark options applicable to write operations */
-  public interface TransactionOption {}
```
olavloite, line 134: This is a marker interface. It should not contain any code or logic.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
```diff
@@ -131,7 +135,29 @@ public interface UpdateAdminApiOption extends AdminApiOption {}
...
+        TransactionOption[] primaryOptions, TransactionOption[] spannerOptions) {
+      if (spannerOptions == null
+          || Arrays.stream(primaryOptions).anyMatch(isValidIsolationLevelOption)) {
+        return primaryOptions;
+      } else {
+        return Stream.concat(Arrays.stream(primaryOptions), Arrays.stream(spannerOptions))
+            .toArray(TransactionOption[]::new);
+      }
+    }
+  }
```
olavloite, line 160: I think that this block of code could be significantly simplified by just moving the check to the place where we are actually building the request, and then do something like this there:

```java
IsolationLevel isolationLevel = MoreObjects.firstNonNull(
    options.getIsolationLevel(),
    spannerOptions.getDefaultIsolationLevel());
if (isolationLevel != null) {
  transactionOptions.setIsolationLevel(isolationLevel);
}
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java
```diff
@@ -111,6 +111,10 @@ default String getDatabaseRole() {
    *       applied to any other requests on the transaction.
    *   <li>{@link Options#commitStats()}: Request that the server includes commit statistics in the
    *       {@link CommitResponse}.
+   *   <li>{@link Options#repeatableReadIsolationLevel()}: Request Repeatable Read Isolation Level
+   *       from the backend.
+   *   <li>{@link Options#serializableIsolationLevel()}: Request Serializable Isolation Level from
+   *       the backend.
```
olavloite, line 117: Here and elsewhere, replace with something like:

```
{@link Options#isolationLevel(IsolationLevel)}: The isolation level for the transaction
```

olavloite: Also, specifically for the methods that just execute a blind write: I don't think the isolation level option is very relevant for those. It should still be supported, but explicitly including it in the documentation seems a bit redundant.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1643,6 +1651,52 @@ public Builder setEnableEndToEndTracing(boolean enableEndToEndTracing) {
       return this;
     }
 
+    public static class TransactionOptions {
```
olavloite, line 1654: I'm not convinced that we should create a `TransactionOptions` class here, but if we absolutely want that, then it should be a wrapper around the proto `TransactionOptions` class, instead of something that contains a list of `com.google.cloud.TransactionOption`.

I would suggest one of the following:
1. Just add a `defaultIsolationLevel` property to `SpannerOptions` and remove this entire class.
2. OR: Simplify this class by removing the list of `TransactionOption`, and instead add one property of type `com.google.cloud.spanner.v1.TransactionOptions`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1643,6 +1651,52 @@ public Builder setEnableEndToEndTracing(boolean enableEndToEndTracing) {
...
+
+        public static TransactionOptionsBuilder newBuilder() {
+          return new TransactionOptionsBuilder();
+        }
+
+        public TransactionOptionsBuilder setIsolationLevel(TransactionOption transactionOption) {
+          if (!isValidIsolationLevelOption.test(transactionOption)) {
+            throw SpannerExceptionFactory.newSpannerException(
+                ErrorCode.INVALID_ARGUMENT, INVALID_ISOLATION_LEVEL_MESSAGE);
+          }
```
olavloite, line 1677: This should really not be necessary. IsolationLevel is an enum, so it should be impossible to set any other value (and if anyone sets one of the `UNRECOGNIZED` or `UNSPECIFIED` values, then just let the error that is returned by Spanner propagate to the application)

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplWithTransactionOptionsTest.java
```diff
@@ -0,0 +1,497 @@
+package com.google.cloud.spanner;
```
olavloite, line 1: nit: add copyright header

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplWithTransactionOptionsTest.java
```diff
@@ -0,0 +1,497 @@
...
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public class DatabaseClientImplWithTransactionOptionsTest {
+
+  private static MockSpannerServiceImpl mockSpanner;
```
olavloite, line 47: There's a lot of copy-pasted code here. Can we re-use / share more of it with one or more of the existing test classes?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
sakthivelmanii: @olavloite I suggested #2 because Go follows the similar approach.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplWithTransactionOptionsTest.java
```diff
@@ -0,0 +1,497 @@
...
+      } else if (request instanceof ExecuteBatchDmlRequest) {
+        foundMatchingRequest = true;
+        assertEquals(
+            ((ExecuteBatchDmlRequest) request).getTransaction().getBegin().getIsolationLevel(),
+            isolationLevel);
+      }
+      if (foundMatchingRequest) {
+        break;
+      }
+    }
```
sakthivelmanii, line 495: Can we throw an error if no matching request found? Ideally there should be atleast. For some reason if gRPC call itself is not made, then it should throw an error

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
sakthivelmanii: @olavloite what if we implement merge method Options.java?

olavloite: I don't think we need this merge method at all if we change the implementation of the default options to use a `com.google.cloud.spanner.v1.TransactionOptions`, as you can then use the merge method on that class. (And the same also if we decide to just have a `DefaultIsolationLevel` instead of `DefaultTransactionOptions`).

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
shobhitsg: This is required with the existing implementation as IsolationLevel enum is not the part of the request currently. Thus, I have to validate the correctness of the `com.google.cloud.spanner.TransactionOption` request. It's not about what isolation level is passed, it was more about that TransactionOption shouldn't of any other type except of IsolationLevel type. 
Anyways, this will be removed in the subsequent commit since I'll be making IsolationLevel change in the request as you suggested in the previous comment.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
shobhitsg: I completely agree with your point. However, IMO, including request options that are not yet fully supported by the library, even for short duration(e.g. development if that's even required?), can create confusion among lib users. Though the backend will reject unsupported options, their presence without clear documentation could mislead developers who often rely on request options to determine functionality. Please let me know if I've misunderstood anything and I'd like to know your POV. Here, I've assumed that the proto will always be merged in the main branch before any development could start.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplWithTransactionOptionsTest.java
shobhitsg: Updated.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplWithTransactionOptionsTest.java
shobhitsg: Added

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java
shobhitsg: Updated

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
shobhitsg: Removed.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
shobhitsg: Updated

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplWithTransactionOptionsTest.java
shobhitsg: Added

shobhitsg:
@olavloite @sakthivelmanii I have implemented the suggested revisions. PTAL.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java
```diff
@@ -414,6 +415,8 @@ ServerStream<BatchWriteResponse> batchWriteAtLeastOnce(
    *       applied to any other requests on the transaction.
    *   <li>{@link Options#commitStats()}: Request that the server includes commit statistics in the
    *       {@link CommitResponse}.
+   *   <li>{@link Options#isolationLevelOption(IsolationLevel)}: The isolation level for the
+   *       transaction
```
olavloite, line 419: How relevant is setting an isolation level for blind writes? I would expect it not to really have any effect, so maybe it's better to just remove this from the documentation? The fact that you can still set it through the options is OK, but I would say that we should choose one of the following:
1. Remove this from the documentation of each blind write operation (assuming that it effectively does not do anything)
2. Keep it in the documentation, but then explicitly state what the effect is (even if that is 'It does not have any effect')

Now it feels a bit like non-information.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
```diff
@@ -159,6 +160,13 @@ public static TransactionOption optimisticLock() {
     return OPTIMISTIC_LOCK_OPTION;
   }
 
+  /**
+   * Specifying this instructs the transaction to request {@link IsolationLevel} from the backend.
+   */
+  public static IsolationLevelOption isolationLevelOption(IsolationLevel isolationLevel) {
```
olavloite, line 166: nit: the return type here should be `TransactionOption` instead of `IsolationLevelOption`. The latter is not a public class or interface, but this method is.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionImpl.java
```diff
@@ -193,6 +199,14 @@ void markUsed(Instant instant) {
     sessionReference.markUsed(instant);
   }
 
+  com.google.spanner.v1.TransactionOptions defaultTransactionOptions() {
+    TransactionOptions transactionOptions =
+        this.spanner.getOptions().getDefaultTransactionOptions();
+    return transactionOptions != null
+        ? transactionOptions.getTransactionOptions()
+        : com.google.spanner.v1.TransactionOptions.getDefaultInstance();
```
olavloite, line 207: nit: All of this can be removed if we let the default for `SpannerOptions.defaultTransactionOptions` be a `TransactionOptions.getDefaultInstance()`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionImpl.java
```diff
@@ -246,13 +260,17 @@ public CommitResponse writeAtLeastOnceWithOptions(
...
+            .setReadWrite(com.google.spanner.v1.TransactionOptions.ReadWrite.getDefaultInstance());
     if (options.withExcludeTxnFromChangeStreams() == Boolean.TRUE) {
       transactionOptionsBuilder.setExcludeTxnFromChangeStreams(true);
     }
-    requestBuilder.setSingleUseTransaction(transactionOptionsBuilder);
+    if (options.isolationLevel() != null) {
+      transactionOptionsBuilder.setIsolationLevel(options.isolationLevel());
+    }
+    requestBuilder.setSingleUseTransaction(
+        this.defaultTransactionOptions().toBuilder().mergeFrom(transactionOptionsBuilder.build()));
```
olavloite, line 273: super-nit: remove `this.`

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1643,6 +1649,45 @@ public Builder setEnableEndToEndTracing(boolean enableEndToEndTracing) {
       return this;
     }
 
+    public static class TransactionOptions {
```
olavloite, line 1652: Can we rename this class to either:
1. `ReadWriteTransactionOptions`
2. `DefaultReadWriteTransactionOptions`

Both makes it easier to distinguish from the proto `TransactionOptions` class, and also better reflects what it is used for (e.g. it is not used for read-only or PDML transactions)

Also: This class needs Javadoc.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1643,6 +1649,45 @@ public Builder setEnableEndToEndTracing(boolean enableEndToEndTracing) {
...
+      private com.google.spanner.v1.TransactionOptions transactionOptions;
+
+      private TransactionOptions() {}
+
+      com.google.spanner.v1.TransactionOptions getTransactionOptions() {
```
olavloite, line 1657:
```suggestion
      com.google.spanner.v1.TransactionOptions getProto() {
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1643,6 +1649,45 @@ public Builder setEnableEndToEndTracing(boolean enableEndToEndTracing) {
...
+        return transactionOptions;
+      }
+
+      public static class TransactionOptionsBuilder {
+        private IsolationLevelOption isolationLevelOption;
```
olavloite, line 1662: this builder class should be a simple wrapper around a `com.google.cloud.spanner.v1.TransactionOptions$Builder`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1643,6 +1649,45 @@ public Builder setEnableEndToEndTracing(boolean enableEndToEndTracing) {
...
+
+        public static TransactionOptionsBuilder newBuilder() {
+          return new TransactionOptionsBuilder();
+        }
+
+        public TransactionOptionsBuilder setIsolationLevel(IsolationLevelOption option) {
```
olavloite, line 1668: This method should accept a `com.google.cloud.spanner.v1.IsolationLevel` value instead of an option.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1643,6 +1649,45 @@ public Builder setEnableEndToEndTracing(boolean enableEndToEndTracing) {
...
+              com.google.spanner.v1.TransactionOptions.newBuilder()
+                  .setIsolationLevel(
+                      Options.fromTransactionOptions(isolationLevelOption).isolationLevel())
+                  .build();
+          return transactionOptions;
+        }
+      }
+    }
+
+    /** Sets the default transaction options. */
```
olavloite, line 1685: This needs more elaborate Javadoc. The current string is just a repeat of the method name.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1643,6 +1649,45 @@ public Builder setEnableEndToEndTracing(boolean enableEndToEndTracing) {
...
+    public Builder setDefaultTransactionOptions(TransactionOptions transactionOptions) {
```
olavloite, line 1686: I think that we should rename this method to `setDefaultReadWriteTransactionOptions()`

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplWithTransactionOptionsTest.java
```diff
@@ -0,0 +1,352 @@
...
+  public static void stopServer() throws InterruptedException {
+    server.shutdown();
+    server.awaitTermination();
+  }
+
+  @Before
+  public void setUp() {
+    mockSpanner.reset();
+    mockSpanner.removeAllExecutionTimes();
+    SpannerOptions.Builder spannerOptionsBuilder =
```
olavloite, line 108: nit: I think that you can move the creation of the `Spanner` instances to a `@BeforeClass` method (and the same then also with the `spanner.close()`

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplWithTransactionOptionsTest.java
```diff
@@ -0,0 +1,352 @@
...
+        SpannerOptions.newBuilder()
+            .setProjectId("[PROJECT]")
+            .setChannelProvider(channelProvider)
+            .setCredentials(NoCredentials.getInstance());
+    spanner = spannerOptionsBuilder.build().getService();
+    client = spanner.getDatabaseClient(DatabaseId.of("[PROJECT]", "[INSTANCE]", "[DATABASE]"));
+    clientWithRepeatableReadOption =
```
olavloite, line 115: This client is never closed

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplWithTransactionOptionsTest.java
```diff
@@ -0,0 +1,352 @@
...
+        spannerOptionsBuilder
+            .setDefaultTransactionOptions(
+                TransactionOptionsBuilder.newBuilder()
+                    .setIsolationLevel(RR_ISOLATION_OPTION)
+                    .build())
+            .build()
+            .getService()
+            .getDatabaseClient(DatabaseId.of("[PROJECT]", "[INSTANCE]", "[DATABASE]"));
+    clientWithSerializableOption =
```
olavloite, line 124: This client is never closed

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java
shobhitsg: This doc is only added for `readWriteTransaction` and `transactionManager`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionImpl.java
shobhitsg: Updated

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
shobhitsg: Renamed to `DefaultReadWriteTransactionOptions` and added Javadoc.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
shobhitsg: Added the link to the original configuration option class.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
shobhitsg: Renamed.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplWithTransactionOptionsTest.java
shobhitsg: Fixed.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplWithTransactionOptionsTest.java
shobhitsg: Fixed.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1643,6 +1649,45 @@ public Builder setEnableEndToEndTracing(boolean enableEndToEndTracing) {
       return this;
     }
 
+    public static class TransactionOptions {
+      private com.google.spanner.v1.TransactionOptions transactionOptions;
```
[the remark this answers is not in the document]
shobhitsg: Updated

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
shobhitsg: Not needed, removed.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
shobhitsg: Updated

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
shobhitsg: Updated

google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplWithTransactionOptionsTest.java
shobhitsg: I'll raise a separate PR for this minor refactoring.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java
olavloite: Sorry. You're absolutely right, I got confused by the way that GitHub shows the name of the previous method above this.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
```diff
@@ -159,6 +160,13 @@ public static TransactionOption optimisticLock() {
     return OPTIMISTIC_LOCK_OPTION;
   }
 
+  /**
+   * Specifying this instructs the transaction to request {@link IsolationLevel} from the backend.
+   */
+  public static TransactionOption isolationLevelOption(IsolationLevel isolationLevel) {
```
olavloite, line 166: Sorry, last nit on this: Can we rename this to the following to be consistent with the other options

```suggestion
  public static TransactionOption isolationLevel(IsolationLevel isolationLevel) {
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1645,6 +1651,55 @@ public Builder setEnableEndToEndTracing(boolean enableEndToEndTracing) {
       return this;
     }
 
+    /**
+     * Provides the default read-write transaction options for all databases, limited to setting the
+     * {@link IsolationLevel}. These defaults are overridden by any explicit {@link
```
olavloite, line 1656: nit: remove 'limited to ...', it is typically a thing that otherwise gets forgotten when a new option is added

```suggestion
     * Provides the default read-write transaction options for all databases. These defaults are overridden by any explicit {@link
```

olavloite (approved):
LGTM, but please change the name of the `Options#isolationLevelOption(..)` method to `Options#isolationLevel(..)` before merging.

sakthivelmanii (approved):

olavloite:
Last build errors are unrelated, created #3690

PR #3676, opened 2025-03-05
chore: enable multplexed sessions for experimental host

sagnghos:
Ensure multiplexed sessions are used when client is connecting to an experimental host

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
@@ -24,7 +24,6 @@ import com.google.api.core.ObsoleteApi;
 import com.google.cloud.spanner.SessionPool.Position;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
-import io.grpc.ExperimentalApi;
 import java.time.Duration;
 import java.util.Locale;
 import java.util.Objects;
@@ -121,8 +120,7 @@ public class SessionPoolOptions {
     Boolean useMultiplexedSessionFromEnvVariable = getUseMultiplexedSessionFromEnvVariable();
     this.useMultiplexedSession =
         builder.isExperimentalHost
-            ? true
-            : ((useMultiplexedSessionFromEnvVariable != null)
+            || ((useMultiplexedSessionFromEnvVariable != null)
                 ? useMultiplexedSessionFromEnvVariable
                 : builder.useMultiplexedSession);
     // useMultiplexedSessionForRW priority => Environment var > private setter > client default
@@ -130,8 +128,7 @@ public class SessionPoolOptions {
         getUseMultiplexedSessionForRWFromEnvVariable();
     this.useMultiplexedSessionForRW =
         builder.isExperimentalHost
-            ? true
-            : ((useMultiplexedSessionForRWFromEnvVariable != null)
+            || ((useMultiplexedSessionForRWFromEnvVariable != null)
                 ? useMultiplexedSessionForRWFromEnvVariable
                 : builder.useMultiplexedSessionForRW);
     // useMultiplexedSessionPartitionedOps priority => Environment var > private setter > client
@@ -140,13 +137,12 @@ public class SessionPoolOptions {
         getUseMultiplexedSessionFromEnvVariablePartitionedOps();
     this.useMultiplexedSessionForPartitionedOps =
         builder.isExperimentalHost
-            ? true
-            : ((useMultiplexedSessionFromEnvVariablePartitionedOps != null)
+            || ((useMultiplexedSessionFromEnvVariablePartitionedOps != null)
                 ? useMultiplexedSessionFromEnvVariablePartitionedOps
                 : builder.useMultiplexedSessionPartitionedOps);
     this.multiplexedSessionMaintenanceDuration = builder.multiplexedSessionMaintenanceDuration;
     this.skipVerifyingBeginTransactionForMuxRW =
-        builder.isExperimentalHost ? true : builder.skipVerifyingBeginTransactionForMuxRW;
+        builder.isExperimentalHost || builder.skipVerifyingBeginTransactionForMuxRW;
   }
 
   @Override
@@ -827,7 +823,7 @@ public class SessionPoolOptions {
       return this;
     }
 
-    @ExperimentalApi("https://github.com/googleapis/java-spanner/pull/3676")
+    @InternalApi
     public Builder setExperimentalHost() {
       this.isExperimentalHost = true;
       return this;
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
@@ -826,7 +826,7 @@ public class SessionPoolOptions {
       return this;
     }
 
-    @ExperimentalApi("")
+    @ExperimentalApi("https://github.com/googleapis/java-spanner/pull/3676")
     public Builder setExperimentalHost() {
       this.isExperimentalHost = true;
       return this;
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
@@ -24,6 +24,7 @@ import com.google.api.core.ObsoleteApi;
 import com.google.cloud.spanner.SessionPool.Position;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
+import io.grpc.ExperimentalApi;
 import java.time.Duration;
 import java.util.Locale;
 import java.util.Objects;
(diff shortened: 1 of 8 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -991,7 +987,7 @@ public static class Builder
     private boolean enableBuiltInMetrics = SpannerOptions.environment.isEnableBuiltInMetrics();
     private String monitoringHost = SpannerOptions.environment.getMonitoringHost();
     private SslContext mTLSContext = null;
-    private boolean isExternalHost = false;
+    private boolean isExperimentalHost = false;
```
krishnamoorthy-r, line 990: Is it possible to avoid adding in SpannerOptions?  Also, even if we want, do we want to keep this option as something like "alwaysUseMultiplexSession" or something like that?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionProperties.java
```diff
@@ -197,7 +199,14 @@ public class ConnectionProperties {
           BOOLEANS,
           BooleanConverter.INSTANCE,
           Context.STARTUP);
-
+  static final ConnectionProperty<Boolean> IS_EXPERIMENTAL_HOST =
```
krishnamoorthy-r, line 202: We anyway has dedicated connection string in case of JDBC right? Do we really need to expose query param? Is it not safe to assume that if connection URI has jdbc:spanner ( it is experimental) otherwise cloud spanner or emulator? I am trying to see how we can keep small set of methods /properties to the users.

sagnghos: No we are supporting prefixes spanner and cloudspanner for both hosts. This was decided in a previous PR.

If user is using the the dedicated pattern they won't need to add the parameter. I added this because pgadapter doesn't use the dedicated pattern and just by appending this parameter we can avoid a lot of code changes in the pgadapter.

krishnamoorthy-r: What is the need to support new prefix jdbc:spanner for cloud spanner also? I assume until today, the only prefix works is jdbc:cloudspanner right?

sagnghos: It was a suggestion from @olavloite. He felt requiring one prefix with one version of the URL format, and one with the other would create unnecessary confusion.

krishnamoorthy-r: If we are not using prefix in JDBC to distinguish between Box vs Cloud Spanner, what is the need to have jdbc:spanner prefix? Mostly I am missing something: My understanding is that jdbc:cloudspanner is prefix used today and that is only one supported currently. We added jdbc:spanner: as prefix to state this JDBC URL maps to Box. If this understanding is correct, whenever JDBC drives uses cloud-spanner-sdk, it would invoke automatically setExternalHost(....) when it seems jdbc url has prefix jdbc:spanner (instead of jdbc:cloudspanner). If we are saying both can be used for both box and cloud spanner, I am missing the reason why we need to introduce a new prefix. Please correct me if I am missing some context!

sagnghos: From our end we wanted to keep separate prefix so that someone trying to connect to the Experimental host doesn't associate the experience with cloudspanner.

We would still continue to do that. We will document the same that for connecting to experimental host use prefix jdbc:spanner and for cloudspanner it will remain jdbc:cloudspanner as it is today.

But Knut's point was internally we should support both prefix in either case because developers tend to reuse existing code and they would unnecessarily get errors when they try to reuse the code they had written for one host for the other.

krishnamoorthy-r: I might not fully understand that point, may be will wait for @olavloite response. I don't think the code will work for both experimental and existing cloud spanner without changing connection string. Even in this PR, user need to add ``isExperimentalHost=...`` to connection string. So I am not understanding why it cannot be jdbc:spanner (or if we want we can rename it that jdbc:experimental-spanner)

sagnghos: Just FYI. User does **NOT** need to add the property isExperimentalHost=true to the jdbc url.

Only the pgAdapter and any other tool which use this ConnectionOptions code would require to attach this parameter (I will make that change once this is checked in) because it uses the cloudspanner pattern internally.

From the user's experience they would **never** require to add this parameter as long as they are using the correct url pattern.

For clarity please refer the unit test I have added in ConnectionOptionsTest. 
The first url doesn't have the parameter attached but since it's following the experience host url pattern all the mux session variables are getting set correctly. (This pattern is how the user experience will look)

Whereas the 2nd url needed the parameter attached since it's based on the existing cloudspanner pattern (This is how pgadapter will use ConnectionOptions to connect to experimental Host)

olavloite (commented):
Can we add a test using the mock Spanner server for this option, and also include a test case that covers the situation that this new field tries to fix? (My understanding is that the need for a separate field was that only setting the various `use-mux-sessions` to true was not enough to prevent the `BatchCreateSessions` RPC to be called. This PR does not really do anything other than setting those three fields to true, which makes me believe that this change does not really fix the original problem. Or am I missing something here?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
```diff
@@ -444,6 +450,10 @@ static boolean isEnableTransactionalConnectionStateForPostgreSQL() {
                       USE_PLAIN_TEXT_PROPERTY_NAME,
                       "Use a plain text communication channel (i.e. non-TLS) for communicating with the server (true/false). Set this value to true for communication with the Cloud Spanner emulator.",
                       DEFAULT_USE_PLAIN_TEXT),
+                  ConnectionProperty.createBooleanProperty(
+                      IS_EXPERIMENTAL_HOST_PROPERTY_NAME,
+                      "Set this value to true for communication with a Experimental Host.",
```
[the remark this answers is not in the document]
sagnghos: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
```diff
@@ -981,16 +992,19 @@ && getInitialConnectionPropertyValue(OAUTH_TOKEN) == null
...
       this.sessionPoolOptions = SessionPoolOptions.newBuilder().setAutoDetectDialect(true).build();
     }
 
-    String projectId = "default";
+    String projectId = EXPERIMENTAL_HOST_PROJECT_ID;
     String instanceId = matcher.group(Builder.INSTANCE_GROUP);
-    if (!isExternalHost) {
+    if (!isExperimentalHost() && !isExperimentalHostPattern) {
       projectId = matcher.group(Builder.PROJECT_GROUP);
     } else if (instanceId == null) {
```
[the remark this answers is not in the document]
sagnghos: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1657,7 +1659,7 @@ public SpannerOptions build() {
         this.setChannelConfigurator(ManagedChannelBuilder::usePlaintext);
         // As we are using plain text, we should never send any credentials.
         this.setCredentials(NoCredentials.getInstance());
-      } else if (isExternalHost && credentials == null) {
+      } else if (isExperimentalHost && credentials == null) {
         credentials = environment.getDefaultExternalHostCredentials();
```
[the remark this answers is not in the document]
sagnghos: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -813,6 +827,12 @@ public Builder setWarnAndCloseIfInactiveTransactions() {
       return this;
     }
 
+    @ExperimentalApi("https://github.com/googleapis/java-spanner/pull/3676")
```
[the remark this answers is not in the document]
sagnghos: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -114,26 +120,33 @@ private SessionPoolOptions(Builder builder) {
...
     Boolean useMultiplexedSessionFromEnvVariable = getUseMultiplexedSessionFromEnvVariable();
     this.useMultiplexedSession =
-        (useMultiplexedSessionFromEnvVariable != null)
-            ? useMultiplexedSessionFromEnvVariable
-            : builder.useMultiplexedSession;
+        builder.isExperimentalHost
+            ? true
+            : ((useMultiplexedSessionFromEnvVariable != null)
+                ? useMultiplexedSessionFromEnvVariable
+                : builder.useMultiplexedSession);
```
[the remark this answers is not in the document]
sagnghos: Done

PR #3645, opened 2025-02-13
feat: enable ALTS hard bound token in DirectPath

rockspore:
This change will enable ALTS hard bound token in DirectPath, if the client runs in on GCE/GKE with ComputeEngineCredentials.

If for any reason the runtime environment doesn't support the issuance of such tokens, normal bearer access tokens will be used.

PR #3679, opened 2025-03-10
fix: revert the ALTS bound token enablement

rockspore:
Reverts #3645


rockspore:
@rahul2393 @olavloite @sakthivelmanii Could you please help merge this asap before it gets into a release. We found a bug at the server side so it will be safer to have this disabled for now. Really sorry for the inconvenience.

sakthivelmanii (approved):

olavloite (approved):

PR #3684, opened 2025-03-13
perf: get database dialect using multiplexed session

olavloite:
Get the database dialect using the multiplexed session that is created by the client when multiplexed sessions are enabled. This prevents unnecessarily creating a regular session when MinSessions has been set to zero and multiplexed sessions have been enabled.

### DIFF
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
@@ -3966,7 +3966,7 @@ public class DatabaseClientImplTest {
     try {
       // Simulate session creation failures on the backend.
       mockSpanner.setCreateSessionExecutionTime(
-          SimulatedExecutionTime.ofStickyException(Status.RESOURCE_EXHAUSTED.asRuntimeException()));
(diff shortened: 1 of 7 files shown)


google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
```diff
@@ -3966,7 +3966,7 @@ public void testCreateSessionsFailure_shouldNotPropagateToCloseMethod() {
     try {
       // Simulate session creation failures on the backend.
       mockSpanner.setCreateSessionExecutionTime(
-          SimulatedExecutionTime.ofStickyException(Status.RESOURCE_EXHAUSTED.asRuntimeException()));
+          SimulatedExecutionTime.ofStickyException(Status.PERMISSION_DENIED.asRuntimeException()));
```
olavloite, line 3969: This change is not directly relevant for the change in this PR, but I noticed that this test took ~30 seconds to execute. The reason for that is that `RESOURCE_EXHAUSTED` is now a retriable error, which means that it will automatically be retried by Gax for up to 30 seconds before failing.

harshachinta (approved):

PR #3691, opened 2025-03-18
Integration test for End to End tracing

manu2:
With End to End tracing, customers are able to view server side traces. The new integration tests aims to verify this from a client side pov.
**Note**: Currently the InegrationTestEnv sets up default interceptor which in turns creates a default GlobalOpenTelementry sdk. This sdk is then injected to the gRPC connection builder and stud. So a new opentelemetry sdk has been created in IntegrationTestEnv that sets up the required config for End to End Testing and will be used by default for all integration tests.

### DIFF
--- a/google-cloud-spanner/pom.xml
+++ b/google-cloud-spanner/pom.xml
@@ -17,8 +17,8 @@
     <site.installationModule>google-cloud-spanner</site.installationModule>
     <opencensus.version>0.31.1</opencensus.version>
     <spanner.testenv.config.class>com.google.cloud.spanner.GceTestEnvConfig</spanner.testenv.config.class>
-    <spanner.testenv.instance>projects/gcloud-devel/instances/spanner-testing-east1</spanner.testenv.instance>
-    <spanner.gce.config.project_id>gcloud-devel</spanner.gce.config.project_id>
+    <spanner.testenv.instance>projects/span-cloud-testing/instances/manu-demo1-mr-nam6</spanner.testenv.instance>
+    <spanner.gce.config.project_id>span-cloud-testing</spanner.gce.config.project_id>
     <spanner.testenv.kms_key.name>projects/gcloud-devel/locations/us-east1/keyRings/cmek-test-key-ring/cryptoKeys/cmek-test-key</spanner.testenv.kms_key.name>
   </properties>
 
@@ -455,6 +455,24 @@
       <artifactId>opentelemetry-sdk-testing</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.google.cloud.opentelemetry</groupId>
+      <artifactId>exporter-trace</artifactId>
+      <version>0.33.0</version>
+      <scope>test</scope>
+    </dependency>
+<!--    <dependency>-->
+<!--      <groupId>com.google.cloud</groupId>-->
+<!--      <artifactId>google-cloud-trace</artifactId>-->
+<!--      <version>2.53.0</version>-->
+<!--      <scope>test</scope>-->
+<!--    </dependency>-->
+<!--    <dependency>-->
+<!--      <groupId>com.google.api.grpc</groupId>-->
+<!--      <artifactId>proto-google-cloud-trace-v1</artifactId>-->
+<!--      <version>2.53.0</version>-->
+<!--      <scope>test</scope>-->
+<!--    </dependency>-->
   </dependencies>
   <profiles>
     <profile>
--- a/google-cloud-spanner/pom.xml
+++ b/google-cloud-spanner/pom.xml
@@ -17,8 +17,8 @@
     <site.installationModule>google-cloud-spanner</site.installationModule>
     <opencensus.version>0.31.1</opencensus.version>
     <spanner.testenv.config.class>com.google.cloud.spanner.GceTestEnvConfig</spanner.testenv.config.class>
-    <spanner.testenv.instance>projects/span-cloud-testing/instances/manu-demo1-mr-nam6</spanner.testenv.instance>
-    <spanner.gce.config.project_id>span-cloud-testing</spanner.gce.config.project_id>
+    <spanner.testenv.instance>projects/gcloud-devel/instances/spanner-testing-east1</spanner.testenv.instance>
+    <spanner.gce.config.project_id>gcloud-devel</spanner.gce.config.project_id>
     <spanner.testenv.kms_key.name>projects/gcloud-devel/locations/us-east1/keyRings/cmek-test-key-ring/cryptoKeys/cmek-test-key</spanner.testenv.kms_key.name>
   </properties>
 
@@ -461,18 +461,6 @@
       <version>0.33.0</version>
       <scope>test</scope>
     </dependency>
-<!--    <dependency>-->
-<!--      <groupId>com.google.cloud</groupId>-->
-<!--      <artifactId>google-cloud-trace</artifactId>-->
-<!--      <version>2.53.0</version>-->
-<!--      <scope>test</scope>-->
-<!--    </dependency>-->
-<!--    <dependency>-->
-<!--      <groupId>com.google.api.grpc</groupId>-->
-<!--      <artifactId>proto-google-cloud-trace-v1</artifactId>-->
-<!--      <version>2.53.0</version>-->
-<!--      <scope>test</scope>-->
-<!--    </dependency>-->
   </dependencies>
   <profiles>
     <profile>
--- a/google-cloud-spanner/pom.xml
+++ b/google-cloud-spanner/pom.xml
@@ -3,7 +3,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>com.google.cloud</groupId>
   <artifactId>google-cloud-spanner</artifactId>
-  <version>6.88.0</version><!-- {x-version-update:google-cloud-spanner:current} -->
+  <version>6.88.1-SNAPSHOT</version><!-- {x-version-update:google-cloud-spanner:current} -->
   <packaging>jar</packaging>
   <name>Google Cloud Spanner</name>
   <url>https://github.com/googleapis/java-spanner</url>
@@ -11,7 +11,7 @@
   <parent>
     <groupId>com.google.cloud</groupId>
     <artifactId>google-cloud-spanner-parent</artifactId>
-    <version>6.88.0</version><!-- {x-version-update:google-cloud-spanner:current} -->
+    <version>6.88.1-SNAPSHOT</version><!-- {x-version-update:google-cloud-spanner:current} -->
   </parent>
   <properties>
     <site.installationModule>google-cloud-spanner</site.installationModule>
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/IntegrationTestEnv.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/IntegrationTestEnv.java
@@ -110,6 +110,8 @@ public class IntegrationTestEnv extends ExternalResource {
 
     SpannerOptions options = config.spannerOptions();
     String instanceProperty = System.getProperty(TEST_INSTANCE_PROPERTY, "");
+    instanceProperty = "projects/span-cloud-testing/instances/manu-demo1-mr-nam6";
+    System.out.println("Instance Property Name " + instanceProperty);
     InstanceId instanceId;
     if (!instanceProperty.isEmpty() && !alwaysCreateNewInstance) {
       instanceId = InstanceId.of(instanceProperty);
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/IntegrationTestEnv.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/IntegrationTestEnv.java
@@ -22,14 +22,26 @@ import static org.junit.Assume.assumeFalse;
 import com.google.api.client.util.ExponentialBackOff;
 import com.google.api.gax.longrunning.OperationFuture;
 import com.google.cloud.Timestamp;
+import com.google.cloud.opentelemetry.trace.TraceConfiguration;
+import com.google.cloud.opentelemetry.trace.TraceExporter;
 import com.google.cloud.spanner.DatabaseInfo.DatabaseField;
 import com.google.cloud.spanner.testing.EmulatorSpannerHelper;
 import com.google.cloud.spanner.testing.RemoteSpannerHelper;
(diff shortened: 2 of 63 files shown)


manu2:
Looking through the details of the failure in [Kokoro - Test: Integration with Directpath](https://source.cloud.google.com/results/invocations/f726c30c-1f3d-4a1e-8fff-44f43f0fc76e) using the following [target log](https://btx.cloud.google.com/invocations/f726c30c-1f3d-4a1e-8fff-44f43f0fc76e/targets;showStatuses=failed,passed/cloud-devrel%2Fclient-libraries%2Fjava%2Fjava-spanner%2Fpresubmit%2Fintegration-directpath-enabled;config=default/log)

Many of the tests seem to be facing "com.google.api.gax.rpc.PermissionDeniedException: io.grpc.StatusRuntimeException: PERMISSION_DENIED: The caller does not have permission" when on trace exporter. 
However in case of ITEndToEndTracingTest, we have a assertion on whether the traces are present in cloud trace, which is causing an assertion failure.

So the underlying issue regarding permission denied for trace export is not related to this change but is causing failure in the new test introduced.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/IntegrationTestEnv.java
```diff
@@ -133,6 +166,40 @@ protected void before() throws Throwable {
...
+    TraceConfiguration.Builder traceConfigurationBuilder = TraceConfiguration.builder();
+    if (options.getCredentials() != null) {
+      traceConfigurationBuilder.setCredentials(options.getCredentials());
+    }
+    SpanExporter traceExporter =
+        TraceExporter.createWithConfiguration(
+            traceConfigurationBuilder.setProjectId(options.getProjectId()).build());
+
+    String serviceName =
+        "java-spanner-jdbc-integration-tests-" + ThreadLocalRandom.current().nextInt();
```
sakthivelmanii, line 182:
```suggestion
        "java-spanner-integration-tests-" + ThreadLocalRandom.current().nextInt();
```

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITEndToEndTracingTest.java
```diff
@@ -0,0 +1,166 @@
...
+@RunWith(JUnit4.class)
+public class ITEndToEndTracingTest {
+  public static Collection<TestEnvOptions> testEnvOptions =
+      Arrays.asList(TestEnvOptions.USE_END_TO_END_TRACING);
+  @ClassRule public static IntegrationTestEnv env = new IntegrationTestEnv(testEnvOptions);
+  private static DatabaseClient googleStandardSQLClient;
+
+  static {
+    SpannerOptionsHelper.resetActiveTracingFramework();
+    SpannerOptions.enableOpenTelemetryMetrics();
```
sakthivelmanii, line 70: why do we need to enable metrics?

google-cloud-spanner/pom.xml
```diff
@@ -455,6 +455,24 @@
...
+    </dependency>
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-trace</artifactId>
+      <version>2.51.0</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.google.api.grpc</groupId>
+      <artifactId>proto-google-cloud-trace-v1</artifactId>
```
sakthivelmanii, line 472: why are we adding this?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITEndToEndTracingTest.java
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  @AfterClass
+  public static void teardown() {
+    ConnectionOptions.closeSpanner();
+  }
+
+  @Before
+  public void initSelectValueQuery() {
+    selectValueQuery = "SELECT @p1 + @p1 ";
```
sakthivelmanii, line 94: can you initialize `selectValueQuery` as a static variable?

`private static final String selectValueQuery = "SELECT @p1 + @p1";`

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITEndToEndTracingTest.java
```diff
@@ -0,0 +1,166 @@
...
+        env.getTestHelper().getOptions().getCredentials() == null
+            ? TraceServiceSettings.newBuilder().build()
+            : TraceServiceSettings.newBuilder()
+                .setCredentialsProvider(
+                    FixedCredentialsProvider.create(
+                        env.getTestHelper().getOptions().getCredentials()))
+                .build();
+    try (TraceServiceClient client = TraceServiceClient.create(settings)) {
+      // It can take a few seconds before the trace is visible.
+      Thread.sleep(5000L);
```
sakthivelmanii, line 108: Is there any better way to know that trace is exported? I just wanted to make sure this is not flakky

surbhigarg92: Do a forceflush(), instead of adding a sleep. 

`sdkTracerProvider.forceFlush();`

google-cloud-spanner/src/test/java/com/google/cloud/spanner/IntegrationTestEnv.java
```diff
@@ -133,6 +166,40 @@ protected void before() throws Throwable {
...
+            // Always sample in this test to ensure we know what we get.
+            .setSampler(Sampler.alwaysOn())
+            .setResource(Resource.builder().put("service.name", serviceName).build())
+            .addSpanProcessor(BatchSpanProcessor.builder(traceExporter).build())
+            .build();
+    OpenTelemetrySdk openTelemetry =
+        OpenTelemetrySdk.builder()
+            .setTracerProvider(sdkTracerProvider)
+            .setPropagators(ContextPropagators.create(W3CTraceContextPropagator.getInstance()))
+            .buildAndRegisterGlobal();
```
surbhigarg92, line 194: Do not register the tracerProvider at global, instead inject OpenTelemerty object via spanner options. 

If we register it globally, we might start seeing traces coming from other parallel integration tests as `enableOpenTelemetryTraces` is a static method.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/IntegrationTestEnv.java
```diff
@@ -107,8 +133,15 @@ protected void before() throws Throwable {
...
     this.config.setUp();
-
     SpannerOptions options = config.spannerOptions();
+    if (testEnvOptions.stream()
+        .anyMatch(testEnvOption -> TestEnvOptions.USE_END_TO_END_TRACING.equals(testEnvOption))) {
+      // OpenTelemetry set up for enabling End to End tracing for all integration test env.
+      // The gRPC stub and connections are created during test env set up using SpannerOptions and
+      // are
+      // reused for executing statements.
+      options = spannerOptionsWithEndToEndTracing(options);
```
surbhigarg92, line 143: This configuration should move to `GceTestEnvConfig` class where the `SpannerOptions` is build. 

We should add a systemproperty for "End to end tracing" and use it to initialize and set opentelemetry object with tracer provider. 

Refer [this](https://github.com/googleapis/java-spanner/blob/main/google-cloud-spanner/src/test/java/com/google/cloud/spanner/GceTestEnvConfig.java#L96) to see how we are enabling Direct Path in test cases , based on system property.

surbhigarg92: @sakthivelmanii I think if there is only going to be single test for "End to end Tracing", then we might create a new Spanner Object in the Test directly. 

Or maybe extend `IntegrationTestEnv` class and use it. 

WDYT?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITEndToEndTracingTest.java
```diff
@@ -0,0 +1,166 @@
...
+      boolean foundTrace = false;
+      for (int attempts = 0; attempts < 2; attempts++) {
+        try {
+          Trace clientTrace =
```
surbhigarg92, line 112: Create the traceClient in "before" block

surbhigarg92:
@manu2 I don't understand the need for this Integration test. Do we want to verify server traces as I don't see that in the assert? 
Client traces could be verified with InMemoryExporter

google-cloud-spanner/src/test/java/com/google/cloud/spanner/IntegrationTestEnv.java
sakthivelmanii: Agree. We should probably handle it through system property like how we handled in other places.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITEndToEndTracingTest.java
manu2: removed.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITEndToEndTracingTest.java
manu2: we need to wait for sever side traces which are collected from request on Span FE and asynchronously exported later.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITEndToEndTracingTest.java
```diff
@@ -0,0 +1,166 @@
...
+              client.getTrace(env.getTestHelper().getInstanceId().getProject(), traceId);
+          // Assert Spanner Frontend Trace is present
+          assertTrue(
+              clientTrace.getSpansList().stream()
+                  .anyMatch(
+                      span ->
+                          "CloudSpannerOperation.ExecuteStreamingQuery".equals(span.getName())));
```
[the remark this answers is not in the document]
manu2: updated.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/IntegrationTestEnv.java
manu2: updated.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITEndToEndTracingTest.java
manu2: TraceClient is created before the block. This is actually the trace, not the client. Updated naming in recent commit to avoid confusion.

manu2:
> @manu2 I don't understand the need for this Integration test. Do we want to verify server traces as I don't see that in the assert? Client traces could be verified with InMemoryExporter

The test intents to verify the generation of server side traces when end to end tracing is enabled. Have updated the assertion to validate server side trace.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/IntegrationTestEnv.java
manu2: Using this opentelemetry configuration for all tests is leading to failures in currently running tests. Since this is a blocker for launch, we can track this general availability as a future backlog item.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITEndToEndTracingTest.java
```diff
@@ -0,0 +1,159 @@
...
+                .build();
+    try (TraceServiceClient client = TraceServiceClient.create(settings)) {
+      // It can take a few seconds before the trace is visible.
+      Thread.sleep(10000);
+      boolean foundTrace = false;
+      for (int attempts = 0; attempts < 2; attempts++) {
+        try {
+          Trace trace = client.getTrace(env.getTestHelper().getInstanceId().getProject(), traceId);
+          // Assert Spanner Frontend Trace is present
+          assertTrue(
```
surbhigarg92, line 106: assertTrue will result into error if the trace is not found and the test will be marked as failed, so the for loop will never actually work. Instead of this , we could try something like below and also remove the 10 second sleep before the loop.

```
Trace trace = client.getTrace(env.getTestHelper().getInstanceId().getProject(), traceId);
while (trace.getSpansList().stream()
                  .anyMatch(span -> "Spanner.ExecuteStreamingSql".equals(span.getName())
        && metricsPollingStopwatch.elapsed(TimeUnit.SECONDS) < 30) {
      // Try every 5 seconds
      Thread.sleep(5000);
    }
    assertTrue(trace.getSpansList().stream()
                  .anyMatch(span -> "Spanner.ExecuteStreamingSql".equals(span.getName());
```

google-cloud-spanner/src/test/java/com/google/cloud/spanner/IntegrationTestEnv.java
surbhigarg92: Could you please create a ticket for this backlog item.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITEndToEndTracingTest.java
manu2: Makes sense. If the getTrace calls fails, it throws an APIException. Have kept the catch block to handle it but changed the logic to poll upto 30 sec and sleep for 5 sec if not found.

google-cloud-spanner/pom.xml
manu2: We query trace using trace client which requires this dependency.
" trace = client.getTrace(env.getTestHelper().getInstanceId().getProject(), traceId)"

surbhigarg92 (approved):

google-cloud-spanner/pom.xml
```diff
@@ -455,6 +455,24 @@
       <artifactId>opentelemetry-sdk-testing</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.google.cloud.opentelemetry</groupId>
+      <artifactId>exporter-trace</artifactId>
+      <version>0.33.0</version>
```
surbhigarg92, line 461: Please remove the version from here. It should either fetch from bom or the latest version

sakthivelmanii (approved):

PR #3722, opened 2025-04-01
fix: add default implementations for Interval methods in AbstractStructReader

sagarwaal:
The PR adds default implementation for following methods of AbstractStructReader.

 getIntervalInternal
 getIntervalListInternal

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractStructReader.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractStructReader.java
@@ -67,7 +67,9 @@ public abstract class AbstractStructReader implements StructReader {
 
   protected abstract Date getDateInternal(int columnIndex);
 
-  protected abstract Interval getIntervalInternal(int columnIndex);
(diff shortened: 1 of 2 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractStructReader.java
```diff
@@ -67,7 +67,9 @@ protected String getPgJsonbInternal(int columnIndex) {
 
   protected abstract Date getDateInternal(int columnIndex);
 
-  protected abstract Interval getIntervalInternal(int columnIndex);
+  protected Interval getIntervalInternal(int columnIndex) {
```
harshachinta, line 70: Can you also remove the CLIRR checks added for these methods?

harshachinta (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractStructReader.java
sagarwaal: done.

PR #3835, opened 2025-04-24
feat: support begin with AbortedException for manager interface

harshachinta:
Aborted transactions should ideally be retried in the same transaction manager instance because the client library will ensure to populate the "Prev txn attempt's txn id" on retry (for mux sessions) which helps preserve the lock order(priority) of the transaction in scenarios of high contention thats causing txn wounding. This is achieved by using `resetForRetry()` that automatically takes care of preserving the lock order.

But if the customer application retries aborted transactions on a new TransactionManager instance with out using the `resetForRetry()`, we lose the lock order of the previous attempt.

To address this scenario of preserving the lockorder across transaction manger instance for a retry of the same transaction
Current proposal as per client lib team recommendation:
1. Overload the [Begin](https://github.com/googleapis/java-spanner/blob/9940b66dabe519495a69dcc0020e2587c16f7285/google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionManager.java#L62) API in the TransactionManager to accept an AbortedException argument.
2. Client library will populate the aborted transaction's transaction ID in the [AbortedException](https://github.com/googleapis/java-spanner/blob/efb168015aee1c4cba16f666b89be19e6f04fdbe/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbortedException.java#L27) object which is propagated up the customer's application stack.
3. On the customer end, they will just need to pass in this AbortedException object to the "Begin" API when they retry the transaction on the new transaction manager instance.

Note: This will be a no-op for regular sessions.
b/407037309

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncTransactionManager.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncTransactionManager.java
@@ -170,7 +170,16 @@ public interface AsyncTransactionManager extends AutoCloseable {
    */
   TransactionContextFuture beginAsync();
 
-  TransactionContextFuture beginAsync(AbortedException abortedException);
+  /**
+   * Initializes a new read-write transaction. This method must be called before performing any
+   * operations, and it can only be invoked once per transaction lifecycle.
+   *
+   * <p>This is especially useful in scenarios involving multiplexed sessions and when creating a
+   * new transaction for retry attempts. If {@link #resetForRetryAsync()} is not used, you can pass
+   * the {@link AbortedException} from a previous attempt here to preserve the transaction's
+   * priority.
+   */
+  TransactionContextFuture beginAsync(AbortedException exception);
 
   /**
    * Rolls back the currently active transaction. In most cases there should be no need to call this
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncTransactionManager.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncTransactionManager.java
@@ -178,6 +178,8 @@ public interface AsyncTransactionManager extends AutoCloseable {
    * new transaction for retry attempts. If {@link #resetForRetryAsync()} is not used, you can pass
    * the {@link AbortedException} from a previous attempt here to preserve the transaction's
    * priority.
+   *
+   * <p>For regular sessions, this behaves the same as {@link #beginAsync()}.
    */
   TransactionContextFuture beginAsync(AbortedException exception);
 
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncTransactionManager.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncTransactionManager.java
@@ -171,13 +171,15 @@ public interface AsyncTransactionManager extends AutoCloseable {
   TransactionContextFuture beginAsync();
 
   /**
-   * Initializes a new read-write transaction. This method must be called before performing any
-   * operations, and it can only be invoked once per transaction lifecycle.
+   * Initializes a new read-write transaction that is a retry of a previously aborted transaction.
+   * This method must be called before performing any operations, and it can only be invoked once
+   * per transaction lifecycle.
    *
-   * <p>This is especially useful in scenarios involving multiplexed sessions and when creating a
-   * new transaction for retry attempts. If {@link #resetForRetryAsync()} is not used, you can pass
-   * the {@link AbortedException} from a previous attempt here to preserve the transaction's
-   * priority.
+   * <p>This method should only be used when multiplexed sessions are enabled to create a retry for
+   * a previously aborted transaction. This method can be used instead of {@link
+   * #resetForRetryAsync()} to create a retry. Using this method or {@link #resetForRetryAsync()}
+   * will have the same effect. You must pass in the {@link AbortedException} from the previous
+   * attempt to preserve the transaction's priority.
    *
    * <p>For regular sessions, this behaves the same as {@link #beginAsync()}.
    */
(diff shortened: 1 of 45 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
```diff
@@ -793,6 +793,7 @@ public SpannerException onError(SpannerException e, boolean withBeginTransaction
         long delay = -1L;
         if (exceptionToThrow instanceof AbortedException) {
           delay = exceptionToThrow.getRetryDelayInMillis();
+          ((AbortedException) exceptionToThrow).setTransactionID(this.transactionId);
```
harshachinta, line 796: Should this be 
`this.transactionId` or 
`this.transactionId != null ? this.transactionId : this.getPreviousTransactionId()`

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncTransactionManagerImpl.java
```diff
@@ -76,14 +76,26 @@ public ApiFuture<Void> closeAsync() {
...
   public TransactionContextFutureImpl beginAsync() {
     Preconditions.checkState(txn == null, "begin can only be called once");
-    return new TransactionContextFutureImpl(this, internalBeginAsync(true));
+    return new TransactionContextFutureImpl(this, internalBeginAsync(true, ByteString.EMPTY));
   }
 
-  private ApiFuture<TransactionContext> internalBeginAsync(boolean firstAttempt) {
+  @Override
+  public TransactionContextFutureImpl beginAsync(AbortedException exception) {
+    Preconditions.checkState(txn == null, "begin can only be called once");
```
olavloite, line 84:
```suggestion
    Preconditions.checkState(txn == null, "begin can only be called once");
    Preconditions.checkNotNull(exception, "AbortedException from the previous attempt is required");
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionManager.java
```diff
@@ -61,6 +61,18 @@ enum TransactionState {
    */
   TransactionContext begin();
 
+  /**
+   * Initializes a new read-write transaction. This method must be called before performing any
```
olavloite, line 65: See my comment on `beginAsync`. I think that we should be more explicit in when this method should be used, and not describe it as 'useful in scenarios involving multiplexed sessions'. I think that could lead to wrong usage of this method.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionManagerImpl.java
```diff
@@ -53,8 +53,20 @@ public void setSpan(ISpan span) {
   @Override
   public TransactionContext begin() {
     Preconditions.checkState(txn == null, "begin can only be called once");
+    return begin(ByteString.EMPTY);
+  }
+
+  @Override
+  public TransactionContext begin(AbortedException exception) {
+    Preconditions.checkState(txn == null, "begin can only be called once");
+    ByteString previousAbortedTransactionID =
```
olavloite, line 62: Here also: Add a not-null check for the AbortedException.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionManagerImpl.java
```diff
@@ -53,8 +53,20 @@ public void setSpan(ISpan span) {
...
+        exception.getTransactionID() != null ? exception.getTransactionID() : ByteString.EMPTY;
+    return begin(previousAbortedTransactionID);
+  }
+
+  TransactionContext begin(ByteString previousTransactionId) {
     try (IScope s = tracer.withSpan(span)) {
-      txn = session.newTransaction(options, /* previousTransactionId = */ ByteString.EMPTY);
+      txn = session.newTransaction(options, /* previousTransactionId = */ previousTransactionId);
```
olavloite, line 69: nit: you can remove the comment for previousTransactionId, as that is now clear from the variable name

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
olavloite: I guess `this.transactionId != null ? this.transactionId : this.getPreviousTransactionId()` would cause the transaction priority to be preserved also when the begin call of a transaction fails, right? In that case, I think we should do that.

harshachinta: Yeah. I have updated it, Thanks.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AsyncTransactionManager.java
```diff
@@ -170,6 +170,19 @@ interface AsyncTransactionFunction<I, O> {
...
   TransactionContextFuture beginAsync();
 
+  /**
+   * Initializes a new read-write transaction. This method must be called before performing any
+   * operations, and it can only be invoked once per transaction lifecycle.
+   *
+   * <p>This is especially useful in scenarios involving multiplexed sessions and when creating a
+   * new transaction for retry attempts. If {@link #resetForRetryAsync()} is not used, you can pass
+   * the {@link AbortedException} from a previous attempt here to preserve the transaction's
+   * priority.
```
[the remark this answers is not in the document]
harshachinta: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionManager.java
harshachinta: Done. Thanks for suggesting this

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionManagerImpl.java
harshachinta: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionManagerImpl.java
harshachinta: Done

PR #3851, opened 2025-05-08
fix: change server timing duration attribute to float as per w3c

surbhigarg92:
Currently the Java implementation only parses the [integer part](https://github.com/googleapis/java-spanner/blob/32b2373d62cac3047d9686c56af278c706d7c488/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/HeaderInterceptor.java#L79).

According to the w3c standard, this should be a float (https://www.w3.org/TR/server-timing/#duration-attribute).

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/CompositeTracer.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/CompositeTracer.java
@@ -191,7 +191,6 @@ public class CompositeTracer extends BaseApiTracer {
     }
   }
 
-  @Deprecated
   public void recordGFELatency(Long gfeLatency) {
     for (ApiTracer child : children) {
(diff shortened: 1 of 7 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/CompositeTracer.java
```diff
@@ -191,10 +191,11 @@ public void addAttributes(Map<String, String> attributes) {
     }
   }
 
+  @Deprecated
   public void recordGFELatency(Long gfeLatency) {
```
sakthivelmanii, line 195: nit: Why do we have to mark as deprecated when this class is marked as @InternalApi?

rahul2393 (approved):

PR #3852, opened 2025-05-08
feat: enable AFE and gRPC metrics for DP

surbhigarg92:
Enable gRPC/AFE built-in metrics if:
1. The env var SPANNER_DISABLE_DIRECT_ACCESS_GRPC_BUILTIN_METRICS / SPANNER_DISABLE_AFE_SERVER_TIMING is explicitly set to "false", OR
2. DirectPath is enabled AND the env var is not set to "true". This allows metrics to be enabled by default when DirectPath is on, unless explicitly disabled via env.
   
This PR also skips GFE metrics when DirectPath is used.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
@@ -923,8 +923,16 @@ public class SpannerOptions extends ServiceOptions<Spanner, SpannerOptions> {
 
     @Override
     public boolean isEnableGRPCBuiltInMetrics() {
-      return "false"
-          .equalsIgnoreCase(System.getenv(SPANNER_DISABLE_DIRECT_ACCESS_GRPC_BUILTIN_METRICS));
+      // Enable gRPC built-in metrics if:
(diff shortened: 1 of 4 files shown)


rahul2393 (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1994,7 +1994,8 @@ public ApiTracerFactory getApiTracerFactory() {
   }
 
   public void enablegRPCMetrics(InstantiatingGrpcChannelProvider.Builder channelProviderBuilder) {
-    if (SpannerOptions.environment.isEnableGRPCBuiltInMetrics()) {
+    if (GapicSpannerRpc.isEnableDirectPathXdsEnv()
+        || SpannerOptions.environment.isEnableGRPCBuiltInMetrics()) {
```
sakthivelmanii, line 1998: If DP is enabled, it's not possible to disable grpc metrics

sakthivelmanii: move this logic inside `isEnableGRPCBuiltInMetrics`

sakthivelmanii (approved):

PR #3882, opened 2025-05-19
chore: update benchmarking script

sakthivelmanii:
**Description:**

Following features are added

1. Support running benchmarking for specific time(30m)
2. Support running warm up before running benchmarking
3. Support staleness
4. Support running reads

### DIFF
--- a/benchmarks/pom.xml
+++ b/benchmarks/pom.xml
@@ -34,7 +34,7 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
     <junixsocket.version>2.10.1</junixsocket.version>
-    <opentelemetry.version>1.50.0</opentelemetry.version>
+    <opentelemetry.version>1.47.0</opentelemetry.version>
   </properties>
 
   <dependencies>
@@ -54,7 +54,12 @@
     <dependency>
       <groupId>com.google.cloud.opentelemetry</groupId>
       <artifactId>exporter-metrics</artifactId>
-      <version>0.34.0</version>
+      <version>0.33.0</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-monitoring</artifactId>
+      <version>3.63.0</version>
     </dependency>
     <!-- OpenTelemetry test dependencies -->
     <dependency>
@@ -85,7 +90,7 @@
     <dependency>
       <groupId>io.opentelemetry</groupId>
       <artifactId>opentelemetry-bom</artifactId>
-      <version>1.50.0</version>
+      <version>1.47.0</version>
       <type>pom</type>
       <scope>import</scope>
     </dependency>
@@ -93,11 +98,6 @@
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-spanner</artifactId>
       <version>6.93.0</version>
-    </dependency>
-    <dependency>
-      <groupId>commons-cli</groupId>
-      <artifactId>commons-cli</artifactId>
-      <version>1.9.0</version>
     </dependency>
     <dependency>
       <groupId>com.google.auto.value</groupId>
@@ -140,8 +140,9 @@
         </configuration>
       </plugin>
       <plugin>
-        <groupId>com.coveo</groupId>
+        <groupId>com.spotify.fmt</groupId>
         <artifactId>fmt-maven-plugin</artifactId>
+        <version>2.27</version>
         <executions>
           <execution>
             <goals>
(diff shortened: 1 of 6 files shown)


benchmarks/pom.xml
```diff
@@ -34,7 +34,7 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
     <junixsocket.version>2.10.1</junixsocket.version>
-    <opentelemetry.version>1.50.0</opentelemetry.version>
```
rahul2393, line 37: Are these downgraded intended?

sakthivelmanii: yes. because it's not compatible with spanner library which has lower versions.

benchmarks/src/main/java/com/google/cloud/spanner/benchmark/JavaClientRunner.java
```diff
@@ -171,13 +173,34 @@ private Duration executeTransaction(
...
+    int key = getRandomKey();
+    columns.add("ID");
+    try (ResultSet resultSet =
+        client
+            .singleUse(TimestampBound.ofExactStaleness(staleness, TimeUnit.SECONDS))
+            .read(TABLE_NAME, KeySet.singleKey(Key.of(key)), columns)) {
+      while (resultSet.next()) {
+        for (int i = 0; i < resultSet.getColumnCount(); i++) {
+          if (resultSet.isNull(i)) {
+            numNullValues++;
```
rahul2393, line 190: Should we use AtomicLong for these values? I think these can be incremented concurrently?

benchmarks/src/main/java/com/google/cloud/spanner/benchmark/AbstractRunner.java
```diff
@@ -77,4 +111,25 @@ protected String generateRandomString() {
     ThreadLocalRandom.current().nextBytes(bytes);
     return new String(bytes, StandardCharsets.UTF_8);
   }
+
+  static class TimerConfiguration {
+    private Instant endTime = Instant.now();
```
rahul2393, line 116: Should we use Optional<Instant> to indicate “not set yet”?

benchmarks/src/main/java/com/google/cloud/spanner/benchmark/JavaClientRunner.java
sakthivelmanii: We aren't not using these values anyway.

sakthivelmanii: Will plan to remove this.

benchmarks/src/main/java/com/google/cloud/spanner/benchmark/AbstractRunner.java
```diff
@@ -18,46 +18,80 @@
...
-    int totalOperations = numClients * numOperations;
+    while (!(finishedClients.size() == configuration.getNumOfClients()))
+      for (int i = 0; i < configuration.getNumOfClients(); i++) {
+        TimerConfiguration timerConfiguration =
+            timerConfigurations.getOrDefault(i, new TimerConfiguration());
+        long totalSeconds =
+            ChronoUnit.SECONDS.between(Instant.now(), timerConfiguration.getEndTime());
+        if (completedClients.contains(i)) {
+          if (!finishedClients.contains(i)) {
+            System.out.printf("Client %s: Completed", i);
```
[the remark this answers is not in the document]
sakthivelmanii: why is this required?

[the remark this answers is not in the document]
rahul2393: I see in else we are adding \r to print results,  I am ok to keep this but for me it looks good with this change

benchmarks/src/main/java/com/google/cloud/spanner/benchmark/AbstractRunner.java
sakthivelmanii: I just don't want this to be null. that's why I set to current time. Ideally this value is set the by client. so handling the null may not be required here.

benchmarks/src/main/java/com/google/cloud/spanner/benchmark/AbstractRunner.java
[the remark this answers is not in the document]
rahul2393: A new line also will help here to see when client completes

benchmarks/src/main/java/com/google/cloud/spanner/benchmark/AbstractRunner.java
```diff
@@ -18,46 +18,80 @@
...
+            finishedClients.add(i);
+          }
+        } else {
+          System.out.printf(
+              "Client %s: %s %s Minutes %s Seconds\r",
+              i + 1, timerConfiguration.getMessage(), totalSeconds / 60, totalSeconds % 60);
+        }
+        //noinspection BusyWait
+        Thread.sleep(1000L);
```
rahul2393, line 88: Should we move this sleep to outer loop?

PR #3904, opened 2025-06-04
feat: enable ALTS hard bound token in DirectPath

rockspore:
Same as #3645 that was reverted earlier. The Cloud Spanner backend is now able to process the bound token if sent.


rockspore:
@olavloite @sakthivelmanii Could you PTAL? Thanks. The backend issue has been fixed and verified by the probers. This change is good to be introduced now.

rockspore:
cc @ko3a4ok @HailongWen

PR #3917, opened 2025-06-17
feat: Use multiplexed sessions for read-only transactions

sakthivelmanii:
**Description:**

Multi-use and single-use read-only transactions now use a single multiplexed session, instead of a session from the session pool. Multiplexed sessions can handle any number of queries and read-only transactions at the same time.

This change means that the client library will require fewer regular sessions to run the same workload. How much, depends on the distribution of read-only vs read/write transactions that are executed by the application.

Users do not need to take any specific action to re-configure their Spanner client due to this change. However, users who have configured a large number of Min or Max sessions in the session pool due to executing a large number of read-only transactions, could consider whether they can lower this number.

This change does not affect how read/write transactions are executed.

The use of multiplexed sessions can be disabled by setting the environment variable GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS=false.

See https://cloud.google.com/spanner/docs/sessions#multiplexed_sessions for more background information about multiplexed sessions.


rahul2393 (approved):

harshachinta:
Couple of suggestions - Please ignore if any of these points have already been addressed.
1. Can we update the PR title to highlight the default behavior of using multiplexed sessions is only for read-only transactions and not for any other transaction types.
2. With this change, we no longer run integration tests for read-only transactions using regular sessions. Since customers can still disable multiplexed sessions, we may need to continue verifying integration test coverage for the regular session path. Would it make sense to run these tests in continuous or nightly builds to ensure we catch any regressions?
3. Currently, all client systest configurations use regular sessions, except for one (`cloudclientmultiplexedsession`), which explicitly enables multiplexed sessions. With this change, all systests will now run read-only transactions using multiplexed sessions. To maintain coverage, we may want to modify or add a lightweight systest that explicitly disables multiplexed sessions for read-only transactions, ensuring the regular session flow is still exercised.
(Note: Since all client systests are currently disabled due to an ongoing issue, feel free to defer this if needed.)

sakthivelmanii:
@harshachinta Thanks for your inputs.

1. I have updated it. 
2. Will introduce a new pipeline to run integration tests with regular session as well.
3. Yes. I am waiting for the issues to be fixed in the systest side. I will ensure all the tests are running with mux session where as one systest is running with regular session

PR #3927, opened 2025-06-26
feat: Allow JDBC to configure directpath for connection

sakthivelmanii:
BEGIN_COMMIT_OVERRIDE
chore: Allow JDBC to configure directpath for connection
END_COMMIT_OVERRIDE

**Description:**

Directpath allows GCP VM to connect to Spanner directly instead of going through public Google frontend. It improves the performance of the client library since all the requests are directly flowing to Spanner directly.

If client workload is deployed in non GCP VM, directpath will not work. it will fallback to connecting to Google public frondend.

In JDBC, we will be providing the support where customers can add the property `attemptDirectPath=true` which will enable directpath for the connection.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionProperties.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionProperties.java
@@ -186,11 +186,13 @@ public class ConnectionProperties {
   static final ConnectionProperty<Boolean> ATTEMPT_DIRECT_PATH =
       create(
           "attemptDirectPath",
-          "Automatically configure the connection to try to connect to the Cloud Spanner "
-              + "using DirectPath (true/false). Client will try to connect to Spanner directly "
-              + "instead of connecting to GFE(Google Front End). Directpath will work only if "
-              + "client is trying to establish a connect from GCP VM. Otherwise it will fallback to"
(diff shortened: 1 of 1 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1604,6 +1604,12 @@ public Builder disableLeaderAwareRouting() {
       return this;
     }
 
+    @BetaApi
```
sakthivelmanii, line 1607: Since directpath is not public GA, I have marked as `@BetaApi`

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1000,7 +1000,7 @@ public static class Builder
     private String compressorName;
     private String emulatorHost = System.getenv("SPANNER_EMULATOR_HOST");
     private boolean leaderAwareRoutingEnabled = true;
-    private boolean attemptDirectPath = true;
```
sakthivelmanii, line 1003: Since this configuration wasn't used at all so far, I changed the default behaviour.

olavloite (approved):

olavloite (approved):

PR #3932, opened 2025-06-29
chore(x-goog-spanner-request-id): setup expectations for end-to-end request-id checks

odeke-em:
This change allows us to perform future end-to-end checks to assert on the behaviour of x-goog-spanner-request-id header values.

Updates googleapis/google-cloud-java#12250


odeke-em:
@olavloite @rahul2393 @sakthivelmanii @surbhigarg92 kindly help me run the bots!

google-cloud-spanner/src/test/java/com/google/cloud/spanner/XGoogSpannerRequestIdTest.java
```diff
@@ -157,40 +160,83 @@ private void assertMonotonicityOfIds(String prefix, List<XGoogSpannerRequestId>
...
-      return accumulated.toArray(new methodAndRequestId[0]);
+      return accumulated.toArray(new MethodAndRequestId[0]);
+    }
+
+    public void checkExpectedUnaryXGoogRequestIds(MethodAndRequestId... wantUnaryValues) {
+      MethodAndRequestId[] gotUnaryValues = this.accumulatedUnaryValues();
+      sortValues(gotUnaryValues);
+      for (int i = 0; i < gotUnaryValues.length && false; i++) {
+        System.out.println("\033[33misUnary: #" + i + ":: " + gotUnaryValues[i] + "\033[00m");
+      }
```
olavloite, line 190: I understand that this is not being printed due to the `&& false` condition in the loop, but can we add a TODO that this should be removed in a follow-up PR.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/XGoogSpannerRequestIdTest.java
```diff
@@ -157,40 +160,83 @@ private void assertMonotonicityOfIds(String prefix, List<XGoogSpannerRequestId>
...
+              accumulated.add(new MethodAndRequestId(method, values.get(i)));
             }
           });
-      return accumulated.toArray(new methodAndRequestId[0]);
+      return accumulated.toArray(new MethodAndRequestId[0]);
+    }
+
+    public void checkExpectedUnaryXGoogRequestIds(MethodAndRequestId... wantUnaryValues) {
+      MethodAndRequestId[] gotUnaryValues = this.accumulatedUnaryValues();
+      sortValues(gotUnaryValues);
```
olavloite, line 187: Is it correct that we are sorting these accumulated values here? Doesn't that mean that if we receive the request-ids out of the expected order on the (mock) server, this sort action hides any potential test failures due to the request-ids having been out of order?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/XGoogSpannerRequestIdTest.java
```diff
@@ -157,40 +160,83 @@ private void assertMonotonicityOfIds(String prefix, List<XGoogSpannerRequestId>
...
+      for (int i = 0; i < gotUnaryValues.length && false; i++) {
+        System.out.println("\033[33misUnary: #" + i + ":: " + gotUnaryValues[i] + "\033[00m");
+      }
+      assertEquals(wantUnaryValues, gotUnaryValues);
+    }
+
+    public void checkAtLeastHasExpectedUnaryXGoogRequestIds(MethodAndRequestId... wantUnaryValues) {
+      MethodAndRequestId[] gotUnaryValues = this.accumulatedUnaryValues();
+      sortValues(gotUnaryValues);
```
olavloite, line 196: Same here regarding sorting the accumulated values as above.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionClient.java
```diff
@@ -223,7 +223,7 @@ DatabaseId getDatabaseId() {
   @Override
   public XGoogSpannerRequestId nextRequestId(long channelId, int attempt) {
     return XGoogSpannerRequestId.of(
-        this.nthId, this.nthRequest.incrementAndGet(), channelId, attempt);
+        this.nthId, channelId, this.nthRequest.incrementAndGet(), attempt);
```
olavloite, line 226: The fact that we can make this change without any tests failing (or needing to be changed) seems to indicate that we have a test gap. Will that be covered by the tests that are being added in this PR, but that are currently not verifying anything, as all verifications are in `if (false)` blocks? If not, can we add tests that cover this in a follow-up PR?

olavloite (approved):
Please address any comments in a follow-up PR.

PR #3983, opened 2025-07-09
ci: Add tests for retry_info trailer

sakthivelmanii:
**Description:**

In this PR, we will be adding an Integration test in which we will check if `google.rpc.retryinfo-bin` is sent in the trailers when two transactions tries to modify the same row. 

1. gRPC request should fail with aborted error for the second transaction
2. gRPC trailers should contain the retryinfo trailer and delay

### DIFF
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/ITTransactionRetryTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/ITTransactionRetryTest.java
@@ -60,9 +60,14 @@ public class ITTransactionRetryTest {
         TransactionContext transaction2 = transactionManager2.begin();
         transaction1.executeUpdate(Statement.of("UPDATE Test SET EMPID = EMPID + 1 WHERE ID = 1"));
         transaction2.executeUpdate(Statement.of("UPDATE Test SET EMPID = EMPID + 1 WHERE ID = 1"));
-        transactionManager1.commit();
+
         AbortedException abortedException =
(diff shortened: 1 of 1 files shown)


google-cloud-spanner/src/test/java/com/google/cloud/spanner/ITTransactionRetryTest.java
```diff
@@ -0,0 +1,71 @@
...
+              return null;
+            });
+
+    try (TransactionManager transactionManager1 = databaseClient.transactionManager()) {
+      try (TransactionManager transactionManager2 = databaseClient.transactionManager()) {
+        TransactionContext transaction1 = transactionManager1.begin();
+        TransactionContext transaction2 = transactionManager2.begin();
+        transaction1.executeUpdate(Statement.of("UPDATE Test SET EMPID = EMPID + 1 WHERE ID = 1"));
+        transaction2.executeUpdate(Statement.of("UPDATE Test SET EMPID = EMPID + 1 WHERE ID = 1"));
+        transactionManager1.commit();
```
olavloite, line 63: nit: There is no guarantee that these statements will always succeed. This transaction could also be aborted by Spanner for internal reasons, so you should put this entire code block in a retry loop to ensure that flaky aborts here does not cause flaky test failures.

sakthivelmanii: done

olavloite (approved):

PR #4000, opened 2025-07-23
test: added `getSnapshotTimestamp()` method

skuruppu:
The snapshot timestamp is the timestamp at which all the reads in a transaction ran when using repeatable read isolation. The method is only intended to be used for internal testing purposes, so the snapshot timestamp will be `null` if attempted to be used externally.


skuruppu:
Requires https://github.com/googleapis/java-spanner/pull/3995 to be merged first.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/CommitResponse.java
```diff
@@ -41,6 +42,18 @@ public Timestamp getCommitTimestamp() {
     return Timestamp.fromProto(proto.getCommitTimestamp());
   }
 
+  /**
+   * Returns a {@link Timestamp} representing the timestamp at which all reads in the transaction
+   * ran at, if the transaction ran at repeatable read isolation in internal test environments, and
+   * otherwise returns null.
+   */
+  public @Nullable Timestamp getSnapshotTimestamp() {
```
sakthivelmanii, line 50: nit: Can we add unit test for this method?

skuruppu: That's a good idea. Done.

skuruppu:
Thanks for the review. PTAL @sakthivelmanii.

sakthivelmanii (approved):

PR #4010, opened 2025-08-05
feat: support read lock mode for R/W transactions

shobhitsg:
Samples and integration tests will be added in separate PRs.

### DIFF
--- a/.github/scripts/update_generation_config.sh
+++ b/.github/scripts/update_generation_config.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -e
+set -ex
 # This script should be run at the root of the repository.
 # This script is used to update googleapis_commitish, gapic_generator_version,
 # and libraries_bom_version in generation configuration at the time of running
--- a/.github/workflows/hermetic_library_generation.yaml
+++ b/.github/workflows/hermetic_library_generation.yaml
@@ -37,7 +37,7 @@ jobs:
       with:
         fetch-depth: 0
         token: ${{ secrets.CLOUD_JAVA_BOT_TOKEN }}
-    - uses: googleapis/sdk-platform-java/.github/scripts@v2.60.2
+    - uses: googleapis/sdk-platform-java/.github/scripts@v2.61.0
       if: env.SHOULD_RUN == 'true'
       with:
         base_ref: ${{ github.base_ref }}
--- a/.github/workflows/unmanaged_dependency_check.yaml
+++ b/.github/workflows/unmanaged_dependency_check.yaml
@@ -17,6 +17,6 @@ jobs:
           # repository
           .kokoro/build.sh
       - name: Unmanaged dependency check
-        uses: googleapis/sdk-platform-java/java-shared-dependencies/unmanaged-dependency-check@google-cloud-shared-dependencies/v3.50.2
+        uses: googleapis/sdk-platform-java/java-shared-dependencies/unmanaged-dependency-check@google-cloud-shared-dependencies/v3.51.0
         with:
           bom-path: google-cloud-spanner-bom/pom.xml
--- a/.kokoro/presubmit/graalvm-native-a.cfg
+++ b/.kokoro/presubmit/graalvm-native-a.cfg
@@ -3,7 +3,7 @@
 # Configure the docker image for kokoro-trampoline.
 env_vars: {
   key: "TRAMPOLINE_IMAGE"
-  value: "gcr.io/cloud-devrel-public-resources/graalvm_sdk_platform_a:3.50.2" # {x-version-update:google-cloud-shared-dependencies:current}
+  value: "gcr.io/cloud-devrel-public-resources/graalvm_sdk_platform_a:3.51.0" # {x-version-update:google-cloud-shared-dependencies:current}
 }
 
 env_vars: {
--- a/.kokoro/presubmit/graalvm-native-b.cfg
+++ b/.kokoro/presubmit/graalvm-native-b.cfg
@@ -3,7 +3,7 @@
 # Configure the docker image for kokoro-trampoline.
 env_vars: {
   key: "TRAMPOLINE_IMAGE"
-  value: "gcr.io/cloud-devrel-public-resources/graalvm_sdk_platform_b:3.50.2" # {x-version-update:google-cloud-shared-dependencies:current}
+  value: "gcr.io/cloud-devrel-public-resources/graalvm_sdk_platform_b:3.51.0" # {x-version-update:google-cloud-shared-dependencies:current}
 }
 
 env_vars: {
--- a/.kokoro/presubmit/graalvm-native-c.cfg
+++ b/.kokoro/presubmit/graalvm-native-c.cfg
@@ -3,7 +3,7 @@
 # Configure the docker image for kokoro-trampoline.
 env_vars: {
   key: "TRAMPOLINE_IMAGE"
-  value: "gcr.io/cloud-devrel-public-resources/graalvm_sdk_platform_c:3.50.2" # {x-version-update:google-cloud-shared-dependencies:current}
+  value: "gcr.io/cloud-devrel-public-resources/graalvm_sdk_platform_c:3.51.0" # {x-version-update:google-cloud-shared-dependencies:current}
 }
 
 env_vars: {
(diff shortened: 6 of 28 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1724,6 +1727,13 @@ public DefaultReadWriteTransactionOptionsBuilder setIsolationLevel(
           return this;
         }
 
+        public DefaultReadWriteTransactionOptionsBuilder setReadLockMode(
```
sakthivelmanii, line 1730: Do we need this at a client level as well?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
```diff
@@ -155,9 +156,44 @@ public static TransactionOption commitStats() {
...
+  /**
+   * Returns a {@link TransactionOption} to set the desired {@link ReadLockMode} for a read-write
+   * transaction.
+   *
+   * <p>This option controls the locking behavior for read operations and queries within a
+   * read-write transaction. It works in conjunction with the transaction's {@link IsolationLevel}.
+   *
+   * <ul>
+   *   <li>{@link ReadLockMode#PESSIMISTIC}: Read locks are acquired immediately on read. This mode
+   *       primarily applies to transactions with {@code SERIALIZABLE} isolation.
```
sakthivelmanii, line 176: nit: Should we say this is a default ReadLockMode for SERIALIZABLE isolation level?

sakthivelmanii (dismissed):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
shobhitsg: As discussed offline, it's not needed at client level for now as users rarely need to switch lock modes for a single isolation level. Instead, they prefer to set the lock mode once at a global level and not modify it afterward.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
olavloite, line 176: What does this sentence mean?

```
This mode primarily applies to transactions with {@code SERIALIZABLE} isolation.
```

That it has no effect on RR? Or that it has a strong effect on Serializable?

What I'm aiming at is: Can we make this more concrete? As a customer, I would have no idea what this means.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
```diff
@@ -155,9 +156,44 @@ public static TransactionOption commitStats() {
...
+   *   <li>{@link ReadLockMode#OPTIMISTIC}: Locks for reads within the transaction are not acquired
+   *       on read. Instead the locks are acquired on a commit to validate that read/queried data
```
olavloite, line 178: nit:

```suggestion
   *       on read. Instead the locks are acquired on commit to validate that read/queried data
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
```diff
@@ -155,9 +156,44 @@ public static TransactionOption commitStats() {
...
+   *       has not changed since the transaction started. If a conflict is detected, the transaction
+   *       will fail. This mode applies to transactions with {@code SERIALIZABLE} isolation.
+   *   <li>{@link ReadLockMode#READ_LOCK_MODE_UNSPECIFIED}: This is the default if no mode is set.
+   *       The locking behavior depends on the isolation level:
+   *       <ul>
+   *         <li>For {@code REPEATABLE_READ} isolation: Locking semantics default to {@code
+   *             OPTIMISTIC}. However, validation checks at commit are only performed for reads
+   *             within queries using {@code SELECT FOR UPDATE}, statements with {@code
```
olavloite, line 186:
```suggestion
   *             OPTIMISTIC}. However, validation checks at commit are only performed for
   *             queries using {@code SELECT FOR UPDATE}, statements with {@code
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
```diff
@@ -155,9 +156,44 @@ public static TransactionOption commitStats() {
...
+   *             LOCK_SCANNED_RANGES} hints, or DML statements. <br>
```
olavloite, line 187:
```suggestion
   *             LOCK_SCANNED_RANGES} hints, and DML statements. <br>
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
```diff
@@ -155,9 +156,44 @@ public static TransactionOption commitStats() {
...
+  @Deprecated
   public static TransactionOption optimisticLock() {
-    return OPTIMISTIC_LOCK_OPTION;
+    return Options.readLockMode(ReadLockMode.OPTIMISTIC);
+  }
+
+  /**
+   * Returns a {@link TransactionOption} to set the desired {@link ReadLockMode} for a read-write
+   * transaction.
+   *
```
olavloite, line 170: Can we also add some generic advice when to use what? (Otherwise, I think that most users will automatically assume that 'optimistic is always better than pessimistic').

So for example something like this:

```suggestion
   * Optimistic locking can improve throughput, as transactions wait with taking locks until they try to commit,
   * but increases the probability that Spanner needs to abort a transaction due to lock conflicts. Pessimistic
   * locking instructs transactions to take locks earlier, which means that concurrent transactions might have
   * to wait for other transactions to release locks before proceeding. Pessimistic locking reduces the probability
   * that Spanner needs to abort a transaction due to lock conflicts.
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionImpl.java
```diff
@@ -283,6 +284,10 @@ public CommitResponse writeAtLeastOnceWithOptions(
     if (options.isolationLevel() != null) {
       transactionOptionsBuilder.setIsolationLevel(options.isolationLevel());
     }
+    if (options.readLockMode() != null) {
+      transactionOptionsBuilder.setReadWrite(
+          ReadWrite.newBuilder().setReadLockMode(options.readLockMode()));
```
olavloite, line 289: It does not really matter now, as there are no other read/write options being set here, but I see a potential for a bug in the future: Can we update the existing ReadWrite object on the transactionOptionsBuilder, instead of setting a new one?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1724,6 +1727,13 @@ public DefaultReadWriteTransactionOptionsBuilder setIsolationLevel(
...
+            ReadLockMode readLockMode) {
+          transactionOptionsBuilder.setReadWrite(
+              ReadWrite.newBuilder().setReadLockMode(readLockMode));
```
olavloite, line 1733: Here also: Can we update the existing ReadWrite object on transactionOptionsBuilder, instead of setting a new one? I foresee potential bugs in the future if we leave this as-is.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Options.java
shobhitsg: I initially added some [advice](https://github.com/googleapis/java-spanner/commit/1c38bc5e3849ac55e176a83f5c50f4ed3a6e3153#diff-992d480774d874d97423da09934b0395ef460130558e2fd6f7b1461ba32535c0R173-R185) to the document but removed it because I felt it deserved a more detailed explanation in a separate link. I've now added the advice back, and we will update this document once the RR + Pessimistic feature is launched.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionImpl.java
shobhitsg: Agreed. Updated.

shobhitsg:
@sakthivelmanii  @olavloite: I've addressed the review comments. PTAL.

sakthivelmanii (approved):

PR #4029, opened 2025-08-28
fix: skip session delete in case of multiplexed sessions

rahul2393:


rahul2393:
When trying to delete multiplexed session, backend returns 
```
com.google.cloud.spanner.SpannerException: INVALID_ARGUMENT: com.google.api.gax.rpc.InvalidArgumentException: io.grpc.StatusRuntimeException: INVALID_ARGUMENT: Multiplexed sessions may not be deleted
```
Which is not expected by customers who does cleanup() for batch transactions.

PR #4030, opened 2025-09-01
test: Integration Tests setup for experimental host

sagnghos:
Configure integration tests to run against an experimental host across 4 different setups
- Plain Text
- TLS
- mTLS
- TLS + access token (generated from username/password login)
 
Introduced ExperimentalHostHelper to manage all the different setups for the test

Changes required:
- Disabling features like Backups and Database Roles which are yet to be available in the host
- Disabling features like data boost / instance administration...etc which are not applicable for this host
- Disabling 3 tests which are failing specific for this host with tracking bugs : 
     1.  b/441255724 (drop db protection failure)
     2. b/441255097 (invalid json values timing out instead of graceful error)
     3. b/442339325 (validation of allow_timestamp column type in updateddl present specifically for Cloud Spanner hosts)

### DIFF
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITCommitTimestampTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITCommitTimestampTest.java
@@ -229,7 +229,9 @@ public class ITCommitTimestampTest {
 
   @Test
   public void invalidColumnType() throws Exception {
-    assumeFalse("Validation not available in experimental host mode", isExperimentalHost());
+    assumeFalse(
+        "Validation currently not available in experimental host mode - tracked via b/442339325",
+        isExperimentalHost());
     // error_catalog error OptionErrorList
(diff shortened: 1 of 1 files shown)


google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITCommitTimestampTest.java
```diff
@@ -227,6 +229,7 @@ public void invalidColumnOptionValue() throws Exception {
 
   @Test
   public void invalidColumnType() throws Exception {
+    assumeFalse("Validation not available in experimental host mode", isExperimentalHost());
```
sakthivelmanii, line 232: why are we skipping it?

sagnghos: Added a tracking bug in the skip statement.

Currently the validation is specifically done for Cloud Spanner hosts on server side. And the changes required to add the validation for this experimental host was not simple enough. After discussion we had decided to skip it for the time being

PR #4196, opened 2025-11-06
chore: add CALL keyword to list of query keywords

olavloite:
Add the `CALL` keyword to the list of query keywords, so this can be used in the Connection API (and JDBC+PGAdapter).


olavloite:
cc @thiagotnunes

PR #4263, opened 2025-12-12
feat: include RequestID in requests and errors

olavloite:
- Send a RequestID to Spanner for each request
- Make sure that the attempt number of the RequestID is increased if the RPC is retried.
- Include the RequestID in every error that is thrown due to an error that is returned by Spanner.

This change removes most of the manual handling of RequestIDs, and instead lets the `SpannerRpc` implementation and a couple of interceptors handle RequestIDs.


google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -825,7 +827,7 @@ public OperationFuture<ResponseT, MetadataT> call() {
             isRetry = true;
 
             if (operationName == null) {
-              GrpcCallContext context = newCallContext(null, instanceName, initialRequest, method);
+              GrpcCallContext context = newAdminCallContext(instanceName, initialRequest, method);
```
olavloite: This isn't directly related, but a small cleanup to prevent each of these methods to invoke `newCallContext(null, ...)`, and instead use a more descriptive method.

PR #4288, opened 2026-01-06
fix: Retry creation of multiplexed session

sakthivelmanii:
**Description:**

Currently when **multiplexed** session fails with any error, we are storing the exception in the session reference and re-throwing that error to all the subsequent requests. This will cause the library to stall since no further requests will be processed successfully. It's a general expectation that all RPC requests are expected due to CPU, Network, GFE and other factors.

**Proposed solution:**
We will be retrying creation of multiplexed session for the duration of `waitForMinSessions`. If a client is having a cold start issues(higher CPU/memory during initial start), they can increase the `waitForMinSessions` time for the client to retry

### DIFF
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClientMockServerTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClientMockServerTest.java
@@ -285,7 +285,7 @@ public class MultiplexedSessionDatabaseClientMockServerTest extends AbstractMock
                     .setUseMultiplexedSession(true)
                     .setUseMultiplexedSessionForRW(true)
                     .setUseMultiplexedSessionPartitionedOps(true)
-                    .setWaitForMinSessionsDuration(Duration.ofSeconds(1))
+                    .setWaitForMinSessionsDuration(Duration.ofMillis(200))
                     .setFailOnSessionLeak()
                     .build())
             .build()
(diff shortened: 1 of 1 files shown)


google-cloud-spanner/src/test/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClientMockServerTest.java
```diff
@@ -210,6 +210,271 @@ public void testMaintainerMaintainsMultipleClients() {
...
+                        "CallOptions deadline exceeded after 22.986872393s. "
+                            + "Name resolution delay 6.911918521 seconds. [closed=[], "
+                            + "open=[[connecting_and_lb_delay=32445014148ns, was_still_waiting]]]")
+                    .asRuntimeException(),
+                Status.DEADLINE_EXCEEDED
+                    .withDescription(
+                        "CallOptions deadline exceeded after 22.986872393s. "
+                            + "Name resolution delay 6.911918521 seconds. [closed=[], "
+                            + "open=[[connecting_and_lb_delay=32445014148ns, was_still_waiting]]]")
+                    .asRuntimeException())));
```
olavloite: If I understand this correctly, then it will wait for 600ms and then return a DEADLINE_EXCEEDED error based on the exceptions. This will again trigger a retry, which will time out. Can we:
1. Lower the wait time to keep the tests as quick as possible
2. Use a different retryable error code for the exceptions to show that the timeout error is really coming from the retry and not the exceptions that are being returned

PR #4292, opened 2026-01-08
feat: add SsFormat encoding library

rahul2393:
This commit adds the foundational SsFormat class that provides sortable string format (ssformat) encoding utilities. This encoding is used by Spanner for key ordering and routing.

Key features:
- Composite tag encoding for interleaved tables
- Signed/unsigned integer encoding (increasing/decreasing)
- String and bytes encoding with proper escaping
- Double encoding with proper sign handling
- Timestamp and UUID encoding
- Null value markers with configurable ordering
- TargetRange class for key range representation

This is part of the experimental location-aware routing for improved latency.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SsFormat.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SsFormat.java
@@ -93,7 +93,7 @@ public final class SsFormat {
   // Offset to make negative timestamp seconds sort correctly
   private static final long TIMESTAMP_SECONDS_OFFSET = 1L << 63;
 
-  public static void appendCompositeTag(GrowableByteArrayOutputStream out, int tag) {
+  public static void appendCompositeTag(UnsynchronizedByteArrayOutputStream out, int tag) {
     if (tag == K_OBJECT_EXISTENCE_TAG || tag <= 0 || tag > K_MAX_FIELD_TAG) {
       throw new IllegalArgumentException("Invalid tag value: " + tag);
     }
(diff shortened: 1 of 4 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SsFormat.java
```diff
@@ -0,0 +1,371 @@
...
+                ? ASCENDING_ZERO_ESCAPE
+                : ASCENDING_ZERO_ESCAPE); // After inversion, 0xFF becomes 0x00. Escape for 0x00
+        // (inverted) is F0.
+        // If increasing, 0x00 -> 0x00 F0.
+      } else if (unsignedByte == 0xFF) {
+        out.write((byte) 0xFF);
+        out.write(
+            decreasing
+                ? ASCENDING_FF_ESCAPE
+                : ASCENDING_FF_ESCAPE); // After inversion, 0x00 becomes 0xFF. Escape for 0xFF
```
olavloite: Same as above: Either remove the ternary operation, or add a test that fails with this implementation, and succeeds after fixing this.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SsFormat.java
```diff
@@ -0,0 +1,371 @@
...
+ * limitations under the License.
+ */
+
+package com.google.cloud.spanner.spi.v1;
+
+import com.google.protobuf.ByteString;
+import java.io.ByteArrayOutputStream;
+import java.nio.charset.StandardCharsets;
+
+public final class SsFormat {
```
olavloite: nit: mark as `@InternalApi`

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SsFormat.java
```diff
@@ -0,0 +1,371 @@
...
+  }
+
+  private SsFormat() {}
+
+  private static final int IS_KEY = 0x80;
+  private static final int TYPE_MASK = 0x7f;
+
+  // HeaderType enum values (selected)
+  private static final int TYPE_UINT_1 = 0;
+  private static final int TYPE_UINT_9 = 8;
```
olavloite: (here and for many of the other constants here): Can we either add some comments, or otherwise change the names of these constants so they are easier to understand. What is `TYPE_UINT_1`? Why is that `0`? And why is `TYPE_UINT_9` equal to `8`?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SsFormat.java
```diff
@@ -0,0 +1,371 @@
...
+    buf[3] = (byte) (hi >> 32);
+    buf[4] = (byte) (hi >> 24);
+    buf[5] = (byte) (hi >> 16);
+    buf[6] = (byte) (hi >> 8);
+    buf[7] = (byte) hi;
+    buf[8] = (byte) (lo >> 24);
+    buf[9] = (byte) (lo >> 16);
+    buf[10] = (byte) (lo >> 8);
+    buf[11] = (byte) lo;
+    return buf;
```
olavloite: You can simplify this to:

```
    byte[] buf = new byte[12];
    ByteBuffer.wrap(buf).order(ByteOrder.BIG_ENDIAN).putLong(hi).putInt(nanos);
    return buf;
```

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SsFormat.java
```diff
@@ -0,0 +1,322 @@
...
+  public static void appendBoolDecreasing(GrowableByteArrayOutputStream out, boolean value) {
+    // BOOL uses decreasing unsigned int encoding: false=0, true=1, then inverted
+    // For values 0 and 1, payload is always 1 byte
+    int encoded = value ? 1 : 0;
+    out.write(
+        (byte) (IS_KEY | TYPE_DECREASING_UINT_1)); // Header for 1-byte decreasing unsigned int
+    out.write((byte) ((~encoded & 0x7F) << 1)); // Inverted payload
+  }
+
+  private static void appendIntInternal(
```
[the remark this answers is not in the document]
rahul2393: Changed to appendInt64Internal

PR #4300, opened 2026-01-12
chore: remove session pool and only use multiplexed sessions

rahul2393:
This PR removes the legacy SessionPool from the Java Spanner client, transitioning to multiplexed sessions as the only session management mechanism. Multiplexed sessions allow a single session to support multiple parallel transactions, eliminating the need for session pooling.


rahul2393:
Closing this, since we are moving ahead with original PR

PR #4311, opened 2026-01-21
Secure parameters

olavloite:
(Creating Draft PR so it's easier to leave comments)


google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ClientContextMockServerTest.java
```diff
@@ -0,0 +1,244 @@
...
+          mockSpanner
+              .getRequestsOfType(CommitRequest.class)
+              .get(0)
+              .getRequestOptions()
+              .getClientContext());
+    }
+  }
+
+  @Test
+  public void testBeginTransaction_PropagatesClientContext() {
```
olavloite: This test seems to be a bit confused about what is the correct/expected behavior.

PR #4316, opened 2026-01-28
feat: ClientContext and secure parameters support

aseering:
Add support for `ClientContext` and secure parameters.


google-cloud-spanner/src/test/java/com/google/cloud/spanner/connection/ClientContextMockServerTest.java
```diff
@@ -0,0 +1,330 @@
...
+      mockSpanner.abortNextStatement();
+
+      connection.setClientContext(CLIENT_CONTEXT);
+      connection.beginTransaction();
+      connection.executeUpdate(INSERT_STATEMENT);
+
+      // We expect multiple ExecuteSqlRequests.
+      // 1. The first one fails with ABORTED. This request includes the BeginTransaction option.
+      // 2. The retry.
+      // Note: precise count depends on Gax retry logic vs Spanner retry logic interaction.
```
olavloite: No, Gax is not involved in this retry logic, and the number of `ExecuteSqlRequests` from this test should always be 2. However, due to the other test in this method above, the number of requests here will be 3, as the requests on the mock server are only cleared after the test method. You can force-clear the requests by calling `mockSpanner.clearRequests()`

PR #4340, opened 2026-02-11
chore: route multi-use read-only txn reads via location-aware cache

rahul2393:
- Added support for multi-use read-only transaction routing so reads/queries are sent to the cached routed server (based
    on routing hints/key ranges) instead of being pinned by transaction affinity/default endpoint.
  - Introduced read-only transaction tracking in KeyAwareChannel to:
      - identify read-only txn IDs from BeginTransaction responses
      - preserve strong/stale leader preference for subsequent requests
      - bypass affinity lookup/recording for those txn IDs
  - Extended ChannelFinder with overloads that allow explicit preferLeader routing for ReadRequest and ExecuteSqlRequest.
  - Ensured cleanup removes both affinity and read-only tracking state when transaction is closed.
  - Added unit tests in KeyAwareChannelTest for:
      - independent per-request routing within the same read-only txn
      - no affinity recording for read-only txns
      - cleanup path coverage.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/KeyAwareChannel.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/KeyAwareChannel.java
@@ -18,6 +18,8 @@ package com.google.cloud.spanner.spi.v1;
 
 import com.google.api.core.InternalApi;
 import com.google.api.gax.grpc.InstantiatingGrpcChannelProvider;
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
 import com.google.protobuf.ByteString;
 import com.google.spanner.v1.BeginTransactionRequest;
 import com.google.spanner.v1.CommitRequest;
@@ -51,6 +53,7 @@ import javax.annotation.Nullable;
  */
 @InternalApi
 final class KeyAwareChannel extends ManagedChannel {
+  private static final long MAX_TRACKED_READ_ONLY_TRANSACTIONS = 100_000L;
   private static final String STREAMING_READ_METHOD = "google.spanner.v1.Spanner/StreamingRead";
   private static final String STREAMING_SQL_METHOD =
       "google.spanner.v1.Spanner/ExecuteStreamingSql";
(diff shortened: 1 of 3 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/KeyAwareChannel.java
```diff
@@ -67,6 +67,9 @@ final class KeyAwareChannel extends ManagedChannel {
   private final Map<String, SoftReference<ChannelFinder>> channelFinders =
       new ConcurrentHashMap<>();
   private final Map<ByteString, String> transactionAffinities = new ConcurrentHashMap<>();
+  // Maps read-only transaction IDs to their preferLeader value.
+  // Strong reads → true (prefer leader), Stale reads → false (any replica).
+  private final Map<ByteString, Boolean> readOnlyTransactions = new ConcurrentHashMap<>();
```
olavloite: nit: could we give the field a more descriptive name? Something like `readOnlyTxPreferLeader`?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/KeyAwareChannel.java
```diff
@@ -325,7 +352,12 @@ public void sendMessage(RequestT message) {
             finder = parentChannel.getOrCreateChannelFinder(databaseId);
             endpoint = finder.findServer(reqBuilder);
           }
-          allowDefaultAffinity = true;
+          if (reqBuilder.hasOptions() && reqBuilder.getOptions().hasReadOnly()) {
```
olavloite: I think that we should add similar logic for `ExecuteSqlRequest` and `ReadRequest`, as those can also include an inlined `BeginTransactionOption`. That might not be something that we use today for read-only transactions, but it is likely to change in the future. (Alternatively, we should have a test that fails if we see an `ExecuteSqlRequest` with an inlined begin. That way we get a signal if this part of the implementation changes in the future, and are forced to fix it here then)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/KeyAwareChannel.java
```diff
@@ -184,12 +187,34 @@ private void clearAffinity(ByteString transactionId) {
       return;
     }
     transactionAffinities.remove(transactionId);
+    readOnlyTransactions.remove(transactionId);
```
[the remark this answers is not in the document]
rahul2393: added bounded cache for this map, if we go beyond it request will follow the default host.

PR #4371, opened 2026-03-02
feat: ability to update credentials on long running client

ldetmer:
This feature introduces the ability to change Service Account credentials on a long running client without having to restart the client.

Usage:
1) Create a MutableCredentials instance wrapping initial Service Account Credentials.
2) Create a client with MutableCredentials created in step 1 
3) Whenever necessary call MutableCredentials.updateCredentials with updated credentials.  The next API call on the client will utilize the new credentials.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/MutableCredentials.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/MutableCredentials.java
@@ -50,7 +50,8 @@ public class MutableCredentials extends Credentials {
     this(credentials, SpannerOptions.SCOPES);
   }
 
-  public MutableCredentials(@Nonnull ServiceAccountCredentials credentials, @Nonnull Set<String> scopes) {
+  public MutableCredentials(
+      @Nonnull ServiceAccountCredentials credentials, @Nonnull Set<String> scopes) {
     Objects.requireNonNull(credentials, "credentials must not be null");
     Objects.requireNonNull(scopes, "scopes must not be null");
     if (scopes.isEmpty()) {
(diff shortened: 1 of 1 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/MutableCredentials.java
```diff
@@ -0,0 +1,84 @@
...
+
+  @Override
+  public String getAuthenticationType() {
+    return delegate.getAuthenticationType();
+  }
+
+  @Override
+  public Map<String, List<String>> getRequestMetadata(URI uri) throws IOException {
+    return delegate.getRequestMetadata(uri);
+  }
```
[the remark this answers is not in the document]
ldetmer: ugh, sorry amatuer hour, thanks for bringing up, added these

samples/snippets/pom.xml
```diff
@@ -45,6 +45,7 @@
     <dependency>
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-spanner</artifactId>
+      <version>6.111.1</version>
```
ldetmer: Is this correct? Should I manually add this and then remove once the bom gets released? Or do I want to add samples until this code is released?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/MutableCredentials.java
```diff
@@ -0,0 +1,112 @@
...
+  /**
+   * Replaces the current delegate with a newly scoped credentials instance.
+   *
+   * <p>The provided {@link ServiceAccountCredentials} is scoped using the same scopes that were
+   * supplied when this {@link MutableCredentials} instance was created.
+   *
+   * @param credentials the new base service account credentials to scope and use for client
+   *     authorization.
+   */
+  public void updateCredentials(ServiceAccountCredentials credentials) {
```
lqiu96: Do you think we should we check to ensure that a non-null SA creds is passed in?

lqiu96 (approved):
Thanks! I think this looks good to me. Should cover being able to rotate the creds properly

PR #4392, opened 2026-03-25
chore: add routing hints and cache updates for begin/commit

rahul2393:
## Summary

  This change extends key-aware routing to include mutation-based `BeginTransaction` and `Commit` request payloads, and teaches the response path to consume `CacheUpdate` from `Transaction` and `CommitResponse`.

  ## What changed

  - Populate `RoutingHint` on `BeginTransaction` when a `mutationKey` is present
  - Populate `RoutingHint` on `Commit` from the first mutation in the request
  - Keep existing transaction-affinity routing for `Commit`/`Rollback`
  - Apply `CacheUpdate` from:
    - `Transaction`
    - `CommitResponse`
  - Add focused `KeyAwareChannel` tests for the new request/response behavior

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/ChannelFinder.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/ChannelFinder.java
@@ -94,40 +94,36 @@ public final class ChannelFinder {
   }
 
   public ChannelEndpoint findServer(BeginTransactionRequest.Builder reqBuilder) {
-    if (!reqBuilder.hasMutationKey()
-        || !recipeCache.computeKeys(
-            reqBuilder.getMutationKey(), reqBuilder.getRoutingHintBuilder())) {
+    if (!reqBuilder.hasMutationKey()) {
       return null;
     }
-    return fillRoutingHint(
+    return routeMutation(
+        reqBuilder.getMutationKey(),
         preferLeader(reqBuilder.getOptions()),
-        KeyRangeCache.RangeMode.COVERING_SPLIT,
-        DirectedReadOptions.getDefaultInstance(),
         reqBuilder.getRoutingHintBuilder());
   }
 
-  public void fillRoutingHint(CommitRequest.Builder reqBuilder) {
+  public ChannelEndpoint fillRoutingHint(CommitRequest.Builder reqBuilder) {
     if (reqBuilder.getMutationsCount() == 0) {
-      return;
+      return null;
     }
     Mutation mutation = reqBuilder.getMutations(0);
-    if (!recipeCache.computeKeys(mutation, reqBuilder.getRoutingHintBuilder())) {
-      return;
-    }
-    fillRoutingHint(
-        /* preferLeader= */ true,
-        KeyRangeCache.RangeMode.COVERING_SPLIT,
-        DirectedReadOptions.getDefaultInstance(),
-        reqBuilder.getRoutingHintBuilder());
+    return routeMutation(mutation, /* preferLeader= */ true, reqBuilder.getRoutingHintBuilder());
   }
 
-  private ChannelEndpoint fillRoutingHint(
-      TransactionSelector transactionSelector,
-      DirectedReadOptions directedReadOptions,
-      KeyRangeCache.RangeMode rangeMode,
-      RoutingHint.Builder hintBuilder) {
+  private ChannelEndpoint routeMutation(
+      Mutation mutation, boolean preferLeader, RoutingHint.Builder hintBuilder) {
+    recipeCache.applySchemaGeneration(hintBuilder);
+    TargetRange target = recipeCache.mutationToTargetRange(mutation);
+    if (target == null) {
+      return null;
+    }
+    recipeCache.applyTargetRange(hintBuilder, target);
     return fillRoutingHint(
-        preferLeader(transactionSelector), rangeMode, directedReadOptions, hintBuilder);
+        preferLeader,
+        KeyRangeCache.RangeMode.COVERING_SPLIT,
+        DirectedReadOptions.getDefaultInstance(),
+        hintBuilder);
   }
 
   private ChannelEndpoint fillRoutingHint(
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/ChannelFinder.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/ChannelFinder.java
@@ -19,8 +19,10 @@ package com.google.cloud.spanner.spi.v1;
 import com.google.api.core.InternalApi;
 import com.google.spanner.v1.BeginTransactionRequest;
 import com.google.spanner.v1.CacheUpdate;
+import com.google.spanner.v1.CommitRequest;
 import com.google.spanner.v1.DirectedReadOptions;
 import com.google.spanner.v1.ExecuteSqlRequest;
+import com.google.spanner.v1.Mutation;
 import com.google.spanner.v1.ReadRequest;
 import com.google.spanner.v1.RoutingHint;
 import com.google.spanner.v1.TransactionOptions;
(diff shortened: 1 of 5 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/ChannelFinder.java
```diff
@@ -92,23 +94,31 @@ public ChannelEndpoint findServer(ExecuteSqlRequest.Builder reqBuilder, boolean
...
+      return;
+    }
+    fillRoutingHint(
+        /* preferLeader= */ true,
+        KeyRangeCache.RangeMode.COVERING_SPLIT,
+        DirectedReadOptions.getDefaultInstance(),
+        reqBuilder.getRoutingHintBuilder());
   }
 
   private ChannelEndpoint fillRoutingHint(
```
olavloite: Maybe not directly related to this change, but I noticed that this method is unused. Can we remove it?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/KeyAwareChannel.java
```diff
@@ -367,11 +371,19 @@ public void sendMessage(RequestT message) {
...
-            transactionIdToClear = request.getTransactionId();
+          CommitRequest.Builder reqBuilder = ((CommitRequest) message).toBuilder();
+          String databaseId = parentChannel.extractDatabaseIdFromSession(reqBuilder.getSession());
+          if (databaseId != null) {
+            finder = parentChannel.getOrCreateChannelFinder(databaseId);
+          }
+          if (finder != null) {
+            finder.fillRoutingHint(reqBuilder);
+          }
+          if (!reqBuilder.getTransactionId().isEmpty()) {
```
olavloite: TransactionId is empty if the `CommitRequest` has a single-use read/write transaction. Meaning that we skip this for 'write-at-least-once' transactions. Is that intentional? Normally, such a `CommitRequest` would contain mutations, meaning that we would be able to route in using the same logic as a `BeginTransactionRequest` that contains a mutation key, right?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/ChannelFinder.java
```diff
@@ -92,23 +94,31 @@ public ChannelEndpoint findServer(ExecuteSqlRequest.Builder reqBuilder, boolean
   }
 
   public ChannelEndpoint findServer(BeginTransactionRequest.Builder reqBuilder) {
-    if (!reqBuilder.hasMutationKey()) {
+    if (!reqBuilder.hasMutationKey()
+        || !recipeCache.computeKeys(
```
olavloite: I find some of this code very hard to read. E.g. the fact that `computeKeys(..)` returns a boolean and has the side-effect that it modifies the `RoutingHintBuilder` that it gets (but only if it returns true, but also sometimes when it returns false, see also below) makes this very hard to understand. I think it would be good if we would take a 'readability sweep' of both this method, but also the code in general.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/KeyRecipeCache.java
```diff
@@ -230,6 +230,23 @@ public void computeKeys(ExecuteSqlRequest.Builder reqBuilder) {
     }
   }
 
+  boolean computeKeys(Mutation mutation, RoutingHint.Builder hintBuilder) {
+    if (!schemaGeneration.isEmpty()) {
+      hintBuilder.setSchemaGeneration(schemaGeneration);
```
olavloite, line 225: This updates the hint builder, but the method might still return false if the `TargetRange` below is null. That seems weird. I would expect this method only to modify the hint builder if it also returns true. Can we otherwise at least add some documentation to the method that explains when it returns true/false, and when it modifies the hint builder?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/KeyAwareChannel.java
```diff
@@ -367,11 +371,19 @@ public void sendMessage(RequestT message) {
           }
           message = (RequestT) reqBuilder.build();
         } else if (message instanceof CommitRequest) {
-          CommitRequest request = (CommitRequest) message;
-          if (!request.getTransactionId().isEmpty()) {
-            endpoint = parentChannel.affinityEndpoint(request.getTransactionId());
-            transactionIdToClear = request.getTransactionId();
+          CommitRequest.Builder reqBuilder = ((CommitRequest) message).toBuilder();
```
olavloite: This could potentially be a very heavy operation, as a `CommitRequest` could contain a very large number of mutations (e.g. during a bulk load). Converting it back to a builder means copying all of those mutations. I think that we should only call `toBuilder()` if we really know that we need the builder (i.e. if we know that we will be setting a routing hint).

This is also true for the other types of requests here, but those are less likely to be as heavy as a `CommitRequest`.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/spi/v1/KeyAwareChannelTest.java
```diff
@@ -274,6 +279,124 @@ public void resultSetCacheUpdateRoutesSubsequentRequest() throws Exception {
     assertThat(harness.endpointCache.callCountForAddress("routed:1234")).isEqualTo(1);
   }
 
+  @Test
+  public void beginTransactionWithMutationKeyAddsRoutingHint() throws Exception {
```
olavloite, line 286: Can we add a test for `CommitRequest` with a single-use read/write transaction?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/spi/v1/KeyAwareChannelTest.java
```diff
@@ -274,6 +279,124 @@ public void resultSetCacheUpdateRoutesSubsequentRequest() throws Exception {
...
+    beginCall.sendMessage(
+        BeginTransactionRequest.newBuilder().setSession(SESSION).setMutationKey(mutation).build());
+
+    @SuppressWarnings("unchecked")
+    RecordingClientCall<BeginTransactionRequest, Transaction> beginDelegate =
+        (RecordingClientCall<BeginTransactionRequest, Transaction>)
+            harness.defaultManagedChannel.latestCall();
+
+    assertThat(beginDelegate.lastMessage).isNotNull();
+    assertThat(beginDelegate.lastMessage.getRoutingHint().getDatabaseId()).isEqualTo(7L);
```
olavloite: nit: use `assertEquals(..)` (and in general try to stick to JUnit assertions)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/KeyRecipeCache.java
```diff
@@ -230,6 +230,23 @@ public void computeKeys(ExecuteSqlRequest.Builder reqBuilder) {
     }
   }
 
+  boolean computeKeys(Mutation mutation, RoutingHint.Builder hintBuilder) {
```
olavloite: This method in general is quite hard to read, as it both returns a value that indicates whether the method did something or not, and has a potential side-effect of (maybe) populating the hintBuilder. Is there a better way to do this?

rahul2393: removed that helper from the mutation routing flow.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/KeyAwareChannel.java
```diff
@@ -371,19 +371,24 @@ public void sendMessage(RequestT message) {
...
-          if (finder != null) {
-            finder.fillRoutingHint(reqBuilder);
+          CommitRequest.Builder reqBuilder = null;
+          if (finder != null && request.getMutationsCount() > 0) {
+            reqBuilder = request.toBuilder();
+            endpoint = finder.fillRoutingHint(reqBuilder);
+            request = reqBuilder.build();
+          }
+          if (!request.getTransactionId().isEmpty()) {
+            endpoint = parentChannel.affinityEndpoint(request.getTransactionId());
```
olavloite: This overrides the endpoint that was filled above, even in the case if `parentChannel.affinityEndpoint(..)` returns null, which means that it will fall back to the default endpoint. Is that intentional? (I'm not sure if it can happen in real life, though....)

olavloite (approved):

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/KeyAwareChannel.java
rahul2393: In case when transactionID is present but no affinity, we should just forward it to the default channel, example

 For ExecuteSql / Read, if the first statement falls back to the default endpoint, we do not record default-host affinity because allowDefaultAffinity is false on that path. That means a later CommitRequest can carry a valid transactionId while affinityEndpoint still returns null, in this case we should just forward to default host.

